PR #2, opened 2015-10-04
python: Allow filtering of objects

mlyle:
This greatly simplifies looking at a range of data during a navigation
or RTH segment.


tracernz:
Looks good, merging. :D

PR #16, opened 2015-10-07
AQ32 - Add logging to OpenLog on AQ32 target.

jihlein:
Since the AQ32 board does not have an SPI flash memory chip for logging, I added the capability to log to an OpenLog device.

Code has been tested on uart4 only as I currently have uart3 and uart6 slated for other functions.  It correctly wrote logs to the OpenLog, which where then successfully parsed with the matching LogConvert.m file.  As a side note, the GNU Octave-4.0.0 download runs the Matlab m-file just fine.  Easier than the python scripts IMHO.

Here's a sample plot from my bench testing, generated from Octave:

image

The modified logging file has not been tested to log on an SPI flash chip, that would be a good test if someone can run it.

I also have the OpenLog code working on F3 targets in my private Naze32Pro and Dodo branch.


mlyle:
@jihlein Woah, really cool!

Hey, could you maybe look at porting aq32 to the PIOS_HAL stuff and moving the stuff to hwshared?  If that's too much of a pain I can do it.  Just the benefit is that it becomes easier to add it to different ports/targets.

jihlein:
Sure, I'll take a stab at it. Is there a target I can follow for a template?  Shouldn't that work be under a separate PR?

mlyle:
Much appreciated!

Yah, probably a separate PR.  Ideally it would land first.  The three targets that use it are sparky2, revomini, and pipxtreme (but pipxtreme is slightly weird).

The code to put it in sparky2 was here: https://github.com/TauLabs/TauLabs/pull/1759/files#diff-fddc7ccc67186b59e95c2e594ed8468bL37

mlyle:
jenkins, test this please

mlyle:
The other issue is the pending logging PR in taulabs that adds lots of performance for high rate logging.

mlyle:
nevermind, taulabs/taulabs#1928 is not quite merge ready

mlyle:
So I've read this.

My main concern is that openlog mutually exclusive with flash logging.  A target can log to high speed onboard flash or to openlog.  It would be nice if you could do either-- use the onboard log or use a big SD card through openlog.

jihlein:
The PR has been modified to allow choosing between the on board SPI flash or an attached OpenLog device.  It's an either or selection, you can not log to both simultaneously.  There was some creep into other targets to get a clean build, the minimal mods where done such that all targets build, and hopefully run.  PR's will need to be opened against Colibri, Quanton, and Spark2 to finish the OpenLog integration on the affected targets.

I tested on AQ32, and it logs to the OpenLog as expected.  I also tested on Quanton, and it logs to SPI flash as expected.  I can't test on the other affected targets.

Regarding the AQ32 PIOS_BOARD.c file, I have two comments/questions:

1)Is there anyway to include line 257 in the general module initializations in line 256, or is it better to leave those as separate function calls?

2)Lines 724-728 are for illustration only as the AQ32 does not have an SPI flash chip.  I intend to delete those lines before this PR gets merged.  They serve as an example of some of the mods required to the other targets that have on board SPI flash chips.

PR #17, opened 2015-10-07
AQ32 - Reformat AQ32 target to use PIOS_HAL.

jihlein:
First cut at this.  Not fully functional, uart2 (configured for GPS) does not seem to work, but uart6 (configured for DSM) does.  Continuing to troubleshoot.  Scope of PR kind of grew when I had to add PWM RX support to PIOS_HAL.c

### DIFF
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -341,6 +341,7 @@ static void PIOS_HAL_ConfigureHSUM(const struct pios_usart_cfg *usart_hsum_cfg,
  * @param[out] i2c_id ID of I2C peripheral if operated in I2C mode
  * @param[in] i2c_Cfg Adapter configuration/registers for I2C mode
  * @param[in] ppm_cfg Configuration/registers for PPM mode
+ * @param[in] pwm_cfg Configuration/registers for PWM mode
  * @param[in] led_id LED to blink when there's panics
  * @param[in] usart_dsm_hsum_cfg usart configuration for DSM/HSUM modes
  * @param[in] dsm_cfg DSM configuration for this port
@@ -355,6 +356,7 @@ void PIOS_HAL_ConfigurePort(HwSharedPortTypesOptions port_type,
 	uint32_t *i2c_id,
 	const struct pios_i2c_adapter_cfg *i2c_cfg,
 	const struct pios_ppm_cfg *ppm_cfg,
+	const struct pios_pwm_cfg *pwm_cfg,
 	uint32_t led_id,
 /* TODO: future work to factor most of these away */
 	const struct pios_usart_cfg *usart_dsm_hsum_cfg,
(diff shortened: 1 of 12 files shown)


mlyle:
Eek.  I'm sorry.

Thanks for doing this.  Whenever you feel is appropriate for me to review I will.

mlyle:
jenkins: test this please

mlyle:
@jihlein -- just FYI, if you join the d-ronin organization I think it will automatically test your PRs. (at https://github.com/d-ronin )

jihlein:
@mlyle, if you want to give this a quick once over, I could use the help.  Uart2 is not getting configured, and I can't figure out why.  I put an endless loop in the GPS config, and if I set Uart2 to GPS, it never hits that loop.  If I set Uart3 to GPS, it hits that endless loop.

I have verified Uart4 and Uart6 operates on DSM, so at least we know Uart3, 4, and 6 are  being configured.

mlyle:
Ah.  One more change needed-- I'd done it for all other targets but since aq32 was coming in at the same time it didn't get done.  The ports need parent types defined in the XML, like

```
                <field name="MainPort" units="function" type="enum" elements="1" parent="HwShared.PortTypes" defaultvalue="Disabled">
```

jihlein:
Thanks @mlyle, that last edit fixed it, GPS is now functioning properly.  I need to dig out a pwm RX and test that, then I think it's ready to merge.

jihlein:
With hardware on hand, I was able to verify DSM, PWM RX, Sync PWM at 666 Hz, GPS, and external mag.  Looks like it's ready to review.

flight/PiOS/Common/pios_hal.c
```diff
@@ -486,6 +507,8 @@ void PIOS_HAL_ConfigurePort(HwSharedPortTypesOptions port_type,
 #if defined(PIOS_INCLUDE_FRSKY_SENSOR_HUB)
 			PIOS_HAL_ConfigureCom(usart_port_cfg, 0, PIOS_COM_FRSKYSENSORHUB_TX_BUF_LEN, com_driver, &port_driver_id);
 			target = &pios_com_frsky_sensor_hub_id;
+			if (sbus_toggle)
```
mlyle: Is this appropriate?  The code I moved from other targets didn't have this-- and since you need an output inverter too I'm not sure why this is here.

mlyle:
Looks very good, other than we need to figure out about the FRSKY_SENSOR_HUB inversion.

jihlein:
AQ32 hardware has hardware inverters on both the input and output of Uart3. Both inverters are driven by the same GPIO pin.

mlyle:
I see.  Maybe we need another flag then.  Because this diff will break existing setups on other platforms where they invert both directions.

jihlein:
Might be best to move the inverter control code back in PIOS_BOARD.c then.  There's too many ways to implement the inverter in hardware, and too many things that use the inversion differently, SBUS, Sensor Hub, SPort.  That might be the cleanest solution.

jihlein:
Latest changes pushed move the inverter control back to PIOS_BOARD.c, eliminating the problem discussed above.  Verified to work with D4R-II reciever on PPM with FrSky Sensor Hub on Uart3.

mlyle:
Cool!  Thank you so much.

PR #20, opened 2015-10-08
MOTOLAB target for d-Ronin

fujin:
* Supposedly pin-compatible for Tornado f3, MotoFC, based on minimal
  modifications to the sparky1


fujin:
woops

PR #24, opened 2015-11-24
uavobjgenerator:  Ensure rebuild of flight when necessary

mlyle:
This code had one remaining weakness-- on non-windows platforms, if you
edited a file and then put it back UAVobjs might not get rebuilt due to
moving to an "old" timestamp.  Now we touch the version file to ensure
that this is OK.


mlyle:
jenkins, test this please

tracernz:
That makes sense. I'm happy to merge this ASAP.

PR #27, opened 2015-11-24
mavlink: reduce memory usage by a couple hundred bytes

mlyle:
Whether enabled or not.  Remove a serial-transmit-buffer when enabled
and save an unnecessary copy.  Only allocate the very large message
structure when the task is running.


mlyle:
jenkins, test this please

mlyle:
jenkins, test this please

PR #28, opened 2015-11-24
flight: remove double precision flight code

mlyle:


mlyle:
jenkins, test this please

flight/Libraries/WorldMagModel.c
```diff
@@ -328,7 +328,7 @@ int WMM_AssociatedLegendreFunction(WMMtype_CoordSpherical * CoordSpherical, uint
 {
 	float sin_phi = sinf(CoordSpherical->phig * DEG2RAD);	/* sin  (geocentric latitude) */
 
-	if (nMax <= 16 || (1 - fabs(sin_phi)) < 1.0e-10)	/* If nMax is less tha 16 or at the poles */
+	if (nMax <= 16 || (1 - fabsf(sin_phi)) < 1.0e-10f)	/* If nMax is less tha 16 or at the poles */
```
tracernz: This is on the limit for single precision right?

flight/Libraries/math/coordinate_conversions.c
```diff
@@ -221,13 +221,13 @@ uint8_t RotFrom2Vectors(const float v1b[3], const float v1e[3], const float v2b[
 
 	// The first rows of rot matrices chosen in direction of v1
 	mag = VectorMagnitude(v1b);
-	if (fabs(mag) < 1e-30)
+	if (fabsf(mag) < 1e-30f)
```
tracernz: Can we go this small with single precision?

flight/Modules/Airspeed/gps_airspeed.c
```diff
@@ -114,7 +114,7 @@ void gps_airspeedGet(float *v_air_GPS)
 	float cosDiff=(Rbe[0][0]*gps->RbeCol1_old[0]) + (Rbe[0][1]*gps->RbeCol1_old[1]) + (Rbe[0][2]*gps->RbeCol1_old[2]);
 	
 	//If there's more than a 5 degree difference between two fuselage measurements, then we have sufficient delta to continue.
-	if (fabs(cosDiff) < cos(5.0f*DEG2RAD)) {
+	if (fabsf(cosDiff) < cosf(5.0f*DEG2RAD)) {
```
tracernz: The sign is opposite of the comment above...

flight/Libraries/math/coordinate_conversions.c
mlyle: The value in either case is originally a float so there is no gain of precision by promotion to double in fabsf-- just a bunch of 0 stuffing.

https://en.wikipedia.org/wiki/Single-precision_floating-point_format#Exponent_encoding

The smallest normal value in single precision floating point (and we shouldn't be using subnormals) is ~ 10^-38.

PR #33, opened 2015-11-25
flight: remove support for original coptercontrol (not CC3D)

mlyle:


dustin:
Oh cool.  I started some work on this, but it was a bit difficult to tell them apart.  Happy to see it go.

mlyle:
Not promising it's 100%.  But yes, let's make progress on CC3D again.

pug398:
I have an original CC but it hasn't been in flight for over a year. Won't miss it.

tracernz:
Yeah, it's pretty much impossible to support it since no devs (AFAIK) have one and they're not available.

mlyle:
I've got one.  It's really nifty and a great accomplishment for the time.

But given a choice between supporting CC & CC3D and having both suck, or good CC3D support, the choice is clear IMO.

PR #4, opened 2015-10-04
makefile: eliminate silliness

mlyle:
This is mostly here to play with PR building.  Jenkins isn't configured to do notification through github/doesn't have credentials yet.  But will it try to build?

PR #36, opened 2015-11-25
Add BrainFPV (+history -branding)

fujin:
I have extracted the entire Brain history, minus any branding-specific changes, during the time that Brain was maintained out of tree and TauLabs was operational (up to the fork point, next.. some time, along Endo.)

Any branding changes can be re-made (I have them on another branch) by BrainFPV LLC for branding-specific/marketing release, would be my thought. I'm certainly happy to help.

### ~~TODO~~
- [x] Gather all branding related changes (gcs, android gcs, java uavobjects stuff.) onto another branch
- [x] Fix indentation for .ui files
- [x] Fix indentation for .xml files


README.md
```diff
@@ -1,4 +1,17 @@
-# Tau Labs
+# Tau Labs - Brain FPV Flight Controller
```
tracernz: Nope :-P
Also, I don't know if we should really keep all the brain stuff below since we don't detail other targets.

flight/Modules/Stabilization/inc/virtualflybar.h
```diff
@@ -43,4 +43,4 @@ int stabilization_virtual_flybar_pirocomp(float z_gyro, float dT);
 /**
  * @}
  * @}
- */
+ */
```
tracernz: Should add this newline back to comply with C + POSIX standards.

tracernz:
There are a few indentation issues, and unrelated things that seem to have slipped in but on the whole this looks good once those are addressed. I haven't reviewed any of the brain specific code in detail since it isn't new code and has prior test.

tracernz:
It's a pain that Github won't show all the diff so you have to drill through individual commits to add comments. :-(

dustin:
I'm generally dissatisfied with github's code review.  It's slightly better than it used to be, but I use good stuff every day. :/

mlyle:
jenkins, test this please

README.md
tgreer: Do we need to merge this if we've got a new readme.md from #4 ?

fujin: Will drop this out. Just an artefact of the rebase.

flight/Modules/Stabilization/inc/virtualflybar.h
fujin: What standard is that?

tracernz: 5.1.1.2 of ISO C 1999, see here for other reasons (like breaking various unix utils): http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

fujin: > A source file that is not empty shall end in a new-line character,
which shall not be immediately preceded by a backslash character before any such
splicing takes place.

For anyone playing along at home :)

fujin: fixed in 64b92bf

tracernz: :D

mlyle:
```pios_board.c: In function 'PIOS_Board_Init':
pios_board.c:472:2: warning: implicit declaration of function 'EventDispatcherInitialize' [-Wimplicit-function-declaration]
  EventDispatcherInitialize();
  ^
 CC          flight/targets/brain/fw/pios_usb_board_data.c
 CC          flight/Libraries/alarms.c
 CC          flight/UAVTalk/uavtalk.c
make[1]: *** No rule to make target `/var/lib/jenkins/jobs/dron/workspace@3/Nodes/master/flight/UAVObjects/eventdispatcher.c', needed by `/var/lib/jenkins/jobs/dron/workspace@3/Nodes/master/build/fw_brain/eventdispatcher.o'.  Stop.
```

needs eventdispatcher.c removed from Makefile

fujin:
@mlyle got it thx, looking

fujin:
@mlyle fixed in 6637c41, sorry about that. slipped in :)

mlyle:
Was conflict from another PR since, so, not even your fault :)

fujin:
fixed the sparkybgc conflict..

flight/PiOS/Common/printf2.c
```diff
@@ -0,0 +1,633 @@
+/*******************************************************************************
```
mluessi: This is currently used by the OSD. Compared to the other printf in, it has floating point support. However, there are some weird bugs that can lead to a crash, so we should find a better solution at some point. c.f. 

https://github.com/BrainFPV/TauLabs/commit/acf781fd9977b9f86c39b48678af700efea5ee80

mluessi:
This looks good to me. Thanks for the hard work @fujin. Does this mean d-ronin is open to have closed hw boards in-tree? Having in-tree support for Brain (and future boards) would definitely be a good reason for BrainFPV to adopt d-ronin as official firmware.

mlyle:
:) Yes we want to merge brain as soon as you give the OK.

It would be really nice if you could give us some kind of cut-down schematic --- just what you're comfortable with --- to make it easier to maintain.  I know it's not essential but some people (me) have an easier time keeping track of I/O's/assignments when looking at a picture.

mluessi:
@mlyle yes, I'm fine with merging. I will click the button then :).

I can add the schematic that shows everything except the OSD part, it will make it easier to develop for the board.

jihlein:
Brain not built as part of make package, at least for Windows build environment.

mlyle:
@jihlein thanks for the report.  package and branding is going to pretty much completely change in the next few days so it may not be worth any effort.

PR #39, opened 2015-11-25
Doxygen: Update configs, add to Makefile

tracernz:
Updated configs with doxygen -u, fixed paths


mlyle:
jenkins, test this please

Issue #46, opened 2015-11-25
F3 inverters for FrSky protocol reported to be broken


mlyle:
by PIOS_HAL

jihlein:
I have a fix for this, it adds another argument to the setup parameters, if that's okay.  Flight tested on Naze32Pro.  Can create clean pull requests if this method is acceptable.

Date: Wed, 25 Nov 2015 13:13:50 -0800

mlyle:
I can live with it for interim and would really appreciate it!

tgreer:
Can this issue be closed now #57 has been merged?

tgreer:
Closing. Whoop Whoop!

PR #57, opened 2015-11-26
F3frsky - Restore operation under PIOS_HAL

jihlein:
Code builds correctly.  Have not tested it yet.  I can try running it on Sparky, Quanton, and AQ32, but most likely not until the weekend.

Figured we could at least start looking it over for the review.


mlyle:
Fixes #46

mlyle:
This is our week of livin' dangerously.  Merge now, test later. :>

Thanks for the fix.

jihlein:
Confirmed patch works on Sparky with D4R-II on PPM, FrSky Sensor Hub selected on main port.  Don't have cables for flex port, but should be fine there too.

jihlein:
Confirmed operation on Quanton with D4R-II on PPM, FrSky Sensor Hub on Uart4 with external inverter.

PR #59, opened 2015-11-26
GCS: Trim out unused filesearch, qtconcurrent extensions

tracernz:
Extensions were only used by filesearch, which is unused.


mlyle:
Adios unused code.

PR #60, opened 2015-11-26
UAVO: Add description element to fields, add tooltips in UAVO Browser

### LINKED ISSUE #41: UAVOs should be self-documenting
There's a lot of need to use the UAVO browser but it's not clear what fields mean without looking at the code.  It would be nice to have documentation strings in the XML that python & the GCS can use. (for tooltips etc).

tracernz:
I was going to sleep but then I kept thinking about this :-P. Fixes #41.
UAVOs can now have an optional description element inside each field. This field is used in the GCS object browser to display tooltips on each UAVO field. No field descriptions are actually defined in this PR (an issue will be created once merged). To test you will need to add a description to one or more fields, e.g. in https://github.com/d-ronin/dronin/blob/next/shared/uavobjectdefinition/actuatorsettings.xml#L4:
```xml
        <field name="TimerUpdateFreq" units="Hz" type="uint16" elements="6" defaultvalue="50">
            <description>How often your spinny things get a fresh command.</description>
        </field>
```

### DIFF
--- a/ground/uavobjgenerator/generators/gcs/uavobjectgeneratorgcs.cpp
+++ b/ground/uavobjgenerator/generators/gcs/uavobjectgeneratorgcs.cpp
@@ -148,6 +148,7 @@ bool UAVObjectGeneratorGCS::process_object(ObjectInfo* info)
     QString propertySetters;
     QString propertyNotifications;
     QString propertyNotificationsImpl;
+    QString fieldDescriptionsStrings;
 
     //to avoid name conflicts
     QStringList reservedProperties;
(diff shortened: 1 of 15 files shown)


ground/uavobjgenerator/generators/gcs/uavobjectgeneratorgcs.cpp
```diff
@@ -480,3 +497,12 @@ bool UAVObjectGeneratorGCS::process_object(ObjectInfo* info)
...
     return true;
 }
+
+/**
+ * Escapes a raw string so it can be used in generated C/C++ source
+ * Whitespace will be gobbled
+ */
+QString UAVObjectGeneratorGCS::escape_raw_string(QString raw)
+{
+    return raw.replace("\\", "\\\\").replace("\"", "\\\"").simplified();
```
tracernz: The QString::simplified function replaces all whitespace chars with a single space, and gobbles consecutive whitespace into a single space.

mlyle:
I've read but not tested this.  Probably at least a couple token desc's would be nice to merge as part of this PR. 

Also need to verify that it doesn't break the python API

tracernz:
Rebased to next.

mlyle:
jenkins, test this please

mlyle:
You or me should do the python stuff as part of this PR.

tracernz:
Added python code to parse descriptions, rebased on next.

mlyle:
pending jenkins

PR #63, opened 2015-11-26
Refactor alarm stringing for reuse.

dustin:
This currently moves the MSP variant of string (and boot)
stringification into a common location.  The intention is to share the
code with BrainFPV and MSP (and anyone else who wants to string an
alarm).

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/63)

### DIFF
--- a/flight/Libraries/alarms.c
+++ b/flight/Libraries/alarms.c
@@ -238,6 +238,157 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
     return 0;
 }
 
+static const char alarm_names[][9] = {
+	[SYSTEMALARMS_ALARM_OUTOFMEMORY] = "MEMORY",
+	[SYSTEMALARMS_ALARM_CPUOVERLOAD] = "CPU",
+	[SYSTEMALARMS_ALARM_STACKOVERFLOW] = "STACK",
+	[SYSTEMALARMS_ALARM_SYSTEMCONFIGURATION] = "CONFIG",
+	[SYSTEMALARMS_ALARM_EVENTSYSTEM] = "EVENT",
+	[SYSTEMALARMS_ALARM_TELEMETRY] = {0}, // ignored
+	[SYSTEMALARMS_ALARM_MANUALCONTROL] = "MANUAL",
+	[SYSTEMALARMS_ALARM_ACTUATOR] = "ACTUATOR",
+	[SYSTEMALARMS_ALARM_ATTITUDE] = "ATTITUDE",
+	[SYSTEMALARMS_ALARM_SENSORS] = "SENSORS",
+	[SYSTEMALARMS_ALARM_STABILIZATION] = "STAB",
+	[SYSTEMALARMS_ALARM_PATHFOLLOWER] = "PATH-F",
+	[SYSTEMALARMS_ALARM_PATHPLANNER] = "PATH-P",
+	[SYSTEMALARMS_ALARM_BATTERY] = "BATTERY",
+	[SYSTEMALARMS_ALARM_FLIGHTTIME] = "TIME",
+	[SYSTEMALARMS_ALARM_I2C] = "I2C",
+	[SYSTEMALARMS_ALARM_GPS] = "GPS",
+	[SYSTEMALARMS_ALARM_ALTITUDEHOLD] = "A-HOLD",
+	[SYSTEMALARMS_ALARM_BOOTFAULT] = "BOOT",
+	[SYSTEMALARMS_ALARM_GEOFENCE] = "GEOFENCE",
+	[SYSTEMALARMS_ALARM_TEMPBARO] = "TEMPBARO",
+	[SYSTEMALARMS_ALARM_GYROBIAS] = "GYROBIAS",
+	[SYSTEMALARMS_ALARM_ADC] = "ADC",
+};
+
+// If someone adds a new alarm, we'd like it added to the array above.
+DONT_BUILD_IF(NELEMENTS(alarm_names) != SYSTEMALARMS_ALARM_NUMELEM, AlarmArrayMismatch);
+
+static const char config_error_names[][15] = {
+	[SYSTEMALARMS_CONFIGERROR_STABILIZATION] = "CFG:STAB",
+	[SYSTEMALARMS_CONFIGERROR_MULTIROTOR] = "CFG:MULTIROTOR",
+	[SYSTEMALARMS_CONFIGERROR_AUTOTUNE] = "CFG:AUTOTUNE",
+	[SYSTEMALARMS_CONFIGERROR_ALTITUDEHOLD] = "CFG:AH1",
+	[SYSTEMALARMS_CONFIGERROR_POSITIONHOLD] = "CFG:POS-HOLD",
+	[SYSTEMALARMS_CONFIGERROR_PATHPLANNER] = "CFG:PATHPLAN",
+	[SYSTEMALARMS_CONFIGERROR_DUPLICATEPORTCFG] = "CFG:DUP PORT",
+	[SYSTEMALARMS_CONFIGERROR_NAVFILTER] = "CFG:NAVFILTER",
+	[SYSTEMALARMS_CONFIGERROR_UNSAFETOARM] = "CFG:UNSAFE",
+	[SYSTEMALARMS_CONFIGERROR_UNDEFINED] = "CFG:UNDEF",
+	[SYSTEMALARMS_CONFIGERROR_NONE] = {0},
+};
+
+// DONT_BUILD_IF(NELEMENTS(CONFIG_ERROR_NAMES) != SYSTEMALARMS_CONFIGERROR_NUMELEM, AlarmArrayMismatch);
+
+static const char manual_control_names[][13] = {
+	[SYSTEMALARMS_MANUALCONTROL_SETTINGS] = "MAN:SETTINGS",
+	[SYSTEMALARMS_MANUALCONTROL_NORX] = "MAN:NO RX",
+	[SYSTEMALARMS_MANUALCONTROL_ACCESSORY] = "MAN:ACC",
+	[SYSTEMALARMS_MANUALCONTROL_ALTITUDEHOLD] = "MAN:A-HOLD",
+	[SYSTEMALARMS_MANUALCONTROL_PATHFOLLOWER] = "MAN:PATH-F",
+	[SYSTEMALARMS_MANUALCONTROL_UNDEFINED] = "MAN:UNDEF",
+	[SYSTEMALARMS_MANUALCONTROL_NONE] = {0},
+};
+
+// DONT_BUILD_IF(NELEMENTS(MANUAL_CONTROL_NAMES) != SYSTEMALARMS_MANUALCONTROL_NUMELEM, AlarmArrayMismatch);
+
+static const char boot_reason_names[][16] = {
+	[SYSTEMALARMS_REBOOTCAUSE_BROWNOUT] = "BOOT:BROWNOUT",
+	[SYSTEMALARMS_REBOOTCAUSE_PINRESET] = "BOOT:PIN RESET",
+	[SYSTEMALARMS_REBOOTCAUSE_POWERONRESET] = "BOOT:PWR ON RST",
+	[SYSTEMALARMS_REBOOTCAUSE_SOFTWARERESET] = "BOOT:SW RESET",
+	[SYSTEMALARMS_REBOOTCAUSE_INDEPENDENTWATCHDOG] "BOOT:INDY WDOG",
+	[SYSTEMALARMS_REBOOTCAUSE_WINDOWWATCHDOG] = "BOOT:WIN WDOG",
+	[SYSTEMALARMS_REBOOTCAUSE_LOWPOWER] = "BOOT:LOW POWER",
+	[SYSTEMALARMS_REBOOTCAUSE_UNDEFINED] = "BOOT:UNDEFINED",
+};
+
+#define LONGEST_MESSAGE 17
+
+DONT_BUILD_IF((LONGEST_MESSAGE <= sizeof(*config_error_names)
+	       || LONGEST_MESSAGE <= sizeof(*manual_control_names)
+	       || LONGEST_MESSAGE <= sizeof(*boot_reason_names)),
+	      InsufficientBufferage);
+
+const char *AlarmBootReason(uint8_t reason) {
+	if (reason >= NELEMENTS(boot_reason_names)) {
+		return (const char*)&boot_reason_names[SYSTEMALARMS_REBOOTCAUSE_UNDEFINED];
+	}
+	return (const char*)&boot_reason_names[reason];
+}
+
+int32_t AlarmString(SystemAlarmsData *alarm, char *buf, size_t buflen, bool blink, uint8_t *state) {
+	*state = SYSTEMALARMS_ALARM_OK;
+	buf[0] = '\0';
+	int pos = 0;
+
+	// TODO(dustin): sort the alarms by severity.  Alarm messages
+	// will get truncated.  We want the most urgent stuff to show
+	// up first, having warnings show up only if there's space.
+	for (int i = 0; i < SYSTEMALARMS_ALARM_NUMELEM; i++) {
+		if (((alarm->Alarm[i] == SYSTEMALARMS_ALARM_WARNING) ||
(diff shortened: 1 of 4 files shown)


dustin:
Just reworked this to work similarly to Brain's.  This should allow a nice uniform stringing of all the things.

STATUS:  Compiles, ship it.

(need to put it on a board at some point)

flight/Libraries/alarms.c
```diff
@@ -238,6 +238,143 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
...
+				strncpy(current_msg, (const char*)alarm_names[i], buflen);
+				buf[sizeof(*alarm_names)] = '\0';
+			}
+
+			int this_len = strlen(current_msg);
+
+			if (pos + this_len + 2 >= buflen)
+				break;
+
+			if ((alarm.Alarm[i] != SYSTEMALARMS_ALARM_WARNING) && !blink) {
```
tracernz: The blink seems like it's a little out of the scope of this function (to generate an alarm string). It would be more suited in the OSD code IMO.

dustin: I do agree.  The problem is that it needs to blink only error/critical strings *inside* the result (which may also include warnings).  I don't like that all that much.

Alternatively, this could return a collection of structured and labeled strings and let the other side manage the concatenation/blinking (since this is really Brain-specific).

Is it worth having a [x]alarmStrings that gets populated from the call instead?

tracernz: Not a deal-breaker for me. I can see why it would be done this way.

tracernz:
This looks good to me, I haven't tested it though.

flight/Libraries/alarms.c
```diff
@@ -238,6 +238,143 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
...
+	[SYSTEMALARMS_MANUALCONTROL_SETTINGS] = "MAN:SETTINGS",
+	[SYSTEMALARMS_MANUALCONTROL_NORX] = "MAN:NO RX",
+	[SYSTEMALARMS_MANUALCONTROL_ACCESSORY] = "MAN:ACC",
+	[SYSTEMALARMS_MANUALCONTROL_ALTITUDEHOLD] = "MAN:A-HOLD",
+	[SYSTEMALARMS_MANUALCONTROL_PATHFOLLOWER] = "MAN:PATH-F",
+	[SYSTEMALARMS_MANUALCONTROL_UNDEFINED] = "MAN:UNDEF",
+	[SYSTEMALARMS_MANUALCONTROL_NONE] = {0},
+};
+
+// DONT_BUILD_IF(NELEMENTS(MANUAL_CONTROL_NAMES) != SYSTEMALARMS_MANUALCONTROL_NUMELEM, AlarmArrayMismatch);
```
mlyle: Why are these commented out?  They seem necessary to have to land.

dustin: I think the ones that are commented out are the ones I wished for, but couldn't find (e.g., no `SYSTEMALARMS_MANUALCONTROL_NUMELEM`).

mlyle:
Not really expected to be related, but anomaly seen on this code on rough landing + disarm-- flight controller reset, reason "independent watchdog".  Could be this (unlikely), event system (moderately likely), or something mechanical (the USB connector on FC was loose/flapping about-- moderately likely).

dustin:
Also slightly suspicious -- I didn't see any warnings during flight.  The battery was low and the OSD was complaining (presumably overriding what the flight controller was doing), but more testing to do here.

flight/Libraries/alarms.c
```diff
@@ -238,6 +238,150 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
     return 0;
 }
 
+static const char alarm_names[][8] = {
```
mluessi: This uses 8 bytes of flash for every alarm entry. I think you could conserve some space by using the following construct:

    static const char * alarm_names[] = {
        [SYSTEMALARMS_ALARM_OUTOFMEMORY] = "MEMORY",
        [SYSTEMALARMS_ALARM_CPUOVERLOAD] = "CPU",
        ...
    }

flight/Libraries/alarms.c
```diff
@@ -238,6 +238,150 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
...
+}
+
+int32_t AlarmString(SystemAlarmsData alarm, char *buf, size_t buflen, bool blink, uint8_t *state) {
+	*state = SYSTEMALARMS_ALARM_OK;
+	buf[0] = '\0';
+	int pos = 0;
+
+	// TODO(dustin): sort the alarms by severity.  Alarm messages
+	// will get truncated.  We want the most urgent stuff to show
+	// up first, having warnings show up only if there's space.
```
mluessi: Good idea. This should definitely be done. Maybe the function could have a parameter whether we want warnings, errors, ot both.

flight/Libraries/alarms.c
```diff
@@ -238,6 +238,150 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
...
+	for (int i = 0; i < SYSTEMALARMS_ALARM_NUMELEM; i++) {
+		if (alarm_names[i][0] != 0 &&
```
mluessi: I would put the `alarm_names[i][0] != 0` inside the loop and use `continue`, so you can put a comment that this means ignored

flight/Libraries/alarms.c
```diff
@@ -238,6 +238,150 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
...
+	      InsufficientBufferage);
+
+const char *AlarmBootReason(uint8_t reason) {
+	if (reason >= NELEMENTS(boot_reason_names)) {
+		return (const char*)&boot_reason_names[SYSTEMALARMS_REBOOTCAUSE_UNDEFINED];
+	}
+	return (const char*)&boot_reason_names[reason];
+}
+
+int32_t AlarmString(SystemAlarmsData alarm, char *buf, size_t buflen, bool blink, uint8_t *state) {
```
mluessi: Add some documentation about parameters (e.g. `state` is non-obvious)

flight/Libraries/alarms.c
mluessi: See comment above. If the function returned either errors, warnings, or both. The OSD could handle to display them however it pleases (e.g. blink errors, don't blink warnings)

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -700,83 +701,30 @@ void draw_flight_mode(int x, int y, int xs, int ys, int va, int ha, int flags, i
...
-	for (uint8_t pos = 0; pos < sizeof(ALL_ALRARMS); pos++)
-	{
-		if ((alarm.Alarm[ALL_ALRARMS[pos]] == SYSTEMALARMS_ALARM_WARNING) ||
-			(alarm.Alarm[ALL_ALRARMS[pos]] == SYSTEMALARMS_ALARM_ERROR) ||
-			(alarm.Alarm[ALL_ALRARMS[pos]] == SYSTEMALARMS_ALARM_CRITICAL)){
-			this_len = strlen(ALL_ALRARM_NAMES[pos]);
-			if (str_pos + this_len + 2 >= sizeof(temp))
-				break;
+	// Boot alarm for a bit.
+	if (PIOS_Thread_Systime() < BOOT_DISPLAY_TIME_MS) {
```
mluessi: It may be better to display this in the welcome screen (?)

flight/Libraries/alarms.c
mlyle: That'll use 4 bytes of flash per entry to save at most 7 bytes of flash per entry.  Not sure it's a win.  Of course, then the null would always be there.

tracernz: Having the null always there is a win.

mlyle: Yah, I really think we should allocate it [9] and remove the magic to cope with the null not being there below.  Costs a few bytes of flash up here, but saves a few bytes of program code below.

dustin:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[flight/Libraries/alarms.c, line 241 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4jH9lTSkNHGz5i17Rp-r2-241:-K4jIPo8UhbOE1qKdPa0:-43986286) ([raw file](https://github.com/d-ronin/dronin/blob/465fe1204d77c24a99b3509bdb2b83db20d70a99/flight/Libraries/alarms.c#L241)):*
The downside of making it [9] is that I don't know how to statically verify that adds a NULL at the end in all cases.  Any ideas?

---

*[flight/Modules/OnScreenDisplay/onscreendisplay.c, line 713 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4jH9lURmT9lDFB-zLa-r2-713:-K4jIgrWMMUqLECtSEXx:1761503941) ([raw file](https://github.com/d-ronin/dronin/blob/465fe1204d77c24a99b3509bdb2b83db20d70a99/flight/Modules/OnScreenDisplay/onscreendisplay.c#L713)):*
I did it in several places on MWOSD -- don't want it to get lost.  Doing the welcome screen separately is a good idea.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63)*

mlyle:
Reviewed 3 of 3 files at r1, 1 of 1 files at r2.

---

*[flight/Libraries/alarms.c, line 241 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4jH9lTSkNHGz5i17Rp-r2-241:-K4jS1NVXho0IU5eY_08:1495410534) ([raw file](https://github.com/d-ronin/dronin/blob/465fe1204d77c24a99b3509bdb2b83db20d70a99/flight/Libraries/alarms.c#L241)):*
You can't, but it'll at least be safe / not crash (because it won't overrun the buffer).  It will just look bad if the null is missing.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63)*

dustin:
I like your idea and think we can do something better.  As mentioned in another comment, I would like to do this a bit more incrementally since this fits the current code a little better.

---

---

*[flight/Libraries/alarms.c, line 241 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4jH9lTSkNHGz5i17Rp-r2-241:-K4srsAj6m1Vn5KeicLF:381820083) ([raw file](https://github.com/d-ronin/dronin/blob/465fe1204d77c24a99b3509bdb2b83db20d70a99/flight/Libraries/alarms.c#L241)):*
I decreased all of them until they failed to compile and then added two.  There's probably a better way to confirm they're correct.  Unit tests would be pretty nice here.

---

*[flight/Libraries/alarms.c, line 323 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4jH9lTSkNHGz5i17Rp-r2-323:-K4szqFqIbkVMKLdoXSd:-499146009) ([raw file](https://github.com/d-ronin/dronin/blob/465fe1204d77c24a99b3509bdb2b83db20d70a99/flight/Libraries/alarms.c#L323)):*
Done.

---

*[flight/Libraries/alarms.c, line 330 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4jH9lTSkNHGz5i17Rp-r2-330:-K4szycK4f4fD2-10IIG:1927715575) ([raw file](https://github.com/d-ronin/dronin/blob/465fe1204d77c24a99b3509bdb2b83db20d70a99/flight/Libraries/alarms.c#L330)):*
I want to get a first pass of some work here and decide how nice it is.  I think if we're going to be able to ask for particular levels of severity, we should also be able to display them separately (e.g. I want critical right under the crosshair on Brain, but warnings can be in the bottom left).

---

*[flight/Libraries/alarms.c, line 332 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4jH9lTSkNHGz5i17Rp-r2-332:-K4svLusTPxiIgipJblH:-499146009) ([raw file](https://github.com/d-ronin/dronin/blob/465fe1204d77c24a99b3509bdb2b83db20d70a99/flight/Libraries/alarms.c#L332)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63#-:-K4t04-gqAKFkTUpZL1u:1205724194)*

dustin:
This seems to be working well on Sparky/MWOSD.  I'm going to give it a shot.

-status/work-in-progress 
-status/requires-test 

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63#-:-K4tO0zE-lFh_qtMaR_V:293621540)*

mlyle:
Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[flight/Libraries/alarms.c, line 253 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4tPCSJYKiYvuW1fREH:-K4tPCSJYKiYvuW1fREI:-1116581652) ([raw file](https://github.com/d-ronin/dronin/blob/608c8d1da12366c89a24a631365d9c7262aa8d4b/flight/Libraries/alarms.c#L253)):*
Suggest P-FOLLOW, P-PLAN.. Totally at your discretion.

---

*[flight/Libraries/inc/alarms.h, line 53 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4tPjEMfqxCuwZoalC8:-K4tPjEMfqxCuwZoalC9:-1036098683) ([raw file](https://github.com/d-ronin/dronin/blob/608c8d1da12366c89a24a631365d9c7262aa8d4b/flight/Libraries/inc/alarms.h#L53)):*
To date the practice has been to document the code, not the headers.  I actually prefer the reverse (as you've done).  We should discuss convention.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63)*

dustin:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[flight/Libraries/alarms.c, line 253 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4tPCSJYKiYvuW1fREH:-K4trdsQHp8qBBxfoWEP:-499146009) ([raw file](https://github.com/d-ronin/dronin/blob/608c8d1da12366c89a24a631365d9c7262aa8d4b/flight/Libraries/alarms.c#L253)):*
Done.

---

*[flight/Libraries/inc/alarms.h, line 53 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4tPjEMfqxCuwZoalC8:-K4tZhOygR1M1ked4H8I:680075197) ([raw file](https://github.com/d-ronin/dronin/blob/608c8d1da12366c89a24a631365d9c7262aa8d4b/flight/Libraries/inc/alarms.h#L53)):*
Should I move this stuff?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63)*

mlyle:

---

*[flight/Libraries/inc/alarms.h, line 53 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4tPjEMfqxCuwZoalC8:-K4tt2C3wj73JYFmb7LY:677872702) ([raw file](https://github.com/d-ronin/dronin/blob/608c8d1da12366c89a24a631365d9c7262aa8d4b/flight/Libraries/inc/alarms.h#L53)):*
nah.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63)*

dustin:
+status/ready-for-merge 

---

---

*[flight/Libraries/inc/alarms.h, line 53 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/63#-K4tPjEMfqxCuwZoalC8:-K4twy46x6bmUlG9kL0d:1910388516) ([raw file](https://github.com/d-ronin/dronin/blob/608c8d1da12366c89a24a631365d9c7262aa8d4b/flight/Libraries/inc/alarms.h#L53)):*
OK, then this is what I've been flying...

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/63#-:-K4tx3OijOXJN-d9gLVc:1654777037)*

PR #89, opened 2015-11-29
ground: hitl: remove support for aerosim and il2

### LINKED ISSUE #71: Remove support for AerosimRC?
Doubt this is used anymore, along with HITL in general.  Would be nice to get HITL down to a short list of targets but to actually make it work on them.

mlyle:
Leave in x-plane and flightgear as these we are likely to use going
forward.

Fixes #71.


mlyle:
Not tested beyond OSX build and GCS launch.

PR #95, opened 2015-11-30
flight: disable stack checking on targets that were missed

mlyle:
Not sure why FlyingF3 was missed; Brain was missed because it was not-yet-in-tree.


fujin:
Looks fine, I think I had this in a brain branch but missed it from the big one. Sorry :(

tracernz:
No prob, review didn't pick it up either. ;)

mlyle:
This is probably the worst thing with duplicating target code-- if there's any kind of skew with outside forks we lose this way.

PR #96, opened 2015-11-30
GCS:ConnectionMgr: Messagebox on connect failed

tracernz:
If the attempted connection fails, we fail silently except for a message to the debug log, not super friendly to users who will see that nothing happened when they clicked connect. This adds a popup message box.


mlyle:
Works.  Boy that red FAILED is angry :)

PR #100, opened 2015-11-30
autotune: buffering and performance improvements

mlyle:
Significantly improves autotune performance and reliability.  

With this change, autotune is able to run without dropping points on F1 (though DSM makes it drop points).  Also, it looks at every point coming from the sensors, so it is more deterministic and reliable and not subject to jitter.

Fixes #100


mlyle:
Todo:

* [ ] circqueue: Add doxygen strings
* [ ] circqueue: Improve unit test (empty/full scenarios on "edge of buffer"
* [ ] F1: investigate free heap w/ autotune enabled for buffer sizing
* [ ] autotune: register for synchronous callbacks
* [ ] autotune: store synchronous values for gyros, actuator and time to circqueue
* [ ] autotune: get points from circular queue

dustin:
OMG.  Is this done?!

mlyle:
Run on both F1 and F4 hardware on bench.

Pros: it keeps up @ 333Hz on F1.  19992 points processed, 0 points spilled from queue, 20000 points expected.  (And there are race conditions at beginning/end so getting exactly 20k is not expected).

Cons: there is some kind of weird jitter on the "period" variable.  On F4, seen to toggle between 1-3.5ms when 1.5 is expected at 666Hz rate.  on F1, seems reasonable and then jumps to 50000 (!).    So something is sick.  Hopefully not circqueue.

Also, 73025a9 is an important fix not just for this PR.

mlyle:
http://forum.taulabs.org/viewtopic.php?f=27&t=268&start=130#p8480

mlyle:
Some points still spilled on the tune, probably because of overhead of DSM.

mlyle:
Flew 3 autotune flights @1KHz on Sparky2/F4.  Got consistent results.

tracernz:
Tunes on Naze32 (2x) and Sparky (1x): http://forum.taulabs.org/viewtopic.php?p=8487#p8487

mluessi:
@tracernz surprising that your second tune on Naze had a higher tau. I would expect it to be lower with a fresh battery.

flight/Libraries/circqueue.c
```diff
@@ -0,0 +1,149 @@
...
+	 */
+
+	/* This is declared as a uint32_t for alignment reasons. */
+	uint32_t contents[];		/**< Contents of the circula queue */
+};
+
+/** Allocate a new circular queue.
+ * @param[in] elem_size The size of each element, as obtained from sizeof().
+ * @param[in] num_elem The number of elements in the queue.  The capacity is
+ * one less than this (it may not be completely filled.
```
mluessi: missing `)`

flight/Modules/Autotune/autotune.c
```diff
@@ -163,6 +228,8 @@ static void UpdateStabilizationDesired(bool doingIdent) {
 	StabilizationDesiredSet(&stabDesired);
 }
 
+#define MAX_PTS_PER_CYCLE 4
```
mluessi: Ideally, this should depend on the gyro rate, i.e., the module can currently handle 2kHz, but wouldn't work correctly at 3kHz. Correct? Maybe it would make sense for the gyro driver to export the gyro rate used.

mluessi:
Very nice :). I like the circular buffer approach.

flight/Modules/Autotune/autotune.c
mlyle: It inhibits the sleep at the bottom if it doesn't complete the loop, so it'll work at faster rates.  It's there so we spend an appropriate amount of effort looking at manualcontrolcommand since we're in the control critical path.

mlyle: One thing that would be nice, that I didn't do.. would be to quantize the dTs to multiples of the sensor interval.  It would cut down noise.

mlyle:
Thanks.  :) updated from feedback and found a couple other typo-ish things.

dustin:
":100:" -- @fujin

mlyle:
ffs i quit ;(

flight/Modules/Autotune/autotune.c
```diff
@@ -207,25 +273,21 @@ static void AutotuneTask(void *parameters)
...
-		ManualControlCommandThrottleGet(&throttle);
-				
 		switch(state) {
 			case AT_INIT:
-
 				last_update_time = PIOS_Thread_Systime();
 
 				// Only start when armed and flying
-				if (flightStatus.Armed == FLIGHTSTATUS_ARMED_ARMED && throttle > 0) {
+				if (flightStatus.Armed == FLIGHTSTATUS_ARMED_ARMED) {
```
fujin: You removed the throttle>0 from here which looks like prevented the system from gathering points while armed and not flying, was that on purpose? If so, are there any downsides? Were there little stray points in general?

shared/uavobjectdefinition/hwnaze.xml
```diff
@@ -11,7 +11,7 @@
 
 		<field name="GyroRange" units="deg/s" type="enum" elements="1" options="250,500,1000,2000" defaultvalue="1000"/>
 		<field name="AccelRange" units="*gravity m/s^2" type="enum" elements="1" options="2G,4G,8G,16G" defaultvalue="8G"/>
-		<field name="MPU6050Rate" units="" type="enum" elements="1" options="200,333,500" defaultvalue="500"/>
+		<field name="MPU6050Rate" units="" type="enum" elements="1" options="200,333,500" defaultvalue="333"/>
```
fujin: Does it work at 500Hz?

fujin:
Looks awesome, I haven't had a chance to run any Brain tunes with it yet. :100: from me though presumably others have by now.

shared/uavobjectdefinition/hwnaze.xml
mlyle: The board barely does-- cpu warning.  And you definitely can't autotune @ 500Hz.

flight/Modules/Autotune/autotune.c
mlyle: The logic that was here was somewhat costly and in places incorrect.  If you were armed and in the middle of autotune (later in the code) it'd discard points when you have throttle=0 in the middle of the tune.... which you really shouldn't have and would render the tune completely untrustworthy.

Basically, what this block would do, is if you were armed and toggled to autotune when throttle is down, start the autotune 2 seconds after throttle up instead of 2 seconds after toggling to the mode.  IMO this is weird.

tracernz:
@mluessi I'm pretty sure I applied the first tune before doing the second tune on the Naze, and I think that's why the Tau changed. It seems dependent on the controller gains in use which means it's not really Tau, but "Tau". :-P

flight/UAVObjects/uavobjectmanager.c
```diff
@@ -1827,6 +1827,9 @@ static int32_t sendEvent(struct UAVOBase * obj, uint16_t instId,
 		.instId = instId
 	};
 
+	pending_events[num_pending].obj_data = obj_data;
+	pending_events[num_pending].len = len;
+
```
tracernz: Nice. :D

mlyle: Yah that's what bricked my little brushed board. :P

tracernz:
I like it. Has been flight tested and shown to allow F1 autotune so that's a massive win in my books.

PR #81, opened 2015-11-28
flight: brain: add -Werror to compile flags

mlyle:
exposed by #78 ; need #80 to successfully build with this.

PR #80, opened 2015-11-28
OSD: Fix UAVO event callback usage

tracernz:
I haven't tested this because I don't yet have a Brain but this fixes the build.

Fixes #78

PR #102, opened 2015-12-01
OSD: Fix UAVO callback usage

mluessi:
Use new callback system for OSD. After this, #81 should be good to merge.


mluessi:
Never mind.. #80 already does the same thing. LOL

PR #106, opened 2015-12-01
gcs: module configuration: remove duplicate txpid

mlyle:
txpid shows up on the module tab twice.  It looks like one was intended
for the on-screen display, but that currently doesn't have a module
flag.


mluessi:
Thanks!

PR #110, opened 2015-12-02
flight: alterations to gyro settings.

mlyle:
1) 444Hz never makes sense on Sparky, so remove it.  (it only works with
the DLPF disabled in which case you're guaranteed aliasing).
2) Set all 666Hz/DLPF disabled targets to 500Hz/188Hz by default--
except colibri.  We know enough about the platform being flown there to
know those defaults work out OK.
3) Change targets with gyrorange of 500 deg/s default to 1000deg/s.


tracernz:
Like. :+1:

mluessi:
LGTM

mluessi:
Merging

Issue #38, opened 2015-11-25
Structured app for autotune results


dustin:
Instead of a long free-form forum thread with a bunch of text and hard to link to things that require login to even look at, let's create a simple web app (I like GAE) to which we can post autotune results, then index them by various fields, aggregate them, query, look for anomalies, alert on stuff, etc...

SystemsGuy:
Database would be cool - "show me all 250's with Cobra 2204/2300's" ;-)

tgreer:
This sounds good.

tracernz:
Are you happy to handle the app end of it @dustin?

dustin:
Yeah, I can make something simple to get started.

VArcht:
Where are we going to store these? Probably be some other uses for persistent data for other parts of the website too.

dustin:
I plan to make a gae app and use datastore.  Export is important as well.
Last time I did something like this, I had it feeding data out to a public
couchdb that anyone could subscribe to.

VArcht:
an attempt at the json https://gist.github.com/VArcht/25cc3de6f0d7cf695a58

tracernz:
I already had the JSON code mostly written as per IRC :-P. This is what I've got, combining a few of the fields from your format. I have removed the user entered aircraft description string since we already cover the aircraft characteristics in other fields.
```json
{
    "dataVersion": 1,
    "identification": {
        "pitch": {
            "bias": -0.041178399999999997,
            "gain": 9.0118799999999997,
            "noise": 139.03700000000001
        },
        "roll": {
            "bias": 0.0073014200000000003,
            "gain": 9.3093800000000009,
            "noise": 151.97
        },
        "tau": 0.028000000000000001
    },
    "tuning": {
        "computed": {
            "derivativeCutoff": 18,
            "gains": {
                "outer": {
                    "kp": 5.3632600000000004
                },
                "pitch": {
                    "kd": 6.4892699999999996e-05,
                    "ki": 0.0096096500000000008,
                    "kp": 0.0032906699999999999
                },
                "roll": {
                    "kd": 4.8193899999999999e-05,
                    "ki": 0.0071367999999999996,
                    "kp": 0.0024438799999999998
                }
            },
            "naturalFrequency": 5.7999999999999998
        },
        "parameters": {
            "damping": 1.1000000000000001,
            "noiseSensitivity": 0
        }
    },
    "uniqueId": "0x2f0037000157325237313620",
[...]

Issue #12, opened 2015-10-06
There's code to integrate with the TL forum that should point at something of ours


mlyle:
* [x] Autotune sharing
* [x] Forum panel on welcome screen
* [x] Crash dumps

fujin:
Autotune, notably

VArcht:
We could use my gist idea.

Have it post anonymous gists with a thumbprint that we can find back.

Then we can have a web page to search them by size, motor, whatever .

Can have a tau hall of fame, some peeps might like that... :)

tracernz:
That would be quite cool actually. Gist for crash dumps would also be good.

mlyle:
I never thought about gist for autotune.  That sounds nice.

VArcht:
Yeah. crash dumps is where I first thought to use them but guil was afraid people would post fake ones to mess with him.

bonus, could replace this stuff with something sane like json https://github.com/d-ronin/dronin/blob/next/ground/gcs/src/plugins/config/configautotunewidget.cpp#L124

mlyle:
While we're at it-- we should add pulldowns for vehicle size and how well it flew with the autotuned settings.. so we can search for good tunes for "200-300 size quads".

VArcht:
yeah, nice to have known enums

dustin:
#38 should hopefully be quick and easy.  We can own and query the data.  Maybe make a usable interface (I can make a non-usable interface to start).

tracernz:
Nothing pointing at Tau Labs anymore.

PR #119, opened 2015-12-03
GCS: Replace autotune forum share

tracernz:
Now allows the user to copy plain text to clipboard or push to the new database @dustin created. Previously entered data is saved for next time, board and vehicle type are automatic and not able to be overridden if a board is connected, otherwise they're editable boxes. Doesn't yet save the previously entered data per board (using the board UUID).
Does the GCS side of #38.
screenshot from 2015-12-03 02 16 55
screenshot from 2015-12-03 02 19 45

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/119)


tracernz:
Also fixes part of #12.

tracernz:
I'm considering add a confirmation dialog before sending to the database informing the user that uniquely identifying data from their board is sent (the STM32 unique device ID). Thoughts?

dustin:
I guess it's the right thing to do.  It's not exactly PII in that we can't
do too much correlation there.

mlyle:
we probably should uniquely identify their GCS configuration too.  I don't know we need a separate dialog, the thing with the form could note that the ID will be sent.

fujin:
I tested this today (required #127 to build gcs, though.) submitted my results to @dustin's database thing as well as copied to pastebin https://www.irccloud.com/pastebin/I9FDpn80/

the tune looks really good, the new ui stuff is nice as well :) :+1: :100: from me.

ground/gcs/src/plugins/coreplugin/generalsettings.cpp
```diff
@@ -189,10 +189,14 @@ void GeneralSettings::readSettings(QSettings* qs)
...
     m_proxyUser = qs->value(QLatin1String("proxyuser"),m_proxyUser).toString();
     m_proxyPassword = qs->value(QLatin1String("proxypassword"),m_proxyPassword).toString();
-    m_forumUser = qs->value(QLatin1String("forumuser"),m_forumUser).toString();
-    m_forumPassword = qs->value(QLatin1String("forumpassword"),m_forumPassword).toString();
-    m_aircraft = qs->value(QLatin1String("aircraft"),m_aircraft).toString();
-    m_observations = qs->value(QLatin1String("observations"),m_observations).toString();
+    m_observations = qs->value(QLatin1String("observations"), "").toString();
+    m_vehicle = qs->value(QLatin1String("vehicle"), "").toString();
+    m_board = qs->value(QLatin1String("board"), "").toString();
+    m_weight = qs->value(QLatin1String("weight"), 0).toInt();
```
fujin: I have a feeling I entered '225mtm' in here and it turned it into '0', if that's intended.. might be better to string, then parse/atoi/trim..

fujin: Yea.. if you enter something like `225mtm` here, it gets turned into `0`, just confirmed.



```
Flight controller:	Brain
Firmware tag:		next_fujin
Firmware commit:	dc4d0a8
Firmware date:		20151203 07:23 UTC

Aircraft description:
Type:			QuadX
Weight (AUW):	0 g
Size:			0
Battery:		1S
Motors:			
ESCs:			

Observations:


Measured properties:
		Gain		Bias		Noise
Roll:	10.5324		-0.00423568	24.5134
Pitch:	10.2604		0.00246422	20.4178
Tau:	14.7 ms

Tuning aggressiveness:
Damping:			1.1
Noise sensitivity:	1 %
Natural frequency:	13.5

Computed values:
			RateKp		RateKi		RateKd
Roll:		0.00197253	0.0135056	2.13666e-05
Pitch:		0.00258912	0.0177273	2.80456e-05
Outer Kp:	12.592		-			-
Derivative cutoff:	46
```

fujin:
Can we get a field for prop size/pitch/number of blades? or just a string field? @dustin can you add something on your end as well? :) :heart:

dustin:
I'm storing it in the raw right now.  I can pull stuff out and parse or
whatever.  Lots of ways for people to get that wrong, though.

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -85,127 +117,128 @@ void ConfigAutotuneWidget::saveStabilization()
...
+    request.setHeader(QNetworkRequest::ContentTypeHeader, "application/json; charset=utf-8");
+    QNetworkAccessManager *manager = new QNetworkAccessManager(this);
+
+    connect(manager, SIGNAL(finished(QNetworkReply*)),
+                this, SLOT(onShareToDatabaseComplete(QNetworkReply*)));
+
+    manager->post(request, json.toJson());
+}
+
+void ConfigAutotuneWidget::onShareToDatabaseComplete(QNetworkReply *reply)
```
PTDreamer: Memory leak here, reply never gets deleted

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -85,127 +117,128 @@ void ConfigAutotuneWidget::saveStabilization()
...
-    ExtensionSystem::PluginManager *pm=ExtensionSystem::PluginManager::instance();
-    Core::Internal::GeneralSettings * settings=pm->getObject<Core::Internal::GeneralSettings>();
-    if (forumInteractionForm->getSaveCredentials()) {
-        settings->setForumPassword(forumInteractionForm->getPassword());
-        settings->setForumUser(forumInteractionForm->getUserName());
-    } else {
-        settings->setForumPassword("");
-        settings->setForumUser("");
+    Q_UNUSED(value);
+    autotuneShareForm->deleteLater();
```
PTDreamer: This will probably lead to a segfault on exit. Here you say you want to delete the form as soon as the flow returns to the main eventloop. But you also set "this" as parent when you create the form instance making it a child of "this". Since "this" inherits QObject it will call all of its children destructors from its own destructor so you will try to delete something which was already deleted.

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -11,16 +38,21 @@
...
 #include "modulesettings.h"
+#include "systemsettings.h"
 #include "coreplugin/generalsettings.h"
-#include "utils/phpbb.h"
 #include "extensionsystem/pluginmanager.h"
-#include <QMessageBox>
+#include "coreplugin/iboardtype.h"
+
+
+const QString ConfigAutotuneWidget::databaseUrl = QString("http://dronin-autotown.appspot.com/storeTune");
```
PTDreamer: Can this be fetched from a central location so it can be easier to maintain sync across projects?

ground/gcs/src/plugins/config/autotuneshareform.cpp
```diff
@@ -0,0 +1,178 @@
+/**
+ ******************************************************************************
+ * @file       autotuneshareform.h
+ * @author     dRonin, http://dronin.org, Copyright (C) 2015
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2014
```
PTDreamer: This looks like a new file, no need to give credits

ground/gcs/src/plugins/config/config.pro
```diff
@@ -47,7 +48,8 @@ HEADERS += calibration.h \
     configmodulewidget.h \
     configosdwidget.h \
     convertmwrate.h \
-    expocurve.h
+    expocurve.h \
+    autotuneshareform.h
```
PTDreamer: Config is getting into a massive library with no separation between logic and UI which is a shame.

ground/gcs/src/plugins/config/configautotunewidget.cpp
tracernz: Thanks, fixed, will push in a bit. :)

ground/gcs/src/plugins/config/configautotunewidget.cpp
tracernz: Yeah, that was silly. Thanks for catching it.

ground/gcs/src/plugins/config/configautotunewidget.cpp
tracernz: Could be, I'll have a think on it. I might move the JSON stuff to a class in utils so it can be shared with crash reporter like the phpBB stuff was.

ground/gcs/src/plugins/config/autotuneshareform.cpp
tracernz: It's actually an old file moved because I think it's more appropriate here. Perhaps git lost the ref during rebase. :-(

ground/gcs/src/plugins/config/config.pro
tracernz: Yeah. Perhaps at some point refactoring each of the sections into separate libs could be a good idea?

ground/gcs/src/plugins/config/configautotunewidget.cpp
PTDreamer: I was actually referring to the url string only. I think the online DB is a good approach which can be followed by others.

tracernz:
Okay, I've done some git surgery and left the phpBB removal out of this PR as the crashreporter depends on it. I'll do the crashreporter stuff in another PR since the server side is not yet working.

ground/gcs/src/plugins/config/configautotunewidget.cpp
tracernz: Thinking about this... I actually want to not set the parent and call deleteLater on it here right? Otherwise pressing the button again will create another instance and they won't be deleted until the main autotune config widget goes away.

ground/gcs/src/plugins/config/configautotunewidget.cpp
tracernz: Yeah, I was thinking in addition to moving the URL I'd like to wrap some JSON request functionality into a util class. ;-)

ground/gcs/src/plugins/config/configautotunewidget.cpp
PTDreamer: You can also create a single instance of the form on the contructor and only show/close when needed.
One other way is to use the Qt::WA_DeleteOnClose attribute on the form dialog.

mluessi:
Very nice. Is the plan to save the vehicle information in the GCS? It would be nice if the user doesn't have to re-enter the data on repeated tunes.

tracernz:
@mluessi Currently the data is all saved in some fields in general settings, so the user only needs to change the vehicle data when he/she changes to another vehicle. What I would like to do in future is save the vehicle data per-vehicle attached to the STM32 UUID so the user can have multiple different vehicles and still not need to re-enter data.

mluessi:
@tracernz using the UUID for this is a good idea. We should make sure that the UUID doesn't get shared publicly without the users permission. A manufacturer may collect this information, so it would be possible to associate a person with the UUID.

VArcht:
It would be preferable for analysis if we had drop downs for esc and motor. Would add some maintenance load but can be minimized by having some catch alls like 
 generic simonk, generic blHeli, generic other. Similar for motors.

tracernz:
I'd be fine if it was just "SimonK, BLHeli, Kiss, Other", but I don't want to go any more specific than that for hardcoded stuff. For motors what do we think is relevant, kv? I'd be fine with making it an int field, but again, I don't want to make a hardcoded list.

DTFUHF:
Instead of a dropdown box, multiple fields for motor size, manufacturer, kV, purchase date, etc. would encourage the entry of useful feedback data.

tracernz:
I'm also unhappy with the number of fields already there, keep it simple and all that. :/

VArcht:
there are only a handfull of esc's and motors that "matter currently :)

tracernz:
Currently being the keyword. ;-) There's already more than enough maintenance required in GCS for a team several times the size of ours.

mlyle:
Uploaded one.  I like it.  I am concerned the UI isn't super intuitive, though.

PR #125, opened 2015-12-03
GCS: Change release tags to Release-yyyymmdd

tracernz:
~~Untested.~~

### DIFF
--- a/ground/gcs/src/plugins/uavobjectutil/uavobjectutilmanager.cpp
+++ b/ground/gcs/src/plugins/uavobjectutil/uavobjectutilmanager.cpp
@@ -705,8 +705,9 @@ bool UAVObjectUtilManager::descriptionToStructure(QByteArray desc, deviceDescrip
        struc.uavoHash.clear();
        struc.uavoHash=desc.mid(60,20);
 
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/uavobjectutil/uavobjectutilmanager.cpp
```diff
@@ -705,8 +705,8 @@ bool UAVObjectUtilManager::descriptionToStructure(QByteArray desc, deviceDescrip
        struc.uavoHash.clear();
        struc.uavoHash=desc.mid(60,20);
 
-       // The git tag format used for releases is yyyymmdd
-       QRegularExpression certifiedPattern("^20[0-9]{6}$");
+       // The git tag format used for releases is Release-yyyymmdd
+       QRegularExpression certifiedPattern("^Release-20[0-9]{6}$");
```
mlyle: can we add `(\.[0-9])?`

That way I can fail to do a release/tag properly and try again on the same day :)

tracernz:
Change to accept the optional .[0-9] suffix, tested that it works with and without the suffix and rejects other tags.

PR #127, opened 2015-12-03
Add an explicit uint bitsWithZoom

fujin:
* Checked against sizeof(some ridiculous shit) which is uint.

### DIFF
--- a/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp
+++ b/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp
@@ -379,9 +379,9 @@ namespace internals {
             while( (1 << (numBits-1)) < projection->TileSize().Width()){
                 numBits++;
             }
-			
+            unsigned int bitsWithZoom = numBits + (quint8)maxzoom;
             //Ensure that no matter what the zoom can never exceed the number of bits required to display it
-            if (numBits + (quint8)maxzoom > sizeof(((core::Point *) 0)->X())*8 - 1){ //Remove one because of the sign bit.
+            if (bitsWithZoom > sizeof(((core::Point *) 0)->X())*8 - 1){ //Remove one because of the sign bit.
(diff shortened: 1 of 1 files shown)


fujin:
@tracernz boop?

fujin:
verified this works ok..

mlyle:
Does it work instead to put 'u' after the 8 and the 1?

e.g.

` if (numBits + (quint8)maxzoom > sizeof(((core::Point *) 0)->X())*8u - 1u){ `

fujin:
@mlyle still seems the same with that.
```
compiling /home/aj/uav/dronin/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp
/home/aj/uav/dronin/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp: In member function ‘void internals::Core::SetMapType(const core::MapType::Types&)’:
/home/aj/uav/dronin/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp:383:43: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
             if (numBits + (quint8)maxzoom > sizeof(((core::Point *) 0)->X())*8u - 1){ //Remove one because of the sign bit.
```

fujin:
missed the one after the '1'

fujin:
adding the 'u' suffix to the 8/1 doesn't seem to do it.
```
compiling /home/aj/uav/dronin/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp
/home/aj/uav/dronin/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp: In member function ‘void internals::Core::SetMapType(const core::MapType::Types&)’:
/home/aj/uav/dronin/ground/gcs/src/libs/tlmapcontrol/internals/core.cpp:383:43: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
             if (numBits + (quint8)maxzoom > sizeof(((core::Point *) 0)->X())*8u - 1u){ //Remove one because of the sign bit.
                                           ^
```

PR #130, opened 2015-12-04
GCS:CrashReporter: Use JSON service instead of forum

tracernz:
@dustin has completed the other end so this can be pushed now. I'd like to write JSON services class in utils at some point so the code can be shared with autotune and perhaps some new services in future.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/130)


tracernz:
So the crashreporterapp works fine on OSX, it just never gets called by the crash handler. The file I sent as a dummy dump came out looking proper at the other end too @dustin. :+1:

PTDreamer:
Is this being testes on a DMG or APP Bundle?

tracernz:
App bundle in the build directory.

PTDreamer:
Is the crash handler binary on the same dir?

tracernz:
crashreporter.app and dronin GCS.app are both in build/ground/gcs/bin.

tracernz:
Ahh, there's also the crashreporterapp binary in dronin GCS.app/Content/MacOS

PTDreamer:
Yeh the .app shouldn't be there, no idea why it is, can't reproduce it.
The one that should get called is the ...GCS.app/Contents/MacOS/crashreporterapp does it work if you manually run it?

tracernz:
Yes, it runs fine if I pass a filename as an argument, and submits the file successfully.

PTDreamer:
Sorry I can't help more since I can't reproduce.
Debugging should be simple just check what is being called on the LaunchUploader function libcrashreporter-qt/libcrashreporter-handler/Handler.cpp. The new process gets spawned on the execl call.

tracernz:
Yep, I've only just looked at it while testing this as I don't usually run GCS on OSX. I'll look into it more later today. Will let you know if I get to the bottom of it.

mlyle:
This has accumulated conflict.

tracernz:
Rebased and removed phpBB stuff, since the autotune PR has already merged nothing else uses phpBB anymore.

mlyle:
WORKSFORME

Issue #61, opened 2015-11-26
Add ability to only import non-default settings


mluessi:
Currently, when UAVO settings are exported and later imported, they overwrite all the settings. This leads to problems, e.g. the VTOL path follower defaults have been tweaked to work better, but users who import settings during an upgrade won't benefit from the new defaults.

I think it would be good to have the ability to only import settings that are different from the default. One way to do this is to indicate in the exported xml file which settings are different from the default. When importing, the user could then select to either import all settings or just the the ones that are different from default (the latter should be the default choice).

kubark42:
+1

It would be even cooler if default changes could be accompanied by a little text that helpsthe user understand if they potentially want to stick with the old default.

mlyle:
I think this shouldn't be done.  There are settings changes where moving a default in combination with user's settings that they did change that could be dangerous.

tracernz:
My thinking was a table with a 2-way checkbox for each thing that will/might need to change, options are "keep old value", "use new default".

mlyle:
I think I like our current practice-- changing the name of things that we should pick up new defaults for; otherwise not picking up new defaults for existing setups.

We could do something smart with a lot more UI work and explanation.. because it would be really unclear this time around, for example, whether people should pick up my derivGamma change or not.

PR #132, opened 2015-12-04
GCS:Config: Ignore lack of StateEstimation object

### LINKED ISSUE #131: Usability: warning saving sensor settings on Naze32
This threw me until I realized what I think is happening.  After doing a 6 point cal and hitting apply with a Naze32 I get a yellow warning.  Eventually I got the tooltip to stick around long enough to read it and saw that it was complaining about the "StateEstimation" UAVO not being present on the hardware.

Which had me scratching my head until I realized that that's on Filter Settings which is another tab on the same screen.  So I'm guessing that apply/save handles all tabs at once.  Unless this is a bug on the Sensor Settings tab.

For someone just coming from *flight this could be scary to do the configs and get a somewhat cryptic warning like that.  If this isn't a bug and is just the way GCS works then we should probably document it as part of #123 that the warning can safely be ignored on Naze32 boards.

tracernz:
F1 boards don't have this object and it generates a warning when not present and the Attitude tab is saved.
Fixes #131.

### DIFF
--- a/ground/gcs/src/plugins/config/configattitudewidget.cpp
+++ b/ground/gcs/src/plugins/config/configattitudewidget.cpp
@@ -173,6 +173,9 @@ ConfigAttitudeWidget::ConfigAttitudeWidget(QWidget *parent) :
     m_ui->levelingStart->setEnabled(true);
     m_ui->levelingAndBiasStart->setEnabled(true);
 
+    // F1 boards don't have this object
+    setNotMandatory("StateEstimation");
+
     refreshWidgetsValues();
 }
 


tracernz:
The side-effect of this is that if you have a board with the full attitude filter, and the UAVOs mismatch GCS so that the StateEstimation object is unavailable, you won't get a warning about the object not being present when you save the Attitude tab. In this case the filter settings will still be greyed out so you can't changed them/expect to save them anyway.

jhitesma:
I like the simplicity of the fix.  But something about the possible side-effect is ringing alarm bells for me.  Not loudly...but my gut is telling me that it should be limited to only affect targets which should never have a StateEstimation object.  

If that's really hard to implement (which it sounds like it might be) I guess this makes sense as a solution - but then we should probably consider revisiting the idea of being able to identify objects which are never available for various targets rather than just "there or not there for any targets" in the future.

mlyle:
What would be really nice is if we would only save objects that are actually dirty / have local changes.  Then this wouldn't trigger, I think.

tracernz:
@mlyle You know I don't think that should actually be particularly hard.

mluessi:
@mlyle what you are suggesting is somewhat related to #61

mlyle:
per IRC, mergin'

PR #138, opened 2015-12-04
OSGEarthView: remove.

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/138)


mluessi:
I guess the fact that I don't know what this does probably means it can be removed ;). Have you had a chance to verify that this causes no problems in the GCS?

mlyle:
It's not built by default.  It's an optional component to do google-earth kinds of things... poorly.

PR #140, opened 2015-12-04
GCS:UAVTalk: Clear queues properly

tracernz:
[The empty() function just returns whether the queue is empty](http://doc.qt.io/qt-5/qlist.html#empty), this was not the intent of the code. I have tested this briefly to make sure I can still connect/disconnect/reconnect/flash firmware/reboot/etc. Seems to be great.


mlyle:
Looks nice, good test on it.  Mergin'.

kubark42:
Went through the code and superficially confirmed that there are no other mistakes of this kind.

PR #144, opened 2015-12-04
Update STM32 Libraries

mluessi:
The STM32 libraries we are using are very old, this updates the libraries for the STM32F4 to the latest version included in the [STM32 USB Libraries V 2.2.0](http://www.st.com/web/catalog/tools/FM147/CL1794/SC961/SS1743/LN1734/PF257882) (released November 2015). This update is needed to support new targets based on newer STM32F4 devices (STM32F411, STM32F446, etc.)

Status: It compiles and it seems to be working (can connect to GCS, etc). However, for some reason the flash controller is in an error state, i.e., `FLASH_GetStatus()` returns `FLASH_ERROR_PROGRAM` in the bootloader before writing to flash is attempted. I have been trying to track this down and the problem seems to be somewhere in the USB Library. Any help with tracking this down is highly appreciated.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/144)


tracernz:
It returns FLASH_ERROR_PROGRAM at the first call to FLASH_GetStatus, before it even attempts the erase, which fails silently (see #147) so GCS tries to go on to programming.

tracernz:
Confirmed it works fine without the updated USB lib related commits. :/

tracernz:
The flash program error gets set somewhere along the line while doing this, before even the first full message is received: https://github.com/mluessi/dronin/blob/3bb29a107f2f795f2688a73e3cb542f59cccef68/flight/PiOS/Common/pios_com_msg.c#L169

tracernz:
Pretty confident it's interrupt related. I put assert checks on the flash status throughout the hid receive function and the assertion fails at different points each time.

tracernz:
The flash status register seems to go bad during this line: https://github.com/mluessi/dronin/blob/stm32_lib_update/flight/PiOS/STM32F4xx/Libraries/STM32_USB_OTG_Driver/src/usb_dcd_int.c#L543

tracernz:
Finally tracked it down, fix submitted in mluessi/dronin#1

mluessi:
@tracernz awesome job :). I suspected it was something weird like that. Unfortunately there are no warnings if you pass a constant into a function that expects a variable that can be modified.

mlyle:
What a waste of ram.

Also curious with the const-- should follow that through the code.  Usually there is a warning:

```
void funny(float *x) {
        *x = 4;
}
...
        const float power = 3;

        funny(&power);
Michaels-MBP:~ mlyle$ gcc -o testy testy.c
testy.c:46:8: warning: passing 'const float *' to parameter of type 'float *'
      discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        funny(&power);
```

lilvinz:
Did you consider removing st libraries completely and using ChibiOS HAL instead? IMHO this is the logical next step after using ChibiOS kernel.

tracernz:
@lilvinz We'd first need to migrate the remainder of the targets to Chibi. It would be nice to do so although we did have some reservations about moving F1 to Chibi due to the increased stack usage we've seen in some places.

tracernz:
@mluessi @mlyle It's this line here that does it: https://github.com/mluessi/dronin/blob/stm32_lib_update/flight/PiOS/STM32F4xx/Libraries/STM32_USB_Device_Library/Core/src/usbd_req.c#L399
Not there in the previous driver:
https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/STM32F4xx/Libraries/STM32_USB_Device_Library/Core/src/usbd_req.c#L392

mluessi:
Rebased. I have been testing / flying this with a prototype that uses an STM32F446\* and have observed no ill effects. We should test this with the current F4 targets. 

 *Some additional changes are required to make this chip work with ChibiOS, will open a separate PR.

mlyle:
```
In file included from /var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/PiOS/pios.h:53:0,
                 from /var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/targets/bu/common/main.c:32:
/var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/PiOS/STM32F4xx/Libraries/CMSIS/Device/ST/STM32F4xx/Include/stm32f4xx.h:112:3: error: #error "Please select first the target STM32F4xx device used in your application (in stm32f4xx.h file)"
  #error "Please select first the target STM32F4xx device used in your application (in stm32f4xx.h file)"
```

mluessi:
Ah.. forgot the bu target, should work now.

Issue #112, opened 2015-12-02
DSM + MotorInputOutputCurveMapping spills autotune points on F1


mlyle:
GCS receiver + MotorInputOutputCurveMapping doesn't spill points
DSM + MotorInputOutputCurveMapping.B=1 doesn't spill points.

powf / expf must be made cheaper for these features to work well on F1.

mlyle:
Just a thought-- may not be DSM, may be my use of MotorInputOutputCurveMapping / powf

jihlein:
FWIW, I looked at the Naze code.  It's an easy fix to separate PPM and DSM.  Will work on getting a pr written and tested.

mlyle:
:) Cool, thanks!

mlyle:
Unfortunately running just SERIAL does not prevent spilled points.  The search continues.

PR #151, opened 2015-12-05
flight: f1: use expf/powf approximations to hopefully drop fewer points

mlyle:
This also removes the constant-multiplication (a term) of MotorInputOutputCurveFit.

Connect #112

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/151)

### DIFF
--- a/flight/Libraries/math/misc_math.h
+++ b/flight/Libraries/math/misc_math.h
@@ -78,6 +78,94 @@ static inline bool IS_NOT_FINITE(float x) {
 	return (!isfinite(x));
 }
 
+/* Following functions from fastapprox https://code.google.com/p/fastapprox/
+ * are governed by this license agreement: */
+
+/*=====================================================================*
+ *                   Copyright (C) 2011 Paul Mineiro                   *
+ * All rights reserved.                                                *
+ *                                                                     *
+ * Redistribution and use in source and binary forms, with             *
+ * or without modification, are permitted provided that the            *
+ * following conditions are met:                                       *
+ *                                                                     *
+ *     * Redistributions of source code must retain the                *
+ *     above copyright notice, this list of conditions and             *
+ *     the following disclaimer.                                       *
+ *                                                                     *
+ *     * Redistributions in binary form must reproduce the             *
+ *     above copyright notice, this list of conditions and             *
+ *     the following disclaimer in the documentation and/or            *
+ *     other materials provided with the distribution.                 *
+ *                                                                     *
+ *     * Neither the name of Paul Mineiro nor the names                *
+ *     of other contributors may be used to endorse or promote         *
+ *     products derived from this software without specific            *
+ *     prior written permission.                                       *
+ *                                                                     *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND              *
+ * CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,         *
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES               *
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE             *
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER               *
+ * OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,                 *
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES            *
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE           *
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR                *
+ * BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF          *
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT           *
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY              *
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE             *
+ * POSSIBILITY OF SUCH DAMAGE.                                         *
+ *                                                                     *
+ * Contact: Paul Mineiro <paul@mineiro.com>                            *
+ *=====================================================================*/
+
+static inline float 
+fastlog2 (float x)
+{
+	union { float f; uint32_t i; } vx = { x };
+	union { uint32_t i; float f; } mx = { (vx.i & 0x007FFFFF) | (0x7e << 23) };
+	float y = vx.i;
+	y *= 1.0f / (1 << 23);
+
+	return y - 124.22551499f
+		- 1.498030302f * mx.f 
+		- 1.72587999f / (0.3520887068f + mx.f);
+}
+
+static inline float
+fastpow2 (float p)
+{
+	float offset = (p < 0) ? 1.0f : 0.0f;
+	int w = p;
+	float z = p - w + offset;
+	union { uint32_t i; float f; } v = { (1 << 23) * (p + 121.2740838f + 27.7280233f / (4.84252568f - z) - 1.49012907f * z) };
+
+	return v.f;
+}
+
+
+static inline float
+fastpow (float x, float p)
+{
+	return fastpow2 (p * fastlog2 (x));
+}
+
+static inline float
+fastexp (float p)
+{
+	  return fastpow2 (1.442695040f * p);
+}
+
+#ifdef SMALLF1
+#define powapprox fastpow
+#define expapprox fastexp
+#else
+#define powapprox powf
+#define expapprox expf
+#endif
+
 #endif /* MISC_MATH_H */
 
 /**
--- a/flight/Libraries/math/misc_math.h
+++ b/flight/Libraries/math/misc_math.h
@@ -121,6 +121,12 @@ static inline bool IS_NOT_FINITE(float x) {
  * Contact: Paul Mineiro <paul@mineiro.com>                            *
  *=====================================================================*/
 
+#ifdef __cplusplus
+#define cast_uint32_t static_cast<uint32_t>
+#else
+#define cast_uint32_t (uint32_t)
+#endif
+
 static inline float 
 fastlog2 (float x)
 {
@@ -140,7 +146,7 @@ fastpow2 (float p)
 	float offset = (p < 0) ? 1.0f : 0.0f;
 	int w = p;
 	float z = p - w + offset;
-	union { uint32_t i; float f; } v = { (1 << 23) * (p + 121.2740838f + 27.7280233f / (4.84252568f - z) - 1.49012907f * z) };
+	union { uint32_t i; float f; } v = { cast_uint32_t ((1 << 23) * (p + 121.2740838f + 27.7280233f / (4.84252568f - z) - 1.49012907f * z)) };
 
 	return v.f;
 }
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -625,8 +625,7 @@ static float mix_channel(int ct, ActuatorDesiredData *desired,
 
 		if (val > 0) {
 			// Apply curve fitting, mapping the input to the propeller output.
-			val = actuatorSettings.MotorInputOutputCurveFit[ACTUATORSETTINGS_MOTORINPUTOUTPUTCURVEFIT_A] *
-					powf(val, actuatorSettings.MotorInputOutputCurveFit[ACTUATORSETTINGS_MOTORINPUTOUTPUTCURVEFIT_B]);
+			val = powapprox(val, actuatorSettings.MotorInputOutputCurveFit);
 		} else {
 			// Idle throttle
 			val = 0.0f;
--- a/shared/uavobjectdefinition/actuatorsettings.xml
+++ b/shared/uavobjectdefinition/actuatorsettings.xml
@@ -23,8 +23,8 @@
             <description>When enabled, the motors will spin at the ChannelNeutral command when armed with zero throttle.</description>
         </field>
 
-        <field name="MotorInputOutputCurveFit" units="-" type="float" elementnames="A,B" defaultvalue="1,1">
-            <description>Actuator mapping of input in [-1,1] to output on [-1,1], using power equation of type a*x^b.</description>
+        <field name="MotorInputOutputCurveFit" units="-" type="float" elements="1" defaultvalue="1">
+            <description>Actuator mapping of input in [-1,1] to output on [-1,1], using power equation of type x^value.</description>
         </field>
 
         <access gcs="readwrite" flight="readwrite"/>
--- a/flight/Modules/Attitude/acro/attitude.c
+++ b/flight/Modules/Attitude/acro/attitude.c
@@ -581,19 +581,26 @@ static void updateTemperatureComp(float temperature, float *temp_bias)
 		temp_accum = 0;
 		temp_counter = 0;
 
-		// Compute a third order polynomial for each chanel after each 500 samples
-		temp_bias[0] = sensorSettings.XGyroTempCoeff[0] + 
-		               sensorSettings.XGyroTempCoeff[1] * t + 
-		               sensorSettings.XGyroTempCoeff[2] * powf(t,2) + 
-		               sensorSettings.XGyroTempCoeff[3] * powf(t,3);
-		temp_bias[1] = sensorSettings.YGyroTempCoeff[0] + 
-		               sensorSettings.YGyroTempCoeff[1] * t + 
-		               sensorSettings.YGyroTempCoeff[2] * powf(t,2) + 
-		               sensorSettings.YGyroTempCoeff[3] * powf(t,3);
+		// Compute a third order polynomial for each channel after each 500 samples
+#if 0
+		for reference :
 		temp_bias[2] = sensorSettings.ZGyroTempCoeff[0] + 
 		               sensorSettings.ZGyroTempCoeff[1] * t + 
 		               sensorSettings.ZGyroTempCoeff[2] * powf(t,2) + 
 		               sensorSettings.ZGyroTempCoeff[3] * powf(t,3);
+#endif
+		temp_bias[0] = sensorSettings.XGyroTempCoeff[0] + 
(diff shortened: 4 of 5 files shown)


LICENSE.txt
```diff
@@ -16,6 +16,49 @@ reference.
...
+ * following conditions are met:                                       *
+ *                                                                     *
+ *     * Redistributions of source code must retain the                *
+ *     above copyright notice, this list of conditions and             *
+ *     the following disclaimer.                                       *
+ *                                                                     *
+ *     * Redistributions in binary form must reproduce the             *
+ *     above copyright notice, this list of conditions and             *
+ *     the following disclaimer in the documentation and/or            *
+ *     other materials provided with the distribution.                 *
```
tracernz: Hmmmmm...... :/

mluessi:
Do we know how good these approximations are? I don't think we have to be super accurate, just curious.

LICENSE.txt
mlyle: 3 clause berkeley license.  No big deal.

mlyle:
@mluessi I did a bunch of tests like these:

https://gist.github.com/mlyle/0637dacaf320214dc7ba

Paying attention to mostly the derivative (right two columns).  It's pretty good.  Basically it takes the pow exponent and folds it into the exponent in the floating point number and fixes up.

dustin:
Plotting the error of those runs...

diff

diff2

mlyle:
Works for me on brushed quad.  No dropped points / fixes #112 .  CPU use 96% during tune on F1 with inputoutputcurvefit.

mlyle:
Additional test appreciated.

tracernz:
Reviewed 4 of 6 files at r1, 1 of 1 files at r2.
Review status: 5 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[flight/Modules/Attitude/smallf1/attitude.c, line 586 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/151#-K4rGNB8N2oiev3uF3i6:-K4rGNB8N2oiev3uF3i7:-425298509) ([raw file](https://github.com/d-ronin/dronin/blob/5af5c17596f7d764374dcc60c94c8755981d21c6/flight/Modules/Attitude/smallf1/attitude.c#L586)):*
I would prefer a comment stating the actual math like below rather than this block TBH, the old could is less meaningful.
```c
// Note:
// a + t * (b + t * (c + t * d)) = a + b*t + c*t^2 + d*t^3
```

---

*[LICENSE.txt, line 35 \[r1\]](https://reviewable.io:443/reviews/d-ronin/dronin/151#-K4l9Idi2Fpjz2CuVmnX-r1-35:-K4rH8SpSL64inzBtpFr:-1738217010) ([raw file](https://github.com/d-ronin/dronin/blob/4fb44fcd22efd0f3f31e3bd8b9ee4521739b0106/LICENSE.txt#L35)):*
We don't provide this stuff with our distribution.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/151)*

flight/Libraries/math/misc_math.h
```diff
@@ -78,6 +78,100 @@ static inline bool IS_NOT_FINITE(float x) {
 	return (!isfinite(x));
 }
 
+/* Following functions from fastapprox https://code.google.com/p/fastapprox/
```
kubark42: Should this be a separate file? Arguments for: 1) the copyright makes more sense when it covers the entire file, 2) this isn't really a miscellaneous math function so much as it is an optimization for pre-existing functions.

flight/Modules/Actuator/actuator.c
```diff
@@ -620,8 +620,7 @@ static float mix_channel(int ct, ActuatorDesiredData *desired,
 
 		if (val > 0) {
 			// Apply curve fitting, mapping the input to the propeller output.
-			val = actuatorSettings.MotorInputOutputCurveFit[ACTUATORSETTINGS_MOTORINPUTOUTPUTCURVEFIT_A] *
-					powf(val, actuatorSettings.MotorInputOutputCurveFit[ACTUATORSETTINGS_MOTORINPUTOUTPUTCURVEFIT_B]);
+			val = powapprox(val, actuatorSettings.MotorInputOutputCurveFit);
```
kubark42: The PR references F1, but this affects all targets. Not sure there's any downside to using ```powapprox()``` here, though.

kubark42:
I dunno. I thought about this and realized I don't think this should be applied globally. I realize there's a large appetite to keep F1 running because there's a large legacy base, and I'm excited that people are willing to support these old chips, but it doesn't make sense to me to see the compromises be made globally. They're going to be very hard to track down and these kind of numerical inaccuracies *do* add up. We're using Analog Devices IMUs on our pro-level drones, as well as everything else you might thing (RTK GPS, UWB, lasers, etc...).  We have enough sensor resolution to notice the difference between doubles and floats, esp. as we try to look at points several miles away from our drone.

mlyle:
@kubark42 

1. It only applies to F1 targets.
2. The curve fit is within 2^-12 on the actuator curve mapping.  We have 10 bits of resolution on what we send to motors.

flight/Libraries/math/misc_math.h
```diff
@@ -78,6 +78,100 @@ static inline bool IS_NOT_FINITE(float x) {
...
+	return fastpow2 (p * fastlog2 (x));
+}
+
+static inline float
+fastexp (float p)
+{
+	  return fastpow2 (1.442695040f * p);
+}
+
+#ifdef SMALLF1
```
mlyle: @kubark42 see here.

kubark42:
10-bits is for the PWM timer circuit, right? All three of our drones-- the PARC, Pocket Flyer, and LVL 1-- use in-house ESCs which are commanded over a digital bus and have speed resolution down to the ZCD, i.e several times the mechanical rotation speed. 

I think that the actuator module is in need of a revamping to account for all the new and neat actuators out there, especially since this module hasn't really been proven to be effective for much more than spinning motors. Even servos it's not really suited for at (the autopilot beats the hell out of them). How about splitting off the F1 stuff into a separate F1-specific actuator module, like is done with Attitude?  Then it can be tightly tuned for the F1's memory and processor needs, e.g. removing the actuator curve fitting. And we're free to evolve the new one.

mlyle:
F1 works fine with actuator curve fitting with the approximation.  It works fine with actuator curve fitting with pow=1 without the approximation.  There is no good justification to split if off.

I am working on going the other direction-- to remove copy & pasted and duplicated code that can be better handled with compile-time defines.

>  On the really tiny, high speed motors for Pocket Flyer, this actually exceeds 16-bits of resolution!

And how closely do you actually hit these control targets?  In dynamic flight you have errors of less than 1 part in 10,000, such that these fitting approximations would matter?

Far cruder approximations were used by the space shuttle to orbital insertion and rendezvous, for airliners navigating over courses of of thousands of miles inertially, for Tomahawk missiles doing vision to hit smallish targets on a pair of m68k's without CPU.  I think we'll be OK.  And it only applies to F1.

mlyle:
Just to super super confirm, @kubark42 --- you do realize the approximation is only being used on F1, right?

kubark42:
> There is no good justification to split if off.

I think this PR is good justification. Our experiences with our industrial vehicles points toward the need to increase accuracy.

> And how closely do you actually hit these control targets? In dynamic flight you have errors of less than 1 part in 10,000, such that these fitting approximations would matter?

It's the sum of the parts, right? All these 1:100000 numerical inaccuracies add up. I don't think we can point to any one spot and say "this must be double". We believe a lot has to do with numerical accuracy in the estimation filter, but without a ground truth source at 400' in the air it's hard to **know**, know what I mean? Our default approach is to use highest precision everywhere so we reduce factors that could be having secondary effects.

> Just to super super confirm, @kubark42 --- you do realize the approximation is only being used on F1, right?

Yup, yup. However, the change in the UAVO is system-wide, right?

shared/uavobjectdefinition/actuatorsettings.xml
```diff
@@ -23,8 +23,8 @@
             <description>When enabled, the motors will spin at the ChannelNeutral command when armed with zero throttle.</description>
         </field>
 
-        <field name="MotorInputOutputCurveFit" units="-" type="float" elementnames="A,B" defaultvalue="1,1">
-            <description>Actuator mapping of input in [-1,1] to output on [-1,1], using power equation of type a*x^b.</description>
+        <field name="MotorInputOutputCurveFit" units="-" type="float" elements="1" defaultvalue="1">
```
kubark42: Might have been better to have had this change be proposed in a second PR. It's actually system-wide, and doesn't affect only F1 targets.

mlyle: We're free to discuss it here.  Multiplication commutes with two levels of control terms and introduces weird clipping behavior (that the previous levels don't).  Multiplication by a constant can't, by definition, improve linearity.

I know I've asked you about this before but we simply can't come up with a justification for this.

mlyle:
> I think this PR is good justification. 

I don't understand how?  Currently, without duplicating code we get something that works well on F1 and higher end targets.  It also is functionally identical (other than removing the vestigial a term) with the previous code.  Duplicating code has been a big problem with maintainability.

IMO 3-5 lines of ifdef are better than duplicating a 760 line module.

shared/uavobjectdefinition/actuatorsettings.xml
kubark42: <shrug> It's okay here, but in general I think we should all advocate for keeping PRs atomic. From my perspective, this makes it harder for me to track important changes in dRonin that affect what I'm doing. I would have had a surprising settings conflict if I wouldn't have looked into this merge request. As I am this very second trying to track down why a merge conflict broke our internal sensor board build, I'm kinda sensitive to this. :p

I think the justification is to keep things atomically separate. I completely agree that the math works out the same since the scalar factors out. However, the process becomes more complex than it needs to. For instance, the curve fits are better with a second term. So ideally I would still do a two-term curve fit, and then I would neglect the A term in the settings object. Workable, but wonky. 

If it were for some great reason, I'd be all for it. But it's really just to save memory and CPU, right? This is the point where I think it's better to spin off the extremely resource constrained platforms into their own frozen module. I really like https://github.com/TauLabs/TauLabs/issues/1970, but I don't see how that fits into an F1 world.

mlyle: What does "better" mean?  Are the curve fits more linear?

mlyle: Also, it's not to save memory and CPU.  It is to remove stuff that no one who's been around at dRonin has been able to see a purpose for?

mlyle: (You still haven't been able to explain what this is for-- is it deliberate for the clipping behavior?  Are you using this to avoid the upper end of the actuator range where there's saturation effects?)   Without some kind of justification of what this is trying to do, it doesn't belong here.  And if you're using it to deliberately clip, rather than as a pure multiplication operation, that needs to be documented and justified.

kubark42: I'm not sure what you mean by more linear in the sense of fitting an exponential, but from our test stand data the fit is better for normal purposes. The data seems to show that the "meat" of the curve, i.e. the center point where the propeller spends most of its time, is a better fit with two parameters. That leads to slightly worse fits at the upper and lower ends (because theoretically the ```A``` term should be identically ```1```), but if we're shooting for maximal precision in standard flight this seems like a good compromise.

mlyle: > The data seems to show that the "meat" of the curve, i.e. the center point where the propeller spends most of its time, is a better fit with two parameters.

Whatever output setting results in a hover does not change by multiplication.

Multiplying a function by something does not make the corresponding "output" part of the function more linear.

The shape of the curve where hover occurs (or a given speed of forward flight) does not change by virtue of multiplication.

kubark42: Agreed in principle. But the common tools for curve fitting will return different results if identical data sets are fitted to ```e^b``` vs. ```a*e^b```.

mlyle: OK, then it's vestigial at best and harmful at worst.  We know it can't make the output any more linear.  And it can introduce clipping at the extremes, and clipping is bad.

mluessi: @kubark42 don't we need the constant if you want the output to be real world units? So it e.g. fits the test stand data? In terms of control it shouldn't matter though (i.e. it just has to linearize it).

kubark42: That's a fair point.

kubark42: @mluessi The constant can be identically 1 and include the  proper unit conversions. :D Happens all the time in mathematics (and really drives me crazy because it's actually important to decide why a unit-shifting constant should be 1, *a la* this conversation).

kubark42: P.S. @mlyle the point I was trying to make about atomicity is important here, because I think it took us a little while to sync up on this issue. I mistook the motivation for the UAVO change because it was framed in the topic of saving RAM on F1.

mlyle: Yes, that's fair.

I can't promise smaller, more atomic PRs-- it would hard for me to justify serializing two related changes in this module (or especially commits, at this point-- we're having some degree of internal debate about squashing-down most PRs into individual commits).  But I'll endeavor to note things like this in the comments.

And I'd be lying if efficiency wasn't a small concern- when I'm cleaning up performance for F1, if I see something that I have always thought was vestigial, I'm inclined to take it away.  But one multiply does not make or break this PR.  Odds are that even the change to the actuator half is optional.

P.S. It's possible that the a term could be actually useful, if one desired to deliberately clip or cut off part of the ESC range.  I would think this would belong in output settings though (or in the case of the ESC, setting a smaller maximum thrust/RPM target to correspond to '1.00' == full effort.

tracernz:
Reviewed 1 of 6 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/151)*

mlyle:
In renatus for now, but doubtful

mlyle:
For this to land, it needs to get 2-3 good tunes on F1 on additional aircraft (or, failing that, we could hack it into F3/F4 temporarily so that other vehicles could fly it but it's not a perfect test).

LICENSE.txt
hernick-qc: This license doesn't belong in LICENSE.txt. If you search for "Redistributions" in the git, you'll see hundreds of other files are covered by similar licenses and they do get a mention in the LICENSE.txt:

"The dRonin code is licensed under the GPLv3. There are a few minor exceptions
to this so please see the headers of all source code for copyright and license
information."

Nothing more necessary. Code is open source, and binaries link to the public source code, which qualifies as "other materials provided".

Just keep the 3-clause in the misc_math.h file

flight/Libraries/math/misc_math.h
hernick-qc: No, it shouldn't be a separate file. There are hundreds of files already in the code that include a 3-clause BSD license, this one shouldn't be treated any differently. Just put the boilerplate in the file and the general disclaimer in LICENSE.txt covers it.

kubark42: @hernick-qc Good to know about the license. I'm not sure that we can arrive at the conclusion yet, since the second point is that an argument could be made that this isn't really miscellaneous math so much as it is optimization math. This code doesn't introduce new functionality, it improves on existing functions.

I like the notion of having an optimization library, but it's up to the author.

mlyle:
Fixes #112

mlyle:
Connect #112

ufoDziner:
Rate mode was used to test after applying the autotune settings with no adjustment to the sliders.  The quad would return to level smoothly.  It did not hold the commanded attitude with the stick returned to center.  It feels soft.

ufoDziner:
My son flew with it today.  Had to adjust the sliders quite a lot to get a decent feel, but once the PIDs were high enough, it flew really well.  He said that it felt as good or better than any tune yet, and that was reinforced by how well he flew.  So, overall great, just not with the slider defaults.

mlyle:
@ufoDziner flew TPR2 and had better, but still anomalous results without this PR.  This PR is probably dead for Renatus -- unless, like, 3 people decide to show up and fly it with great results.

dustin:
Just test flew this on my [big tri](http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAgOuingoM).  AFAICT, it's doing great.

mlyle:
build artifacts please

mlyle:
build artifacts please

mlyle:
@ufodziner flew this in 3 successive tunes, getting these results

http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAgOvdiAoM
http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAgN2oiAoM
http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAgOvdiAoM

mlyle:
The taus and gains are radically different, showing that at least on this craft pr151 does not produce deterministic results.  We don't know whether this is a problem at baseline, but at the least this is a significant concern.

dustin:
It'd be very interesting to know if that's related to this change.

mlyle:
The conclusion is.. inconclusive.  :P  That particular craft does not seem to autotune well from defaults.

The set of 3 tunes gathered without 151 were very slightly better.

Still, there's not much cost from deferring '151.  It just means we can't recommend actuator curve fitting to people flying F1 craft.  It's time to derisk.

jhitesma:
Ran three back to back tunes on this on my 250 tonight and got similar but not identical Tau's on all three.  Tau got slightly worse each tune but I'm guessing that was from my pack gettiing lower - by the end I was at the end of the pack.  Flew the last tune briefly and it felt good in the air.

mlyle:

mlyle:
Rebased out conflicts.

dustin:
My last test included this an #457 on naze32: http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAgILMuAgM

mlyle:
Closing in favor of #457 which now supersedes

PR #152, opened 2015-12-05
Revo/Naze: New board graphics for uploader

tracernz:
Also turned on smooth pixel transform when resizing the board pics, before/after:
screenshot from 2015-11-04 18 44 32screenshot from 2015-11-04 18 43 10
[Altium PcbDoc](https://github.com/d-ronin/dRonin/blob/next/flight/targets/revomini/hw/Revolution.PcbDoc) used for the Revo image.
[Diptrace file](https://github.com/d-ronin/dRonin/blob/next/flight/targets/naze32/hw/OpenNaze.dip) with some additional 3D models used for the OpenNaze image.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/152)


tracernz:
Also kills a null pixmap warning when using the revo since it didn't have a pic before.

mluessi:
LGTM

DTFUHF:
Are there any pics of actual naze? Now that closed source targets are allowed unabashedly.

tracernz:
@DTFUHF: No source files available, so difficult to get some, and the original author isn't exactly co-operative on the matter so unlikely that we can use his artwork. We could use another photo but my feeling is that they always look worse than the renders.

tracernz:
We could go to the trouble of trying the recreate a 3D model of the Naze32 which would be a matter of just laying the components on the top layer without any routing for this job but it's a bit of work.

Issue #126, opened 2015-12-03
Consider updating to Qt 5.5.1 or 5.6


tracernz:
Even if we decide not to follow Qt 5.6, staying up-to-date on the 5.5 branch is still a good idea. Relevant changes:
- http://code.qt.io/cgit/qt/qtbase.git/tree/dist/changes-5.5.1
   * [QTBUG-46379] Tooltips on OS X are now transparent for mouse events.
 - [QTBUG-47775][Darwin] Fixed Info.plist creation when bundle name
   contains spaces.
 - ?
- http://code.qt.io/cgit/qt/qtquickcontrols.git/tree/dist/changes-5.5.1
- ?
- http://code.qt.io/cgit/qt/qtserialport.git/tree/dist/changes-5.5.1
   * [QTBUG-46875] Fixed crash while enumerating available ports
     on OS X 10.10
   * [QTBUG-47127] Available serial port information is not duplicated
     anymore on Windows.
- https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/dist/changes-3.5.1.md?h=3.5
 * Fixed crashes with code completion (QTCREATORBUG-14991, QTCREATORBUG-15020)
 * Added support for MSYS2 compilers and debuggers

jhitesma:
So I've apparently been building with 5.5.1 on windows all week inadvertently and haven't noticed any issues (other than the one command missing a .exe extension when building standalone from #119 - but that issue doesn't exist building just gcs or building release.)

![gcs_ver](https://cloud.githubusercontent.com/assets/2054286/11609345/9c53d080-9b41-11e5-8bb0-21c48781ace1.jpg)

PR #153, opened 2015-12-05
Build: Update default windows env bash_profile

DTFUHF:
Current Qt version available for download is 5.5.1 which resides in a different directory than 5.5.0. This changes allows new development environment setups to work out of the box. I suggest also changing the wiki to link directly to a specific version, or even the desired .exe as there are a lot to choose from and only one is the correct one.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/153)

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -581,7 +581,7 @@ endif
 
 # OPENSSL download URL
 ifdef WINDOWS
-  openssl_install: OPENSSL_URL  := https://slproweb.com/download/Win32OpenSSL-1_0_2d.exe
+  openssl_install: OPENSSL_URL  := https://slproweb.com/download/Win32OpenSSL-1_0_2e.exe
 
 openssl_install: OPENSSL_FILE := $(notdir $(OPENSSL_URL))
 OPENSSL_DIR = $(TOOLS_DIR)/win32openssl
--- a/make/winx86/bash_profile
+++ b/make/winx86/bash_profile
@@ -1,7 +1,7 @@
 # bashrc for TauLabs development on Windows
 # Michael Lyle - 2015
 
-QT_VERSION_FULL=5.5.1
+QT_VERSION_FULL=5.5.0
 QT_VERSION=`echo "$QT_VERSION_FULL" | cut -d '.' -f1-2`
 QT_BASEDIR="/C/Qt"
 
--- a/make/winx86/bash_profile
+++ b/make/winx86/bash_profile
@@ -1,7 +1,7 @@
 # bashrc for TauLabs development on Windows
 # Michael Lyle - 2015
 
-QT_VERSION_FULL=5.5.0
+QT_VERSION_FULL=5.5.1
 QT_VERSION=`echo "$QT_VERSION_FULL" | cut -d '.' -f1-2`
 QT_BASEDIR="/C/Qt"
 


DTFUHF:
Updated this PR to include an update to the OpenSSL installer

jhitesma:
Not opposed, just trying to enlighten myself...I just setup my environment from scratch this week and got 5.5.1, it's been working just fine without this change.  It looks like the directory is still qt/5.5 for both .0 and .1 - in fact doesn't the next line of the profile strip off the trailing .x?

New openSSL must have just pushed...it worked for me two days ago but confirmed the d version is gone.

I say merge, just not sure if the QT version change is really "necessary" but I don't see any reason not to do it.

DTFUHF:
Qt installed in C:\Qt\Qt5.5.1 for me, which was the default.

I don't know much about Linux commands so I can't help there. But, bash couldn't find the make executable without this change.

DTFUHF:
Ah, check out the next lines 8-10, there is an additional conditional if /Qt5.5/ doesn't exist

jhitesma:
Missed those lines :)

I am on windows but working in git_bash where things look unixy and I was abbreviating the path.  Sorry for the confusion.

I had to install QT on my D: drive (C: is getting kind of full) and tweak the Makefile for that, but that was just line 6 QT_BASEDIR I left all the directories default when installing QT other than changing it to D: drive instead of C: 

```
jhitesma@backroom MINGW64 /d/qt
$ ls
5.4/            Examples/            MaintenanceTool.dat   Tools/
5.5/            Extras/              MaintenanceTool.exe*  update.rcc
components.xml  InstallationLog.txt  MaintenanceTool.ini   vcredist/
Docs/           Licenses/            network.xml
```

My .bash_profile
```
QT_VERSION_FULL=5.5.0
QT_VERSION=`echo "$QT_VERSION_FULL" | cut -d '.' -f1-2`
QT_BASEDIR="/D/Qt"

if [ -d "$QT_BASEDIR/Qt$QT_VERSION_FULL" ] ; then
  QT_BASEDIR="$QT_BASEDIR/Qt$QT_VERSION_FULL"
fi
```

And my resulting QT_BASEDIR
```
jhitesma@backroom MINGW64 /d/qt
$ echo $QT_BASEDIR
/D/Qt
```

So it looks like line 6 is setting the basedir, then lines 8-10 try to change it if the full version path exists...but since I only have /d/qt/5.5 it's not changing it and leaving me with /d/Qt and I can confirm things are working.

I just checked my other windows computer I setup this week where I did install QT in the default location on C and it's the same only c:\qt\5.5

I just updated my .bash_profile to be 5.5.1 and then restarted bash - and it still works for me.  So the change doesn't break anything and if it fixes things in other cases I say go for it.

I'm just trying to figure out how we got different directories since we both apparently just installed it within the past few days and appear to be using the same version :D

mlyle:
The big installer vs. the one that dynamically downloads puts things in different places-- the .bash_profile is set up to handle both.

I think I'd prefer to update the wiki to give the path of the 5.5.0 installer.  All other platforms build with 5.5.0 and if we move to 5.5.1 it should be a deliberate thing, not our developers skewing on Windows to 5.5.1.

mlyle:
Made the wiki change to point to 5.5.0.    Also relevant, #126.

DTFUHF:
GCS seems to be working OK building with 5.5.1 tools on windows. :)

I can easily roll back my installation to use 5.5.0 and remove my change, leaving only the openssl change. Then the makefile would be addressed as part of #126

mlyle:
@DTFUHF :) That's a helpful report.

> I can easily roll back my installation to use 5.5.0 and remove my change, leaving only the openssl change. Then the makefile would be addressed as part of #126

Sounds like a plan.

jhitesma:
Yeah, I've been building with 5.5.1 and didn't realize we were supposed to be on 5.5.0, haven't seen any issues from it yet with my GCS builds other than that issue with the missing .exe on the one packaging tool I experiened buiding #119 

Thanks for clarifying how we wound up with different installations.  Didn't know they had two different installer packages, was just rushing through following the wiki instructions.

DTFUHF:
Reverted makefile change

tracernz:
We should probably change our download links to point here so people don't end up with 5.5.1 http://download.qt.io/archive/qt/5.5/5.5.0/

tracernz:
I have opened an issue #126 already to consider 5.5.1.

mlyle:
@tracernz yah above I already updated the wiki.

DTFUHF:
As an aside - the instruction to download DD doesn't say where to install it to (if anywhere), or when it's needed. Everything I've done hasn't required it...

tracernz:
That's a good point, I wonder what we do actually use it for.

tracernz:
@DTFUHF We use dd to make the entire flash firmware images.

DTFUHF:
I've done that too, without downloading dd as instructed :-\

tracernz:
dd is probably already included in something then. I guess just remove it from the wiki.

PR #168, opened 2015-12-06
Make: Add EF target for Naze, make Intel hex of EF image

tracernz:
This allows a complete Intel hex image for the Naze to be produced, which can then be flashed with other utilities like baseflight configurator. This should be useful until GCS is able to flash these boards. Intel hex files are now produced for all EF targets.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/168)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -897,8 +897,8 @@ all_$(1)_clean: $$(addsuffix _clean, $$(filter ef_$(1), $$(EF_TARGETS)))
 endef
 
 # Some boards don't use the bootloader
-NOBL_BOARDS    := naze32
 FW_BOARDS      := $(ALL_BOARDS)
+NOBL_BOARDS    := $(strip $(foreach BOARD, $(ALL_BOARDS),$(if $(filter no,$($(BOARD)_bootloader)),$(BOARD))))
 BL_BOARDS      := $(filter-out $(NOBL_BOARDS), $(ALL_BOARDS))
 BU_BOARDS      := $(BL_BOARDS)
 EF_BOARDS_NOBL := $(NOBL_BOARDS)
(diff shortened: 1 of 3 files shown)


jhitesma:
Even though this is how I've had to flash my one Naze board (because stm32flash and that particular board are not good buddies) I'm not sure if this is really a good idea due to the hex files lacking meta data.

It seems that the primary use case for the .hex would be new users, precisely the people who would benefit most from having that meta data to make sure their FW and GCS match.

If we can add the meta data to the hex file then I'm all for this, since I love the idea of people switching from *flight being able to use the tools they're used to get away from *flight.  And there are apparently boards like mine that for whatever reason simply refuse to talk to stm32flash.  But without the meta data in the hex I'm worried that this is going to lead to a lot of support issues and frustrated new users.

mlyle:
I think this hex file (the EF file) will have metadata.

Downside is it should not be used after the first flash, because it'll erase configuration.  (it writes -everything-)

mluessi:
@tracernz unrelated but another makefile / openocd challenge: make "fw_*_program` write the FW version information. At the moment, this is only written when using USB with the bootloader to program a target.

tracernz:
@jhitesma The whole idea of this PR is to create hex files with the metadata. ;-) @mlyle That's true. @mluessi That's an interesting one, I should pick up my OpenOCD work when I find some time.

tracernz:
@mlyle What do you think about calculating the firmware post-pad only up to the EE_BANK_BASE? This would affect all targets that use emulated EEPROM.

jhitesma:
I thought it looked like it was going to include the meta but wasn't quite sure.  Thanks for clarifying!

jhitesma:
Just gave it a try.  Confirmed that it builds with meta data included and can be flashed with CF configurator :+1:

Issue #170, opened 2015-12-06
Re-evaluate 60 second autotune period


droidicus:
With the new buffer based autotune that @mlyle has done, we think that a shorter autotune period would provide converged autotune data.

I will take an initial pass at looking into this and report back when I have more.

droidicus:
#172 will allow the SystemIdent object to be logged whenever it changes. This should be at a rate of ~4 Hz, so setting the logging to 10Hz is sufficient to get every data point.

Below is the data from an autotune run on my 150mm Diatone Blade, starting with default PID values-

Noise:
![autotune_noise](https://cloud.githubusercontent.com/assets/6981482/11615582/3b7f1902-9c33-11e5-82ba-4214f526a4e6.png)

Beta:
![autotune_beta](https://cloud.githubusercontent.com/assets/6981482/11615579/3b7e02a6-9c33-11e5-9474-b6a1a4e1ca12.png)

Bias:
![autotune_bias](https://cloud.githubusercontent.com/assets/6981482/11615580/3b7ed7a8-9c33-11e5-8bda-90313c649bb2.png)

Tau:
![autotune_tau](https://cloud.githubusercontent.com/assets/6981482/11615581/3b7f093a-9c33-11e5-9730-2c15f0744abb.png)

droidicus:
Results from @fujin and his 250 racing quad

Noise:
![fujin250_autotune_noise](https://cloud.githubusercontent.com/assets/6981482/11616920/15385dd6-9c55-11e5-8ff0-1fea25860f90.png)

Beta:
![fujin250_autotune_beta](https://cloud.githubusercontent.com/assets/6981482/11616919/15374d2e-9c55-11e5-9db0-3453d7c04b04.png)

Bias:
![fujin250_autotune_bias](https://cloud.githubusercontent.com/assets/6981482/11616921/1538e9d6-9c55-11e5-8e23-46e98a0586ad.png)

Tau:
![fujin250_autotune_tau](https://cloud.githubusercontent.com/assets/6981482/11616922/1539a506-9c55-11e5-8a29-0336c2fed60e.png)

droidicus:
@fujin 250mm race quad, second tune starting with the PID values from the previous run.

Noise:
![fujin250second_autotune_noise](https://cloud.githubusercontent.com/assets/6981482/11617045/acfac0f8-9c57-11e5-9bf9-471d853a896b.png)

Beta:
[...]

PR #172, opened 2015-12-06
Adding SystemIdent object to flash logging

droidicus:
Adding SystemIdent object to flash logging, only logged on change (currently once every 256 autotune cycles)

Related to #170

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/172)

### DIFF
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -172,7 +172,8 @@ static void loggingTask(void *parameters)
 
 	// Connect callbacks for UAVOs being logged on change
 	FlightStatusConnectCallbackCtx(UAVObjCbSetFlag, &flightstatus_updated);
-	SystemIdentConnectCallbackCtx(UAVObjCbSetFlag, &systemident_updated);
+	if (SystemIdentActiveHandle())
+		SystemIdentConnectCallbackCtx(UAVObjCbSetFlag, &systemident_updated);
 	if (WaypointActiveHandle())
 		WaypointActiveConnectCallbackCtx(UAVObjCbSetFlag, &waypoint_updated);
 
@@ -333,7 +334,7 @@ static void loggingTask(void *parameters)
 				flightstatus_updated = false;
 			}
 			
-			if (systemident_updated){
+			if (systemident_updated && SystemIdentActiveHandle()){
 				UAVTalkSendObjectTimestamped(uavTalkCon, SystemIdentHandle(), 0, false, 0);
 				systemident_updated = false;
 			}
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -7,7 +7,7 @@
  *
  * @file       logging.c
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2014
- # @author     dRonin, http://dronin.org Copyright (C) 2015
+ * @author     dRonin, http://dronin.org Copyright (C) 2015
  * @brief      Forward a set of UAVObjects when updated out a PIOS_COM port
  * @see        The GNU Public License (GPL) Version 3
  *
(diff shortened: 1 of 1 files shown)


mlyle:
dronin, test this please

flight/Modules/Logging/logging.c
```diff
@@ -169,6 +172,7 @@ static void loggingTask(void *parameters)
 
 	// Connect callbacks for UAVOs being logged on change
 	FlightStatusConnectCallbackCtx(UAVObjCbSetFlag, &flightstatus_updated);
+	SystemIdentConnectCallbackCtx(UAVObjCbSetFlag, &systemident_updated);
```
BrainFPV: I think you need to use an `if SystemIdentHandle()` (like for waypoint) otherwise it will crash if the UAVO is not present.

flight/Modules/Logging/logging.c
```diff
@@ -327,7 +332,12 @@ static void loggingTask(void *parameters)
 				UAVTalkSendObjectTimestamped(uavTalkCon, FlightStatusHandle(), 0, false, 0);
 				flightstatus_updated = false;
 			}
-
+			
+			if (systemident_updated){
```
BrainFPV: Same here

flight/Modules/Logging/logging.c
mluessi: Oops: used wrong github account ;)

droidicus:
@mluessi, please review changes

tracernz:
Reviewed 1 of 1 files at r2.

---

*[flight/Modules/Logging/logging.c, line 338 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/172#-K4titsbUlw-6M1nWvaM:-K4titsbUlw-6M1nWvaN:1337222452) ([raw file](https://github.com/d-ronin/dronin/blob/fb02af87621d2153267de651932628e5b1be9a87/flight/Modules/Logging/logging.c#L338)):*
Should we be directly calling the UAVTalk layer here?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/172)*

droidicus:

---

*[flight/Modules/Logging/logging.c, line 338 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/172#-K4titsbUlw-6M1nWvaM:-K4tk8mFBqvRJ9ihHsq-:1351458498) ([raw file](https://github.com/d-ronin/dronin/blob/fb02af87621d2153267de651932628e5b1be9a87/flight/Modules/Logging/logging.c#L338)):*
I am not sure, I followed the pattern that was used to log the other UAVOs, so if this should be changed the whole logging scheme would need to be reviewed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/172)*

droidicus:
I broke something, standby.

mlyle:
I think we prefer taulabs/taulabs#2031 ... but things like this are proof we need to get that landed sooner rather than later.

droidicus:
Should be un-broken now. @mlyle, agree, this is a stopgap if anything.

mluessi:
@droidicus looks good. Depending on how long it takes for #180 to land, we should maybe wait and the UAVO can just be added to the default profile.

droidicus:
@mluessi, agree

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/172#-:-K4x1X69k2VwrkJhCSFl:701738457)*

mlyle:
Closing in anticipation of taking #180

droidicus:
Closing in favor of #180

PR #187, opened 2015-12-08
branding: gcs: change one last string for Windows only

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/187)


mlyle:
Short circuited per dsal in PtDreamer#1

PR #191, opened 2015-12-08
Remove ccache spam from fw makefile on Windows build env

droidicus:
I do not have much experience with makefiles, so please let me know if there is a better way to accomplish this.

Tested on Windows build env, not tested on other platforms.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/191)

### DIFF
--- a/make/firmware-defs.mk
+++ b/make/firmware-defs.mk
@@ -5,14 +5,19 @@ CCACHE :=
 
 ifeq ($(FLIGHT_BUILD_CONF), debug)
 export DEBUG:=YES
-CCACHE := $(shell which ccache)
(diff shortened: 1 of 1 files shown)


mlyle:
Can you open an issue instead?

What I want to do is find ccache in the top level makefile and set an appropriate variable that is inherited by child makefiles.. and is selectively taken based on FLIGHT_BUILD_CONF.  Problem is I have lots of pending makefile changes.

The reason I like this approach is that it consolidates all the logic to find ccache and it also lowers the cost of each recursed makefile having to look for it.  It also leaves the option open to use ccache when we get appropriate bins made that can tolerate our long command lines.

mlyle:
Superseded #196

PR #193, opened 2015-12-08
PiOS: Use same math for system time regardless of RTOS

### LINKED ISSUE #98: ChibiOS uptime with GCS connected is limited to 4295 seconds/1hr 11 min
ChibiOS:
https://github.com/d-ronin/dronin/blob/next/flight/PiOS/Common/pios_thread.c#L206
```c
#define ST2MS(n) (((((n) - 1UL) * 1000UL) / CH_FREQUENCY) + 1UL)
```
 Overflows at (2^32-1)/1000
 (2^32-1)/1000/1000(ms) = 4295 s = approx 1 hr 11 mins

FreeRTOS:
https://github.com/d-ronin/dronin/blob/next/flight/PiOS/Common/pios_thread.c#L43
```c
#define TICKS2MS(t) ((t) * (portTICK_RATE_MS))
```
portTICK_RATE_MS is 1 for our targets (i.e. 1000 Hz tick frequency) thus overflows at 2^31-1
 (2^32-1)/1000(ms) = 4294967 s = approx 1193 hours

I am assuming an assert somewhere causes a reboot when we rollover PIOS_Thread_Systime because my Chibi boards reboot at about this time.

fujin:
Avoids system time rolling over at 4295 seconds. May be worse on Chibi for other tick frequencies but we use 1000 Hz across the board so no issue anticipated.
Fixes #98.

(it looks like @tracernz forgot to open a PR here.)

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/193)

### DIFF
--- a/flight/PiOS/Common/pios_thread.c
+++ b/flight/PiOS/Common/pios_thread.c
@@ -31,17 +31,19 @@
 #error "pios_thread.c requires PIOS_INCLUDE_FREERTOS or PIOS_INCLUDE_CHIBIOS"
 #endif
 
+// portTICK_RATE_MS is in [ms/tick].
+// See http://sourceforge.net/tracker/?func=detail&aid=3498382&group_id=111543&atid=659636
+// This is used on both FreeRTOS and ChibiOS because the equivalent ChibiOS macro rolls over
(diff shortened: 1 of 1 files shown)


fujin:
?? seriously, closing because people think opening PRs is bad. stupid af

PR #197, opened 2015-12-09
AQ32 - Fix rescue mode.

jihlein:
A while back I noticed that the AQ32 board could not be "rescued" by the GCS.  After reading some of the code review comments for the Lumenier F3 board and it's lack of a vsense input from the usb port, I changed the AQ32 bootloader delay from 2 to 500, and now the board can be rescued.

Tested on AQ32 hardware, board can now be halted, booted, and rescued from the GCS.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/197)


tracernz:
:lgtm_strong: once Jenkins gives the go-ahead.

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/197#-:-K53qUR5J6b1_RwPYAeW:1962039119)*

PTDreamer:
This makes no sense to me, so the board seems to have the VSense pin connected according to this changeset and the delay was increased as a result? This is the only board present on the codebase with this combination. 
Also FYI this will break spektrum binding.

mlyle:
It looks like on the schematic that VBUS goes through a diode to a linear regulator and nowhere else.

I know it breaks Spektrum binding, but for targets without a bus sense I don't think we have a choice.

PTDreamer:
> It looks like on the schematic that VBUS goes through a diode to a linear regulator and nowhere else.

So why does this PR define VBUS_SENSING_ENABLED?

mlyle:
I believe it comments out that line.

PTDreamer:
> I believe it comments out that line.

Please disregard the above convo, I'm an idiot.

PR #199, opened 2015-12-09
GCS:  Welcome Page update HITL Button to point to Advanced view

VArcht:
HITL button on the welcome page did not do anything when clicked now it
will navigate to advanced view

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/199)


ground/gcs/src/plugins/welcome/qml/main.qml
```diff
@@ -90,8 +90,8 @@ Rectangle {
 
                 WelcomePageButton {
                     baseIconName: "hitl"
```
tracernz: Should really rename this too.

VArcht: Breaks it, if it is important I'll have to dig into it, wasn't immediately obvious how it works.

tracernz: I imagine it points to the image filename?

VArcht: Yeah probably, I did consider that before I got lazy. I'll give it a look in the morning, thx.

tracernz: Yeah, it does: https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/welcome/qml/WelcomePageButton.qml#L16

VArcht:
I addressed the icon name.

PR #200, opened 2015-12-09
Remove temperature bias infrastructure

### LINKED ISSUE #198: GCS/Flight: Remove temperature compensation infrastructure
Only works on F1, poorly tested, uses up a tab in GCS.

tracernz:
Fixes #198.
Will probably clash with #151.
Probable that I missed some stuff.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/200)


mluessi:
Is it a good idea to remove this? I wonder if some people are actually using it to compensate for gyro bias when doing navigation stuff. Maybe just disable it for F1 targets?

tracernz:
#198 says it only works on F1.

mluessi:
That doesn't seem right, the code is used on all targets. That being said, I did some tests a while ago which made me question how effective the temp comp really is:

http://forum.taulabs.org/viewtopic.php?f=17&t=366&start=10

kubark42:
I still use this. It's been very helpful when transitioning from indoors to outdoors in the Boston winter. We've done thermal chamber tests from -40C to +80C and the results were linearly temperature dependent and-- to a certain extent-- repeatable across multiple sensors. 

The magnetometer in particular was surprisingly susceptible to temperature change. In the future, I expect to be doing more thermal compensation, not less.

mlyle:
@tracernz It seems I was wrong and the F3/F4 version of the temperature compensation is done in sensors.c.

mlyle:
@tracernz I'm sorry that I wasted your effort here :/  It is a much less clear cut case to remove it with it being possible on modern targets.

tracernz:
No problem. Looking at @mluessi's data I think we should keep this, at least on F3/F4.

PR #201, opened 2015-12-09
GCS: Tweak basic tab

### LINKED ISSUE #133: GCS:Stab Tab: Tweaks to basic page
We should add derivative to the basic page and give it a suitable scaling. This is a good page for ex-MultiWii users with this change. 

Should also review the limits as most people report that autotuned integral gains show up red.

Also to consider:
* Add full stick rate and expo.

tracernz:
Added derivative (with nice scaling), added max. rate and expo. Moved zero integral at zero throttle to advanced tab. Also increased the maximum settable rate from 500 to 1000 deg/sec.

(screenshot is pre-increased max rates)
screenshot from 2015-12-09 21 21 27

Fixes #133

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/201)

### DIFF
--- a/shared/uavobjectdefinition/stabilizationsettings.xml
+++ b/shared/uavobjectdefinition/stabilizationsettings.xml
@@ -11,9 +11,9 @@
 	<field name="HorizonExpo" units="%" type="uint8" elementnames="Roll,Pitch,Yaw" defaultvalue="30,30,30" limits="%BE:0:100,%BE:0:100,%BE:0:100"/>
 	<field name="PoiMaximumRate" units="degrees/sec" type="float" elementnames="Roll,Pitch,Yaw" defaultvalue="30,30,30" limits="%BE:0:500,%BE:0:500,%BE:0:500"/>
 
-	<field name="RollRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.002,0.0015,0,0.3" limits="%BE:0:0.01,%BE:0:0.01,, "/>
-	<field name="PitchRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.002,0.0015,0,0.3" limits="%BE:0:0.01,%BE:0:0.01,, "/>
-	<field name="YawRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.0035,0.0035,0,0.3" limits="%BE:0:0.01,%BE:0:0.01,, "/>
+	<field name="RollRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.002,0.0015,0,0.3" limits="%BE:0:0.01,%BE:0:0.05,, "/>
+	<field name="PitchRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.002,0.0015,0,0.3" limits="%BE:0:0.01,%BE:0:0.05,, "/>
+	<field name="YawRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.0035,0.0035,0,0.3" limits="%BE:0:0.01,%BE:0:0.05,, "/>
 	<field name="RollPI" units="" type="float" elementnames="Kp,Ki,ILimit" defaultvalue="2.5,0,50" limits="%BE:0:10,%BE:0:10,"/>
 	<field name="PitchPI" units="" type="float" elementnames="Kp,Ki,ILimit" defaultvalue="2.5,0,50" limits="%BE:0:10,%BE:0:10,"/>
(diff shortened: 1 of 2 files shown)


tracernz:
Need to tweak the buttongroups and plumb the default button still...

tracernz:
Okay, fixups done. I also tweaked the integral limit up to 0.05 (base on autotown db) so it doesn't go red with autotune values.

mluessi:
Nice. The values getting red when autotune is used needs to be addressed for all PID parameters, as it gives the user the impression something is wrong. Or have we done this already?

shared/uavobjectdefinition/stabilizationsettings.xml
```diff
@@ -4,16 +4,16 @@
 	<field name="RollMax" units="degrees" type="uint8" elements="1" defaultvalue="55" limits="%BE:0:180"/>
 	<field name="PitchMax" units="degrees" type="uint8" elements="1" defaultvalue="55" limits="%BE:0:180"/>
 	<field name="YawMax" units="degrees" type="uint8" elements="1" defaultvalue="35" limits="%BE:0:180"/>
-	<field name="ManualRate" units="degrees/sec" type="float" elementnames="Roll,Pitch,Yaw" defaultvalue="150,150,150" limits="%BE:0:500,%BE:0:500,%BE:0:500"/>
+	<field name="ManualRate" units="degrees/sec" type="float" elementnames="Roll,Pitch,Yaw" defaultvalue="150,150,150" limits="%BE:0:1000,%BE:0:1000,%BE:0:1000"/>
```
mluessi: Should we use 1440? It's a nice multiple of 360 and it's also what the OSD menu uses as maximum.

tracernz: Can do. The max possible is 2000*0.85=1700 but increasing the max in the UAVO increases the step size of the slider in the UI.

tracernz:
I didn't see any P parameters higher than 0.01 (the current limit), and there is no limit on the derivative.

kubark42:
Nice work. Would it make sense to have the slider limits be a function of the UAVO settings?

Also, IIRC historically the derivative was kept out of the "Basic" tab because of the propensity for the derivative term to make the model unflyable as well as the fact that there's no convenient range that is correct to an order of magnitude. (The maximum derivative term we can (stably) apply is more than 10 times greater if we vibration isolate our IMU.)

shared/uavobjectdefinition/stabilizationsettings.xml
jhitesma: I think the big issue is that the red looks like a big scary error but isn't well documented.  Maybe instead of changing the point at which it triggers we should just add a more obvious explanation of WHY it's being flagged.  Seems like it's good functionality, just poorly explained.

jhitesma:
Love it, finally stick scaling on the basic tab!

tracernz:
@kubark42 The limits already come from the UAVO, the functionality provided by the UAVO->widget relation framework is very nice!

The main reason I wanted to get the derivative gain on there was because we anticipate more users coming over from Multiwii derivatives (sorry :-P) and they will be very confused by either no derivative gain, or a derivative gain with a magnitude of approximately e-5. I have resized the spinboxes so they can handle extra digits without clipping.

kubark42:
Cool. No objections to it moving. Just thought I'd share some past.

tracernz:
Appreciated, it's always better to understand the reasons for things being done a certain way in the past. That kind of information often gets lost, leading us to re-invent the wheel.

tracernz:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[shared/uavobjectdefinition/stabilizationsettings.xml, line 7 \[r2\]](https://reviewable.io:443/reviews/d-ronin/dronin/201#-K54rd2xidv9SLOEpR9--r2-7:-K5Ah7TbN9BYTvHX8lGd:-2110686521) ([raw file](https://github.com/d-ronin/dronin/blob/b68b31ac146e4c5b09ede9b2c5b403738fa7c686/shared/uavobjectdefinition/stabilizationsettings.xml#L7)):*
In the case of the rate though, it really just sets the highest value you can drag the slider to.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/201)*

tracernz:
Bumped the max roll rate to 1440 deg/sec as suggested. I'd also like to look at upping the default rate of 150 deg/sec at some point. I've found it to be dangerously low in the past.

mlyle:
Looks great-- I like the values you've selected for the derivative range, too-- fairly small derivatives look like moderate sized numbers (44) to me

mlyle:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/201)*

mlyle:
:lgtm_strong: LGTM

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/201#-:-K5CQ45Q2l2EHHmaTPpa:88857767)*

PR #210, opened 2015-12-10
GCS: Branding TX wizard graphic

VArcht:
Replaced Tau Labs logo with dRonin style.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/210)


VArcht:
tx

tracernz:
Looks great! It might be a good idea to pull the branding from a separate file in the ./branding subdirectory though?

VArcht:
I like the idea of breaking out the branding. Not trivial though and still a lot of branding tasks that are more important, maybe best to wait till it is updated to QML?

VArcht:
Oh, forgot this hadn't landed yet, Ok to combine?

VArcht:
screen_setup_wizard

mluessi:
LGTM

tracernz:
I prefer the green of the existing setup wizard button, it guides new users in the right direction and makes the page a little less drab with some colour.

jhitesma:
Wasn't a fan of the old one...but agree that the big friendly green button did make it obvious where a noob should click first.  I do like the new design, but I think the green (or better yet a nicer green) was a good thing.

VArcht:
I don't think it being green helped much, read several comments of people missing it. It would be better to have it pulse or a balloon on a time saying "Start Here" or something.

jhitesma:
I'm not a fan of animation...but it would help draw attention.

I think the big problem is the old design didn't look like a button.  In fact it was a month or so after I started using Tau that I realized those things were buttons.  I thought they were just decoration or "advertising" for the various features.

tracernz:
I don't think the new one addresses that, infact it looks the same as the static logo on the left side.

jhitesma:
Yeah, I should have been more clear and agree totally.  I was hoping to inspire varcht to come up with some fancy graphics that look more buttony.  I'm not a fan of glossy ios looking buttons but maybe a bit of a frame/border and some shading would make it more obvious http://www.clker.com/clipart-245700.html  (I know we have the rollover...but you have to actually mouse over it to see that.)

DTFUHF:
I didn't know these were buttons until today.

VArcht:
We all know the ui needs lots of help but it is not going to get done before the January release. This wizard button looks better than the existing one and does not make a significant difference in UX. There is still a lot that should get done for January and I have a lot of ideas for improvements will be open for discussion after that but meh, not if minor things are like this are going to be a struggle...

jhitesma:
I'm perfectly fine with it merging as it is.  I didn't mean to come across as implying we should iterate over something this minor until it's perfect.  Like you say there's a lot that's going to be changed abut the GCS and I know there's even been talk about doing something different than a wizard.  I was just trying to give constructive criticism based on my experiences.  

This is a an improvement and I'm all for going with it.  But I'm also interested in discussion options for future improvements.  My apologies if this was the wrong place/time to do that or if my feedback was taken other than how I intended it.

@VArcht you've been rocking and I love what you're doing.  The last thing I want to do is make it more difficult for you to keep rocking or discourage you from continuing to bless us with your skills.

mlyle:
@VArcht I really appreciate all the work you're doing and I think this is an improvement.

I think we need to come up with a plan for what we're doing with UI/UX in the near term and how high the bar is and how much we should be planning to iterate.

Basically we have a choice for this short release (and probably next one, too) of making a few high profile things much better or quick haphazard (I don't mean this in a bad way) things across the whole UI that are each small improvements but don't perfect any specific mechanism.

There's good arguments to be made for either approach.  Measured by the first criteria, this PR is troublesome-- it looks good but slightly deepens a persistent UX issue.  Then again, almost anything done to the welcome screen falls short because we know we need to overhaul the initial user experience.  Measured by the second it's great-- it looks a little better and if it's one of many things we do it'll make the UI feel nicer to new users.

mluessi:
+1 for merging. I think it's better than what we currently have but we need to come up with a plan to drastically improve the user experience of the GCS in the long term.

VArcht:
Sure,  but at worse this is a net zero and the work is already done, silly place to draw a red line.

PR #221, opened 2015-12-10
Remove unused FaultSettings UAVO

tracernz:
Not used anywhere, IIRC the firmware code was CC only and removed some time ago.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/221)


mlyle:
The Fault module is still in tree-- we could remove it as part of the same PR...

I plan to ultimately use Fault in DiscoveryF4, but it's OK if it leaves the tree in the meantime.  Your call.

tracernz:
So it is, although unused. Not sure why I didn't catch it with grep, maybe a typo. We can leave it if you're going to use it.

PR #225, opened 2015-12-10
manualcontrolsettings: don't put "manual" in defaults

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/225)

### DIFF
--- a/shared/uavobjectdefinition/manualcontrolsettings.xml
+++ b/shared/uavobjectdefinition/manualcontrolsettings.xml
@@ -129,7 +129,7 @@
 
 		<!-- Note these options values should be identical to those defined in FlightMode -->
 		<field name="FlightModeNumber" units="" type="uint8" elements="1" defaultvalue="3"/>
-		<field name="FlightModePosition" units="" type="enum" elements="6" options="Manual,Acro,Leveling,MWRate,Horizon,AxisLock,VirtualBar,Stabilized1,Stabilized2,Stabilized3,Autotune,AltitudeHold,PositionHold,ReturnToHome,PathPlanner,TabletControl,AcroPlus" defaultvalue="Manual,Stabilized1,Stabilized2,Stabilized3,ReturnToHome,PositionHold"/>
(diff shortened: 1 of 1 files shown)


jhitesma:
Looks good to me, but are Stabilized 1-3 the best top defaults?  Wouldn't it make more sense to set predefined modes like:  Leveling, Horizon, Acro or something instead of the Stabilized modes as defaults?

When I first tried Tau the Stabilized modes confused (but entertained) me.  Trying to figure out why someone may want to set different modes on roll and pitch.  It did get me thinking about other use cases for the firmware....but they just don't strike me as good defaults.

tracernz:
Yeah, I think Stabilized are actually more advanced/flexible modes that are aimed more at power-users, we should make the default basic flight modes.

mlyle:
Should I duplicate leveling/leveling/leveling then?  Because 3 position switches are common and I want to be ... unsurprising if someone misses this.

tracernz:
Sounds reasonable.

jhitesma:
Sounds boring...but like the safest option.

tracernz:
Will leave this another day to give anybody else a chance to comment since it's a change to defaults.

mluessi:
I usually start in rate mode then run autotune etc. Is it safe to assume that leveling with default PIDs works OK for most setups? I.e., are we recommending switching to AT from leveling mode?

mlyle:
Autotune effectively is an excited leveling mode, so if you can fly AT you can fly leveling by definition.

(Rate is a "safer" starting point from a controls perspective, but not everyone is going to be able to fly rate LoS-- so I figured this is a better default).

jhitesma:
I also usually start in rate mode, and by default *flight starts in rate.  But one of the questions I see over and over and over about *flight is "It flies...but how to I get it to autolevel?"  So as long as our default PID's are safe for leveling I'm thinking it's a good default.  (though part of likes just dumping people into rate and encouraging them to sink or swim with it.)

I always switch to a leveling mode (usually horizon) before switching to AT just as a sanity check, pretty sure I read somewhere in the AT stuff that it's recommended to switch in from a leveling mode.

I'd suggest Leveling, Horizon, Rate as a default - but there is so much variety in how people may have their mode switch setup I think it's safest to just pick one mode and set the 3 top defaults to be the same.

Which is just a long way of saying I'm in favor of leveling for the defaults.

pug398:
I'm for leveling on all as default too. Rate can give a first timer a big surprise.

PR #247, opened 2015-12-12
Unbreak release build

tracernz:
The USBInfo fix isn't my favourite work. There doesn't seem to be any way to init all the fields of a struct to a default value without doing so explicitly(and thus writing code that breaks every time the struct definition changes) with GCC & C++11?

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/247)

### DIFF
--- a/ground/gcs/src/plugins/coreplugin/iboardtype.h
+++ b/ground/gcs/src/plugins/coreplugin/iboardtype.h
@@ -57,6 +57,14 @@ public:
      *       is relevant.
      */
     struct USBInfo {
+        USBInfo() : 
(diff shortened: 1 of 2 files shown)


mlyle:
:lgtm: 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/247#-:-K5JQyf9ldh4egHohUWy:980077423)*

mlyle:
Per @tracernz-- revert because #253 does this more nicely.

Issue #227, opened 2015-12-10
Sparky reboots on inputs setup


jhitesma:
Using recent builds of next I'm getting reboots on my sparky board when setting up inputs with a PPM RX.

I've tried two different RX's (though admittedly both homemade FlySky RX's so I'm not 100% sure there's not a hardware issue going on.) and tried powering the RX both through the sparky and directly off the BEC.  

Either way my board is rebooting near the end of the inputs wizard (between the center the sticks step to just after the confirm failsafe step) or if I try manual setup it's rebooting almost as soon as I try to change anything.

mlyle:
@jhitesma and @dustin have both seen various sparky1 anomalies

* This
* Freeze on attempt to arm
* On landing/disarm, a reset

I don't know why sparky1 would be different, and there's other possible explanations for each of these, but the evidence of badness is adding up.

jhitesma:
Ok, I can cause this to happen really easily now :)  Starting Manual config does it every time with a PPM RX setup and working.  If the RX is disconnected or not powered - no crash.  RX powered - insta-crash on manual calibration.  

Reboot Cause is independent watchdog timer

I watched TaskInfo-Stack->StackRemaining and didn't see anything jump...but it crashed so quick something could have changed before I got an update.  ManualControl was at 240...it is at 464 after a boot and before I enable manual calibration.  After the crash it stays at 240.

SystemSats I'm seeing the following right after boot and no change at or after crash (just the Flight Time resets)
HeapRemaining 14160 
CPULoad 42
IRQStackRemaining 512

mlyle:
Can you please screenshot stackremaining just at baseline, before manual calibration? (but with RX configured)

mlyle:
This seems to happen on Sparky1 with PPM and input calibration wizard.  With the event system changes reverted the functionality works.  This is the stacks at the end of that:

![](https://usercontent.irccloud-cdn.com/file/a0Onu7gh/Screenshot%20from%202015-12-10%2023%3A25%3A37.png)

So telemetry is very close.  But turning up telemetry stack size before didn't seem to help.

jhitesma:
Just to clarify it happens with both the wizard and the manual setup.  In fact the manual setup it happens almost immediately every time.  With the wizard it happens at variable points - sometimes as soon as the center all sticks page, sometimes not until the correct backwards channels page.  

Seems like it's happening when the GCS starts reading the stick positions faster is my impression.

mlyle:
[...]

PR #248, opened 2015-12-12
uavobjectmanager: correct some bugs

mlyle:
- Reentrancy requires things that can change under the course of a loop
  to be volatile
- The logic to check num_pending before checking for a duplicate event
  type is obsoleted

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/248)


mlyle:
could relate to #227

jhitesma:
So I'm no longer reproducing #227 with this.  But @tracernz said he was still repoing it.

mlyle:
@jihlein if you can easily try this i'd really appreciate it.

mlyle:
Nevermind.  Problem came back for @jhitesma

jihlein:
Was going to try anyway, but after rebasing to the current next, my build environment is no longer working.  I can't build a working GCS in the Windows environment.

mlyle:
What error do you get?  

There's two things I can think of in recent changes that are relevant:

* The build environment only looks inside tools/ for arm compilers and no longer searches for system installations.
* The build environment doesn't look on its own for a QT install, expecting you to be running from an environment with Qt already in the path-- as described here -- https://github.com/d-ronin/dRonin/wiki/Development-Windows-setup :

```
Start a shell using the "Git bash" program shortcut in the right-click context menu.

... 

Type this command to install a bash_profile suitable for developing the project: cp dronin/make/winx86/bash_profile ~/.bash_profile. Next, ensure that the bash_profile is executable by typing chmod 755 ~/.bash_profile . You may have to edit bash_profile if you have changed any installation paths from the default. Then exit from the terminal and start a new "Git bash" instance.
```

The old way of sh.cmd is deprecated because it resulted in a lot of drift between everyones' build environments that sometimes caused problems, this is easier to set up for new users, and it gives a real terminal.

mlyle:
P.S. Sorry this broke-- I'm coalescing a lot of things and a lot of legacy from the build instructions.

jihlein:
Okay, I can build again, but either due to problems with the environment, or this patch, things are not working.

First, the even system WCA box, while normally green, occasionally flashes yellow.  Second, I the boxes for configuration wizard and manual calibration are greyed out so there is no way to attempt calibration of the PPM receiver.  I'm going to back out this patch as see where that leads.

mlyle:
Really strange.  Others have experienced this not fixing anything, but not helping anything (except @jhitesma got through the wizard with this once).

I will have PPM receivers in hand Tues (want to avoid dismantling the couple things I have PPM on for testing).

jihlein:
This GCS build is a little odd too.  The map display initially comes up, but after halting-flashing-booting the target, the map is blank and won't recover without a GCS restart.  The GCS also take a long time to establish communication with the target.  Files with a .bin extension have a black icon when viewed thru GCS windows.

After reverting the patch, the GCS connects and generally responds better, and the input calibration icons are active.  All other issues remain.  I'm going to delete my build folder and rebuild the whole thing to see if that changes anything.

jihlein:
Rebuilt after deleting build folder.  Still getting solid black icons for bin files, but map display seems stable.  Implemented this patch, reloaded on AQ32, and was able to complete PPM Rx setup wizard.  Flight software seems stable, but I still get the feeling the GCS is not quite right.  Applying and saving settings, on many of the pages, results in a red circle with an X in it, but the changed objects have saved correctly.  Will reset board to default settings and try complete setup from scratch tomorrow.

mlyle:
Doesn't fix key bug, and I think this should get cleaned up in various ways.  I'd still like this change eventually, but marked WIP, moved post-renatus.

mlyle:
I read this again,  Further reflection makes me think this is a bad idea.  withdrawing.

PR #254, opened 2015-12-12
gcs: performance/correctness: use const in foreach

mlyle:
Don't copy UAVO information so much.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/254)

### DIFF
--- a/ground/gcs/src/plugins/uavobjects/uavobjectfield.cpp
+++ b/ground/gcs/src/plugins/uavobjects/uavobjectfield.cpp
@@ -115,10 +115,10 @@ void UAVObjectField::limitsInitialize(const QString &limits)
         return;
     QStringList stringPerElement=limits.split(",");
     quint32 index=0;
-    foreach (QString str, stringPerElement) {
+    foreach (const QString &str, stringPerElement) {
         QStringList ruleList=str.split(";");
         QList<LimitStruct> limitList;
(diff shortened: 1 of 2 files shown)


mlyle:
It seems like some of this could end up copying the entire data structure of UAVOs for most UAVO accesses.  Also puts the allocator through a lot of work.

mlyle:
Makes the GCS noticeably snappier in test.

PTDreamer:
This is great but unexpected. Most of the classes handled are shared implicitly which means that a new object is only created if written to.
http://doc.qt.io/qt-5/implicit-sharing.html

mlyle:
While not all contents of the object might not get copied if it's a Q container or other implicitly shared object, there's still a shim object allocated/constructed for each one if it's not handled with a const reference.  We have 120 UAVOs, so each Get triggers this about 60 alloc/constructors/deletes.

tracernz:
Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/254)*

tracernz:
:lgtm_strong: Looks good, works well.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/254#-:-K5O4pOOKhSH8D_-cGr5:1207925138)*

PR #262, opened 2015-12-13
Fix maximum yaw_scale to match the tested value

droidicus:
Fix maximum yaw_scale to match the tested value

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/262)

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -558,7 +558,7 @@ static void stabilizationTask(void* parameters)
 						if (pitch_scale > 0.25f)
 							pitch_scale = 0.25f;
 						if (yaw_scale > 0.25f)
-							yaw_scale = 0.2f;
+							yaw_scale = 0.25f;
 
 						switch(ident_iteration & 0x07) {
 							case 0:


mluessi:
LGTM. I assume it's safe to merge this (?).

mlyle:
It'd be nice if it could see one third-party flight test, because I can't rule out that "bouncing" off the edge of the scale has some useful characteristic.  Or a good argument from plots that we get nowhere close to that on normal runs.

mlyle:
@droidicus do you have a few plots of yaw_scale from past flights?

droidicus:
Here is a typical plot from one of Fujin's runs on his 225mm race quad:
image

This would cause an increase in actuation during the initial convergence (0.2 -> 0.25), but no difference after ~7 seconds. I do not think that the step-decrease when going from a 0.25 to 0.251 is intentional, there doesn't seem to be a rational reason to have a discontinuity in the actuation curve.

mlyle:
I don't understand the plot.  Isn't it supposed to be limited to 0.2/0.25?  How'd it get up to 1.3?

mlyle:
Ah, this is just on the stabilization side.  I hadn't read this before.  No big deal.

PR #267, opened 2015-12-15
sparky: get serial telemetry working again

mlyle:
Fix regression spotted by @jhitesma that occurred in
taulabs/taulabs#1982 .

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/267)

### DIFF
--- a/flight/targets/sparky/board-info/pios_board.h
+++ b/flight/targets/sparky/board-info/pios_board.h
@@ -106,7 +106,7 @@ extern uint32_t pios_i2c_flexi_id;
 //
 // See also pios_board.c
 //-------------------------
-extern uintptr_t pios_com_telem_rf_id;
+extern uintptr_t pios_com_telem_serial_id;
 extern uintptr_t pios_com_gps_id;
 extern uintptr_t pios_com_telem_usb_id;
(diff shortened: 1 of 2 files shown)


jhitesma:
Able to get a connection with this fix.  But RX only.  PFD responds but never get a link.

mluessi:
What @jhitesma is observing sounds like TauLabs/TauLabs#1418. Can you try increasing the `.bInterval` (e.g. 8ms) and see if it fixes things? Using 1ms on F4 targets was working fine for me but stopped working when I updated my laptop to a 4.X Linux kernel. Users seeing this problem on Windows can sometimes fix it by switching to the "high performance profile".

mluessi:
Wait, this is for serial telemetry, not USB. Maybe we have a race condition in the telemetry module or something?

mluessi:
@mlyle how does this fix things? Isn't it just renaming a variable?

mlyle:
@mluessi PIOS_HAL directly writes to pios_com_telem_serial_id for serial telemetry-- rf_id is for actual RF links.  Various older targets abused this, so this is part of the necessary change for PIOS_HAL.

PR #272, opened 2015-12-15
flight: brain: make CDC serial telem work/obey hw settings

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/272)


mluessi:
Oh wow.. it seems like no one ever used the other USB modes. Thanks!

Issue #68, opened 2015-11-28
TX wizard: make robust to small movements/jitter


mlyle:

jhitesma:
This actually worked much better for me this time than the last time I tried doing it about 6-9 months ago.  However I did still experience some issues (Using 20150928 Brain release) with channels being mis-recognized when running the input wizard the first time. 

Thought that could be related to issue #69 since it worked after I changed the port from PPM to PPM+Outputs and re-ran the wizard

Video for reference: https://youtu.be/NTS6S9icEVI?t=4m21s  (same vid as in issue #69)

jhitesma:
And just flashed my Naze and am having a hell of a time setting up the inputs because of this.  Same basic RX I used on my Brain a few days ago (DIY FlySky PPM which is showing about 2-3 jitter on the signals)

wizard totally failed - detected things before I could touch anything.

And manual setup was impossible until I set the deadband to 3.0...even then it was jumpy despite the values only shifting 2-3.

jhitesma:
Just to update on this I determined that the RX I was using was having issues which were making the problems more pronounced.  

However even on my other RX's that aren't randomly cutting out but still have 2-3ms of jitter I still have issues with things like channels being "detected" in the setup wizard before I touch anything and with screens in my OSD switching even though I don't touch the switch I have assigned to do that.

PR #273, opened 2015-12-16
flight: run config_check asynchronously

mlyle:
This is just too expensive.  Input wizard can create wildly toggling
modes-- both from transmitter and GCS object changes-- that back up the
event system.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/273)

### DIFF
--- a/flight/Modules/System/systemmod.c
+++ b/flight/Modules/System/systemmod.c
@@ -110,6 +110,8 @@ static uint32_t idleCounterClear;
 static struct pios_thread *systemTaskHandle;
 static struct pios_queue *objectPersistenceQueue;
 
+static volatile bool config_check_needed;
+
 // Private functions
 static void systemPeriodicCb(UAVObjEvent *ev, void *ctx, void *obj_data, int len);
 static void objectUpdatedCb(UAVObjEvent * ev, void *ctx, void *obj, int len);
@@ -280,6 +282,13 @@ static void systemPeriodicCb(UAVObjEvent *ev, void *ctx, void *obj_data, int len
 	// Update the modem status, if present
 	updateRfm22bStats();
 
+#ifndef NO_SENSORS
+	if (config_check_needed) {
+		configuration_check();
+		config_check_needed = false;
+	}
+#endif
+
 #ifndef PIPXTREME
 	// Update the system statistics
 	updateStats();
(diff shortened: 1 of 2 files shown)


mlyle:
Believed to solve #227 .

jhitesma:
I'm no longer able to repo #227 with this applied

flight/Modules/System/systemmod.c
```diff
@@ -280,6 +282,14 @@ static void systemPeriodicCb(UAVObjEvent *ev, void *ctx, void *obj_data, int len
 	// Update the modem status, if present
 	updateRfm22bStats();
 
+#ifndef NO_SENSORS
+	if (config_check_needed) {
+		configuration_check();
+	}
+
+	config_check_needed = false;
```
mluessi: should be inside if

mluessi:
Interesting.. I will test this on brain later today. I think a better fix would be to find a way to prevent the rapidly switching modes (in addition to this).

mluessi:
Confirmed no crash on brain :)

jhitesma:
I speculated that the rapid mode switching may be related to issue #68 but @mlyle seemed to feel it was unrelated.  I did test adding a fairly large deadband but that didn't help (and again @mlyle explained that switches are dealt with differently than flight controls for deadband.)

Here's what he said in IRC last night about the rapid mode switches:
10:04 PM <icee> not just the jumpy switch thing
10:04 PM <icee> we go and hear the changes
10:04 PM <icee> and reprogram manualcontrolsettings in the gcs
10:04 PM <icee> which then can make them change again
10:04 PM <icee> way faster than actual transmitter frames can do
10:04 PM <icee> i believe.
10:05 PM <icee> now i'm systematically neutering things relating to changing flightmode

However @mluessi - this also reminds me of a similar issue I've had with my brain that I've been meaning to tell you about.  I can't use multiple pages in my OSD because the 2-3ms jitter from my TX is enough to trigger a change between them even if I don't touch the switch.  

I should probably just open a new issue about switches...

flight/Modules/System/systemmod.c
mlyle: Thanks-- I'm used to doing it outside but without synchronization it's really important to get it in.

PR #279, opened 2015-12-16
vtolpathfollower: immediately (slowly) climb to target alt

### LINKED ISSUE #157: Sparky2:RTH sequencer
Perhaps a little more work can be done on the RTH sequencer. Specifically the pause and climb if below minimum altitude. How about this. Engage RTH. Pause for 1 second. Climb at a sane velocity to minimum altitude if not there already. Pause additional time (some felt 10 seconds too long) and then start return path home. Currently it immediately climbs a few meters, pauses, climbs additional difference in minimum altitude and immediately starts return path. I am afraid to a new user it would appear broken.

mlyle:
And no longer dwell at target alt (For now).

Intended to fix #157

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/279)

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -58,8 +58,9 @@
 
 // Various navigation constants
 const static float RTH_MIN_ALTITUDE = 15.f;  //!< Hover at least 15 m above home */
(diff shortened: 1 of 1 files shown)


mlyle:
Works well in simulator; next step flight test.

pug398:
Tested this and it works well.

mluessi:
Merging based on good flight test by @pug398 Thanks!

PR #282, opened 2015-12-17
gcs: work around different paths in build tree vs. pkg

mlyle:
Without this, Linux and Windows executables work in a package but not directly run from the build tree.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/282)


mlyle:
Rebased;  works for me on Windows 10.

PR #286, opened 2015-12-17
gcs: hide splash when asking what config to use

### LINKED ISSUE #237: Widget to choose a configuration often shows up behind splash screen
When you don't have a config, the dialog that prompts you to create one is often in a bad place.

mlyle:
Otherwise the splash screen can be in front, preventing interaction with the dialog.

Fixes #237 

This required a lot of plumbing.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/286)

### DIFF
--- a/ground/gcs/src/plugins/coreplugin/mainwindow.cpp
+++ b/ground/gcs/src/plugins/coreplugin/mainwindow.cpp
@@ -344,10 +344,17 @@ void MainWindow::extensionsInitialized()
         }
         if(showDialog)
         {
+            // This has often ended up behind the splash screen, which looks
+            // bad.
+            emit hideSplash();
+
             importSettings * dialog=new importSettings(this);
             dialog->loadFiles(directory.absolutePath());
             dialog->exec();
             filename=dialog->choosenConfig();
+
+            emit showSplash();
+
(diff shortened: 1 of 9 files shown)


ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -344,10 +344,17 @@ void MainWindow::extensionsInitialized()
         }
         if(showDialog)
         {
+            // This has often ended up behind the splash screen, which looks
```
mlyle: This is the actual change.  The rest is just pumping signals to slots up the way.

mluessi:
For some reason the GCS crashes on startup (Linux) but it isn't obvious to me what in this branch could be causing it (current next works fine). 

    Failed to load core: Could not find 'Core.pluginspec' in /home/martin/data/brain_fpv/code/dRonin/build/ground/gcs/lib/plugins,/home/martin/data/brain_fpv/code/dRonin/build/ground/gcs/Plugins

mlyle:
@mluessi Ah, sounds like you need #282 -- i've updated the description.

mluessi:
Confirmed to work on Linux with #282

PR #310, opened 2015-12-19
sysalarms: only look at correct uavo field

mlyle:
Fixes #296 

Previously we'd iterate over all the stuff in SystemAlarms instead of just the stuff that represents alarms.

This would trigger on various configuration errors and other things "0 module in warning state".  With Kenn's change to report reset reason, it would always fire on normal powerup.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/310)

PR #285, opened 2015-12-17
gcs: remove disused source files in coreplugin

mlyle:
Remove fileiconprovider, mimedatabase, rightpane, settingsdatabase,
sidebar, tabpositionindicator, variablemanager.

No doubt this saves a few seconds of compile time and some bytes and
some startup time.   Miniscule bits, but there's further fertile
ground for cleanup here.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/285)

PR #311, opened 2015-12-19
gcs: don't load lots of emptygadgets when restoring config

### LINKED ISSUE #297: Lots of plugins show "loading emptygadget" or whatever on splash
This would be nice to run down because if the correct name was shown we can get a better feeling for what is taking time on startup.

mlyle:
Previously, when we were instantiating splitter views, we'd briefly
instantiate an EmptyGadget until we've gone and recursed and restored
that configuration.  Now, in that case, skip doing that as the
appropriate gadget widget will be selected soon.

Fixes #297 and noticeably improves GCS startup time.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/311)

### DIFF
--- a/ground/gcs/src/plugins/coreplugin/uavgadgetmanager/splitterorview.cpp
+++ b/ground/gcs/src/plugins/coreplugin/uavgadgetmanager/splitterorview.cpp
@@ -39,11 +39,11 @@
 using namespace Core;
 using namespace Core::Internal;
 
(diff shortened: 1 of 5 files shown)


mlyle:
Tested scenarios:

* Normal startup
* Startup with invalid gadgets in configuration file
* Splitting and unsplitting views

droidicus:
LGTM, tested on next with PR #310, #311, #285, #312

mluessi:
Tested on Linux, works beautifully. Merging.

PR #312, opened 2015-12-19
configgadgetwidget: defer creating subwidgets for 0.5s

mlyle:
This allows the app to open a couple seconds quicker.  The trade off is,
if you try to go to config gadget widget too fast, you encounter a
couple second delay while we catch up and instantiate all the widgets.
But not everyone will go to config gadget first, and not all will go
there extremely quickly after loading.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/312)

### DIFF
--- a/ground/gcs/src/plugins/config/configgadgetwidget.cpp
+++ b/ground/gcs/src/plugins/config/configgadgetwidget.cpp
@@ -66,7 +66,7 @@ ConfigGadgetWidget::ConfigGadgetWidget(QWidget *parent) : QWidget(parent)
     help = 0;
     chunk = 0;
 
-    QTimer::singleShot(3000, this, SLOT(deferredLoader()));
+    QTimer::singleShot(500, this, SLOT(deferredLoader()));
 }
 
 void ConfigGadgetWidget::deferredLoader()
@@ -96,7 +96,7 @@ void ConfigGadgetWidget::deferredLoader()
     ftw->insertTab(ConfigGadgetWidget::aircraft, qwd, *icon, QString("Vehicle"));
     break;
 
-    	case 3:
+        case 3:
     icon = new QIcon();
     icon->addFile(":/configgadget/images/input_normal.png", QSize(), QIcon::Normal, QIcon::Off);
     icon->addFile(":/configgadget/images/input_selected.png", QSize(), QIcon::Selected, QIcon::Off);
--- a/ground/gcs/src/plugins/config/configgadgetwidget.cpp
+++ b/ground/gcs/src/plugins/config/configgadgetwidget.cpp
@@ -48,7 +48,7 @@
 #include <QTextEdit>
 #include <QVBoxLayout>
 #include <QPushButton>
-
+#include <QTimer>
 
 
 ConfigGadgetWidget::ConfigGadgetWidget(QWidget *parent) : QWidget(parent)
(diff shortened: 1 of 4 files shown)


droidicus:
If I try to go to a config gadget as soon as it is possible to do so, clicking on it doesn't seem to do anything for a few seconds. At first I thought that the program was frozen. That being said, I had to stare at the screen to try and click ASAP.

Other then that LGTM, tested on next with PR #310, #311, #285, #312

mluessi:
I observe the same thing as @droidicus Maybe reducing it to 2s would make it less noticeable.

mlyle:
what would be best is triggering it on the main window opening, but that doesn't seem to be super easy to do.

It's important that the time we "sleep" be enough for everything else to finish loading, else we're just fighting with other stuff trying to load in the event thread.

It's up to you guys whether this is a net win.  I kinda think you need to work to make it to config gadget fast enough / see this.

droidicus:
I personally think that it is a net win, I doubt that I would encounter an issue in normal usage.

tracernz:
I think only developers/testers would notice a second or two faster statup but users will notice if the UI hangs/is unresponsive after startup.

mlyle:
OK, I've made another commit to improve configwidget load times.

This is what it looks like if you rush and win the race before the background loading happens:

https://www.youtube.com/watch?v=fJ6f4I-6V1s

It is not that bad, I don't think.

mlyle:
(Also, I think this is a little moot-- because there's no way that there is a telemetry connection done in this timeframe / the config panes are going to reload and do weird stuff anyways).

tracernz:
That doesn't look too bad, at least it looks like something is happening while it's still loading.

mlyle:
Rebased, set to .5s instead of 3s, because that plays nice enough after the other change.

mluessi:
Awesome, thanks!

ground/gcs/src/plugins/config/configgadgetwidget.cpp
```diff
@@ -63,76 +63,113 @@ ConfigGadgetWidget::ConfigGadgetWidget(QWidget *parent) : QWidget(parent)
...
+	case 11:
     icon = new QIcon();
     icon->addFile(":/configgadget/images/osd_normal.png", QSize(), QIcon::Normal, QIcon::Off);
     icon->addFile(":/configgadget/images/osd_selected.png", QSize(), QIcon::Selected, QIcon::Off);
     qwd = new ConfigOsdWidget(this);
     ftw->insertTab(ConfigGadgetWidget::osd, qwd, *icon, QString("OSD"));
+    break;
 
-    ftw->setCurrentIndex(ConfigGadgetWidget::hardware);
+        case 12:
```
kubark42: Looks like there are some formatting problems in this ```switch()..case```

ground/gcs/src/plugins/config/configgadgetwidget.cpp
```diff
@@ -146,8 +183,24 @@ ConfigGadgetWidget::ConfigGadgetWidget(QWidget *parent) : QWidget(parent)
...
+    QTimer::singleShot(0, this, SLOT(deferredLoader()));
+}
+
+void ConfigGadgetWidget::paintEvent(QPaintEvent *event)
+{
+    (void) event;
+
+    if (chunk < 12) {
+	// jumpstart loading.
+        deferredLoader();
```
kubark42: Is it intentional to have ```deferredLoader()``` twice?

mlyle: Yes.  That way we do 2 steps per chunk in the event loop (or more, if the deferred loading has already started).  This effectively means "hurry up."

Issue #143, opened 2015-12-04
Record Symbols for Breakpad


dustin:
In order to use breakpad, we have to have symbols exported from the build and available for the minidump processor.  As we're having crashes go to autotown, it makes sense to send the symbol tables up as well.  Perhaps this can also lead to analysis automatically occurring there.

dustin:
Putting this comment here since I don't know where else to put it.  :/

https://chromium.googlesource.com/breakpad/breakpad/+/master/docs/getting_started_with_breakpad.md

tracernz:
Yep, if we can keep this nice and simple and get it hooked into Jenkins it would be grand.

dustin:
I don't know what the thing looks like, but if it's just a file, I can make a tool that can be placed on builders along with a file for auth that hands it over.

dustin:
I hope this is easy.  I have a feeling we might regret having ~1/3 of the crash infrastructure, but having the symbols be useless.

tracernz:
We keep the git hash so I think that should allow us to build symbols for each dump.

mlyle:
We need to do it at original build time.  We have no guarantee for
byte-for-byte, address-for-address reproducibility just by having the same
code.

tracernz:
Yeah, but it will give us a shot at using our existing crash-dump data until we get something together to dump symbols from releases. ;-)

mlyle:
@tracernz have you seen this work?  Because I doubt it will work.

tracernz:
Nope, haven't tried. Will look into getting this sorted in the coming week if nobody else is?

mlyle:
If you have time, a recipe I could use as part of the package build to get things would be nice.

mlyle:
Needs appropriate documentation as part of build process and/or hooking it into the build.

mlyle:
Won't be further improving this before samsara; deferring.

mlyle:
This was done for OSX, but Windows still needs to have its path forward determined.

mlyle:
Closed in favor of #290 (more specific bug)

Issue #294, opened 2015-12-18
packaging artifact paths, and directories within archives... inconsistent between platforms


mlyle:

mlyle:
`build/package-preview-20151218/dronin_osx_amd64-preview-20151218.dmg`

mlyle:
`build/flight-preview-20151218.zip`

mlyle:
`build/preview-20151218.zip`

`build/package-preview-20151218/dRonin-preview-20151218-install.exe`

mlyle:
`build/preview-20151218.tar.xz`

mlyle:
depends upon #300

Issue #290, opened 2015-12-17
Windows Breakpad Support


dustin:
We don't have breakpad support for Windows, so crashes on that platform are not reported to us.

tracernz:
Got MSVC build working again this morning and confirmed breakpad works with it. See #1138.

mlyle:
This is all good now-- closin'

PR #327, opened 2015-12-21
packaging and symbols export

PTDreamer:
This had lots of conflicts, the project moves too fast :), I haven't tested this after the rebase.

This PR was supposed to fix windows packaging, but I now see reports that it was already fixed?
It also includes a EXPORT_SYMBOLS=YES make option to use with all ground packaging targets. It only works for linux and windows and produces a directory with the symbol files needed for breakpad. 
EXPORT_SYMBOLS must be run with the DUMP_SYMBOLS_TOOL=path_to_dump_syms, no enforcing is implemented, but should be in the future.

EDIT: Added some documentation https://github.com/d-ronin/dRonin/wiki/Google-Breakpad-(WIP)

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/327)

Edited by mlyle: connects to #143


mlyle:
Marked into renatus for now.  Hopeful, but it's a big change.

mlyle:
Also relates to #290.

It's not really feasible to change our build for Renatus to use MSVC...  There is https://bugs.chromium.org/p/google-breakpad/issues/detail?id=436 as an option if we really wanted to shoehorn crash reporting in for this release, but we're pretty low on time for that.

P.S. The windows packaging fix turned out to have a lot of overlap with the Mac fix I was looking at, so I did both.

package/linux_extract_debug_symbols.sh
```diff
@@ -0,0 +1,42 @@
...
+  ${DUMP_SYMBOLS_TOOL} ${SOURCE_DIR}/${1} > ${debugfile}
+  echo "Striping debug information from $1"
+  strip --strip-debug --strip-unneeded "${SOURCE_DIR}/${1}"
+}
+
+interesting_files=$(cd ${SOURCE_DIR} && find . -executable -type f -not -path "./share/*" )
+for file in $interesting_files; do
+    generateSymbols $file
+done
+moveSymbols
```
mlyle: newline at end of file.

package/Makefile.winx86
```diff
@@ -7,19 +7,19 @@ NSIS_OPTS     := /V3
 NSIS_DIR      := $(ROOT_DIR)/package/winx86
 NSIS_SCRIPT   := $(NSIS_DIR)/dronin.nsi
 NSIS_TEMPLATE := $(NSIS_DIR)/dronin.tpl
-NSIS_HEADER   := $(BUILD_DIR)/ground/gcs/dronin.nsh
+NSIS_HEADER   := $(PACKAGE_DIR)/$(GCS_PACKAGE_NAME)/dronin.nsh
 
-standalone: 
+standalone: gcs
```
mlyle: I've been trying to get away from packaging-related targets depending on build, so that we can have separate build steps and package afterwards.  Your thoughts?

Makefile
```diff
@@ -1013,7 +1013,8 @@ endif
 # Packaging components
 #
 ##############################
-PACKAGE_TARGETS = package_installer package_all_compress
+PACKAGE_TARGETS = package_installer package_ground package_matlab
```
mlyle: Do we need all these?  I think it's unlikely we'll ship matlab separately, for instance...

package/Makefile.winx86
PTDreamer: packaging is tagged with sha, if you don't use that dependency you can be naming package_something_sha to something you built ages ago.

Makefile
PTDreamer: with package_matlab I agree, with others I think they all serve a purpose.

package/Makefile.winx86
mlyle: Yah, I know it will let you do something bad / you need to be careful.  We also have dependency problems in various places in the code, so unless you do a make clean you are not guaranteed a good package build either.

The Jenkins configuration we have guarantees clean artifact builds with matching githash.  It's a much simpler configuration/packaging than before.  I'd rather keep packaging as a separate step that depends on completing a proper build before (so we can use parallelism properly in the earlier steps, etc).

One thing that might be a good idea (I'd be willing to do it) would be to export the githash as a symbol in the GCS executable-- we could assert that it's correct as part of the packaging process.

Makefile
mlyle: OK

package/Makefile.winx86
PTDreamer: I understand what you are saying, but for me, as someone who is building his own code and not jenkins I prefer to have a make target that does everything for me while I take a cup of coffee (and at least @tracernz from IRC seems to share the idea), even if I have to clean the build first. We can have one target with and other without dependencies making everyone happy though.

tracernz: Yeah, I think adding a target that completes all necessary steps would be worthwhile, even if Jenkins doesn't use it.

mlyle: Yah, I have no issue with there being further targets like that.  We could even have it invoke the "packaging halves" in a non-parallel fashion explicitly.

mlyle:
When feedback is touched up and it's rebased ... I can feed to Jenkins to see if output artifacts are OK / compatible with what's there.

PTDreamer:
> When feedback is touched up and it's rebased ... I can feed to Jenkins to see if output artifacts are OK / compatible with what's there.

What does jenkins run and what does it expect? Packages naming seems to have changed with #334 which conflicted badly with this so I'm a bit lost with what works and what doesn't.

mlyle:
http://jar.lyle.org:8080/job/dronin/18/  This is an example Jenkins build

This is the jenkins configuration for the nightly builder, so far: https://gist.github.com/mlyle/c1ba8eefacdb03b89746
Right now it runs on just Windows and Linux but Mac would work with these settings too, I believe.

It basically does (in different steps:

```
git clean -f -d -x -e tools

rm -f tools

if [ &quot;$NODE_NAME&quot; = &quot;winx86&quot; ] ; then
cmd //c mklink //j tools /dronin-tools
else 
ln -s /dronin-tools tools
fi

git reset --hard

# Note-- this is going to copy the flight package to subsidiary builders soon
GCS_BUILD_CONF=release FLIGHT_BUILD_CONF=release make -j2 all package_flight

if [ &quot;$NODE_NAME&quot; = &quot;master&quot; ] ; then
  make all_ut_run
fi

make package_all_compress

make package_installer

mkdir -p build/artifacts
mv build/dronin*.zip build/artifacts || true
mv build/dronin*.xz build/artifacts || true
mv build/dRonin*.exe build/artifacts || true
mv build/dronin*.dmg build/artifacts || true
mv build/dronin*.deb build/artifacts || true
```

mlyle:
P.S. I'm sorry that we've had some conflict on the packaging stuff.  I was about halfway done when your PR showed up and I kinda.. smashed our work together. heh.

PTDreamer:
So one thing that keeps going back and forth is the location of the files produced by the package targets. You seem to want them all to be built under "build" while the code I have been pushing places them one step deeper in the directory path "build/package_label". Although this  seems true there is still targets which write on the deeper directory. 
Also as it is now installer packages seem to depend on the standalone target which was never meant to be used for that. The code I have submitted and was squashed makes the installer depend on the ground_package. One of the difference is that like it is now you should be getting packages full of uneeded stuff like .lib and .exp files on win.
I just want to understand the desired end state, and try to fix things, the packaging code right now is really messed up and probably non functional for a few targets.

mlyle:
> You seem to want them all to be built under "build" while the code I have been pushing places them one step deeper in the directory path "build/package_label". Although this seems true there is still targets which write on the deeper directory. 

I opened an issue #294 because half of the installer/compressed artifacts were in build/ and half were in build/package-* , and then moved them to be consistent in build/ in #334.  If you want them to go the other place that's fine--  I only preferred build/ because it's easier to avoid installers ending up in archives.  If you want to move them to the package directory I can reconfigure Jenkins to add an appropriate additional wildcard to gathering artifacts.

> Also as it is now installer packages seem to depend on the standalone target which was never meant to be used for that. The code I have submitted and was squashed makes the installer depend on the ground_package. One of the difference is that like it is now you should be getting packages full of uneeded stuff like .lib and .exp files on win.

I don't think this happens-- the .zip from last night's build http://jar.lyle.org:8080/job/dronin/19/Nodes=winx86/artifact/build/artifacts/dronin-20151223-07442cdc.zip seems to be OK.   But if the other way is preferred I'm fine with it-- just we need to figure it out soon because we hope to formally release in mid-January.  (And it seems like the packages we are getting right now are fine).

mlyle:
This isn't expected to work, because of the conflict.  I'm curious what it will do.  build artifacts please

mlyle:
build artifacts please

mlyle:
I hate jenkins.  It selects the wrong branch when there's no appropriate merge.  build artifacts please.

mlyle:
last try.  build artifacts please

Makefile
mlyle: I just tried this.. package_all_compress no longer works at toplevel so Jenkins is not happy.

mlyle:

package/osx_extract_debug_symbols.sh
```diff
@@ -0,0 +1,48 @@
...
+  filename=$(basename "$1")
+  debugfile="$debugdir/$filename.sym"
+  if [ ! -d "${debugdir}" ] ; then
+    echo "creating dir ${debugdir}"
+    mkdir -p "${debugdir}"
+  fi
+  echo "Generating symbols for $1, putting them in ${debugfile}"
+  ${DUMP_SYMBOLS_TOOL} ${SOURCE_DIR}/${1} > ${debugfile}
+  echo "Striping debug information from $1"
+  strip -S "${SOURCE_DIR}/${1}"
```
PTDreamer: OSX binaries seem to be too big when symbols are generated. I'm not familiar with "strip" on OSX so I'm not sure what to use here.

mlyle:
I will read this (much) later when kids are asleep :)

mlyle:
Review status: 0 of 19 files reviewed at latest revision, 6 unresolved discussions.

---

*[package/osx_extract_debug_symbols.sh, line 35 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K74HY-IptWMwh0cHQ1B-r3-35:-K7CsMqCtZJTiPnyRbzC:36149131) ([raw file](https://github.com/d-ronin/dronin/blob/a96fcf70fbd38e0eb9ba2bf7af146e12b2692e0a/package/osx_extract_debug_symbols.sh#L35)):*
Yah.  I don't know offhand what is correct-- stripping libraries is already weird.    strip -S on OSX is supposed to do this.

---

*[package/osx_extract_debug_symbols.sh, line 44 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtOyAuu40xXHaRvSn:-K7CtOyAuu40xXHaRvSo:-293002909) ([raw file](https://github.com/d-ronin/dronin/blob/a96fcf70fbd38e0eb9ba2bf7af146e12b2692e0a/package/osx_extract_debug_symbols.sh#L44)):*
gfind?  I suggest find -perm -0100 or similar to not create additional dependencies

---

*[package/winx86/dronin.nsi, line 207 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtwDfs2Qqr5Y17Nc7:-K7CtwDfs2Qqr5Y17Nc8:-1358885893) ([raw file](https://github.com/d-ronin/dronin/blob/a6dc1e80d2d045932b09c5dc5a2ed180f191f02b/package/winx86/dronin.nsi#L207)):*
This is good stuff to remove.

---

*[package/winx86/resources/dronin.ico, line 0 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7Ctnhg0wkY43gyjgLm:-K7Ctnhg0wkY43gyjgLn:-1546044007) ([raw file](https://github.com/d-ronin/dronin/blob/a96fcf70fbd38e0eb9ba2bf7af146e12b2692e0a/package/winx86/resources/dronin.ico#L0)):*
What's the change here?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/327)*

PTDreamer:
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions.

---

*[package/osx_extract_debug_symbols.sh, line 44 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtOyAuu40xXHaRvSn:-K7DcjwGfA7qNcHTZmcL:5173665) ([raw file](https://github.com/d-ronin/dronin/blob/a96fcf70fbd38e0eb9ba2bf7af146e12b2692e0a/package/osx_extract_debug_symbols.sh#L44)):*
OK

---

*[package/winx86/dronin.nsi, line 207 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtwDfs2Qqr5Y17Nc7:-K7DcytS6GvlultFtjjQ:1461413556) ([raw file](https://github.com/d-ronin/dronin/blob/a6dc1e80d2d045932b09c5dc5a2ed180f191f02b/package/winx86/dronin.nsi#L207)):*
The OSX has similar copy, I can remove it although I'm not sure what is the intention regarding FW release distro.

---

*[package/winx86/resources/dronin.ico, line 0 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7Ctnhg0wkY43gyjgLm:-K7DdLYsX_HUi7q_FISJ:218054867) ([raw file](https://github.com/d-ronin/dronin/blob/a96fcf70fbd38e0eb9ba2bf7af146e12b2692e0a/package/winx86/resources/dronin.ico#L0)):*
This was moved to the branding dir.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/327)*

mlyle:
Review status: 0 of 19 files reviewed at latest revision, 1 unresolved discussion.

---

*[package/osx_extract_debug_symbols.sh, line 44 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtOyAuu40xXHaRvSn:-K7Dg0E34YIekzen-HBh:-2137399250) ([raw file](https://github.com/d-ronin/dronin/blob/a96fcf70fbd38e0eb9ba2bf7af146e12b2692e0a/package/osx_extract_debug_symbols.sh#L44)):*
Thanks

---

*[package/winx86/dronin.nsi, line 207 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtwDfs2Qqr5Y17Nc7:-K7Dfnt4wo5vgdjXM-_5:-1371253819) ([raw file](https://github.com/d-ronin/dronin/blob/a6dc1e80d2d045932b09c5dc5a2ed180f191f02b/package/winx86/dronin.nsi#L207)):*
I don't really see the point in shipping the matlab stuff in packages--- if people want to do developer-stuff they should just check out the source.

---

*[package/winx86/resources/dronin.ico, line 0 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7Ctnhg0wkY43gyjgLm:-K7DfkS80QjpGVFpmdoa:1462820874) ([raw file](https://github.com/d-ronin/dronin/blob/a96fcf70fbd38e0eb9ba2bf7af146e12b2692e0a/package/winx86/resources/dronin.ico#L0)):*
I see.  Github /reviewable don't show this very well.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/327)*

mlyle:
I need to really run this.  Sorry I've not gotten to it yet.

mlyle:
Review status: 0 of 19 files reviewed at latest revision, 2 unresolved discussions.

---

*[package/Makefile.osx, line 24 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7DjYOsvQEu1Miel0Bp:-K7DjYOsvQEu1Miel0Bq:-1011947724) ([raw file](https://github.com/d-ronin/dronin/blob/a25a7bdd47458b5e7d4fee7b405c0d90dedfa6a8/package/Makefile.osx#L24)):*
Tested on OSX-- looks good.  Only complaint is this-- I believe this ends up changing the package name on OSX to include gcs_...

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/327)*

tracernz:
The MATLAB stuff can be useful to people doing stuff with pico-c while running a release. They would avoid having to learn git etc. (people who don't git find it very confusing)

mlyle:
The windows installer is missing from the list of above artifacts-- is the path still different?

mlyle:
Deb is missing too

mlyle:
The actual portable packages work OK

PTDreamer:
Installers are built with package_installer, is that what you have tried? I believe currently they aren't built with package_all either 

---

Review status: 0 of 19 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/327#-:-K7DxXVoyU6vTksdUlWV:562901231)*

PTDreamer:
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions.

---

*[package/Makefile.osx, line 24 \[r4\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7DjYOsvQEu1Miel0Bp:-K7DyfeOr6RNet9X8T-b:1570720649) ([raw file](https://github.com/d-ronin/dronin/blob/a25a7bdd47458b5e7d4fee7b405c0d90dedfa6a8/package/Makefile.osx#L24)):*
It does

---

*[package/winx86/dronin.nsi, line 207 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtwDfs2Qqr5Y17Nc7:-K7DywH_5pSC01rsoFkh:-823017919) ([raw file](https://github.com/d-ronin/dronin/blob/a6dc1e80d2d045932b09c5dc5a2ed180f191f02b/package/winx86/dronin.nsi#L207)):*
Hum, I'm lost, the diff I'm looking at above is about the firmware packaging together with the installer.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/327)*

PTDreamer:
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions.

---

*[package/winx86/dronin.nsi, line 207 \[r3\]](https://reviewable.io:443/reviews/d-ronin/dronin/327#-K7CtwDfs2Qqr5Y17Nc7:-K7DzZV3ZFPLirZk1S3v:-436982013) ([raw file](https://github.com/d-ronin/dronin/blob/a6dc1e80d2d045932b09c5dc5a2ed180f191f02b/package/winx86/dronin.nsi#L207)):*
Ok now I see how this thing works and this dialog makes more sense. Yes the matlab packages aren't copied anymore and there is no package_matlab anymore

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/327)*

mlyle:
jenkins build artifacts please

mlyle:
I'll change jenkins to look in the package path for the exe / deb, even though I don't really like it that way compared to the build path.

Can the name of the OSX package change to not be gcs-* but instead dRonin-*?  Then this could merge.

mlyle:
jenkins build artifacts please

mlyle:
jenkins build artifacts please

PR #330, opened 2015-12-22
Sparky: Add XTAL & Baro to BoM

tracernz:
As per title, details were missing. Also noted that the USB termination resistors are only present on rev. 1.0.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/330)


tracernz:
Turns out the F3 includes USB matching impedance internally so the external series termination resistors are unnecessary. :)

mluessi:
Looks like you forgot to fill in the "Supplier Subtotal" column

tracernz:
Oops, fixed, thx. :)

mluessi:
Thanks!

PR #342, opened 2015-12-24
pios_servo/oneshot: Ensure dead time.

### LINKED ISSUE #337: Need minimum deadtime on syncpwm output
re: issue #323 

My proposed fix is, around here:

```
        if (output_channel_mode[servo] == SYNC_PWM_TRUE) {
                TIM_Cmd(chan->timer, DISABLE);
        }
```

To ensure that the count value is past max+10us.  This implicitly assumes the max values for the things in the same bank are the same.

It's super unfortunate that we're not using hardware oneshot/single count functionality which I think exists on all of our targets.

mlyle:
Now max pulse length is passed into pios_servo, and we make sure we're
"past" that point by a defined margin before we'll disable the timer.

pios_servo should get factored common in the future and non-HPWM
variants of the code should get removed in that refactor.

This depends on all channels in a oneshot bank being defined with the
same "max" time to be fully protective.

Fixes #337

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/342)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -562,15 +562,16 @@ static bool set_channel(uint8_t mixer_channel, float value)
 			}
 		}
 		PIOS_Servo_Set(mixer_channel,
-					buzzOn ? actuatorSettings.ChannelMax[mixer_channel] : actuatorSettings.ChannelMin[mixer_channel]);
+					buzzOn ? actuatorSettings.ChannelMax[mixer_channel] : actuatorSettings.ChannelMin[mixer_channel],
+					actuatorSettings.ChannelMax[mixer_channel]);
 		return true;
 	}
 	case ACTUATORSETTINGS_CHANNELTYPE_PWM:
 #if defined(PIOS_INCLUDE_HPWM)
 		// The HPWM method will convert from us to the appropriate settings
-		PIOS_Servo_Set(mixer_channel, value);
+		PIOS_Servo_Set(mixer_channel, value, actuatorSettings.ChannelMax[mixer_channel]);
 #else
-		PIOS_Servo_Set(mixer_channel, value);
+		PIOS_Servo_Set(mixer_channel, value, actuatorSettings.ChannelMax[mixer_channel]);
 #endif
 		return true;
 	default:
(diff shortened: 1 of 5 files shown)


mlyle:
Status: untested, compiles

mlyle:
I updated this to remove another kind of glitch the previous code allowed-- don't update the timer if we didn't disable it.

droidicus:
Bench tested, no scope available right now, eliminated the motor ticks that I get with the current next.

jihlein:
I'll try to get a before and after logic analyzer trace this afternoon.  Any specific setup you'd like to see for continuity with your tests?

mlyle:
@jihlein I think the jitter manifesting may require logging on Sparky2 (which has slow flash).

jihlein:
@mlyle, ok, don't have a Sparky2 to help with that....

mlyle:
@jihlein if you're willing-- it should be possible to artificially create this situation.. if you run a 1KHz flight loop on a board with oneshot and set the PWM length to a maximum of 1100us...  the old code should do bad pulse-merge things and the new code should just end up outputting 500Hz PWM.

jihlein:
@mlyle, sure, no problem.

jihlein:
Here's the setup I used on an AQ32 board.

Hardware:
pwmhardware

PWM Outputs:
pwmoutput

Next PWM Output:
nextpwm

Next + PR342 PWM Output:
nextpwm342

mlyle:
Perfect-- thank you very much.  Looks exactly as expected.

droidicus:
It looks like this is reducing the pulse rate of channels even if they would not be effected by the pulse merge. Is this intended?

mlyle:
@droidicus yes.  We compare to the 'max' value when deciding whether to disable banks.  If we're not past the max (plus margin) we don't reset.

mluessi:
Code LGTM. With the margins used, the max. frequency with OneShot125 (250us max) is 3.77kHz, which is plenty. I think it would make sense to verify this on a F1 and F3 target before merge.

pug398:

pug398:
Tested on F1 and rate falls back as expected when pulse is to long

mlyle:
@pug398 Cool, thank you :)

mluessi:
Thanks!

PR #347, opened 2015-12-25
from TL/Cyphy: contextual menus on scopes

mlyle:
Corresponds to taulabs/taulabs#2116.

Needs test and review here.  Ability to copy lots will be nice.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/347)

### DIFF
--- a/ground/gcs/src/plugins/scope/plotdata.h
+++ b/ground/gcs/src/plugins/scope/plotdata.h
@@ -83,6 +83,7 @@ public:
     virtual bool readAndResetUpdatedFlag() = 0;
     virtual void plotNewData(PlotData *, ScopeConfig *, ScopeGadgetWidget *) = 0;
     virtual void deletePlots(PlotData *) = 0;
+    virtual void clearPlots() = 0;
 
     QwtScaleWidget *rightAxis;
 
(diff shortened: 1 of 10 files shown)


mlyle:
:lgtm:

---

Reviewed 10 of 10 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/347#-:-K6PYjwnED4kYyq6Wxvt:1417088721)*

mlyle:
http://imgur.com/5ww83qY
http://imgur.com/ilwyyJj
Is the result of a copy and then paste to imgur.  Really nice @kubark42

kubark42:
Thank you! But @m-thread deserves the credit as it was his original idea. 

P.S. If someone has time to run the pathology to ground before I do, that'd be cool. If I don't have time to figure it out before reviewing is completed, I'm considering disabling the options page feature since I find it's more work to have to resize the dialog box every time than it is just to navigate to it via the menu. (Although on the other hand, it provides easy access to a configuration option many people might not have known they had.)

mlyle:
I get the same size of dialog for the configuration whether i come via menu or this.  (It's not a great size, but equivalent)

mlyle:
https://www.youtube.com/watch?v=7meAalCYbWs

kubark42:
I think the youtube video is demonstrating the pathology, although I can't be sure because you didn't resize the configuration dialog, close it, and then reopen it via the menu. Note, I don't believe this is a regression, the dialog size stays constant as long as I only access it via the menu. So prior workflows aren't changed.

It's only accessing it from the ```Core::ICore::instance()->showOptionsDialog();``` method that creams the preferred size. It's probably a bug that was there all along and we never noticed because it was never called outside of CorePlugin.

P.S. The default size for that dialog could probably grow. It's that small on every system starting out, and I think it's easy to argue that it's way too small. I think it might have been really cramped in order to work on netbooks by default, but that was quite some time ago and it's worth a refresh.

PR #349, opened 2015-12-25
Package/OSX strip binaries to decrease size

mlyle:
For consideration.  Not having symbols could be annoying at times.  But it shrinks down the package size by a few megs, which is fairly large.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/349)

### DIFF
--- a/package/osx/libraries
+++ b/package/osx/libraries
@@ -12,7 +12,7 @@ QMLDIR="${APP}/Contents/Resources/welcome"
 
 pushd "${QT_SDK_BIN_PATH}"
 
-./macdeployqt "${APP}" -verbose=2 -no-strip -always-overwrite -qmldir="$QMLDIR"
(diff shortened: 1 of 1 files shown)


mlyle:
I could go either way on this one.  Someone flip a coin and decide whether to merge ;)

mluessi:
Does not having symbols mean we will get less informative crash dumps? If so, I would prefer the larger package size.

mlyle:
It means we'd need to retain the original build tree to get useful crash info.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/349#-:-K6t1o6-GZyYQhN5IoYG:771857039)*

PTDreamer:
You usually strip after you extract the symbols. It's how it is done in
nix, at least it will after I clean the PR
Em 31/12/2015 17:19, "Michael Lyle" <notifications@github.com> escreveu:

PR #358, opened 2015-12-28
manualcontrol: prevent spurious disarms in altitude hold mode

mlyle:
Altitude hold mode is an "autonomous flight mode" for the purpose of the
test in transmitter_control.  That is, low throttle values for extended
times can be legal.  If the "don't disarm on low throttle when
autonomous" is set, don't disarm under these circumstances in altitude
hold mode.  Else, in a prolonged descent, copter may drop out of the
sky.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/358)


mluessi:
LGTM

PR #359, opened 2015-12-28
gcs: settings: increase size to 780x525

### LINKED ISSUE #351: GCS: preferences dialog opens at a dumb initial size
Should be much bigger.

mlyle:
Previous size was 697x496.  It's still a little cramped.

Rationale for this size was:  fit in 800x600 with margin at the bottom.
At 1024x768, be small-ish compared to screen.

Sort-of fixes #351

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/359)

### DIFF
--- a/ground/gcs/src/plugins/coreplugin/dialogs/settingsdialog.ui
+++ b/ground/gcs/src/plugins/coreplugin/dialogs/settingsdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>697</width>
-    <height>476</height>
+    <width>780</width>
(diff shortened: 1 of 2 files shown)


mlyle:
Here are a couple of panes that look the worst now.  Summary: They're not perfect but they're usable.  After a release, it should be safe to assume 1366x768 and pick a size like 950x650 that completely cures this.  But the extra 83 pixels of width helps.

image
image

mlyle:
Adjusted "UAVO data multiplier" on both of those panes to further improve things.

mlyle:
How these look now:

image

image

PR #366, opened 2016-01-01
tracking-taulabs: squash all the history

### LINKED ISSUE #350: "Upstream" tracking: need a magical merge commit
After #348 and #349's fates are known, we should do a merge from TL `-s ours` to "ignore" everything that has drifted since in future rebases.

Those commits, right now, are:

```
pick 92ebf05 GCS/OSX: updates from comments
     Not really applicable
pick 893ef25 Ground-Updated license template
     Different project / separate changes
pick cdfe9cc Fix project credits file
     Ditto
pick de0413c Fix project credits file
    Ditto
pick 0a505b4 More granular alarm messages (subtypes).
    We have this independently / factored cleaner
pick 30f4286 FlyingF4: Use new ADC
    N/A removed target
pick e56e200 GCS: USBInfo remove unused fields
    We have a different solution
pick b32e4a1 VehicleConfigHelper: add default value
    We already have this
pick d382b81 UAVObjects Plugin: Update copyright
    N/A
pick b6db52e Main: Update copyrights
    N/A
pick ce31fc0 SysAlarms: Update copyright
    N/A
```

mlyle:
Simplify / speed up future rebases back and forth, by getting all the changes we're not taking into our history.

fixes #350 

@dustin Is this a good idea, or should we punt on the idea?

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/366)


dustin:
Hmm...  It seems like a good idea, but I'd guess we wouldn't find out if it's bad until we actually wanted to merge a change on top of something "we already did" but didn't.  Technically, the history should be there, so it *should* be resolvable.  I think it's worth having a formal maker for how caught up we are.  If this works, it'd automate everything pretty nicely.

mlyle:
I've tried to rebase a hypothetical Tau change on top of this branch, and end up just seeing that change in the interactive rebase.

mlyle:
(I don't know if merged one level deeper if all will be well, but I'd think so)

dustin:
That's great.  I think this could be a pretty nice way to track upstream.

mluessi:
I'm not sure if this is a good way to do it, as it may be hard to maintain in the future. If there is a TL PR that we want, why not just rebase the PR branch on top of our next or cherry pick the commits?

mlyle:
If we rebase the PR branch on top of our next, it wants to bring all the commits from prior unincluded work in and they need to be manually removed from the list.  Cherry-picking works fine, as long as it's a PR with a small number of commits.

The main advantage of this, though, is being able to audit at intervals what commits TL has that we don't.

mluessi:
You can use `--onto` when rebasing to put the commits starting with the first commit in the PR on top of our next, it's basically like cherry picking all the commits from the PR.

mluessi:
I.e.,

    git rebase --onto next COMMIT~1

where COMMIT is the first commit of the PR.

mlyle:
Yah.  If you're really opposed we don't need to do this, but I'm curious what you think the downside is.

I like being able to do log messages (`git log ..tl/next`) from us to TL to see "what we're missing" to systematically review, but unless we reset the slate it's going to rapidly become less useful.

mluessi:
@mlyle it isn't clear to me what this PR works. Is the idea to have a separate branch that contains all the TL commits we don't want and then when merging a PR from TL we automatically remove them? I guess this should work, but if TL becomes more active again it would be quite a tedious to maintain in the long run.

mlyle:
No, the idea is that it is a merge commit with all the TL stuff we don't want as ancestors on one side, that in the actual merge junction it flattens it out to 0 diff.  Future logs, rebases, merges, and other revision-based-comparisons will now ignore these things.

If Tau picks up again, every 2-3 weeks we can audit the set of changes since we last did one of these, make PRs from them, and accept them.. and then do this to squash the rest of the history down to nothing.  (This was produced with merge -s ours).

mluessi:
Makes sense. Merging.

PR #374, opened 2016-01-02
TxPID stack and Init fix

### LINKED ISSUE #370: Enabling TxPID module causes boot loop
After successful autotune session (and disabling the module) I wanted to tune yaw using TxPID module, but I was unable to do so, because after enabling TxPID module GCS was not recognising the board anymore. I had to rescue it using LibrePilot GCS to make it boot again.

dRonin GCS was reporting 3 unsuccessful boots alert.

droidicus:
This PR dynamically allocates memory for the large memory variables used in the code, preventing the stack from being depleted. Also fixes a problem where the Init function could try to construct a periodic callback before that system was running.

Tested on Sparky2 with TxPID on and off, also tested functionality of TxPID with FrSky S.PORT X4R-SB RX.

Fixes #370

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/374)

### DIFF
--- a/flight/Modules/TxPID/txpid.c
+++ b/flight/Modules/TxPID/txpid.c
@@ -96,14 +96,19 @@ static float scale(float val, float inMin, float inMax, float outMin, float outM
  */
 int32_t TxPIDInitialize(void)
 {
+	bool module_enabled = false;
+
 #ifdef MODULE_TxPID_BUILTIN
-	if 1 {
+	module_enabled = true;
 #else
 	uint8_t module_state[MODULESETTINGS_ADMINSTATE_NUMELEM];
 	ModuleSettingsAdminStateGet(module_state);
 	if (module_state[MODULESETTINGS_ADMINSTATE_TXPID] == MODULESETTINGS_ADMINSTATE_ENABLED) {
+		module_enabled = true;
+	}
 #endif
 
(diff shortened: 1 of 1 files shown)


mlyle:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[flight/Modules/TxPID/txpid.c, line 119 \[r1\]](https://reviewable.io:443/reviews/d-ronin/dronin/374#-K7-oG8mqsMTsEjy1ybP:-K7-oG8mqsMTsEjy1ybQ:1781009433) ([raw file](https://github.com/d-ronin/dronin/blob/e07f7c3aa2f50ec9e13fcec3f8e0f904ff3002af/flight/Modules/TxPID/txpid.c#L119)):*
Start will start the module in this case, with txpid_data set to NULL.

We could arguably just use txpid_data being not-null as the flag as to whether to start it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/374)*

mlyle:
:lgtm:

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/374#-:-K72Pl3ofUOEQgR2wO4L:1417088721)*

PR #378, opened 2016-01-03
gcs: attempt to make config tab icons work better

### LINKED ISSUE #260: GCS:Configuration: Scrollbar obscures tabs
This was reported in TauLabs/TauLabs#1146 and is relevant on Linux as well (OS X is okay due to the native scrollbar style used). The issue is becoming worse as more and more tabs get added. Should be easy enough to fix by applying a CSS style.
screenshot from 2015-12-13 23 05 24

### LINKED ISSUE #371: Camera stabilisation icon misaligned
One of the options on the left hand side in GCS seems to have a too long title, which misaligns the icon and enables horizontal scrolling.

image

Edit: Sorry, I probably should've provided setup details - this is on osx.

mlyle:
Fixes #371 . Fixes #260 .

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/378)

### DIFF
--- a/ground/gcs/src/libs/utils/mytabbedstackwidget.cpp
+++ b/ground/gcs/src/libs/utils/mytabbedstackwidget.cpp
@@ -58,15 +58,16 @@ MyTabbedStackWidget::MyTabbedStackWidget(QWidget *parent, bool isVertical, bool 
     }
 
     if (m_iconAbove && m_vertical) {
-        m_listWidget->setFixedWidth(90); // this should be computed instead
+        m_listWidget->setFixedWidth(91); // this should be computed instead
         m_listWidget->setWrapping(false);
+        m_listWidget->setStyleSheet(QString("QScrollBar:vertical { width: 12px; }"));
     }
 
     toplevelLayout->setSpacing(0);
     toplevelLayout->setContentsMargins(0, 0, 0, 0);
-    m_listWidget->setContentsMargins(2, 0, 2, 0);
+    m_listWidget->setContentsMargins(0, 0, 4, 0);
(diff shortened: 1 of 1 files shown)


mlyle:
I need to test this on windows to see how it looks, etc.  The pegged sizes are unfortunate but it's no worse than the prior practices here.

droidicus:
Tried this on Win10, looks better here. One comment (and this was true before and after this PR), if I make the window short, and this causes a vertical scroll bar for the config tab icons, then the scrollbar is just overlaid on top of the box, with no change to the width or position of the icons and the box. Since that sentence made virtually no sense to me immediately after writing, here are a couple pictures...
tall
short

mlyle:
@droidicus -- thanks for the test.  The scrollbar thing is something I don't plan to totally fix but it's on my radar for trying to tune this better.

mlyle:
https://github.com/TauLabs/TauLabs/issues/1146  was the historical issue @ Tau, and it looks like we're already a little better.

tracernz:
@droidicus, see #260

mlyle:
image
image

mlyle:
I am happy with this now on Mac and Windows.  It is not pixel perfect but it is much better.

mluessi:
Tested on Linux, the icons are no longer centered for some reason:

dr_icons

mlyle:
Are they centered without the scrollbar?  e.g. see the 2 pictures above on Windows.

I have not worked on getting the icons to "jump left" appropriately when the scrollbar comes up, instead just focusing on lowering the scrollbar-icon interference.

mluessi:
Confirmed that the icons are only not centered if the scroll bar is shown and this was also the case before this PR. Merging.

PR #384, opened 2016-01-03
GCS: Remove 50Hz PWM defaults

### LINKED ISSUE #383: "Standard PWM" confuses everyone
We need to default to Turbo, both in UAVOs and .. appropriately caution people in the wizard.

mlyle:
This is not a configuration that we ever want someone to use, and it's been pretty much the "default".  Hence it's something a lot of users trip over.  Fix that.

Fixes #383 

Note that the default is now "400" on all channels, but the GCS sets unused banks in the wizard to 50 to better support analog servos.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/384)

### DIFF
--- a/ground/gcs/src/plugins/setupwizard/pages/outputpage.cpp
+++ b/ground/gcs/src/plugins/setupwizard/pages/outputpage.cpp
@@ -65,10 +65,8 @@ bool OutputPage::validatePage()
         // This is safe to do even if they are wrong. Normal ESCS
         // ignore oneshot.
         setOneshotTimings();
-    } else if (ui->rapidESCButton->isChecked()) {
-        getWizard()->setESCType(SetupWizard::ESC_RAPID);
     } else {
-        getWizard()->setESCType(SetupWizard::ESC_LEGACY);
+        getWizard()->setESCType(SetupWizard::ESC_RAPID);
     }
 
     return true;
--- a/ground/gcs/src/plugins/setupwizard/pages/outputpage.ui
+++ b/ground/gcs/src/plugins/setupwizard/pages/outputpage.ui
@@ -56,50 +56,6 @@ p, li { white-space: pre-wrap; }
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout">
      <item>
-      <widget class="QToolButton" name="defaultESCButton">
-       <property name="font">
-        <font>
-         <pointsize>10</pointsize>
-        </font>
-       </property>
-       <property name="toolTip">
-        <string>Standard ESC 50Hz</string>
-       </property>
-       <property name="styleSheet">
-        <string notr="true">QToolButton { border: none }</string>
-       </property>
-       <property name="text">
-        <string>Standard ESC</string>
-       </property>
-       <property name="icon">
-        <iconset resource="../wizardResources.qrc">
-         <normaloff>:/setupwizard/resources/bttn-ESC-up.png</normaloff>
-         <normalon>:/setupwizard/resources/bttn-ESC-down.png</normalon>:/setupwizard/resources/bttn-ESC-up.png</iconset>
-       </property>
-       <property name="iconSize">
-        <size>
-         <width>200</width>
-         <height>100</height>
-        </size>
-       </property>
-       <property name="checkable">
-        <bool>true</bool>
-       </property>
-       <property name="checked">
-        <bool>false</bool>
-       </property>
-       <property name="autoExclusive">
-        <bool>true</bool>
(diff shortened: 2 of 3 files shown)


droidicus:
Tested on Win10 and Sparky2. 400Hz PWM and OneShot are showing up. If neither is selected, then 400Hz PWM is default.

dustin:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/384#-:-K79LIf1cLs4sG-X9DlN:1417088721)*

mluessi:
I was under the impression that 50Hz was chosen because servos often burn out when running at 400Hz, so using it as default may result in many unhappy users.

jihlein:
Same impression here, I feel 50 Hz is a safer default, even if 99.9% of users change it.

mlyle:
On small multirotors 50Hz does not have enough bandwidth to control the vehicle and is basically ensured to crash.  There's basically no ESCs extant that we should send stuff at 50Hz to.  The second commit makes it so when people use the wizard all non-ESC banks are programmed at 50Hz.  This is a very frequent support issue.

mlyle:
(So I could accept reverting the first commit of this, I don't think the second should be).

mlyle:
(There's also a question whether Min should be 0 or 1000, because on PWM/brushed setups motors spin  when default actuatorsettings are loaded).

DTFUHF:
I agree, and was surprised to see this commit. I think it is short sighted.
The first think I thought was oh OK, dronin will soon drop fixedwing
support completely.

DTFUHF:
A solution might be to disable outputs entirely by default, using the
frequency field. This would cover the brushed/direct-drive case too.

mluessi:
Using a OFF default sounds like a good idea to me. It would make sure that all users touch the setting (so MR users wouldn't be flying with 50Hz unless they are silly) and servos won't die. @mlyle WDYT?

dustin:
Off make sense.  50Hz isn't any more of a right answer than 400Hz.

The sky isn't falling here WRT to servos.  We want to ease users into a successful path for whatever their use cases are.  The previous state was pretty terrible presenting 50Hz as the "standard" for ESCs for multirotor use.  I think a few of us learned that some users power up planes early in setup without configuring their outputs properly.

It would be good to have some care put into more setup workflows.  I don't know that anyone's even offered to do any testing on fixed-wing, which likely means other changes that are improving some areas might be considered regressions there.

pug398:
Since I have always had to touch the output settings on every setup anyway setting them to off would be no different. Not exactly helping in ease of setup for first time users though. Perhaps I misunderstood but doesn't the second PR set motor outputs to 400hz and servo outputs to 50hz?

DTFUHF:
As I understand it, new users are supposed to use the vehicle setup wizard
which will set these outputs appropriately

mluessi:
The problem is with changing the default in the UAVO definition, as it will go to that value when upgrading or first flashing a board. OFF would be the best default IMO, as things clearly wouldn't work when settings are not touched as opposed to working poorly (50Hz on MR).

pug398:
So before you ever set the board up all outputs are on and at 400hz after flash? Just trying to figure out how it would change since the outputs are currently disabled after a flash and you have to select a vehicle type to turn them on.

mlyle:
Outputs actually (unfortunately) default on.  @mluessi is right that we should set the actuator defaults to turn them off.

This has surprised people on brushed motors because the default motor 1us / 50Hz timing is enough to spin them.  It also has surprised people by "turning off" oneshot when the actuator settings uavo has changed.  Better to do the safe thing.

SystemsGuy:
I'm a bit late to the party - happy 2016 everyone - but I think I'm in the camp of "default to off" - I've toasted a servo on a tricopter by not paying attention to what I was doing..

mlyle:
@SystemsGuy Just so I can better understand-- how did that happen, given we defaulted to 50Hz before?

(Was it that it was driven to 1000us so "to the far stop"?)

SystemsGuy:
mlyle - no, was bone headedness on my part - I enabled the servo channels on 400hz - more confirmation that is a bad thing :)

pug398:
My mistake in thinking they were off by default then. I thought the motor channels are disabled until you set them to a particular output channel either by the vehicle tab or the wizard does it for you. Definitely don't want a channel enabled to 400hz after a flash that someone might unknowingly plug a servo into just messing around.

PR #394, opened 2016-01-05
flight: printf2 stack alignment change

mlyle:
Source: brainFPV/TauLabs#40 / @droney

Converted to diff and reformatted

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/394)

### DIFF
--- a/flight/PiOS/Common/pios_heap.c
+++ b/flight/PiOS/Common/pios_heap.c
@@ -249,6 +249,14 @@ void PIOS_heap_increase_size(size_t bytes)
 #endif	/* PIOS_INCLUDE_FREERTOS || defined(PIOS_INCLUDE_CHIBIOS) */
 }
 
+/* Provide an implementation of _sbrk for library functions.
+ * Right now it returns failure always.
+ */
+void *_sbrk(int incr) {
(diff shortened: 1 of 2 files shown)


mlyle:
I have not yet reviewed or vetted this.

(Not a fan of maintaining our own printf implementation-- is there a reason we can't just use the one in newlib?)

@droney thanks for the code/fix.

droney:
I attempted to close this pull request as it was intended to be directed at brainfpv specifically since that fork currently uses printf2. Taulabs eliminated this module two months ago. Please close this pull request.... my attempt apparently did not work.

mlyle:
@droney dRonin is a fork of TauLabs that supports BrainFPV, so I picked up your PR because of that.  (It may become the official BrainFPV firmware).

droney:
I agree with you that newlib is the right long term solution. newlib has changed enough where I no longer see the stack alignment code in the printf function. I suspect when compiling with newlib for EABI platforms, stack alignment is handled for all public function calls. However I am not familiar enough with newlib to confirm this.

mlyle:
Attribution fixed.

mluessi:
Bench tested, works.

PR #397, opened 2016-01-06
uavos: logging: log a few more objects by default in custom profiles

mlyle:
AttitudeActual/StabilizationDesired are the minimum needed to have some
idea what's going on... ManualControlCommand doesn't hurt.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/397)

### DIFF
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -43,6 +43,7 @@
 
 #include "accels.h"
 #include "actuatorcommand.h"
+#include "actuatordesired.h"
 #include "airspeedactual.h"
 #include "attitudeactual.h"
 #include "baroaltitude.h"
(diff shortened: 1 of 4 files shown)


mluessi:
@mlyle for them to be logged with the default profile, they would need to be added in the code. This just makes them being logged with the custom profile.

mlyle:
Thanks for pointing that out.  I assumed that the actual default profile had both of these, but it was slightly different.

mluessi:
LGTM. Did you have a chance to test this? I think it should be fine to merge without test.

mlyle:
No test, unfortunately.

ekaszubski:
@mluessi @mlyle I can test this after lunch today

ekaszubski:
Fetching and testing now

ekaszubski:
@mluessi  @mlyle Confirmed logging periods are set correctly on sparky2.

mluessi:
@ekaszubski thanks for the test!

Issue #425, opened 2016-01-10
S.Bus control safety anomaly: throttle up, unresponsive to disarm switch


fujin:
I experienced a very concerning safety event yesterday at the field where I had my goggles on, armed, applied some throttle and immediately lost control of the RC, though the OSD did not indicate failsafe (by moving into ST1).

I heard throttle starting to increase, flipped my disarm switch, the machine did not respond, throttle continued to increase. I kicked the quad off the ammo can I was launching from, held it upside down with my shoe and ripped the xt60 plug out.

I'm pulling the DVR footage. We could potentially audit s.bus failsafe path / s.bus frame receipt code paths as well.

fujin:
not sure where to put this.. this was the 'flight stats' anomaly/disarm in-flight today on shrike. 

https://www.youtube.com/watch?v=IKdDf3exeeE

pug398:
I have had zero sbus issues on 5 fc but I do not use an arming switch. I think the switch code needs serious auditing. One may crash next flight but no control problems since sbus was adopted for me.

mluessi:
It seems like the s.bus driver sets the channels to defaults if no valid s.bus frame is received for 100ms. When switch arming is used, this can mean it acts like the switch has been switched to disarm (if the channel defaults correspond to switch in disarm state). This seems to have happened in the video by @fujin, i.e., it disarms (stats are shown) and then immediately arms again when the s.bus data is valid again.

pug398:
Considering my mostly sedate flying a 0.1 second glitch every now and then could go undetected I suppose.

fujin:
https://youtu.be/hzaeqhbNsFo

mlyle:
Closing; we'll keep a record of this but there's nothing actionable here and we've made substantial sbus improvements.

Issue #357, opened 2015-12-28
Switch arming is completely broken in multiple ways


mlyle:
In renatus originally, but probably won't stay here.

Things to think about:

* Failsafe and what happens to armed state (especially in conjunction with +Delay modes)
* Funny interaction with low throttle timeout-- especially in conjunction with +Delay modes)
* Funniness of ARMED_STILL_HOLDING state
* Edge vs. level triggered behavior
* ManualControl's enclosing state machine

mlyle:
Also of note-- it should take more than 1 frame to arm/disarm via switch.  I know we want to switch arm/disarm quickly... but we don't want to get fooled by glitches.

henn1001:
As someone who is used to mw/cleanflight the failsafe behavior here seems to be a total mess. It´s very confusing. 
I tested Stick arming with Arming timeout of 2 seconds.

Switch+Throttle:
Turning off or unplugging my receivers disarms the quad after 2 seconds. (good)
After arming and even in flight  the quad will disarm after 2 seconds if i don´t raise my throttle over the neutral position (wtf - why) ?

Switch+Throttle+Delay:
Turning off or unplugging my receivers disarms the quad after 2 seconds. (good)
After arming, the quad can sit there indefinitely without disarming (good). After i raise my throttle and go back below the neutral position it disarms again after 2 seconds... and then it rearms itself (double wtf D: )

This whole auto disarm while below throttle neutral seems very alien to me. For the meantime i probably use really low throttle neutral + some kind of idle up switch that forces my min throttle command above neutral while armed. No disarming + working pids = win

tracernz:
The armed timeout defaults to 30 seconds, did you change it to 2 seconds or has a bug been introduced?

henn1001:
No i changed it to 2 seconds.. First i thought the failsafe behavior is completely broken. Keep in mind i´m using "spin at neutral while armed". In case of a failsafe the motors would spin for 30 seconds at neutral without disarming.. who wants that? Wanna buy new motors if the quad "crashes" into deep grass?

This whole feature seems to be designed around the people who don´t use "spin at neutral".

tracernz:
The default is to not spin while armed. Leaving an armed vehicle with no visible indication that it is armed for an indefinite period is very dangerous, hence the timeout by default.

tracernz:
You can disable the timeout by setting it to 0 I believe (can confirm by looking at the code if necessary).

henn1001:
[...]

Issue #42, opened 2015-11-25
Setup 'new contributor' labels based issue guidance


fujin:
See
![image](https://cloud.githubusercontent.com/assets/6873/11408068/b9d359d2-941b-11e5-8b10-05676b2848d2.png)
 for an example from the InfluxDB project.

It would be great to have category and difficulty labels to provide a way for new contributors to find work they can pick up and do without worrying too much about a far-reaching, multi-target(?) change.

dustin:
I like this idea.  Finding low hanging fruit is good for both newbies and boredom.

fujin:
I started adding a bunch of labels.

tracernz:
@dustin difficulty/fix-cleanflight can be the hardest one ;-P

dustin:
Ha.  I try not to dislike stuff.  The organization is nice here, though.

fujin:
Closing this now :) we have a good bunch of labels and I see others are adding new ones.

PR #428, opened 2016-01-11
flight: sbus: don't attempt to use time to synchronize

mlyle:
Connects to #425 

Background:  We now have 2 users that have had significant anomalies with sbus + brain + arming switch.  Neither is at a rate that's conclusive.  The known switch arming issues mentioned in #357 are a significant confound.

However, in auditing the sbus code I noticed that we rely on pretty specific timing that relates to the Futaba air-interface that third party RX firmwares may not always try so hard to comply with.  This seems bad-- while this timing is probably considerably helpful in ensuring synchronization with genuine Futaba stuff and rejecting illegitimate data, it may work against us in other cases.

Non-openpilot ecosystem sbus decoders do not have this check.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/428)

### DIFF
--- a/flight/PiOS/Common/pios_sbus.c
+++ b/flight/PiOS/Common/pios_sbus.c
@@ -286,7 +286,7 @@ static uint16_t PIOS_SBus_RxInCallback(uintptr_t context,
  * RTC timer is running at 625Hz (1.6ms).
  *
  * Data receive function must clear the receive_timer to confirm new
- * data reception. If no new data received in 100ms, we must call the
+ * data reception. If no new data received in 48ms, we must call the
  * failsafe function which clears all channels.
  */
 static void PIOS_SBus_Supervisor(uintptr_t sbus_id)
@@ -305,8 +305,8 @@ static void PIOS_SBus_Supervisor(uintptr_t sbus_id)
 		state->receive_timer = 0;
 	}
 
-	/* activate failsafe if no frames have arrived in 102.4ms */
-	if (++state->failsafe_timer > 64) {
+	/* activate failsafe if no frames have arrived in 48ms */
+	if (++state->failsafe_timer > 30) {
 		PIOS_SBus_ResetChannels(state);
 		state->failsafe_timer = 0;
 	}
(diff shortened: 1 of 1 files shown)


mlyle:
Completely untested; I wrote the code, and it compiles.  What else can you ask for?

mlyle:
@fujin @hshred @tracernz @dtfuhf @ohmeye comments appreciated.

DTFUHF:
Without reading too much code on my phone, I think its still a good idea to
have a byte timeout. I've seen the case where the loop locks onto a start
byte in the middle of the packet, so only nonsense data is received until a
slight destabilization in timing and it is allowed to lock onto the real
start byte.

mlyle:
So basically to let idle time knock you out of packet processing, but not to require idle time to begin a packet?

Of course, if a tx ever hesitates in the middle of a frame/gets an interrupt/does something else, it will cause desync..

Meh.  It's not a very good protocol.

tracernz:
Is there not a checksum to prevent mixed up packets from decoding correctly?

DTFUHF:
There is in SUMD!

dustin:
Does anyone have any frsky contacts to convince them to make a small RX
with three pins, SUMD, and RSSI injection?

mlyle:
@tracernz nope.  Just a loosely-defined stop packet condition, and parity on the stuff along the way.

tracernz:
Nasty, Futaba really dropped the ball there.

fujin:
Bench tested BrainFPV with next + #42  + #430 per @mlyle in IRC, able to arm/disarm.

mluessi:
I'm not sure if  merging this change is a good idea so close to a release. #430 alone may mask erroneous switch toggles by the s.bus layer. Thoughts?

mlyle:
I think that's the point-- to mask erroneous toggles.  A small bit of electrical noise or a receiver sending a single glitched frame shouldn't cause an in-flight disarm that dooms the vehicle and can't be recovered from.  Every other arming method has an appropriate protection against this.

I think this is a good idea because A) neither APM nor *flight have this check, and B) it relies on timing that I don't trust people writing clone firmwares to maintain.   E.g. OpenLRSng does not mask interrupts during the serial output phase of the firmware and could have short gaps in the frame.

If we don't merge this entire change, we need to at least merge the failsafe-time portion of the change -- else the switch change made in #430 is ineffective.  I agree doing this close to a release is not good.  But I think based on problems we have heard about (which may or may not be new regression) we have no choice but to slip the release and wait a week after we have things like this merged for reports.

PR #430, opened 2016-01-12
flight: manualcontrol: debounce disarm for 100ms

mlyle:
With reports of disarm-in-flight on sbus + switch+throttle arming + brain, it's become clear we're really vulnerable to a single aberrant frame in the switch+throttle / switch+delay modes.  So debounce-- don't listen to a switch disarm for 100ms.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/430)

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -697,12 +697,7 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
  		}
 
   		bool disarm = disarming_position(cmd, settings) && valid;
-		if (disarm && (settings->Arming == MANUALCONTROLSETTINGS_ARMING_SWITCH ||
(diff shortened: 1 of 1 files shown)


tracernz:
I'm weary of anything that puts conditions on disarm. This will also potentially mask underlying issues with our code. We should be cautious.

tracernz:
That said the code looks good and makes switch disarm a bit less of a special case.

mlyle:
@tracernz Yah.  We need error counters and things on this receiver infrastructure so stuff is more obvious.  And unconditional anomaly locking.

fujin:
Bench tested BrainFPV with next + #42  + #430 per @mlyle in IRC, able to arm/disarm.

mluessi:
LGTM. Off topic, but what do the `SWITCHDELAY` and  `SWITCHTHROTTLEDELAY` modes actually do? They seem to be the same as the variants without delay in the name.

mlyle:
They go through the ARMING state which allows zeroing of the gyros.

PR #439, opened 2016-01-14
Linux: improvements to udev

### LINKED ISSUE #437: Lux: udev rule on linux missing.
Also ideally udev rules would go in the linux package.

mlyle:
- Include LUX
- Include the udev rules in the compressed tar.xz.

Fixes #437

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/439)

### DIFF
--- a/package/Makefile.linux
+++ b/package/Makefile.linux
@@ -68,6 +68,7 @@ ground_package_os_specific: ground_package_common
 ifeq ($(EXPORT_SYMBOLS), YES)
 	$(V1) $(ROOT_DIR)/package/linux_extract_debug_symbols.sh $(DUMP_SYMBOLS_TOOL) $(PACKAGE_DIR)/$(GCS_PACKAGE_NAME)
 endif
+	$(V1) cp -v $(ROOT_DIR)/package/linux/deb/_package.udev $(PACKAGE_DIR)/rules.udev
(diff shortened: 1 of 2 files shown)


mlyle:
Needs confirmation

package/Makefile.linux
```diff
@@ -68,6 +68,7 @@ ground_package_os_specific: ground_package_common
 ifeq ($(EXPORT_SYMBOLS), YES)
 	$(V1) $(ROOT_DIR)/package/linux_extract_debug_symbols.sh $(DUMP_SYMBOLS_TOOL) $(PACKAGE_DIR)/$(GCS_PACKAGE_NAME)
 endif
+	$(V1) cp -v $(ROOT_DIR)/package/linux/deb/_package.udev $(PACKAGE_DIR)/rules.udev
```
tracernz: Should probably put a number/prio in front?

mlyle: So I don't have the branding down here (I don't think), so I can't construct a complete name.  I was thinking docs would instruct user what to do?

The deb attempts to do the right thing.

tracernz: Okay, we can just tell the user to do something like `sudo cp ./rules.udev /etc/udev/rules.d/45-dronin.udev && sudo udevadm control --reload-rules`

mlyle: Yes, we have a line like that in current errata.

tracernz:
Works as expected, merging.

One thing I noticed is that the flight-*.zip is included in the tar, this is unnecessary since all the files are already expanded in the firmware_* directory.

PR #440, opened 2016-01-14
GCS:SetupWizard: Rename TurboPWM help link

tracernz:
s/OnlineHelp:-TurboPWM-ESC's/OnlineHelp:-ESC-Refresh-Rate/ 

Relates to #308, wiki page needs renamed too. 

We should do this before the first official release to avoid breaking the help link in old releases.

Also, renamed Oneshot label to Oneshot125 to avoid confusion with the new Oneshot42.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/440)

Issue #219, opened 2015-12-10
Auto-select PWM timer frequency


tracernz:
So user doesn't have to select 1 MHz or 12 MHz when configuring outputs. See TauLabs/TauLabs#2074

PR #442, opened 2016-01-14
Support for Oneshot42 and 24 MHz PWM

tracernz:
With the release of the latest KISS ESC 24A Race Edition ESCs comes Oneshot42. This is a first cut to add support for 24 MHz timer operation and Oneshot42 in the GCS setup wizard. 24 MHz output only tested on Sparky1 with scope so far. I think there are prescaler issues on F4/168 MHz SysClk when using timers on APB1 which I will need to investigate/solve. Possibly other issues I will discover when I delve into datasheets.

Anybody have Oneshot42 ESCs to flight test when this is ready (NOT YET)?

Note: the 2 commits from #440 are included in this PR as it will merge first and would otherwise clash with this. Is that the right thing to do?

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/442)

edit by mlyle: connects to #219


mluessi:
I think this would be automatically solved by #219, WDYT?

tracernz:
@mluessi Good suggestion. I'll include that in this PR and it will take care of my problems.

fujin:
@tracernz I have these ESCs when you are ready for test.

tracernz:
Superseded by #958

PR #445, opened 2016-01-15
flight: linker script changes: sort BSS by alignment …

### LINKED ISSUE #14: Investigate whether sorting bss by alignment requirement is beneficial
`*(SORT_BY_ALIGNMENT(.bss*))` in linker script or similar.

### LINKED ISSUE #369: CC3D memory alarm when autotune is enabled
Enabling autotune module seems to consume significant amount of memory on cc3d - the memory alert is red.

I've managed to work around that problem by disabling flexi port telemetry, which is enabled by default.

mlyle:
Saves 72 bytes of memory on Naze32, 152 bytes of memory on sparky2.

Also mark BSS noload; has no difference now (because all constituent BSS
sections are already marked noload) but a good practice for the future.
Most section scripts had this already, but chibios ones were missing it.

Fixes #14

Fixes #369

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/445)


tracernz:
Good good. :+1:

PR #446, opened 2016-01-15
GCS:Config:Input: Add description of arming timeout

### LINKED ISSUE #427: Arming timeout in GCS very confusing to end users.
We need to improve this.

tracernz:
Fixes #427 

Also changes default input tab to be RC input.

screenshot from 2016-01-15 16 01 38

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/446)

### DIFF
--- a/ground/gcs/src/plugins/config/input.ui
+++ b/ground/gcs/src/plugins/config/input.ui
@@ -29,7 +29,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>2</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="RCInput">
       <attribute name="title">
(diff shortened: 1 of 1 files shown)


mlyle:
Edited top desc

DTFUHF:
I agree this needs to be changed.... but this is far too confusing and
wordy.

Maybe the solution is to use radio buttons for each arming type, with a
tooltip for each one.

mlyle:
We could move arming timeout to its own box with the second paragraph.

tracernz:
Better?
screenshot from 2016-01-15 16 40 04

mlyle:
@dtfuhf Like that more?  I think I do.  (Maybe slightly nudge down the Arming timeout since we have lots o' space)

DTFUHF:
It's a start, but I disagree more with the paragraph of text.

If you don't like radio buttons, maybe the text can be updated to describe the currently selected arming mode. If the user isn't using a switch arming mode, then the text does not have to be complicated with its description.

mlyle:
I agree it's a little TLDR for many of our users.  It's still better than nothing, and I'd prefer solutions that don't incur code and just change .ui files.  Can always improve more in future releases.

tracernz:
I don't want to go changing code and introducing bugs at this point in the release cycle. We can revisit at the next release.

PR #449, opened 2016-01-15
GCS: Add board specific ADC pin names

tracernz:
Board plugins can now provide names for the ADC pins. The way this is implemented is slightly hacky (it just replaces the combobox text set by the UAVO->widget relation) but I think it's the cleanest way I could accomplish this without having to manipulate/update FlightBatterySettings by hand. The usability win is large (this has been a pet hate of mine since I first began using Tau Labs).

Naze32 Example:
screenshot from 2016-01-16 00 50 58

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/449)

### DIFF
--- a/ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
+++ b/ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
@@ -139,3 +139,20 @@ int AQ32::queryMaxGyroRate()
         return 500;
     }
 }
+
+QStringList AQ32::getAdcNames()
+{
+    ExtensionSystem::PluginManager *pm = ExtensionSystem::PluginManager::instance();
+    UAVObjectManager *uavoManager = pm->getObject<UAVObjectManager>();
+    HwAQ32 *hwAQ32 = HwAQ32::GetInstance(uavoManager);
+    Q_ASSERT(hwAQ32);
+    if (!hwAQ32)
+        return QStringList();
+
+    HwAQ32::DataFields settings = hwAQ32->getData();
+    if (settings.ADCInputs == HwAQ32::ADCINPUTS_ENABLED) {
+        return QStringList() << "BM" << "Analog AI2" << "Analog AI4";
+    }
+
+    return QStringList() << "Disabled" << "Disabled" << "Disabled";
+}
--- a/ground/gcs/src/plugins/boards_aeroquad/aq32.h
+++ b/ground/gcs/src/plugins/boards_aeroquad/aq32.h
@@ -45,6 +45,7 @@ public:
     virtual QPixmap getBoardPicture();
     virtual QString getHwUAVO();
     virtual int queryMaxGyroRate();
+    virtual QStringList getAdcNames();
 
 };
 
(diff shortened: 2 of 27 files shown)


mluessi:
Awsome! Is it possible to also use the board specific name in the RSSI channel selection when analog RSSI is selected?

tracernz:
Yeah, I was thinking about that just now, will check it out.

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
```diff
@@ -139,3 +139,20 @@ int AQ32::queryMaxGyroRate()
...
+    ExtensionSystem::PluginManager *pm = ExtensionSystem::PluginManager::instance();
+    UAVObjectManager *uavoManager = pm->getObject<UAVObjectManager>();
+    HwAQ32 *hwAQ32 = HwAQ32::GetInstance(uavoManager);
+    Q_ASSERT(hwAQ32);
+    if (!hwAQ32)
+        return QStringList();
+
+    HwAQ32::DataFields settings = hwAQ32->getData();
+    if (settings.ADCInputs == HwAQ32::ADCINPUTS_ENABLED) {
+        return QStringList() << "BM" << "Analog ADC2" << "Analog ADC4";
```
jihlein: This a great idea.  Can we can we change Analog ADC2 and Analog ADC4 to AI2 and AI4 respectively to match the board silk screens.

tracernz: Yep, that's a mistake on my part. I ended up taking the labels from here instead of the silk: http://aeroquad.com/showwiki.php?title=AeroQuad32-Flight-Control-Board-v2#AeroQuad32-v2-board-pin-assignments

tracernz:
I found the RSSI channel groups on the Input page are subtly broken. The code for the channel setup form (each function/line is it's own form) uses [ManualControlSettings.ChannelGroups](https://github.com/d-ronin/dRonin/blob/next/shared/uavobjectdefinition/manualcontrolsettings.xml#L5) to decide what goes in the channel number dropdown but the RSSI function is passing in [ManualControlSettings.RssiType](https://github.com/d-ronin/dRonin/blob/next/shared/uavobjectdefinition/manualcontrolsettings.xml#L32). For example if None is selected (ManualControlSettings.RssiType.None = 0), it ends up using the [PWM number of channels (8)](https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/config/inputchannelform.cpp#L128) rather than 0 as it should (ManualControlSettings.ChannelGroups.PWM = 0)! Took me a while to figure out why I couldn't find the ADC code...

tracernz:
Fixed the RSSI input type bug and added custom ADC names to it.

tracernz:
Set FrSkyPWM RSSI channel count to 1.
```
<tracernz> hshred: q: for FrskyPWM RSSI, how many channels should be available in the input config form? is the channel number even used?
<hshred> tracernz: the channel number isn't used
```

mlyle:
Is that last comment a pending item, @tracernz?

tracernz:
Nope, done.

mlyle:
:lgtm:

---

Reviewed 23 of 24 files at r1, 4 of 4 files at r2.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/449#-:-K8WvEFkL5WQHtQqasUv:1417088721)_

Issue #450, opened 2016-01-15
Logging: Add logging to OpenLog for other targets


henn1001:
I think @jihlein already added that option to the AQ32 and Quanton target in #16 
Would be cool if this is added for other targets as well :)

I will have a look at this later.

- [x] AQ32
- [x] Brain
- [x] Colibri
- [ ] Discoveryf4
- [x] Flyingf3
- [x] Lux
- [x] Quatnon
- [x] Revolution
- [x] Sparky
- [x] Sparky2

jihlein:
If the targets are on PIOS_HAL, should just be a matter of adding OpenLog to HwXXXX.xml.  PRs are submitted to upgrade the remaining non PIOS_HAL targets FWIW.

henn1001:
@jihlein i´ve got it working on sparky 2 but i have problems parsing the file
https://www.dropbox.com/s/ylvnz7pzrrxmkny/LOG00042.TXT?dl=0
the python script is throwing me different errors - have you had any luck with this?
Any tips?

jihlein:
@henn1001, honestly I've never had much luck with the python scripts.  What I did was download the Octave program (a freeware version of Matlab) and use the LogConvert.m file in the build/matlab folder.  It's worked fine for me.  

Find Octave at: https://www.gnu.org/software/octave/download.html

Let me know if you have any questions and/or how it works for you.

henn1001:
uhh i have matlab - i read something about that but i could not find the .m file in the source folder. I did not know i have to build them explicitly.

I´ll look into that, thanks!

henn1001:
It´s getting parced with no errors - unfortunately every struct is empty :(

jihlein:
Can you post your code changes?

henn1001:
https://github.com/henn1001/dRonin/commits/sparky2OpenLog

jihlein:
Cool, I'll take a look.

jihlein:
On a quick once over it looks correct.  

The M file can parse both GCS logs and Overo logs.  Overo logs are generated by the flight controller logging, either the spi flash or open log.  Make sure the proper argument flags are set:

% To process Overo files call with
% LogConvert(path_to_file, true)

I don't recall if not setting that argument to true will result in empty structures.

I'll dig into it more later.  One question - are you building the flight firmware and Matlab M from the same source?

henn1001:
Build from the same source/fork - yeah.

`LogConvert(path_to_file, true)`

[...]

PR #458, opened 2016-01-16
flight: support openlog on lux

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/458)


mlyle:
just written and compiled.  No test yet.

henn1001:
Seems to work -> #450

mlyle:
@henn1001 You have a Lumenier LUX and tried it on that target?

henn1001:
Nope - sorry i meant it works on sparky2. But it was more or less a copy paste from your implementation.

mlyle:
Please feel free to open an PR :)

henn1001:
My commit depends on some changes in this PR - should i make it standalone or base it on this one?

mlyle:
The only change this has outside the target is to work when there's no SPI flash on a target.  Sparky2 has a SPI flash, so it shouldn't need that part of the change  (that is, the #ifdef in logging.c will not omit any code on sparky2).

Lux is the first target we'd support logging on that doesn't have any logging flash internally.

jihlein:
Actually AQ32 is that way too, it's why I put it on openlog originally.

mlyle:
@jihlein ah, aq32 could probably be slightly cleaned up for how it bypasses that after this PR, then.

mlyle:
Rebased

mlyle:
Confirmed to work nicely on Lux.

PR #463, opened 2016-01-17
More tau labs labels in logging code

droidicus:
Note: This change will modify the log file header signature as well.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/463)


droidicus:
@tgreer it would be nice if this could land in before the release. Since it is changes the log header it would be nice to put this in before wide distribution for compatibility reasons. I am fine if the ultimate decision is to defer though.

tgreer:
@droidicus understood, but we're quite close to release. However we do have a safety bug to be merged before we can release. @mluessi @mlyle @tracernz thoughts?

mlyle:
This should defer, and then we should accept either log signature format going forward.  We want to be able to open old logs, including ones made @ tau.

droidicus:
@mlyle Fair enough, I would suggest that we cut support for OpenPilot logs though, enough divergence has happened that it probably doesn't make any sense anyway.

mlyle:
:lgtm:

---

Review status: 0 of 11 files reviewed at latest revision, all discussions resolved.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/463#-:-K8WuhUmRhBO3Nk5RNgr:1417088721)_

PR #468, opened 2016-01-18
lux: use inverters for frsky telemetry

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/468)


mlyle:
Verified sensor hub works with this change.

PR #339, opened 2015-12-23
Usage statistics plugin

PTDreamer:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/339)

edit by mlyle: fixes #75

PR #469, opened 2016-01-18
GCS: Obfuscate chip serial # in AT results

mluessi:
Use a cryptographic hash to obfuscate the chip serial number in the shared autotune results. 

The rationale for doing this is that the serial number may be used for other things, like warranty claims or registering the controller with a manufacturer, so the serial number should be kept private. While this is not 100% secure (it may be possible to guess a valid serial # due to limited entropy), this is IMO better than not obfuscating the number at all.

Compile but not tested. This should go into the Renatus release, otherwise we will not be able to associate boards with each other in the future.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/469)

### DIFF
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -28,6 +28,7 @@
 
 #include "configautotunewidget.h"
 
+#include <QCryptographicHash>
 #include <QDebug>
 #include <QStringList>
 #include <QWidget>
@@ -506,7 +507,7 @@ QJsonDocument ConfigAutotuneWidget::getResultsJson()
 
     QJsonObject json;
     json["dataVersion"] = 1;
-    json["uniqueId"] = QString(utilMngr->getBoardCPUSerial().toHex());
+    json["uniqueId"] = QString(QCryptographicHash::hash(utilMngr->getBoardCPUSerial(), QCryptographicHash::Sha256).toHex());
 
     QJsonObject vehicle, fw;
     fw["board"] = autotuneShareForm->getBoardType();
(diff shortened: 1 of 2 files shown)


tracernz:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/469)*

tracernz:

---

*[ground/gcs/src/plugins/config/configautotunewidget.cpp, line 510 \[r1\]](https://reviewable.io:443/reviews/d-ronin/dronin/469#-K8JmW4RMuGEhRP9D1Zr:-K8JmW4Stvpd4ewFzyJb:2087645262) ([raw file](https://github.com/d-ronin/dronin/blob/907f15747262b0f907f4ac03f7a01677fa6bbdcc/ground/gcs/src/plugins/config/configautotunewidget.cpp#L510)):*
If the UUID is empty we should probably not hash it as hashing it makes it appear non-empty.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/469)*

tracernz:
Also need to do this in usage tracker? https://github.com/d-ronin/dRonin/pull/339/files#diff-b4957b270a3e8e93cfedbf157d26a687R248

mluessi:
@tracernz added to usage tracker. 

see also https://github.com/d-ronin/autotown/issues/2

mlyle:
We should have done this on the autotown side, but since we're coming up to release it seems like this should merge.

mlyle:
:lgtm:

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/469#-:-K8Lfm-PW_0MiM_9uLts:1417088721)*

mluessi:
@mlyle I think we should do both, send a hash and don't publicly show the hash in autotown. If the hash is public, it may also be abused to submit bogus results that then get associated with a submission by another user (-> bad user experience).

tracernz:
>We should have done this on the autotown side, but since we're coming up to release it seems like this should merge.

Discussed that with @dustin and he thought it would be too difficult to do it there.

dustin:
Not showing data that got sent is possible, but not sending it is better if
we can do that.

PR #473, opened 2016-01-18
flight: make failsafe into flight mode

mlyle:
This refactors most of the handling of failsafe from ManualControl to Stabilization.  It also allows configuring a flight mode for failsafe behavior.  This is going to be useful in testing more advanced failsafe situations.

Note to @ekaszubski since he's doing work in related areas.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/473)


mlyle:
Untested. WIP.  Compiles.

tracernz:
How do you envisage this to work with extendable/configurable failsafe which can do user-configured things like AH+Level, PH and RTH?

mlyle:
@ekaszubski is working on moving AH-- well, "thrust axis" control-- to stabilization.  So that will be relatively clear cut, and we can put sequencing/logic down into here for now for the next release or two.

Ultimately, there's going to need to be a higher level choreography for mode selection.  It's going to involve a big refactor, probably with lots of tasks going away.  At which point some of the logic that we've moved to stabilization module will have to factor out yet again.

mlyle:
(Though probably these actual bits-- the logic for "fall out of sky/last-resort failsafe mode"-- would remain in stabilization).

ekaszubski:
@mlyle this looks good to me in general.

Might be nice to have a table for various alarms or critical events (e.g. loss of RX, geofence, etc) with an option to select a failsafe/recovery flight mode for each so that failsafe behavior can be more intelligent instead of just "drop out of the sky" or "spiral down". For example, on loss of RX, you could have your vehicle go into RTH, or on geofence alarm, you could have your vehicle go into loiter.

I think you also mentioned in the past either having a hierarchy of fallback strategies in the event that a given failsafe/recovery flight mode can't be used (e.g. no gps fix yet) or preventing arming until the conditions to enter all selected failsafe/recovery modes are met.

The downside is that users can configure "risky" failsafe behavior.

mlyle:
Rebased.  Unverified.

mlyle:
Need to change this to display FAILSAFE in flight status.

mlyle:
Tested on naze32, works nice.

ekaszubski:
@mlyle Sorry for the wall of text.

I thought about this some more; IMO to make this a more complete/meaningful PR, the failsafe system should not just be a flight mode, it should be an extra bank of flight mode switch positions or something of equivalent functionality, each of which is activated when different alarms are raised.
- Possible alarms include:
  - Critical failsafe (something is critically broken, no guarantee on vehicle's ability to control itself, should take safest vehicle-specific action to avoid flying away into the sunset)
  - GPS lost
  - Receiver connection lost
  - Battery voltage critical
  - User panic due to loss of LOS, vehicle orientation, etc
  - GeoFence or similar alarm conditioned on position/velocity/orientation/mix (could be min/max altitude, max velocity, crashing trajectory estimate)
- Thus, the user can select exactly what happens under various error scenarios; they can select anything from a flight mode to per-axis stabilization settings (maybe we limit this for now, so that they can only select "safe" flight modes for a given scenario); example for each scenario:
  - Critical failsafe, battery voltage critical: basically fall out of the sky; the user can specify full engine cutoff instead of attitude mode (the latter may be spinning the props as it hits the ground)
  - GPS lost: the user can select fallback to attitude mode or drop out of the sky (not sure what current GPS-lost behavior is)
  - Receiver connection lost, geofence or similar, user panic: loiter (if gps available) or drop out of the sky
- While an alarm is raised, failsafe_control should set the corresponding flight mode via StabilizationDesired (but not the RPYT values), and then the rest of the ManualControl module should continue to run as if nothing has changed (but should not touch the flight mode); if the flight mode set by failsafe_control is able to take user input and the scenario allows for user input (e.g. receiver is still connected) then the user can still control the vehicle (e.g. loiter mode). If it is not able to take user input (e.g. the current failsafe mode) then the behavior will still be correct.
- Each alarm used by the failsafe system should have an explicit priority such that more critical failsafe behaviors will override less critical behaviors
- The flight mode selected by the failsafe system should stick until the user manually clears it, either by landing and disarming, changing flight mode, etc. More specifically, if a failsafe condition is resolved by the system (e.g. receiver was lost but is now reconnected), the user should have to manually "approve" the transition back to the pre-alarm flight mode, to avoid unexpected behavior. This behavior could perhaps be augmented to use timeouts to enter/exit the alarm conditions, such that if the duration of the alarm condition is sufficiently brief, the system will automatically return to the pre-alarm flight mode.
- Setting alarm-specific failsafe modes would also allow for a natural degradation of autonomy as issues arise with the vehicle; for example, if the receiver connection is lost, the vehicle might enter loiter mode, but if GPS is lost or the battery drains too far from being in loiter mode and the receiver is never reconnected, the vehicle may still eventually just fall out of the sky if the corresponding flight modes are set. On the other hand, if the receiver is reconnected, the user could somehow indicate he/she is ready to retake control and continue operating the vehicle, either in the flight mode triggered by the alarm (loiter in this example) or by switching back to a normal flight mode
- The failsafe flight mode added in this PR would still be valid; it would basically be the default flight mode selection for some of the more critical alarms
- In the end, an alarm-based system (instead of a one-size-fits-all failsafe mode) should at worst perform as well as the current implementation, would ideally reduce the likelihood of a crash in some scenarios, and would naturally enable the use of more autonomous vehicle recovery behaviors

mlyle:
What I said before:

> Ultimately, there's going to need to be a higher level choreography for mode selection. It's going to involve a big refactor, probably with lots of tasks going away. At which point some of the logic that we've moved to stabilization module will have to factor out yet again.
> 
> (Though probably these actual bits-- the logic for "fall out of sky/last-resort failsafe mode"-- would remain in stabilization).

mlyle:
This achieves what it was built to-- 2 functional improvements and cleanup towards overall failsafe improvements.
- In onscreen display and GCS, failsafe mode is set.
- For people who have receivers that insist on continuing to generate PPM / PWM with preset values (e.g. some of the Spektrum RX) you can have the rx generate the flightmode switch position for safety without having to play games with control axis ranges, etc.  Safety improvement.

ekaszubski:
> For people who have receivers that insist on continuing to generate PPM / PWM with preset values (e.g. some of the Spektrum RX) you can have the rx generate the flightmode switch position for safety without having to play games with control axis ranges, etc. Safety improvement.

Ah, this definitely seems to be a valuable feature that I didn't realize this PR was providing.

mlyle:
@ekaszubski -- I don't know if I sent you this before but this is roughly the plan I'm following with failsafe stuff.  https://github.com/TauLabs/TauLabs/issues/1774   It explicitly mentions first going with something like this, transitioning to dumbly-holding-modes for very short times to ride through control failures, and then using a combination of alarms and convergence checks for failsafes.

mlyle:
And now with this, use of the switch position works nicely too.  Woot.

ekaszubski:
@mlyle #1774 is new to me; I'm glad we share a similar long-term vision for what this stuff should look like :)

mlyle:
Bench-tested brain.  "FAIL" shows up in the upper left when failsafe happens, either because of RX fail or selecting it on a flight mode switch.

There is an annoying side effect.. CFG:UNSAFE is present when MAN: NO RX is (because it's unsafe to arm in the failsafe flight mode.  Don't quite know how to work around this easily, and I think the balance of benefits from this outweighs the small downside.

mlyle:
Bunch of flight tests, including a couple of deliberate failsafes in flight.

dustin:
Failsafe as a flight mode... what will they think of next?  :lgtm: 

---

Review status: 0 of 12 files reviewed at latest revision, all discussions resolved.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/473#-:-K9nGfFaRm133-O_ix3z:-281095099)_

PR #434, opened 2016-01-13
flight: fixed switch arming bugs

henn1001:
1: SWITCHDELAY got stuck in ARM_STATE_ARMED_STILL_HOLDING which prevented it from auto disarming from low throttle etc. 

2: SWITCHDELAY & SWITCHTHROTTLEDELAY would rearm themselves after a low throttle disarm event. The rising edge detection prevents that.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/434)

PR #474, opened 2016-01-19
flight: don't traverse STILL_HOLDING for switches

mlyle:
This is necessary so that with switch modes you don't need to toggle
twice to rearm.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/474)


mlyle:
Introduced in #434.  Still better than the old behavior but should get corrected before release.

This continues to be evidence for #357 -- switch arming as written is super bad.

tracernz:
:lgtm_strong:  Pulling to test now

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/474#-:-K8Mw4EED5cX53fe6g06:-708429987)*

PR #496, opened 2016-01-22
Removed Manual flight mode; added Manual stabilization mode

### LINKED ISSUE #495: Push "Manual" flight mode into stabilization modes
Manual flight mode is dangerous for most VTOL aircraft. In the cases where manual control is desirable (flybar helicopters, some fixed-wing aircraft, and many ground vehicles), it's often only desirable on a single stabilization axis (e.g. yaw on flybar helicopters equipped with a gyro).

Thus, we should:
- Remove manual flight mode completely?
- Add per-axis manual stabilization mode to allow for better customization of stabilization
- Update alarms/warnings currently looking for manual flight mode to instead look for manual stabilization mode in stabilization1/2/3 flight modes

ekaszubski:
Fixes #495 
- Updated sanity checks
- Pushed transmitter -> actuator logic into stabilization
- Updated/removed references to manual mode in:
  - OnScreenDisplay
  - UAVOHoTTBridge
  - UAVOLighttelemetryBridge
  - UAVOMavlinkBridge
  - Various GCS plugins

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/496)


flight/Libraries/sanitycheck.c
```diff
@@ -101,11 +101,6 @@ int32_t configuration_check()
 
 	for(uint32_t i = 0; i < num_modes; i++) {
 		switch(modes[i]) {
-			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_MANUAL:
```
mlyle, line 104: I think we should keep manual, but we also need to keep manual off any axis of a multirotor.

~~Right now the net effect of this change is to make things less safe-- people can select "manual" thinking it means "rate" on a quadcopter and the alarm that prevents arming is no longer present.~~

edit: nevermind, I see that you have handled that below.   Still, the manual flight mode (and this block) should come back.

ekaszubski, line 104: OK, will add manual mode back. For future reference, we should have a warning for cpheli vehicles using manual mode.

tracernz, line 104: Agree. Nobody flies manual on the tail and you will get a nasty surprise if not expecting to.

tracernz:
I like the idea, but the new stabilization mode almost exactly duplicates STABILIZATIONMODE_NONE, hopefully we can just use it.

---

Reviewed 1 of 15 files at r1, 6 of 14 files at r2.

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1014 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjCveq52RjUY6OtyI:-K8rjCveq52RjUY6OtyJ:-824116423) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1014)):_
Indentation (of existing lines actually).

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1029 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjMoHqckNz8bnVuXk:-K8rjMoHqckNz8bnVuXl:-824116423) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1029)):_
Indentation (of existing lines actually).

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1042 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjNQDdLHrqnlw00W_:-K8rjNQDdLHrqnlw00Wa:-824116423) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1042)):_
Indentation (of existing lines actually).

---

_[ground/gcs/src/plugins/hitl/fgsimulator.cpp, line 152 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rkufTI2fq455kKAOG:-K8rkufTI2fq455kKAOH:-1988726795) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/ground/gcs/src/plugins/hitl/fgsimulator.cpp#L152)):_
Wrong indentation, and it looks like this adds an additional check that wasn't here before (the if(armed) check was only in manual mode before).

---

_[shared/uavobjectdefinition/stabilizationdesired.xml, line 9 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjc3_tyDZobxPQ2IU:-K8rjc3aAPHaz-j3ZDhV:1191792705) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/shared/uavobjectdefinition/stabilizationdesired.xml#L9)):_
This looks like it does the exact same thing as the None mode, can we just use that?

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496#-:-K8rlAR9vEJ_hC61r6Hk:-313602652)_

ekaszubski:
So I asked about STABILIZATIONMODE_NONE before taking this on; none actually implies "do not attempt to control this axis in any way" whereas manual implies "just forward the stick inputs to ActuatorDesired".

Manual also moves the logic related to ActuatorDesired out of the ManualControl module (where it is only referenced once, for manual mode) and puts it into the Stabilization module (where it's already referenced tons of times, for all stabilization modes).

---

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1014 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjCveq52RjUY6OtyI:-K8s1dWWGTXTdWZQnCE0:-334794633) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1014)):_
Most of the lines in this file use tabs for indentation, so I followed that precedent here.

The other lines in this section use 1 tab plus 5 spaces; is this the desired formatting? It seems a bit odd.

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1029 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjMoHqckNz8bnVuXk:-K8s3-pGni-BHVq2suVs:-2069836363) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1029)):_
Same as line 1014

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1042 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjNQDdLHrqnlw00W_:-K8s35OyNgIGdzmvFw_V:-2069836363) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1042)):_
Same as line 1014

---

_[ground/gcs/src/plugins/hitl/fgsimulator.cpp, line 152 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rkufTI2fq455kKAOG:-K8s-8YkiYB7ikpCYnAc:498565685) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/ground/gcs/src/plugins/hitl/fgsimulator.cpp#L152)):_
So in the original implementation, ManualControlCommand is used as the control source when the vehicle is in manual, and AcutatorDesired is used otherwise. However, when the vehicle is in manual mode, ActuatorDesired will convey identical information as ManualControlCommand. Thus, it is pointless to use both sources, and only ActuatorDesired should be used.

As for the armed check, it should either go away entirely (so the control input is always active) or should be condensed to one line; so if( flight mode is not manual OR flight status is armed ) -> set control source from ActuatorDesired. This would have identical functionality to the original implementation but with cleaner code.

Will convert tabs to spaces.

---

_[shared/uavobjectdefinition/stabilizationdesired.xml, line 9 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjc3_tyDZobxPQ2IU:-K8rzhM8XIgoSwj1C5BK:1917329837) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/shared/uavobjectdefinition/stabilizationdesired.xml#L9)):_
This was one of my first questions when starting the PR; none implies no control (autonomous or otherwise), manual implies we should forward sticks to ActuatorDesired.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496#-:-K8ryDmO5i5Bepq63ePc:30933813)_

tracernz:
Heh, the existing code does forward the sticks for none as well doesn't it? Seems silly to create another mode that does exactly the same thing.

---

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1014 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjCveq52RjUY6OtyI:-K8s5UNR_rnH-bZ51W23:-2147104377) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1014)):_
If it were me I'd reformat the existing (incorrectly formatted) code to match the new line you added. The line you added is correct, sorry for not being clear.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496#-:-K8s5MDJS3ZHK3Gpz75Y:758799896)_

tracernz:
FWIW: I found this diff MUCH easier to review on Github and then just enter the comments in Reviewable... I'm not liking it so much, the diff was near unreadable on it.

tracernz:
e.g. 

```
stabilization.Roll = (stab_settings[0] == STABILIZATIONDESIRED_STABILIZATIONMODE_NONE) ? cmd->Roll :
        (stab_settings[0] == STABILIZATIONDESIRED_STABILIZATIONMODE_MANUAL) ? cmd->Roll :
```

^The same.

ekaszubski:
Yep, saw that; basically the same in stabilization too. Seems like the _intended_ behavior of NONE should be clarified, as it does seem like it should be distinct from MANUAL in that it should effectively be disabled / set to 0.

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

_[flight/Modules/ManualControl/transmitter_control.c, line 1014 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rjCveq52RjUY6OtyI:-K8s8-u59s8Q7XDrMA3w:-2146059436) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/ManualControl/transmitter_control.c#L1014)):_
OK will do.

---

_[flight/Modules/Stabilization/stabilization.c, line 749 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8s70YAbimKvLIDDYpB:-K8s70YAbimKvLIDDYpC:245573367) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/flight/Modules/Stabilization/stabilization.c#L749)):_
Yep, you're right, NONE currently does basically the same thing. Is this the intended behavior? I believe the desired behaivor of NONE was described to me differently, i.e. literally no control over the axis.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496#-:-K8s7_7z35Jb6w88qQJj:-1849788660)_

tracernz:
It makes sense to me that no stabilization just passes through the inputs... doesn't that make sense to you? Setting the outputs to zero makes no sense.

tracernz:
Stabilization != none implies that the Stabilization module takes some action to stabilize the craft which it does not in "manual" stabilization mode.

ekaszubski:
> It makes sense to me that no stabilization just passes through the inputs... doesn't that make sense to you? Setting the outputs to zero makes no sense.

IMO none should be almost like failsafe (which @mlyle is turning into a flight mode) in that it overrides the corresponding outputs and forces them to neutral (in comparison, failsafe overrides them and sets them to values that make the vehicle spiral/drop down).

Manual on the other hand should do "manual stabilization" i.e. forward the corresponding manual inputs to ActuatorDesired.

Maybe NONE should be renamed to DISABLED for clarity, but either way I think the distinction between those two flight modes is important.

> Stabilization != none implies that the Stabilization module takes some action to stabilize the craft which it does not in "manual" stabilization mode.

Technically, the stabilization module _is_ taking some action to stabilize the craft, in that it's looking at StabilizationDesired and setting ActuatorDesired. The only difference is that those StabilizationDesired commands are coming directly from the user, so the stabilization function for manual mode is just a copy function (as opposed to a closed-loop controller for the other modes).

IMO the ManualControl module shouldn't be touching ActuatorDesired, only the Stabilization module should be touching ActuatorDesired.

Main questions:
1. Do you think there should be a mode that sets the ActuatorDesired value for a given axis to the neutral value for that axis? If so, I think that mode should be called DISABLED (stabilization) instead of NONE (stabilization)
2. If you say "no" to (1), then I think NONE (stabilization) should be renamed to MANUAL (stabilization) since its functionality is to implement a manual override of the corresponding vehicle axis via the Stabilization module
3. Regardless of the above, I think all references to ActuatorDesired should be removed from the ManualControl module as demonstrated in this PR, and the functionality of the manual _flight mode_ (not stabilization mode) should be implemented by setting the _stabilization modes_ of R/P/Y to the mode that is currently represented by the "none" mode so that the Stabilization module can have exclusive write access to ActuatorDesired (as demonstrated in this PR). See [this graph](http://robotics.usc.edu/~ekaszubski/files/dronin/datapath-nosmallf1.png); this part of the PR removes the awkward/unnecessary read/write between the ManualControl module and the ActuatorDesired UAVO.

ekaszubski:

---

_[ground/gcs/src/plugins/hitl/fgsimulator.cpp, line 152 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K8rkufTI2fq455kKAOG:-K8sIkNp7cWis0zlf1oC:50689049) ([raw file](https://github.com/d-ronin/dronin/blob/2b6bf12f71cfffdf3b15ff0a6362ed88b0c5b82c/ground/gcs/src/plugins/hitl/fgsimulator.cpp#L152)):_
So should I remove the arming check or add flight mode to the check? Note that the flight mode check can be bypassed in the original code by simply setting flight mode to stabilized1/2/3 and setting all stabilization modes to "None". Thus, the original check either doesn't cover all the bases or assigns some special meaning to the manual flight mode which may or may not be appropriate to apply to the equivalent scenario mentioned here (stabilized1/2/3 with all stabilization modes set to "None").

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496)_

ekaszubski:
@tracernz Ready for another review; performed all requested formatting changes. Got some feedback from the IRC channel in terms of naming, consensus was that a flight mode called "None" is ambiguous, it should instead be called "Manual". Thus, I went ahead and renamed all references to "NONE" stabilization mode to "MANUAL" stabilization mode.

@mlyle I added a stabilization mode called "Disabled" that sets corresponding ActuatorDesired values to 0 (and should thus set the corresponding actuators to their neutral positions). However, it seems like adding any stabilization mode will break existing configs. **Question**: do you think this "Disabled" flight mode is worthwhile, and if so, will placing it at the end of the list of stabilization modes in the UAVOs avoid breaking existing configs?

mlyle:
It'll break existing configs, but at least the normal migration strategy of export/import will work fine.  So sure, adding a disabled sounds fine.

tracernz:
Looks good. I'll do a detailed review pass when I get a chance.

ekaszubski:
@mlyle Looks github was down when jenkins tried to post the [build result](http://jar.lyle.org:8080/job/dron/690/console), so github thinks the build is incomplete; what do?

tracernz:
jenkins, test this please

tracernz:
:lgtm: @icee, when you're happy it's good.

---

Reviewed 8 of 10 files at r3, 2 of 2 files at r4.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496#-:-K9EELK93nF2mMHFH7Sp:1421317286)_

mlyle:
Good that we're through review :)

What test has this seen?

ekaszubski:
Literally no test; if you give me a checklist I can run through it with one of my helis.

mlyle:
Ideally I'd like to see normal flight test on this, and ground test of manual control of actuators for "both ways" of configuring manual (maybe with a couple scenarios mixed with an axis set to something else).

Also testing hte multirotor manual lockout thoroughly would be really good.  I can spend some time on some of this eventually, too, but I have a very very big backlog.

ekaszubski:
- [x] Bench test manual flight mode mode on helicopter
  - appears to disable all axes, investigating
  - there was some code in stabilization preventing ActuatorDesired from being updated when in manual flight mode
  - resolved
- [x] Bench test per-axis manual stabilization mode mixed with attitude mode on helicopter
  - success
- [x] Bench test per-axis disabled stabilization mode mixed with attitude mode on helicopter
  - success
- [x] Bench test multirotor manual lockout (FC only)
  - success
- [ ] Flight test with normal config

ekaszubski:
@tracernz in 5ff8dc7 I fixed a bug that was causing failsafe to be raised when in manual flight mode; can you please take a quick look and confirm it looks good?

mlyle:
@ekaszubski one more dumb thing to ask.. can you make sure that the copyright line is up to date for all files you changed, and that the additional clause now at the bottom of licensetemplate.txt is included in your changes?

ekaszubski:
@mlyle yeah will do

ekaszubski:
@mlyle copyright/license updated for this and for #569

mlyle:
Much appreciated.  Sorry for being slow on PRs.

ekaszubski:
@mlyle Looks like I won't have a chance to do a flight test until next week. However, since I can see all control outputs via swash movement and tail rotor pitch, I can confirm that the behavior is correct.

I could go one step further and unplug my motor, arm the heli, and test again (via swash + tail) to confirm correct behavior while armed and non-zero throttle.

tracernz:
Reviewed 9 of 9 files at r6.

---

_[flight/Modules/Stabilization/stabilization.c, line 37 [r6]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K9LPms-VZTHZvOxiqEi:-K9LPms0agO7hTbpPd9g:-906127895) ([raw file](https://github.com/d-ronin/dronin/blob/c35141591145716a11416c352e2981c2444352de/flight/Modules/Stabilization/stabilization.c#L37)):_
Might want to put the additional note on redistribution in.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496)_

ekaszubski:

---

_[flight/Modules/Stabilization/stabilization.c, line 37 [r6]](https://reviewable.io:443/reviews/d-ronin/dronin/496#-K9LPms-VZTHZvOxiqEi:-K9LQL1eZDri10KrfqrE:216812800) ([raw file](https://github.com/d-ronin/dronin/blob/c35141591145716a11416c352e2981c2444352de/flight/Modules/Stabilization/stabilization.c#L37)):_
Oh, not sure how I missed that. Will do.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496)_

tracernz:
Reviewed 1 of 1 files at r7.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/496)_

mlyle:
Flown (including forced-manual-axis-on-multirotor-luls) on naze32 multirotor.  looks good.  skimming code now.

PR #499, opened 2016-01-23
UAVOGen: Fix broken inherited enum defaults

tracernz:
When an enum type was inheriting from a parent field, and the child had a different list of options, the default value would be incorrect

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/499)


mlyle:
lgtm

tracernz:
For completeness, I did check the other generators and the Java one doesn't suffer this issue (it users strings for enums rather than ints), and MATLAB and Wireshark don't use default values.

PR #500, opened 2016-01-23
Lux: New board config plugin, Config UAVO tweaks, Board reboot button

tracernz:
- Board config plugin for Lux
- Reboot button function in ConfigTaskWidget allows reboots from the board config plugin (or anywhere for that matter, although it will return to the board config tab after reboot).
- Removed PicoC option from Lux VCP
- Set a default RxPort type (PPM), was there any reason this wasn't done before?

screenshot from 2016-01-23 13 46 43

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/500)

### DIFF
--- a/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
+++ b/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
@@ -30,6 +30,8 @@
 #include <QWidget>
 #include <QLineEdit>
 #include "uavsettingsimportexport/uavsettingsimportexportfactory.h"
+#include <coreplugin/connectionmanager.h>
+#include <coreplugin/icore.h>
 
 /**
  * Constructor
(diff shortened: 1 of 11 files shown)


tracernz:
Oops, missed out a dep, fixed and force-pushed.

ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
```diff
@@ -902,6 +914,16 @@ void ConfigTaskWidget::addReloadButton(QPushButton *button, int buttonGroup)
     connect(button,SIGNAL(clicked()),this,SLOT(reloadButtonClicked()));
 }
 /**
+ * Adds a button to reboot board
+ * @param button pointer to the reload button
```
mluessi, line 918: reboot button

mluessi:
Code looks good to me. Is the Receiver Port really a UART? If so, why does it only support PPM?

tracernz:
It supports some serial protocols too https://github.com/d-ronin/dRonin/blob/next/shared/uavobjectdefinition/hwlux.xml#L7. The UART is on a different pin to PPM so you need a solder jumper to select whether to use PPM or serial on the RX pin. It was done like that to preserve firmware compatibility with the TBS Colibri Race. :-(

mlyle:
Code looks good to me.  Smoke tests OK.  Will want to put more mileage on Lux in general this release.

PR #503, opened 2016-01-23
CC3D: Rename from coptercontrol, remove legacy CC stuff

### LINKED ISSUE #265: Rename coptercontrol -> CC3D
- Will be clearer for users that it works on CC3D
- Will be clearer for users that it doesn't work on original CC

tracernz:
Fixes #265
- Removes ADXL345
- Removes PFC8591
- Removes ADC driver from CC gyros
- Allows saving settings to flash on CC3D without expert mode enabled
- Numerous small tweaks

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/503)


flight/targets/cc3d/fw/pios_board.c
```diff
@@ -107,14 +107,8 @@ void PIOS_Board_Init(void) {
 	/* Delay system */
 	PIOS_DELAY_Init();
 
-	const struct pios_board_info * bdinfo = &pios_board_info_blob;
```
mlyle, line 110: Are you sure we should remove this?  It seems nice to refuse to start on original CopterControl.

mlyle:
Looks good to me other than that one thing.

flight/targets/cc3d/fw/pios_board.c
tracernz, line 110: Not a bad point.

tracernz:
Will also need to rename the board element in the connection diagram. That would conflict with #546 so maybe a good idea to merge it first.

mlyle:
I tried this and it's good.  Ready for merge after rebase.

mlyle:
I noticed one minor thing on windows:  if you install a release over an old one, the old firmware files are left there.  fw_coptercontrol.tlfw is very tempting to click on and results in very confusing outcomes.

tracernz:
I'm guessing we can set the installer up to clean the firmware directory if it's upgrading?

mlyle:
Yes, that would be sane.   http://nsis.sourceforge.net/Reference/RMDir

tracernz:
Rebased, just the removal of UAVObjects.inc causing the conflict.

tracernz:
Blocked on #546 + I need to make the change to the connection diagram once 546 is in.

tracernz:
Fixed now, even tested the connection diagram to make sure.

flight/targets/cc3d/fw/pios_board.c
tracernz, line 110: This code does nothing though? There's just a comment saying what it should do.

mlyle, line 110: Haha. OK, I'm awesome.

PR #504, opened 2016-01-23
gcs: simplify ccache usage

mlyle:
This formula works on Windows, too... lets the toplevel makefile to decide whether to ccache and where to find it.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/504)

### DIFF
--- a/ground/gcs/gcs.pri
+++ b/ground/gcs/gcs.pri
@@ -121,16 +121,18 @@ unix {
     CONFIG(debug, debug|release):MOC_DIR = $${OUT_PWD}/.moc/debug-shared
     CONFIG(release, debug|release):MOC_DIR = $${OUT_PWD}/.moc/release-shared
 
-    CONFIG(debug, debug|release) {
-# Unfortunately this is ineffective on OSX, due to
-# https://bugreports.qt.io/browse/QTBUG-39417
-# Should probe paths once upstream defect resolved
-        exists(/usr/bin/ccache):QMAKE_CXX="ccache g++"
-    }
-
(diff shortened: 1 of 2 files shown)


mlyle:
So far, this seems to work nicely.   I'm still going to poke at it a bit more.

A ccache.exe is at http://jar.lyle.org/~mlyle/ccache.exe   built from ccache-3.2.3 with a slight makefile tweak so zlib stuff could be found.

mlyle:
`make -j4 all` on a slow windows VM was 10 mins 5 seconds hitting the cache.  Not great, but better than builds were before.

tracernz:
Nice idea. If only it would work on OS X. ;-)

tracernz:
:lgtm_strong: 

---

Reviewed 2 of 2 files at r1.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/504#-:-K8hKUcb_-PJlvat_DO-:-1868698381)_

mlyle:
I spent a few minutes looking at if there was any new developments or anything I could reasonably do... not really. (re: osx)

droidicus:
Tested on Windows 10 with @mlyle provided ccache.exe above.
Before running make all_clean && make all -j4:

```
real    18m55.900s
user    0m0.015s
sys     0m0.016s
```

After running make all_clean && make all -j4 (after first run to load ccache):

```
real    4m49.475s
user    0m0.000s
sys     0m0.000s
```

Still about half the speed of the Linux VM I set up, but a significant speedup for sure.

PR #516, opened 2016-01-24
Config:CCPM: Don't crash due to short mixer vector

### LINKED ISSUE #511: gcs: Default selection for helicopter swashplate config crashes gcs on apply
Steps to reproduce:
- Set vehicle type: helicopter
- Swashplate config defaults to "Custom - Advanced Settings"
- Click apply -> gcs crashes

Issue:
- This swashplate config has no sane defaults for the mixer under the Advanced Settings tab

Solution:
- Manually changing all mixer values to 0 under the Advanced Settings tab allows settings to be applied
- These values should default to something sane

tracernz:
Minimal fix, CCPM config in general needs reworked/cleaned up. Also some style cleanup I can remove if required.

Fixes #511

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/516)

### DIFF
--- a/ground/gcs/src/plugins/config/cfg_vehicletypes/configccpmwidget.cpp
+++ b/ground/gcs/src/plugins/config/cfg_vehicletypes/configccpmwidget.cpp
@@ -917,7 +917,7 @@ void ConfigCcpmWidget::setMixer()
     UpdateMixer();
 
     // Set up some helper pointers
-    decltype(&mixerSettingsData.Mixer1Vector[0]) mixers[8] = {
+    decltype(&mixerSettingsData.Mixer1Vector[0]) mixers[] = {
         mixerSettingsData.Mixer1Vector,
         mixerSettingsData.Mixer2Vector,
         mixerSettingsData.Mixer3Vector,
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/config/cfg_vehicletypes/configccpmwidget.cpp
```diff
@@ -936,33 +938,34 @@ void ConfigCcpmWidget::setMixer()
...
-        *mixerTypes[i] = 0;
+    for (i = 0; i < 8; i++)
+        *(mixerTypes[i]) = 0;
 
     //go through the user data and update the mixer matrix
-    for (i=0;i<6;i++)
+    for (i = 0; i < 6; i++)
     {
-        if (MixerChannelData[i]>0)
+        if (MixerChannelData[i] > 0 && (uint)MixerChannelData[i] < sizeof(mixerTypes) / sizeof(mixerTypes[0]))
```
mlyle, line 953: do we not have an NELEMENTS thing in GCS? :(

tracernz, line 953: Not that I found. I'd like to add one but really I'm not a big fan of how this code works so I intend to rewrite it, hence minimal fix here.

ekaszubski:
@tracernz I can verify that this fixes all the crashes I was experiencing.

Issue #83, opened 2015-11-29
Recurring situations where old configuration causes boot-time reset


mlyle:
We need to run this problem down and kill it with fire.

mlyle:
Relates to #3

tracernz:
Should be fixed by #801

tracernz:
@mlyle, since you're not planning to use upgrader (?) for pipxtreme, this issue is not fixed?

mlyle:
I think it'd be much more challenging to fill pipxtreme flash-- like, OK, technically possible, but...

mlyle:
(only one object stored on pipxtreme)

tracernz:
Assuming that filling the flash is the issue ;-).

mlyle:
;) well, if it's not... the issue has never been seen on pipx, so...

tracernz:
Did that issue the other day turn out to be something else?

mlyle:
I think they reflashed it again and it started working.

PR #521, opened 2016-01-24
Flight: Erase old settings UAVOs on boot

tracernz:
Previously settings UAVOs left over from old firmware versions would be left after an upgrade, taking up slots in the filesystem. It is suspected that the the cause of problems seen where the board fails to boot after an upgrade unless the settings partition is erased, is due to the partition becoming full. The UAVO manager now cycles through each slot in the settings partition, and if an unknown UAVO ID is found it is deleted.

It would be great to find some test cases where we failed to boot before to verify that this fixes it.

Connects to #83

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/521)

### DIFF
--- a/flight/UAVObjects/uavobjectmanager.c
+++ b/flight/UAVObjects/uavobjectmanager.c
@@ -178,6 +178,7 @@ static int32_t connectObj(UAVObjHandle obj_handle, struct pios_queue *queue,
 			uint16_t interval);
 static int32_t disconnectObj(UAVObjHandle obj_handle, struct pios_queue *queue,
 			UAVObjEventCallback cb, void *cbCtx);
+static int32_t cleanupSettingsObjs(void);
 
 // Private variables
 static struct UAVOData * uavo_list;
@@ -229,6 +230,10 @@ int32_t UAVObjInitialize()
 	mutex = PIOS_Recursive_Mutex_Create();
 	if (mutex == NULL)
 		return -1;
+
+	// Clean up old settings objects
+	cleanupSettingsObjs();
+
 	// Done
 	return 0;
 }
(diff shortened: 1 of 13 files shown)


tracernz:
Ooops, forgot to do the simulator linker script.

tracernz:
Well, turns out it doesn't even use it's linker script... sigh

flight/UAVObjects/uavobjectmanager.c
```diff
@@ -2177,6 +2182,42 @@ void UAVObjRegisterNewInstanceCB(new_uavo_instance_cb_t callback)
...
+	do {
+		res = PIOS_FLASHFS_ObjIdAtSlot(pios_uavo_settings_fs_id, slot++, 
+					&curr_obj_id, &curr_obj_inst_id) == 0;
+
+		if (res == -3) {
+			// empty slot, ignore it
+			res = 0;
+			continue;
+		}
+
```
mluessi, line 2203: We should handle other error conditions here (`res` can be negative and the code below gets executed anyways).

tracernz, line 2203: ahhhh, I am dumb.... I overwrote that code when I added this bit and forgot to re-write it.

tracernz, line 2203: Thanks. :)

tracernz:
Force-pushed a fix for that little issue. I'll have to do something for the simulator.

mlyle:
So the thing I don't like about this is that even the small targets end up knowing about all UAVOids and it wastes a bit of flash.  I have ideas on how to do magic and mitigate, though..

tracernz:
Another potential cause of this issue (I haven't read the code to understand it yet): https://bitbucket.org/librepilot/librepilot/pull-requests/142/lp-83-unbrick-revo-and-sparky2/diff

tracernz:
This will be obsoleted by new firmware upgrade work.

tracernz:
This can be left now. Upgrade wizard will render it unnecessary.

PR #522, opened 2016-01-24
gcs: Fix default collective curve being ignored on initial setup

### LINKED ISSUE #401: GCS: Vehicle Setup: Helicopter: Default curve2 should be set for CP helicopters
- Default curve2 (collective pitch) for helicopters only allows for positive collective pitch whereas it should be set to allow for the full range of collective pitch
- Specifically, the points should be [-1,-1][-0.5,-0.5][0,0][0.5,0.5][1,1] instead of [-1,0][-0.5,0.25][0,0.5][0.5.0.75][1,1]

ekaszubski:
When changing vehicle type to helicopter, the collective curve has its range on [0,1] even though the code clearly tries to have it default to [-1,1]. It seems as though refreshAirframeWidgetsValues() triggers some complex behavior that resets the the collective curve range to [0,1] regardless of the desired value. 

Thus, the simple fix presented here configures the collective curve range after the call to refreshAirframeWidgetsValues().

Resolves #401.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/522)

### DIFF
--- a/ground/gcs/src/plugins/config/cfg_vehicletypes/configccpmwidget.cpp
+++ b/ground/gcs/src/plugins/config/cfg_vehicletypes/configccpmwidget.cpp
@@ -130,13 +130,9 @@ ConfigCcpmWidget::ConfigCcpmWidget(QWidget *parent) : VehicleConfig(parent)
 
     }
 
-    //initialize our two mixer curves
+    //initialize our throttle mixer curve
     // mixercurve defaults to mixercurve_throttle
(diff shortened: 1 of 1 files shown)


tracernz:
:lgtm: 

---

Reviewed 1 of 1 files at r1.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/522#-:-K8rl_Mn49CVU6Pwzr9O:980077423)_

PR #526, opened 2016-01-24
Lux: Un-fudge UART configs

tracernz:
To match production version. Please review throughly in-case I'm dumb (quite likely).

```
<ReadError> usart1_tx = PC4
<ReadError> usart1_rx = PC5
<ReadError> usart2_tx = PA14
<ReadError> usart2_rx = PA15
<ReadError> usart3_tx = PB10
<ReadError> usart3_rx = PB11
<ReadError> RX pin on front = usart1_rx(pc5) and ppm (pa8)
```

Also, we may want to allow use of UART1 tx at some point but I'm leaving that off for now since it'd require more invasive changes.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/526)

### DIFF
--- a/flight/targets/lux/board-info/board_hw_defs.c
+++ b/flight/targets/lux/board-info/board_hw_defs.c
@@ -318,7 +318,7 @@ static const struct pios_usart_cfg pios_rcvr_dsm_hsum_cfg = {
 	},
 	.irq = {
 		.init = {
-			.NVIC_IRQChannel = USART2_IRQn,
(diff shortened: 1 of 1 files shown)


mlyle:
This was rather extensive.  Thanks for doing it.  I'll check the couple of cases of serial I'm set up to use on Lux soon.

mlyle:
Spot checked the code.  Tested frsky telem on the board..  I can't fully check until I have final hardware, but this looks good.

PR #528, opened 2016-01-24
Move discoveryf4/coptercontrol to new flash erase paradigm

### LINKED ISSUE #518: CC3D/DiscoveryF4: Compiling with ERASE_FLASH=YES broken
Broken in 2013 by @solidgoldbomb in https://github.com/d-ronin/dRonin/commit/ac53e9d518c94c3d785eeed8466c1cd043c291f8#diff-c64214d8dc83380dce317d7c0aa66a18.
We should remove this on the Disco since F4 has partition manager so it's unnecessary. Need to write some new code to erase all the sectors on CC3D since it doesn't have a partition manager so there's no way to erase settings that cause a reboot loop.

mlyle:
Fixes #518

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/528)

### DIFF
--- a/flight/targets/coptercontrol/fw/pios_board.c
+++ b/flight/targets/coptercontrol/fw/pios_board.c
@@ -142,6 +142,11 @@ void PIOS_Board_Init(void) {
 
 #if defined(ERASE_FLASH)
 	PIOS_FLASHFS_Format(pios_uavo_settings_fs_id);
+#if defined(PIOS_LED_HEARTBEAT)
+	PIOS_LED_Off(PIOS_LED_HEARTBEAT);
+#endif /* PIOS_LED_HEARTBEAT */
+
+	while (1) ;
 #endif
 
 	/* Initialize the task monitor library */
--- a/flight/targets/coptercontrol/fw/pios_board.c
+++ b/flight/targets/coptercontrol/fw/pios_board.c
@@ -140,6 +140,10 @@ void PIOS_Board_Init(void) {
 	/* Mount all filesystems */
 	PIOS_FLASHFS_Logfs_Init(&pios_uavo_settings_fs_id, &flashfs_settings_cfg, FLASH_PARTITION_LABEL_SETTINGS);
 
+#if defined(ERASE_FLASH)
+	PIOS_FLASHFS_Format(pios_uavo_settings_fs_id);
+#endif
+
 	/* Initialize the task monitor library */
 	TaskMonitorInitialize();
 
--- a/flight/targets/aq32/fw/Makefile
+++ b/flight/targets/aq32/fw/Makefile
@@ -31,7 +31,6 @@ include $(BOARD_INFO_DIR)/board-info.mk
 # Set developer code and compile options
 # Set to YES for debugging
 DEBUG ?= NO
-ERASE_FLASH ?= NO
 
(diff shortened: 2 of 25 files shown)


mlyle:
Still untested.

flight/targets/coptercontrol/fw/pios_board.c
```diff
@@ -140,6 +140,10 @@ void PIOS_Board_Init(void) {
 	/* Mount all filesystems */
 	PIOS_FLASHFS_Logfs_Init(&pios_uavo_settings_fs_id, &flashfs_settings_cfg, FLASH_PARTITION_LABEL_SETTINGS);
 
+#if defined(ERASE_FLASH)
+	PIOS_FLASHFS_Format(pios_uavo_settings_fs_id);
+#endif
+
```
tracernz, line 151: So it no longer just turns the LED off and stops here, it will continue booting firmware? I'm not sure that's a good idea because the user may end up wondering why his board never keeps settings, especially if he sits it down for a few weeks after erasing or something.

mlyle, line 151: I thought about halting.  This is consistent with the other ERASE_FLASH targets, though.

tracernz, line 151: It's not consistent with the other one in this PR. :/

mlyle, line 151: Haha.  touche.  But all the ones that have this in pios_board.c .  And I've not analyzed the safety of halting there in pios_board.

tracernz, line 151: PIOS_Board_Init is not called from an RTOS task on CC3D, it's called from main.

mlyle, line 151: I'd really like to avoid pios_boards further diverging because i'd like to work on factoring them common.  That being said, if you really object I'll make it so CC3D (but none of our other targets) pause after flash erase.

tracernz, line 151: Well we can remove this whole thing from F3/F4 since they have partition manager available in the bootloader. I really don't like that some boards carry on running the firmware after erasing, we should make a point of this not being for regular running IMO which is what was originally done here.

mlyle, line 151: So would you prefer in this PR I rip it out from F3/F4 and then have it -only- in CC3D and pipxtreme?

tracernz, line 151: And Naze32. Sounds good to me if you agree?

mlyle, line 151: Per IRC, no naze32.

mlyle:
Need to cram this into pipxtreme, too.

tracernz:
:lgtm_strong: 

---

Reviewed 1 of 3 files at r1, 23 of 23 files at r2, 1 of 1 files at r3.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/528#-:-K92WV87SkWlv8teIcub:-1868698381)_

tracernz:
Tested on CC3D.

PR #531, opened 2016-01-25
Add Avg hover throttle

droidicus:
Used a scaled integer summation of the throttle, and convert to average throttle on logging or at the end of the autotune run.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/531)

### DIFF
--- a/flight/Modules/Autotune/autotune.c
+++ b/flight/Modules/Autotune/autotune.c
@@ -59,9 +59,9 @@
 enum AUTOTUNE_STATE { AT_INIT, AT_START, AT_RUN, AT_FINISHED, AT_WAITING };
 
 struct at_queued_data {
-	float y[3];	/* Gyro measurements */
-	float u[3];	/* Actuator desired */
-	float t;	/* Throttle desired */
+	float y[3];		/* Gyro measurements */
+	float u[3];		/* Actuator desired */
+	float throttle;	/* Throttle desired */
 
 	uint32_t raw_time;	/* From PIOS_DELAY_GetRaw() */
 };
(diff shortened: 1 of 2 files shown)


shared/uavobjectdefinition/systemident.xml
```diff
@@ -12,5 +12,6 @@
         <telemetrygcs acked="false" updatemode="manual" period="0"/>
         <telemetryflight acked="false" updatemode="onchange" period="1000"/>
         <logging updatemode="manual" period="0"/>
+	<field name="HoverThrottle" units="% / 100" type="float" elements="1" defaultvalue="0"/>
```
tracernz: Should go with the other fields, also indentation is mixed throughout the file but you can ignore that if you want.

flight/Modules/Autotune/autotune.c
```diff
@@ -61,6 +61,7 @@ enum AUTOTUNE_STATE { AT_INIT, AT_START, AT_RUN, AT_FINISHED, AT_WAITING };
 struct at_queued_data {
 	float y[3];	/* Gyro measurements */
 	float u[3];	/* Actuator desired */
+	float t;	/* Throttle desired */
```
tracernz: It wouldn't be a killer to give this another name? t is usually reserved for time and this will be confusing when reading code below.

droidicus:
Applied the previous code review comments, and also changed the variable relay -> system_ident... that had been bugging me...

droidicus:
Tested and verified that it is sending the value to autotown: http://dronin-autotown.appspot.com/api/tune?tune=ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAgMfhogsM

mlyle:
I suggest you use a float accumulator and the Kahan summation algorithm instead--

https://en.wikipedia.org/wiki/Kahan_summation_algorithm

droidicus:
@mlyle, I am looking at this, but am missing one piece of information, especially for F1 processors... Would the Kahan algorithm with flotats be less processor intensive then a simple summation with doubles? Using doubles should have plenty of precision from my estimation.

mlyle:
Another option is just summing up throttle*1000 in an integer.  Cheap on F1, accurate enough.

mlyle:
Thanks much!

mlyle:
Just waiting for Jenkins.

PR #540, opened 2016-01-26
GCS: Quieten down a lot of debug warnings

tracernz:
This disables a lot of (imo unncecessary) debug warnings with defines to re-enabled them on a per-module level as required (following existing GCS code with this). The debug console is now usable for debugging. Also removed the last of the tlfw_resource.qrc stuff.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/540)


mlyle:
This is very nice to quiet some of this stuff down.

I want to eventually build some infrastructure so that pieces can be toggled on and off at runtime.

PR #543, opened 2016-01-26
Sparky2: Allow DAC pin to be used as ADC, cleanup GPIO

tracernz:
Firstly, cleans up a pointless GPIO (as far as I can tell). We were initialising the USB Vsense pin as an input and calling it the current/sonar pin for reasons I do not know.

Mostly, adds the (currently unused) DAC out pin to the list of available ADC inputs. This allows three ADCs for setups like Volt/Current/RSSI. The only other choice would be to use PWM output 5 and/or 6. The DAC pin is the best choice because:
- It's unused
- It's on the same connector as the other ADC pins
- It has a 10 k series resistor on-board (the PWM outs would require this to be done externally)

We can add a configuration option to select whether this should be used as a DAC or ADC later if we do end up adding DAC functionality.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/543)


tracernz:
P.S. Bench tested this already.

tracernz:
Turns out that sonar/current thing was a copy-paste from the Revo.

PR #546, opened 2016-01-27
Revo/CC3D: Fix PPM on revo connection diagram, remove OP branding

### LINKED ISSUE #545: Revo: Connection diagram shows wrong PPM pin
TauLabs/TauLabs#1775 added an 8 to the connection diagram not realising it was shared with the Revo which uses pin 5. If we are to keep the 8 on the diagram I think we need to split the CC3D and Revo into separate diagrams.

tracernz:
This is a potential solution to #545. I added a small white box to the Revo board object that covers the (8) drawn when PPM is selected. To make this work I then raised the z-index of the board object so it appears above the PPM object (and all other objects). The downside of this is we can no longer have anything drawn over top of the board, e.g. wires going to pins in the middle of the board. Do we think we will need to do that? The only example I can think of off-hand is the telemetry pins on the Naze32.

Also removes OpenPilot branding from the vehicle images in the diagram.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/546)

edit: fixes #545

### DIFF
--- a/ground/gcs/src/plugins/setupwizard/connectiondiagram.cpp
+++ b/ground/gcs/src/plugins/setupwizard/connectiondiagram.cpp
@@ -84,10 +84,6 @@ void ConnectionDiagram::setupGraphicsScene()
 
         QList<QString> elementsToShow;
 
-        Core::IBoardType* type = m_configSource->getControllerType();
-        if (type != NULL)
-            elementsToShow << QString("controller-").append(type->shortName().toLower());
-
         switch (m_configSource->getVehicleType()) {
         case VehicleConfigurationSource::VEHICLE_MULTI:
             switch (m_configSource->getVehicleSubType()) {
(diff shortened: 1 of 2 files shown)


ground/gcs/src/plugins/setupwizard/connectiondiagram.cpp
```diff
@@ -141,6 +137,10 @@ void ConnectionDiagram::setupGraphicsScene()
             break;
         }
 
+        Core::IBoardType* type = m_configSource->getControllerType();
+        if (type != NULL)
+            elementsToShow << QString("controller-").append(type->shortName().toLower());
```
mluessi, line 142: There is no functional change; just the code moved down, correct?

tracernz, line 142: It adds the item to the list last so it ends up with a higher z-index later, which makes it appear over top of all other elements instead of under them.

PR #551, opened 2016-01-28
Uploader: Only show firmware images by default in file dialog

tracernz:
Makes it less confusing for new users.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/551)


mlyle:
Very nice compromise around the issues we talked about.

PR #557, opened 2016-01-28
Flight: Allow disabling of mag. for Sparky2

henn1001:
Other targets are capable of that doing that.

Compiled & tested => works

https://github.com/henn1001/dRonin/blob/local/flight/targets/sparky2/fw/pios_board.c#L699

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/557)

### DIFF
--- a/shared/uavobjectdefinition/hwsparky2.xml
+++ b/shared/uavobjectdefinition/hwsparky2.xml
@@ -116,7 +116,7 @@
 			</options>
 		</field>
 
-		<field name="Magnetometer" units="function" type="enum" elements="1" options="Internal,ExternalI2CFlexiPort,ExternalAuxI2C" defaultvalue="Internal"/>
+		<field name="Magnetometer" units="function" type="enum" elements="1" options="Disabled,Internal,ExternalI2CFlexiPort,ExternalAuxI2C" defaultvalue="Internal"/>
 		<field name="ExtMagOrientation" units="function" type="enum" elements="1" parent="HwShared.MagOrientation" defaultvalue="Top0degCW" />
 
 		<access gcs="readwrite" flight="readwrite"/>


mlyle:
jenkins, test this please

mlyle:
What's the benefit of this, out of curiosity?

mlyle:
jenkins, test this please

henn1001:
My inability to properly calibrate the mag haha. I always get a lot of yaw drift in attitude/gcs no matter what i try.

Maybe i don´t understand the logic behind it yet though.. Out of curiosity - is the mag influence for attitude.yaw disabled while armed? Think i´ve read somewhere that the complementary filter doesn´t use mag.

mlyle:
jenkins, test this please --- jenkins was out of disk space, so hopefully fixed

@henn1001 the magnetometer only very, very very slowly affects yaw.  it's mostly affected by gyro.  So it's OK if it's badly calibrated

tracernz:
This isn't needed on S2 since it doesn't share the external I2C bus with an internal mag. There is one I2C flexi-port with nothing on it, and one with just the baro.

henn1001:
But it does not boot if you select either of these options without a available mag on those ports? (it least it does not for me)

tracernz:
Let's do this. I'm hearing reports that the mag causes jitter with MPU-9250 (not confirmed though).

mlyle:
@tracernz that's interesting, would be good to find out root cause.

tracernz:
Need to verify that this won't have bad side effects on the attitude filter etc.

mlyle:
jenkins, test this please

mlyle:
Fine, we'll pick this up for now.  Thanks :)

PR #520, opened 2016-01-24
flight: autotune: always save on disarm.

mlyle:
Before, we'd only save if you landed and disarmed while still in autotune mode.  Now, if you've completed an autotune, switch to another mode, and then disarm.. it'll save the systemident object too.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/520)

PR #411, opened 2016-01-08
Brain: Move to PIOS_HAL

jihlein:
Implement PIOS_HAL on Brain.  Compiles clean, needs testing by someone with Brain FC board.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/411)

PR #139, opened 2015-12-04
uavobjects: conserve stack some and eliminate extra copy

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/139)

PR #563, opened 2016-01-29
systemident: set GCS perms readwrite.

### LINKED ISSUE #561: GCS: Unable to import Systemident values BUG
I can export systemident values http://i.imgur.com/kauRhgI.png but i´m somehow unable to import them successfully.

Even after manually pasting them into the fields http://i.imgur.com/5jfnHqC.png and hitting send/save, reverts them back to 0.

mlyle:
Sometimes it's useful to nudge things around.  Sometimes it's good to be
able to import an old tune from an exported settings set.

Fixes #561

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/563)

### DIFF
--- a/ground/gcs/src/plugins/uavobjectwidgetutils/smartsavebutton.cpp
+++ b/ground/gcs/src/plugins/uavobjectwidgetutils/smartsavebutton.cpp
@@ -2,6 +2,8 @@
  ******************************************************************************
  *
  * @file       smartsavebutton.cpp
+ *
+ * @author     dRonin, http://dRonin.org/, Copyright (C) 2016
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
  *
@@ -25,6 +27,10 @@
  * You should have received a copy of the GNU General Public License along
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ *
+ * Additional note on redistribution: The copyright and license notices above
+ * must be maintained in each individual source file that is a derivative work
+ * of this source file; otherwise redistribution is prohibited.
  */
 #include "smartsavebutton.h"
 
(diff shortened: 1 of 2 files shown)


mlyle:
Unverified / untested.

henn1001:
Tested and i can verify that this fixes the issue - thanks!

mlyle:
@henn1001 Cool, thank you!

mlyle:
Looks good to me too.  Ready for merge?

mlyle:
rebased

mlyle:
Woops! This needs to be held back because it prevents GCS from saving systemident object on autotune pane..... which it really shouldn't be trying to do even.

mlyle:
With this additional change, things seem happy.  Turns out it was unhappy because it doesn't know how to translate some things to the uavo format (e.g. log).  But just as well anyways, this shouldn't be enabled for saving.

mlyle:
Additional test, etc, appreciated

mlyle:
Happy on naze32 brushed quad, through enablement to share.

droidicus:
Tested along with #520, #411, #563, #139 on Sparky2.

Confirmed that with the sequence AT#1, export UAV, AT#2, import UAV from AT#1, power cycle, that the SystemIdent UAVO values are set to the values from AT#1.

ground/gcs/src/plugins/uavobjectwidgetutils/smartsavebutton.cpp
```diff
@@ -158,7 +164,8 @@ void smartSaveButton::processOperation(QPushButton * button,bool save)
 
         if(upload_result==false) {
             qDebug() << "[smartsavebutton.cpp] Object upload error:" << obj->getName();
-            failedUploads.append(obj->getName());
+            if(mandatoryList.value(obj, true))
```
mluessi, line 167: Why was this change necessary?

mlyle, line 167: There was a long-standing bug in Tau where attempting to turn on the autotune module would return a message about being unable to save the autotune object.  This is because there was no function to convert from display format to uavo format (e.g. milliseconds to exponential tau), but we don't want to save it anyways.

I previously fixed this at TL by setting the metadata to readonly, where it catches on the other if condition and suppressed that error.  But it's sometimes nice to be able to change that value--- either to import a uavo file with a previous set of tune measurements or to be able to override inputs to the autotune stuff.  

The mandatory list was checked for permissions but wasn't checked for these other situations.  This PR changes it back to read-write, and changes other places where savings can fail to check the mandatory list.

mluessi, line 167: Thanks, makes sense.

mlyle:
Thanks for merge :)

PR #571, opened 2016-01-29
Flight: Add raw and comp GPS messages to MSP protocol

jihlein:
Bench tested at various home and ownship positions and ranges.  MWOSD display seems to be correct.  

Need someone to check my math.  Also think about what math simplifications might be made given the display resolution.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/571)

### DIFF
--- a/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
+++ b/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
@@ -394,7 +394,7 @@ static void msp_send_comp_gps(struct msp_bridge *m)
 		GPSPositionGet(&gps_data);
 		HomeLocationGet(&home_data);
 		
-		if((gps_data.Status < GPSPOSITION_STATUS_FIX2D) || (home_data.Set == FALSE))
+		if((gps_data.Status < GPSPOSITION_STATUS_FIX2D) || (home_data.Set == HOMELOCATION_SET_FALSE))
 		{
 			data.comp_gps.distance_to_home    = 0;
 			data.comp_gps.direction_to_home   = 0;
--- a/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
+++ b/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
@@ -387,7 +387,7 @@ static void msp_send_comp_gps(struct msp_bridge *m)
 	
 	deltaY *= cosf((float)homeData.Latitude * 1e-7f * (float)DEG2RAD);  // Latitude compression correction
 	
-	data.comp_gps.distance_to_home  = (uint16_t)(sqrt(deltaX * deltaX + deltaY * deltaY) * 1e-4);  // meters
+	data.comp_gps.distance_to_home  = (uint16_t)(sqrtf(deltaX * deltaX + deltaY * deltaY) * 1e-4f);  // meters
 	
 	if ((deltaLon == 0) && (deltaLat == 0))
 		data.comp_gps.direction_to_home = 0;
--- a/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
+++ b/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
@@ -253,10 +253,17 @@ static void msp_send_status(struct msp_bridge *m)
 	// TODO: https://github.com/TauLabs/TauLabs/blob/next/shared/uavobjectdefinition/actuatordesired.xml#L8
 	data.status.cycleTime = 0;
 	data.status.i2cErrors = 0;
-	data.status.sensors =
-		(PIOS_SENSORS_IsRegistered(PIOS_SENSOR_ACCEL) ? MSP_SENSOR_ACC : 0)
-		| (PIOS_SENSORS_IsRegistered(PIOS_SENSOR_BARO) ? MSP_SENSOR_BARO : 0)
-		| (PIOS_SENSORS_IsRegistered(PIOS_SENSOR_MAG) ? MSP_SENSOR_MAG : 0);
+	
+	GPSPositionData gpsData = {};
+	
+	if (GPSPositionHandle() != NULL)
+		GPSPositionGet(&gpsData);
+	
+	data.status.sensors = (PIOS_SENSORS_IsRegistered(PIOS_SENSOR_ACCEL) ? MSP_SENSOR_ACC  : 0) |
+	                      (PIOS_SENSORS_IsRegistered(PIOS_SENSOR_BARO)  ? MSP_SENSOR_BARO : 0) |
+	                      (PIOS_SENSORS_IsRegistered(PIOS_SENSOR_MAG)   ? MSP_SENSOR_MAG  : 0) |
+	                      (gpsData.Status > GPSPOSITION_STATUS_NOFIX    ? MSP_SENSOR_GPS  : 0);
+	
 	data.status.flags = 0;
 	data.status.setting = 0;
 
@@ -264,7 +271,7 @@ static void msp_send_status(struct msp_bridge *m)
 		FlightStatusData flight_status;
 		FlightStatusGet(&flight_status);
 
-		data.status.flags = flight_status.Armed == FLIGHTSTATUS_ARMED_ARMED;
+		data.status.flags = 0x0001; //flight_status.Armed == FLIGHTSTATUS_ARMED_ARMED;
 
 		for (int i = 1; msp_boxes[i].mode != MSP_BOX_LAST; i++) {
 			if (flight_status.FlightMode == msp_boxes[i].tlmode) {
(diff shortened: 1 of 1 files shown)


dustin:
Awesome.  I might see if I can get this going on my naze.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

_[flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c, line 336 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/571#-K9EAnfuwxQ6r3N9hGIG:-K9EAnfvtkqllrNmSxfw:444867413) ([raw file](https://github.com/d-ronin/dronin/blob/0408d06e48f8a16775c85bfe8efcb4774a2ef5b2/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c#L336)):_
These two lines have odd horizontal spacing.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/571#-:-K9EAt51zW52JMew9hII:-302977716)_

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -322,12 +329,72 @@ static void msp_send_ident(struct msp_bridge *m)
...
+	
+	int32_t deltaLon = (homeData.Longitude - gpsData.Longitude);  // degrees * 1e7
+	int32_t deltaLat = (homeData.Latitude  - gpsData.Latitude );  // degrees * 1e7
+	
+	float deltaY = (float)deltaLon * WGS84_RADIUS_EARTH_KM * DEG2RAD;  // KM * 1e7
+	float deltaX = (float)deltaLat * WGS84_RADIUS_EARTH_KM * DEG2RAD;  // KM * 1e7
+	
+	deltaY *= cosf((float)homeData.Latitude * 1e-7f * (float)DEG2RAD);  // Latitude compression correction
+	
+	data.comp_gps.distance_to_home  = (uint16_t)(sqrt(deltaX * deltaX + deltaY * deltaY) * 1e-4);  // meters
```
tracernz: sqrtf okay here rather than double precision sqrt?

jihlein:
Good catch, replaced with sqrtf.  Also should have mentioned my testing has been with an AQ32 board.

mlyle:
Reviewed 1 of 1 files at r2.

---

_[flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c, line 263 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/571#-K9EUZxHkup96Cx7MN5e:-K9EUZxHkup96Cx7MN5f:-289064107) ([raw file](https://github.com/d-ronin/dronin/blob/c907e16adc63b4ca58ef8984a0d52403bcb101d1/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c#L263)):_
These lines are space-indented too

---

_[flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c, line 347 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/571#-K9EUi6VhijmAb3rChXS:-K9EUi6VhijmAb3rChXT:1916723154) ([raw file](https://github.com/d-ronin/dronin/blob/c907e16adc63b4ca58ef8984a0d52403bcb101d1/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c#L347)):_
If there's not GPS, what should we do?  Probably not what we do below, e.g. return 0's..  Should we just return and decline to answer?

---

_[flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c, line 373 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/571#-K9EVV8Yt85PxWUt0vKR:-K9EVV8Yt85PxWUt0vKS:-767890270) ([raw file](https://github.com/d-ronin/dronin/blob/c907e16adc63b4ca58ef8984a0d52403bcb101d1/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c#L373)):_
Even more important here-- if GPS is missing, return the direction to homeloc from 0,0?  Probably should just return.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/571)_

jihlein:
Line 265 covers this, if there is no GPS fix, all the GPS information is suppressed from the display.  We could still return on GPS not found too.

I'm not following the spacing comments, they appear to be the same as the other instances in the file.

dustin:
It's perfectly fine to not return a requested value AFAIK.

---

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/571#-:-K9EYvLgpEK05hleDSSk:1682648746)_

mlyle:

---

_[flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c, line 336 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/571#-K9EAnfuwxQ6r3N9hGIG:-K9EZiiks3P7216PIjM-:-1333159692) ([raw file](https://github.com/d-ronin/dronin/blob/0408d06e48f8a16775c85bfe8efcb4774a2ef5b2/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c#L336)):_
These ones look fine to me actually

---

_[flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c, line 263 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/571#-K9EZpjyPCg4CQankp30:-K9EZpjyPCg4CQankp31:175620115) ([raw file](https://github.com/d-ronin/dronin/blob/c907e16adc63b4ca58ef8984a0d52403bcb101d1/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c#L263)):_
These lines are one tab then 16 spaces to try and align; should be two tabs. (and give up on maintaining alignment).

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/571)_

mlyle:
Yah-- whether or not they're not going to use the info, it's better to not send back garbage information anyways.

tracernz:
[MSP defines an error response](http://www.multiwii.com/forum/viewtopic.php?f=8&t=1516) in case you decline to answer:

```
$M|[0][code][checksum]
1 octet '$'
1 octet 'M'
1 octet '|'
1 octet 0
1 octet [unknown code]
1 octet [checksum]
```

Cleanflight does something different... https://github.com/cleanflight/cleanflight/blob/master/src/main/io/serial_msp.c#L207
And MWOSD doesn't care unless it's a > https://github.com/ShikOfTheRa/scarab-osd/blob/master/MW_OSD/Serial.ino#L824

jihlein:
So the issue becomes when do we display raw GPS data, and when do we have good enough GPS data to display meaningful computed GPS data (distance to home and direction to home).

We have the following GPS states: NoGPS ,NoFix, Fix2D, Fix3D, Diff3D

We could display raw GPS at Fix2D or greater.  Home position is set at Fix3D, so we could start displaying computed GPS data then.  If drops below Fix3D, do we continue to display the computed data?

This is further complicated by the MWOSD software, which seems to display both the raw and computed GPS data once GPS is set in the sensor word.  I need to research this a little further and see if there is a way to independently select raw and computed displays.

mlyle:
@jihlein yes, appropriately masking the data would be even better.  I don't know the answer.

jihlein:
@mlyle, a possible answer is to modify the MWOSD code slightly, to allow for individually masking the raw gps vs the comp gps data.  We could turn on/off the MWOSD mods with a Dronin specific #define.

My thought is to display raw gps if fix >= 2d.  The comp gps won't be displayed until a home position is set, and once displayed, will be displayed as long as fix >= 2d.  Fix < 2d will cause all GPS data to disappear.

It's only a few lines of change to the MWOSD code, so I'm going to try this later today.  It's one more thing to have to load/configure, but at least this aligns more with the GPS paradigm we use.  I don't know how *flight code sets this up, not sure it really matters.

Plus, it's one less excuse to not make a custom character set and have the MWOSD correctly display Dronin flight modes.  But that's for another day/discussion......

mlyle:
@jihlein There's already #ifdef TAULABS in MWOSD.  Just-- in my opinion-- as much as possible of this stuff should work with the plain MWOSD-as-distributed.  So, whatever you send should have the best possible "fallback" behavior on the existing MWOSD code.

jihlein:
Well I reworked the logic to send zeros when gps and/or home position is invalid.

It's not a 100% solution though, the home distance and direction display as 0 when they really should not be displayed.  Standard MWOSD code has raw and computed GPS display tied together.

A 99% solution is new MWOSD code.  You can find the code I'm using at:
https://github.com/jihlein/scarab-osd/tree/dronin

It's not a 100% solution because there is another firmware package that must be loaded into the OSD hardware.

If I'm missing something in the MWOSD firmware, I'm not seeing it.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -322,12 +329,99 @@ static void msp_send_ident(struct msp_bridge *m)
...
+		data.comp_gps.distance_to_home    = 0;
+		data.comp_gps.direction_to_home   = 0;
+		data.comp_gps.home_position_valid = 0;  // Home distance and direction will not display on OSD
+	}
+	else
+	{
+		GPSPositionGet(&gps_data);
+		HomeLocationGet(&home_data);
+		
+		if((gps_data.Status < GPSPOSITION_STATUS_FIX2D) || (home_data.Set == FALSE))
```
tracernz: HOMELOCATION_SET_FALSE :P

mlyle:
LGTM.

PR #572, opened 2016-01-30
Sparky2: Fix ADC config

tracernz:
ADC channels swapped. Checked against schematic, gerber, and tested.

@henn1001 The labels weren't wrong, just the channels. ;)

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/572)

### DIFF
--- a/flight/targets/sparky2/board-info/board_hw_defs.c
+++ b/flight/targets/sparky2/board-info/board_hw_defs.c
@@ -7,25 +7,29 @@
  *
  * @file       pios_board.c 
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2015
- * @author     dRonin, http://dronin.org Copyright (C) 2015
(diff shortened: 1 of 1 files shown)


henn1001:
You sure? GPIO seems to be still messed up.

unbenannt

henn1001:
I think the peripheral / timer was correct.. just the volt/current assignment was switched.
unbenannt

By the way - are we allowed to correct the schematics with photoshop-fu? :d Really confusing

tracernz:
Ahhh, let me double check :P

tracernz:
Ahh, the GPIOs are incorrect still, I shall fix. Everything else, including the schematic, is correct.
capture
(leftmost column is pin number for LQFP64)
capture2
capture3

tracernz:
Note that fixing the GPIOs will have no functional impact since we always init both pins as ADC AF anyway.

tracernz:
Fixed and force-pushed.

PR #573, opened 2016-01-30
flight: osd: further refactor

mlyle:
- Let people turn the centermark on and off independently from the AI
- Make the code a little cleaner

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/573)

### DIFF
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -554,96 +554,88 @@ void hud_draw_linear_compass(int v, int home_dir, int range, int width, int x, i
 #define CENTER_RUDDER     5
 #define PITCH_STEP       10
 void simple_artificial_horizon(float roll, float pitch, int16_t x, int16_t y, int16_t width, int16_t height,
-		int8_t max_pitch, uint8_t n_pitch_steps, bool center_mark)
+		int8_t max_pitch, uint8_t n_pitch_steps, bool show_horizon, bool center_mark)
 {
-	float sin_roll;
-	float cos_roll;
-	int16_t d_x, d_x2; // delta x
-	int16_t d_y, d_y2; // delta y
-	char tmp_str[5];
-
-	int16_t pp_x; // pitch point x
-	int16_t pp_y; // pitch point y
-	int16_t d_x_10, d_y_10, d_x_2, d_y_2;
-
-	int16_t pp_x2;
-	int16_t pp_y2;
-
 	float camera_tilt;
-
-	int pitch_step_offset;
 
 	StabilizationSettingsCameraTiltGet(&camera_tilt);
 
 	width /= 2;
 	height /= 2;
 
-	sin_roll    = sinf(roll * (float)(M_PI / 180));
-	cos_roll    = cosf(roll * (float)(M_PI / 180));
-
 	pitch += camera_tilt;
 
-	pitch_step_offset = pitch / PITCH_STEP;
+	if (show_horizon) {
+		float sin_roll = sinf(roll * (float)(M_PI / 180));
+		float cos_roll = cosf(roll * (float)(M_PI / 180));
 
-	/* how many degrees the "lines" are offset from their ideal pos
-	 * since we need both, don't do fmodf.. */
-	float modulo_pitch = pitch - pitch_step_offset * 10.0f;
+		int pitch_step_offset = pitch / PITCH_STEP;
 
-	// roll to pitch transformation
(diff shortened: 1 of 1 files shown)


mlyle:
Completely untested.

mlyle:
Tested.  Works good.  You can keep centermark now, @ohmeye & @fujin :P

mluessi:
Thanks. Somehow the option to show the centermark separately was removed since the flies were added to dRonin, not sure why. But this is probably better, as it handles the camera tilt.

mlyle:
The previous PR factored the centermark into the AI code so that the centermark can be offset to be on the horizon.

Some people also want to offset the horizon but not the centermark-- I don't include support for that (would have to change settings).

OhmEye:
I don't understand what the centermark has to do with the horizon. I've always considered it to be a mark for the center of the screen and unrelated to the AHI.

mlyle:
@ohmeye Attitude indicators have a picture of an airplane seen from behind on them, relative to the horizon. In the west (e.g. not Russia/China), the horizon moves.  If the horizon moves down, the plane is pointing at "blue" / nose up.  If the horizon moves up, the plane is pointing at "brown" / nose down, etc.

In Russia, it's the same thing, except it's the plane that moves.  That is, the western model is of a pilot being stationary relative to the plane and actions moving the horizon relative to the plane-- it's more like "flying the aircraft" directly and is suitable for superimposing on a HUD, etc.

The eastern model it's like flying an RC plane, you "move it into a right bank" or whatever.

OhmEye:
To me the OSD is more like a HUD, overlaying reality with augmented info, not a panel mounted AHI. This is probably me just being used to how it worked previously and how MWOSD works. We don't have a true HUD flight path marker (FPM) so the centermark kinda serves as that. The main thing I'm not understanding is the benefit of having the centermark not be in the center of the screen. I'm sure that's my deficit. As long as it's possible to make it be in the center, I'll stop complaining :)

The best way I know to describe my opinion is that myself and others have styles that "fly the camera, not the quad" such that the attitude of the quad is important but secondary, it's the camera attitude that we fly relative to.

mlyle:
https://jethead.files.wordpress.com/2013/03/hud-sample.jpg?w=454&h=273

If you ever actually have to rely on the HUD for attitude guidance-- e.g. you're leveling out to land a large quad from a hover, or you're flying a plane and need to keep the nose on the horizon even when sun glare keeps you from being able to need it-- you need to indicate the actual attitude of the craft.   On the HUD above, the attitude is shown through the plane relative to the horizon line and the little arrow at the top showing the bank angle.

> We don't have a true HUD flight path marker (FPM) so the centermark kinda serves as that.

Yah, I understand what you're doing-- .. when we fly full-scale planes in for a landing, or in general, "anything stationary in the view" is where you're going to hit.  If things recede below you, you're going to overshoot them.. if they stretch out away you are going to come up short.  This is how mid-air collisions happen-- planes you're "not going to hit" move in your vision and are really obvious while you can be on a collision course with another plane and you just slowly grow larger to each other until it's too late.  Odds are you could put that centermark almost anywhere reasonable in your view (left, right, up, down), and as long as you "hold on object" on it you will hit/go through that object.  So I can see how having a handy mark at a known place could be nice.

But the attitude reference provided by the OSD is worthless without an indicator of the attitude.

OhmEye:
Needing to indicate the actual attitude of the craft is where we disagree. I agree completely that this is critical for most situations, but it's actually bordering on irrelevant for miniquad racing and freestyle in my opinion. This is the difference between "flying the camera" and "flying the aircraft."

When tilting the camera for a desired FF speed, more often than not the actual horizon is well below camera view if the frame is level. It's very rare that I care about actual attitude of the aircraft, I only care about flight path and don't need to know what angle the frame is at, I'm flying the camera, not the frame. I agree we can use any arbitrary position in the camera view for our "hold on object" but because we are flying within a meter from physical obstacles it's desirable to have as wide a view as possible around the centermark, and typically the best and obvious spot for a reticle is the center of the screen.

Anyway, I feel we're talking in circles at this point. I believe the majority of rate mode proximity style pilots would agree with me but I may be wrong, and I don't disagree with your view in general, only for the style of flying I like to do. So I'm content to agree to disagree if I have a combination of settings that result in placing the centermark at the center of the screen regardless of whether that offsets the AHI. :) My original interest was for more ladder rungs, and I appreciate that. :)

mlyle:
Yes-- no worries.  The AHI needs to handle use cases that aren't really miniquad racing related-- actually being a backup horizon when camera systems aren't helpful, providing video that can be used to troubleshoot navigation, working well for planes, etc.

mlyle:
Flight tested today, works as expected. (screens with both, with centermark only, with none).

PR #576, opened 2016-01-31
Allow automatic calculation of CellCount

### LINKED ISSUE #575: Batterymodule cell-count autodetection [WIP]
Got some barebone code working in my local branch
https://github.com/henn1001/dRonin/commit/b02dc901fb70abdbf0ed9ac08c0fa52884c3096c
Good/bad idea?

The thresholds act now as a limit per cell.

Did some testing and it seems to work fine but need to test some more edge cases.

henn1001:
fixes #575
Needs some more edge case testing and someone has to test the Frsky telemetry.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/576)


flight/Modules/Battery/battery.c
```diff
@@ -149,14 +152,21 @@ static void batteryTask(void * parameters)
 					scaled_voltage = 0.0f;
 					adc_offset_invalid = true;
 				}
+				else if(scaled_voltage > 3.0f && !cells_calculated && (batterySettings.MaxCellVoltage > 0)){
+					cells_calculated = true;
+					cells = ((scaled_voltage / batterySettings.MaxCellVoltage) + 1);
+					flightBatteryData.DetectedCellCount = cells;
```
henn1001: `FlightBatterySettingsNbCellsSet((uint8_t*)&cells);`
Could be used too instead of the new var. in batterystate but should be avoided.

flight/Modules/Battery/battery.c
```diff
@@ -50,6 +50,7 @@ static struct pios_thread *batteryTaskHandle;
 static int8_t voltageADCPin = -1; //ADC pin for voltage
 static int8_t currentADCPin = -1; //ADC pin for current
 static bool battery_settings_updated;
+static bool cells_calculated;
```
tracernz: Scope can be reduced to just the batteryTask function.

flight/Modules/Battery/battery.c
tracernz: We never overwrite settings in firmware though. :) What you've done is good.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
```diff
@@ -400,6 +400,9 @@ static void uavoFrSKYSensorHubBridgeTask(void *parameters)
 			// As long as there is no voltage for each cell
 			// all cells will have the same voltage.
 			// Receiver will know number of cells.
+			if(batState.DetectedCellCount)
+				batSettings.NbCells = batState.DetectedCellCount;
```
tracernz: Even if it takes a little more RAM I think it would be good to assign to a new variable rather than the settings struct, just for clarity.

shared/uavobjectdefinition/flightbatterysettings.xml
```diff
@@ -6,6 +6,9 @@
         <field name="VoltagePin" units="" type="enum" elements="1" options="ADC0,ADC1,ADC2,ADC3,ADC4,ADC5,ADC6,ADC7,ADC8,NONE" defaultvalue="ADC1"/>
 
         <field name="NbCells" units=""  type="uint8"  elements="1" defaultvalue="3"/>
+	<field name="MaxCellVoltage" units="V" type="float" elements="1" defaultvalue="0">
```
tracernz: Nice, but the indentation is now broken :P (surrounding lines use spaces, yours use tabs).

shared/uavobjectdefinition/flightbatterystate.xml
```diff
@@ -1,7 +1,8 @@
 <xml>
     <object name="FlightBatteryState" singleinstance="true" settings="false">
         <description>Battery status information.</description> 
-        <field name="Voltage" units="V" type="float" elements="1" defaultvalue="0.0"/>
+	<field name="Voltage" units="V" type="float" elements="1" defaultvalue="0.0"/>
```
tracernz: Indentation

flight/Modules/Battery/battery.c
```diff
@@ -100,6 +101,7 @@ static void batteryTask(void * parameters)
 	const float dT = SAMPLE_PERIOD_MS / 1000.0f;
 
 	battery_settings_updated = true;
+	cells_calculated = false;
```
tracernz: Just add bool in front of this and delete the declaration at the top to reduce scope.

shared/uavobjectdefinition/flightbatterysettings.xml
tracernz: Also, default value should be maybe 4.25?

tracernz:
This is a nice addition @henn1001, thanks!

jenkins, test this please

shared/uavobjectdefinition/flightbatterysettings.xml
henn1001: Do we want this feature on or off by default?

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
henn1001: Not sure about this one , it might interfere with the logic above https://github.com/henn1001/dRonin/blob/CellDetection/flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c#L268

henn1001:
Do we want this feature on or off by default?

tracernz:
Good question. Battery module is off by default. I'm happy for this to be enabled by default with the battery module.

Another thing that would be cool is if we can detect a less than fully charged pack at startup and generate a warning in that case. It would be great to set the low voltage alarm per cell rather than per pack too, making switching between 3S and 4S much easier. These can come in future PRs.

henn1001:
Update the copyright here as well?

mlyle:
@henn1001 sounds like a good idea.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
```diff
@@ -400,6 +405,10 @@ static void uavoFrSKYSensorHubBridgeTask(void *parameters)
 			// As long as there is no voltage for each cell
 			// all cells will have the same voltage.
 			// Receiver will know number of cells.
+			uint8_t cellCount = batState.DetectedCellCount;
+			if(cellCount)
+				batSettings.NbCells = cellCount;
+
```
henn1001, line 408: @tracernz is this ok?

tracernz, line 408: Looks fine.

tracernz:
jenkins, test this please

flight/Modules/Battery/battery.c
mluessi: I think we should use rounding instead of `((scaled_voltage / batterySettings.MaxCellVoltage) + 1)`. If for example `MaxCellVoltage` is 4.2 and the voltage sensor calibration is slightly off, the reading for a full 3S battery may be 12.65V, the current code would then detect 4 cells.

tracernz:
Reviewed 5 of 6 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

_[flight/Modules/Battery/battery.c, line 158 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/576#-K9KKfr5nXe_YFscKB_a-r1-158:-KAcH3VwRqJwbCDhMuMT:-1482637730) ([raw file](https://github.com/d-ronin/dronin/blob/41a79aca36fb0ba49123e0d595887888772896f6/flight/Modules/Battery/battery.c#L158)):_
Max cell voltage should be greater than 4.2, at least 4.3. It's robust for 6S with 3.6 <= cell < 4.3 when MaxCellVoltage is set to 4.3 V.

---

_[shared/uavobjectdefinition/flightbatterysettings.xml, line 9 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/576#-K9KKfr5nXe_YFscKB_c-r1-9:-KAcIWL9z7FpksfOxSF2:-1866529696) ([raw file](https://github.com/d-ronin/dronin/blob/41a79aca36fb0ba49123e0d595887888772896f6/shared/uavobjectdefinition/flightbatterysettings.xml#L9)):_
On IMO. As above I think we should make default 4.3 V.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/576)_

henn1001:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

_[flight/Modules/Battery/battery.c, line 158 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/576#-K9KKfr5nXe_YFscKB_a-r1-158:-KAdVIQbqHwcHy3Fq7-h:-550304389) ([raw file](https://github.com/d-ronin/dronin/blob/41a79aca36fb0ba49123e0d595887888772896f6/flight/Modules/Battery/battery.c#L158)):_
Rounding would make probably make it too complicated again. The calculation is more or less from *flight and seems to work fine - never heard anyone complain about false detection :) .
But maybe the name should be changed  so people don´t set it to 4.2 because that is their max lipo cell voltage? 

---

_[shared/uavobjectdefinition/flightbatterysettings.xml, line 9 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/576#-K9KKfr5nXe_YFscKB_c-r1-9:-KAdW34J_B1SBclMqdIz:160384640) ([raw file](https://github.com/d-ronin/dronin/blob/41a79aca36fb0ba49123e0d595887888772896f6/shared/uavobjectdefinition/flightbatterysettings.xml#L9)):_
Ok!

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/576)_

mluessi:
_[flight/Modules/Battery/battery.c, line 158 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/576#-K9KKfr5nXe_YFscKB_a-r1-158:-KAdwtmke_bo_LqIMaUO:-371584779) ([raw file](https://github.com/d-ronin/dronin/blob/41a79aca36fb0ba49123e0d595887888772896f6/flight/Modules/Battery/battery.c#L158)):_
By too complicated, do you mean too complex to run on F1? How about adding 0.5 instead of 1? This will essentially do the same thing. I think it would be most natural if we can use 4.2V for the max cell voltage (for lipos).

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/576)_

henn1001:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

_[flight/Modules/Battery/battery.c, line 158 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/576#-K9KKfr5nXe_YFscKB_a-r1-158:-KAeTLnQ4-Jlj4J0wSIB:2139671435) ([raw file](https://github.com/d-ronin/dronin/blob/41a79aca36fb0ba49123e0d595887888772896f6/flight/Modules/Battery/battery.c#L158)):_
Complicated in terms of deciding when to round upwards or downwards. On the other hand  i didn´t spent much thought on it either because this logic  seems to work fine for years now :)

I don´t like 0.5 . takes away too much of the lower boundary 
https://docs.google.com/spreadsheets/d/16li7pCzq_bO6_4cc-A-SCiTmsrLBXzENQckW9kHDSyM/edit?usp=sharing

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/576)_

mlyle:
I like the / 4.3 range in the spreadsheet.  Happy from 2.87V to 4.3V on 3S, 3.22V to 4.3V on 4S, which are the two that most matter for autodetect.  Yes, 4.3V is a little close to 4.2V and offsets / glitches make me nervous. 

I played with the spreadsheet, and floorf(v/4.2f + 0.9f ) or around there seemed best... basically, it seems like to me that we want to have a cell-voltage wide interval... and we want to offset that interval just a little, because low cell voltages are more likely than high cell voltages.  But it's really small differences/bikeshedding IMO.  

IMO autodetect is only important for 2S/3S/4S.  Beyond there things get too ambiguous, and I don't think there's many people flying a setup on either, say, 6S or 8S.

tracernz:
So are we all happy with this subject to a little test? I'd just like to change the default to 4.3 V so it works by default out-of-the-box, the less things requiring configuration the better.

henn1001:
Sure go ahead . I´ve been using this now for a while :)

mluessi:
I would prefer using "+ 0.9" and `MaxCellVoltage=4.2` by default, this will make it more user friendly when using non-lipo cells. But if there is strong objection, I'm fine with the code as is.

Nice work @henn1001, this will be very useful for things like switching stabilization profiles based on detected number of cells :).

tracernz:
I'm okay with adding 0.9 but then it does start to become a little opaque to the user. That can sometimes be a good thing though.

henn1001:
Ok so: +0.9 and limit at 4.2? I´m ok with that. Doesn´t seem to affect the range by that much.

pug398:
So what will a fully charged 6s currently read? I am one of those lowly types :)

henn1001:
min/cell  3.57  
max/cell  4.2693

Limits for 6s detection

pug398:
Ok that is what I was thinking. The first part of the calculation uses v but the offset goes askew as voltage goes up and down. Perhaps for the offset something more dynamic that tracks voltage change as well? Maybe +((v/3.2) - (v/4.2))/2 ??. Not saying that is correct but the offset should scale to the voltage and range as well so that any battery will detect correctly as long as the max voltage per cell sets the range and hopefully each cell is somewhere in that range. Nice work on getting this done by the way. Hopefully I am pointing where I am thinking LOL.

pug398:
The above offset doesn't work out either. So much for my math skills.

pug398:
Continue on with what you were doing before I interrupted. Still no way to get around overlapping ranges above 4s.

mluessi:
Yeah, above a certain number of cells there is nothing you can do to avoid overlap. I think having something that is close to rounding up, like we have now (using "/4.2 + 0.9") is probably best, as we can assume that the batteries are close to fully charged at the beginning of a flight.

mlyle:
(Alas, the most useful thing to do with voltage detection is to detect undercharged batteries at startup and warn the heck out of you... but, what can you do?)

tracernz:
@mlyle, quite true. The other obvious benefit is that we can make the voltage warning be per-cell so you can easily switch between 3S and 4S without changing your alarm configuration.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
mlyle, line 408: I tested this and really read this code now.

I think we should only do this if cellcount is 0, and let you set cellcount to 0 in the UI to indicate autodetection.

flight/Modules/Battery/battery.c
```diff
@@ -149,14 +155,21 @@ static void batteryTask(void * parameters)
...
+				}
+				else if(!cells_calculated)
+					cells = 1;
 			}
 
 			flightBatteryData.Voltage = scaled_voltage;
 
 			// generate alarms and warnings
-			if (flightBatteryData.Voltage < batterySettings.VoltageThresholds[FLIGHTBATTERYSETTINGS_VOLTAGETHRESHOLDS_ALARM])
+			if (flightBatteryData.Voltage < (batterySettings.VoltageThresholds[FLIGHTBATTERYSETTINGS_VOLTAGETHRESHOLDS_ALARM] * cells))
```
mlyle: I think we need some way to turn this on and off.   Use cells specified in UI, or if 0, do autodetect.  Also change UI to indicate (per cell) for the voltages.

It probably also makes sense to rename these parameters so people upgrading don't get bogus settings.

mlyle:
Bench tested this on Lux.  The logic seems to do the right thing overall, but I'm concerned about usability.

flight/Modules/Battery/battery.c
henn1001: Not quite sure what you mean? The autodetection is active as soon as MaxCellVoltage > 0 / off when MaxCellVoltage == 0 .

Renaming the warning thresholds sounds good. Did not thought about that because i wanted to make this feature off by default so that would have been a non-issue. But now it´s on by default and people who import old values will get bogus detection.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
mlyle, line 408: Nvm.  I understand intentions better now.

flight/Modules/Battery/battery.c
mlyle: Yes, renaming the fields sounds good.

I see now there's a way to turn it off, and that it defaults on.

I am a little concerned that there is the cellcount specified in the UI, which is by default ignored.  It seems like it would be good to use the UI cellcount if specified, and if the UI cellcount is 0 use autodetect.

henn1001:
^ Woops sorry. Did a squash&rebase.

Ok, i renamed the settings. I wanted to change the autodetection to nbCells as well but i feel like it would need some description in the GUI and i´m not quite up for that task to change that.
Can someone help me out here?

tracernz:
@henn1001, I would be happy to make necessary UI changes. :-)

henn1001:
@tracernz thanks - can you shoot me a pr to my branch or create a follow up pr?

"I am a little concerned that there is the cellcount specified in the UI, which is by default ignored. It seems like it would be good to use the UI cellcount if specified, and if the UI cellcount is 0 use autodetect." t. mlyle

So change this line 
`else if(scaled_voltage > 3.0f && !cells_calculated && (batterySettings.MaxCellVoltage > 0))`
to
`else if(scaled_voltage > 3.0f && !cells_calculated && (batterySettings.NbCells == 0))`

Change defaultvalue for NbCells to 0

Change description for MaxCellVoltage

Gui changes in batterymodule to make a note about the behavior.

mlyle:
That sounds great, yes.

tracernz:
@henn1001, if I forget remind me on irc when I'm at home.

tracernz:
:lgtm: Should probably get review on GCS parts by someone other than me. Would be nice to get this in before it accumulates conflict.

---

Reviewed 5 of 5 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/576#-:-KCcvANuFBISnMRXsQO6:1636177199)_

henn1001:
Mhh think there´s a bug again - flew it today and battery warning doesn´t seem to work.

flight/Modules/Battery/battery.c
```diff
@@ -133,6 +140,7 @@ static void batteryTask(void * parameters)
 		if (voltageADCPin >= 0) {
 			float adc_voltage = (float)PIOS_ADC_GetChannelVolt(voltageADCPin);
 			float scaled_voltage = 0.0f;
+			unsigned cells = batterySettings.NbCells;
```
henn1001: @tracernz yah so removing the static here broke it. Was there a reason for this change? No statics allowed?

tracernz: Yah, it needs to be static otherwise it gets overwritten on the next iteration?

tracernz: ~~Could instead take away the " = batterySettings.NbCells" and then it can remain on the stack.~~
Could increase it's scope outside the main while loop and then it can remain on the task stack.

henn1001:
increased scope

flight/Modules/Battery/battery.c
```diff
@@ -105,6 +105,9 @@ static void batteryTask(void * parameters)
 
 	battery_settings_updated = true;
 	bool cells_calculated = false;
+	unsigned cells;
+	FlightBatterySettingsNbCellsGet((uint8_t*)&cells);
```
tracernz: I don't think this is actually needed, it's always either detected or set to batterySettings.NbCells later.

henn1001: Haha. I have probably read to much into your change again.

henn1001: Ah wait! it is needed or else it will result in a compiler warning.

tracernz: Just go = 1 then?

henn1001: k :D

tracernz:
Looks good :).

tracernz:
Reviewed 1 of 1 files at r9.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/576)_

tracernz:
:lgtm: 

---

Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/576#-:-KDDeGMVPEVUOKeK4Ix5:980077423)_

mlyle:
This has accumulated conflicts, unfortunately.

No promises, but if you get this rebased, I'll try and get some time to play with this towards merging it for Tanto / the April release.

henn1001:
Fixed and rebased

ground/gcs/src/plugins/config/modules.ui
```diff
@@ -1083,7 +1170,7 @@
            <rect>
             <x>0</x>
             <y>0</y>
-            <width>360</width>
+            <width>392</width>
```
henn1001: @tracernz This created the merge conflict - not sure if this is still the correct value.

mlyle:
Thanks--- sorry we've been slow with this.

tracernz:
@henn1001, heh, I think that just the size of the window when you have it open in Qt creator; don't think it actually does anything useful. :-)

mlyle:
jenkins, test this please

mlyle:

mlyle:

henn1001:
Just two things from irc:
1. There was confusion about the "live detected cell count" in gcs because it only shows the cellcount that gets calculated once on boot. If people connect the fc first with usb and then the lipo it will show only 1cell. Maybe "live" should mean a cell calculation in gcs. Not sure about this one... @tracernz what do you think?
2. Report of a 23V 6s Bat detected as 5s. Maybe a "settle" timer is needed where we wait for 1-2seconds after boot for the correct values from adc.

tracernz:
Unless the code has changed it should only calculate when the battery is connected? Until then it should use the manually set cell count iirc.

henn1001:
Just a small re on this one: 
I added a threshold so usb power won´t trigger the cellcount. (did not commit yet because i wait for some other changes)

Live view being a live view.. would only be sensible if we do the calculation in gcs (Cellcount= (V/maxCell)  +0.9) instead of showing the uav data object. But it´s pretty meh and only cosmetic. A simple hit on save does the same... So either rename the "live" or just leave it like it is.  But again, this is mostly QT stuff where i have no clue :D

Still not sure about the "settle timer" because we already sleep after boot for 500ms and that should be fairly enough for the voltage/adc readings to settle. I also can´t really test it for cells above 4s but the math doesn´t show any errors.
 [Link](https://docs.google.com/spreadsheets/d/16li7pCzq_bO6_4cc-A-SCiTmsrLBXzENQckW9kHDSyM/edit#gid=1161304764)

I´ve been running this now for a few months on 3s & 4s without a single misreading so i dunno..

tracernz:
All other live values in GCS come from the FC and I don't think we should change that pattern here. Shame this missed the release :(.

henn1001:
Yes i think it will be fine after i commit my change to prevent the 1cell detection from USB power. The user was confused because he first connected the USB and then the lipo and it still only showed 1 cell in gcs. 

Other than that i´m not sure what else could be done here. I think i also heard from cleanflight that celldetection above 5S is flaky for them.

A shame indeed but i guess there just wasn´t enough interest in this D:

mlyle:
I don't expect it to be reliable over 5S, etc, because of the thinner margins.  But the two people we had try it ended up confused.  We need to work on the UI for it so that it's not a support problem.

mlyle:
Let's bring this in now.  We can work more on usability as the cycle continues.

tracernz:
Thanks.

henn1001:
Yay <3

PR #569, opened 2016-01-29
Consolidated Throttle/Collective into Thrust

ekaszubski:
Resolves #515.

Updated field names in UAVOs:
- ActuatorDesired
- AltitudeHoldState
- LoiterCommand
- StabilizationDesired

Updated references and throttle/collective logic in various flight modules.

This PR also enables helicopters to use flight modes which stabilize altitude.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/569)

PR #581, opened 2016-01-31
WIP WIP WIP Add units to StabilizationDesired

ekaszubski:
- Add new StabilizationShared for orientation and position units (later for stabilization modes as well)
- Add units to all axes in StabilizationDesired
- Move all axis scaling logic out of ManualControl and into Stabilization

Note: depends on #569

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/581)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -140,13 +140,18 @@ int32_t ActuatorInitialize()
 }
 MODULE_INITCALL(ActuatorInitialize, ActuatorStart);
 
-static float get_curve2_source(ActuatorDesiredData *desired, MixerSettingsCurve2SourceOptions source)
+static float get_curve2_source(ActuatorDesiredData *desired, SystemSettingsAirframeTypeOptions airframe_type, MixerSettingsCurve2SourceOptions source)
 {
 	float tmp;
 
 	switch (source) {
 	case MIXERSETTINGS_CURVE2SOURCE_THROTTLE:
-		return desired->Throttle;
+		if(airframe_type == SYSTEMSETTINGS_AIRFRAMETYPE_HELICP)
+		{
+			ManualControlCommandThrottleGet(&tmp);
+			return tmp;
+		}
+		return desired->Thrust;
 		break;
 	case MIXERSETTINGS_CURVE2SOURCE_ROLL:
 		return desired->Roll;
@@ -158,6 +163,10 @@ static float get_curve2_source(ActuatorDesiredData *desired, MixerSettingsCurve2
 		return desired->Yaw;
 		break;
 	case MIXERSETTINGS_CURVE2SOURCE_COLLECTIVE:
+		if (airframe_type == SYSTEMSETTINGS_AIRFRAMETYPE_HELICP)
+		{
+			return desired->Thrust;
+		}
 		ManualControlCommandCollectiveGet(&tmp);
 		return tmp;
 		break;
@@ -202,6 +211,9 @@ static void actuator_task(void* parameters)
 	ActuatorDesiredData desired;
 	MixerStatusData mixerStatus;
 	FlightStatusData flightStatus;
+	ManualControlCommandData manual_control_command;
+
+	SystemSettingsAirframeTypeOptions airframe_type;
 
 	/* Read initial values of ActuatorSettings */
 	settings_updated = true;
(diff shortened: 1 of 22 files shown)


mlyle:
I want to sequence this one in after #473, so it'll probably need to be based onto that.  (going to be awhile.. tangly tangly long PR queue.. fuuuuu)

ekaszubski:
@d-ronin/gcs-developers Stuff related to unit conversion, would appreciate your input:
- [x] Classify flight modes according to the kinds of units they accept
  - Degrees
    - Attitude
    - SystemIdent
  - Degrees/sec
    - Rate
    - WeakLevelling
    - AxisLock
  - Effort
    - Manual
    - VirtualFlybar
    - CoordinatedFlight
    - POI
    - AcroPlus
    - MWRate
    - Horizon
- [x] Up-conversion via system properties/settings
  - effort -> deg = effort \* max attitude
  - effort -> deg/sec = effort \* manual rate
  - deg/sec -> deg = integrate( deg/sec, t )
- [ ] Down-conversion to effort (maybe bogus)
  - deemed too dangerous for now
  - deg -> effort = deg / max attitude
  - deg/sec -> effort = deg/sec / manual rate
- [ ] Cross-conversion via effort (maybe bogus)
  - deemed too dangerous for now
  - deg -> deg/sec = ( deg / max attitude ) \* manual rate
  - deg/sec -> deg = ( deg/sec / manual rate ) \* max attitude
- [ ] Fallbacks in weird scenarios:
  - Degrees + any rate-based mode
    - [x] fall back to Attitude mode?
    - [ ] or use outer (attitude) PID fed into the given mode?
    - [ ] or convert to effort via degrees / max attitude capped at 1.0 and convert to degrees/sec via effort \* manual rate?
    - [ ] or assume value is actually in degrees/sec?
  - Degrees/sec + any attitude-based mode
    - [ ] convert to degrees by accumulating deg/sec \* time
    - [x] or fall back to rate mode?
    - [ ] or convert to effort via degrees/sec / manual rate capped at 1.0 and convert to degrees via effort \* max attitude?
    - [ ] or assume value is actually in degrees?
  - Degrees + any effort-based mode
    - [x] fall back to attitude mode?
    - [ ] or convert to effort via degrees / max attitude capped at 1.0?
  - Degrees/sec + any effort-based mode
    - [x] fall back to rate mode?
    - [ ] or convert to effort via degrees/sec / manual rate capped at 1.0?
  - Any position-type unit (meters, meters/sec, meters/sec^2) + any stabilization mode on R/P/Y
    - we won't allow position units in orientation fields so this will never happen
  - Any orientation-type unit (degrees, degrees/sec) + any stabilization mode on Thrust (not yet implemented)
    - we won't allow orientation units in position fields so this will never happen

mlyle:
"Fall back to ...." for all of the above please.

Also, I don't think position-type units on R/P/Y make sense.  Position oriented controllers need to sit over this layer and won't ever be interchangeable.  In turn the position oriented controllers can ask this layer for what they want.

ekaszubski:
Updated checklist according to feedback. Conclusion for now is to trust the units, which means all modes setting StabilizationDesired need to be very careful when specifying the units field as setting the wrong units can change the stabilization mode and have very bad consequences.

In general, the proposed procedure is as follows:
- Look at specified unit for each axis
- Check whether selected flight mode for the current axis accepts the given unit
  - If not, fall back to default flight mode for those units
  - First exception is "Effort" units, which can be converted based on system settings to any flight mode's units
  - Second exception (proposed) is "Prescaled" units, which would bypass all unit checks and pass the given StabilizationDesired values directly to the given stabilization mode. This is how the stabilization pipeline currently functions (prior to this PR), but could cause issues if abused (e.g. via laziness or misinterpretations of its functionality, but this true of many parts of the code).

This procedure would be implemented via the following functions:
- `convert_units( input unit, output unit, value )`
  - Return value converted from input units to output units
  - Only valid conversion for now is effort -> other units
  - If conversion cannot be done, throw error
- `get_default_mode( input unit )`
  - Return the default flight mode to be used to stabilize an axis when its setpoint is specified in the given units
- `check_units( input unit, mode-desired unit, value )`
  - Attempt to convert units via `convert_units`
    - If success, update setpoint with converted value
    - If conversion fails, look up fallback stabilization mode via `get_default_mode( input unit )` and update stabilization mode with the result (setpoint would be unchanged)

mlyle:
convert_units should also handle where inputunit == outputunit and properly do a no-op.

ekaszubski:
> convert_units should also handle where inputunit == outputunit and properly do a no-op.

Yep, agreed.

ekaszubski:
@mlyle putting this at the top of my dRonin queue

mlyle:
This should probably close out for now.. can reconsider this / pick it back up again if updated.

PR #586, opened 2016-02-01
flight: OpenLog selectable baudrate

henn1001:
Seems to "work" for me at 250k Baud. Should i add more rates?

Note that i´m using this special firmware: 
https://github.com/cleanflight/blackbox-firmware

Can´t really fly much at the moment but this is a really short log while armed without props. Custom profile and gyro log rate at 1000hz.
[Matlab.zip](https://github.com/d-ronin/dRonin/files/112347/Matlab.zip)

```
wrongSyncByte instances:            85
wrongMessageByte instances:          0

Completed bytes:    245542 of    283010
Est. time remaining, 00h:00m:01s 
Unknown object ID: 0x0457E4D6 appeared 1 times.
Unknown object ID: 0x420A42B9 appeared 1 times.
Unknown object ID: 0x0A5832F6 appeared 1 times.
Unknown object ID: 0xFE418AF6 appeared 1 times.
Unknown object ID: 0x43408D74 appeared 1 times.
283010 records in 1.93 seconds.
```

By the way, is there a limit of bytes which the matlab file can parse? completed bytes x of y ...

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/586)

### DIFF
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -92,6 +92,7 @@ static void register_object(UAVObjHandle obj);
 static void register_default_profile();
 static void logSettings(UAVObjHandle obj);
 static void writeHeader();
+static void updateSettings();
 
 // Local variables
 static uintptr_t logging_com_id;
@@ -113,6 +114,7 @@ int32_t LoggingInitialize(void)
 	if (PIOS_COM_OPENLOG) {
 		logging_com_id = PIOS_COM_OPENLOG;
 		destination_spi_flash = false;
+		updateSettings();
 	}
 	else if (PIOS_COM_SPIFLASH) {
 		logging_com_id = PIOS_COM_SPIFLASH;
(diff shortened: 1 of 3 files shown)


henn1001:
The timestamp seems a bit odd though with the .m file.
If i change 
`timestamp = timestamp + timestampAccumulator;`
to
`timestamp = timestamp;`
in line `2492`
It corrects this issue. Probably because the timestamp resolution is >=1ms which is right on the edge of the logging rate.

mlyle:
jenkins, test this please

mlyle:
I don't think there's a limit on the matlab size.

I don't know the matlab code, but I think that's attempting to deal with timestamp wraparound (I think we just store the low 16bits of the time, and keep an accumulator).  There's code like that in the python API, but I had to fix it before because the prior implementation was broken.

Issue #527, opened 2016-01-24
Sparky: Attempt to free some flash


tracernz:
So new features can be added. Best candidates:

```
Michaels-MacBook-Pro:dronin mike$ ./tools/gcc-arm-none-eabi-4_9-2015q1/bin/arm-none-eabi-nm -S --size-sort --reverse-sort --radix=dec ./build/fw_sparky/fw_sparky.elf 
134359376 00013766 T CovariancePrediction
134454192 00012288 T twiddleCoefQ15
134308400 00005308 t AutotuneTask
134335856 00003068 t uavoHoTTBridgeTask
134296432 00002796 t stabilizationTask
134283824 00002264 t updateAttitudeComplementary
134471168 00002184 t CoeffFile
134292896 00002084 T transmitter_control_update
134324368 00002076 t pathfollowerTask
134452144 00002048 T armBitRevTable
134313856 00002032 t updatePIDs
134281856 00001964 t updateAttitudeINSGPS
134339248 00001932 t uavoFrSKYSensorHubBridgeTask
134407824 00001796 t PIOS_MPU9150_Task
134446864 00001520 T __ieee754_powf
134449488 00001506 T __kernel_rem_pio2f
134328096 00001448 t uavoMavlinkBridgeTask
134278960 00001444 t SensorsTask
134344256 00001288 T PIOS_Board_Init
134329952 00001252 t uavoMSPBridgeTask
134333520 00001218 t generate_warning
134317856 00001216 t pathPlannerTask
134304320 00001164 T parse_ubx_stream
134386432 00001152 T PIOS_INTERNAL_ADC_Init
134306736 00001148 t attitudeUpdated
536871576 00001064 d impure_data
134286096 00001052 t AttitudeTask
```

mlyle:
> 134452144 00002048 T armBitRevTable

This is vibration analysis I think.

> 134335856 00003068 t uavoHoTTBridgeTask

Why is this so much bigger than all the other serial rx protocol things?

Most of the rest is fairly necessary INS / etc stuff.

tracernz:
Yeah but we might be able to squash some of it down without removing it.

PR #589, opened 2016-02-02
Flight: Make external mag handling consistent across targets

jihlein:
The external mag handling is not consistent across targets  This PR will change the external mag handling to be as follows:

1)I2C initialization will no longer generate a panic code upon failure.  Successful initialization will be indicated by the I2C WCA box being displayed in green.  Either an init failure or a checkclear failure will turn the I2C WCA box red.  Both internal and external I2C ports will adhere to this.  One deficiency is that if an I2C bus fails initialization, it will be hard to tell whether the internal or external bus is at fault.  A mitigation here is to disable any active external I2C buses and see if the error condition clears,

2)A failure of the external mag to initialize will no longer generate a panic code upon initialization or test failure, it will turn the sensor WCA box red.

3)An option to disable the use of mag inputs completely will be added.

This initial set of commits moves Sparky to this paradigm.  Other targets to follow.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/589)

### DIFF
--- a/flight/Modules/Sensors/sensors.c
+++ b/flight/Modules/Sensors/sensors.c
@@ -116,7 +116,7 @@ static float z_accel_offset = 0;
 static float Rsb[3][3] = {{0}}; //! Rotation matrix that transforms from the body frame to the sensor board frame
 static int8_t rotate = 0;
 
-#if defined (AQ32)
+#if defined(SUPPORTS_EXTERNAL_MAG)
 // indicates whether the external mag works
 extern bool external_mag_fail;
 #endif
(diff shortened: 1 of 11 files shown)


mlyle:
```
/var/lib/jenkins/jobs/dron/workspace/Nodes/master/build/fw_sparky/fw_sparky.elf section `.data' will not fit in region `flash'
/home/mlyle/dronin/tools/gcc-arm-none-eabi-4_9-2015q1/bin/../lib/gcc/arm-none-eabi/4.9.3/../../../../arm-none-eabi/bin/ld: region `flash' overflowed by 48 bytes
collect2: error: ld returned 1 exit status
```

Wow, looks like we're really down to the wire on sparky again.

jihlein:
Does jenkins use different compiler options?  This built fine for me rebased off the latest next last night.  Tested fine on the bench.

mlyle:
@jihlein No, it doesn't, but we've been within a hundred bytes of being out of flash on sparky for awhile, so it really matters what commits have sequenced in before.  Issue #527

mlyle:
jenkins, test this please

mlyle:
With the new tools, this just squeaks in. :)

PR #590, opened 2016-02-02
revomini: default to no telemetry on mainport

### LINKED ISSUE #568: RFM22B bind wizard should disable "telemetry" from board serial ports
RevoMini defaults to 'telemetry' on mainport; this prevents rfm22b from working.

Alternatively change the revomini mainport default.

Credits to @pug398 for initial report.

mlyle:
fixes #568

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/590)


tracernz:
Makes sense and matches other targets which only have Telemetry on HID by default.

PR #592, opened 2016-02-02
gcs: usability improvements

### LINKED ISSUE #405: GCS Buttons Apply and Save confuse new users
GCS Buttons Apply and Save confuse new users.  It was suggested to remove the "apply" button.  Here is the chat session on the topic.

<dmorgan_> cool.  i'll just follow your advice and use save.
<icee> apply makes the flight controller start using the settings
<icee> save does that, -and- makes the flight controller save it to nonvolatile memory
<icee> you almost always want save
<droidicus> icee, from a user perspective, is there a good argument for "apply" to exist at all?
<icee> it probably should go away
<icee> it's nice to do something "dangerous" and be able to undo it by rebooting
<icee> but that's an edge case
<droidicus> or maybe a dev-profile only option?
<icee> nah, that's harmful in other ways
<OhmEye> yeah, apply should go away then imho
<OhmEye> If I really want to undo something I can use export/import

mlyle:
- Enable "expert mode" in default configs.  Fixes #405
- Eliminate "apply" button on a few more tabs when expert mode is not set.  Fix up associated instruction text.
- Improve autotune instructions in GCS to be a little more minimal/clear.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/592)

### DIFF
--- a/ground/gcs/share/default_configurations/developer.xml
+++ b/ground/gcs/share/default_configurations/developer.xml
@@ -4,7 +4,7 @@
     <AutoSelect>true</AutoSelect>
     <Description>Developer</Description>
     <Details>Developer mode configuration</Details>
-    <ExpertMode>true</ExpertMode>
+    <ExpertMode>false</ExpertMode>
     <OverrideLanguage>en_US</OverrideLanguage>
     <SaveSettingsOnExit>true</SaveSettingsOnExit>
     <StyleSheet>default</StyleSheet>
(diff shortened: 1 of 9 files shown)


tracernz:
:lgtm_strong: 
Maybe make the vimeo URL a link though while you're at it?

---

Reviewed 9 of 9 files at r1.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/592#-:-K9evHbpUSJuE2-JKSAz:-1643967744)_

tracernz:
The expert mode setting change doesn't show/hide the buttons until after GCS is restarted (the show/hide happens when the plugins load). Probably outside the scope of this though.

mlyle:
Yes-- the idea is we'll put users in expertmode to begin with.  Most of tehse things require restart.

Unfortunately I had no success in quick attempts to linkify that.

PR #601, opened 2016-02-03
python: signature: search beginning of file for sig

### LINKED ISSUE #555: Python log tools should be willing to skip a few nulls or other characters to find signature
OpenLog looks like it often gets a little bit of garbage-- usually a few nulls-- before the actual log data, and we only look for the signature at the very first byte.

mlyle:
fixes #555

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/601)


mlyle:
Tested, works well.

tracernz:
LGTM (reviewable is down).

PR #602, opened 2016-02-03
windows installer: refinements

### LINKED ISSUE #529: Windows CDC driver is out-of-date
Needs Sparky2 and Lux VID/PID fixed/added.

mlyle:
- Blow away old firmware files (so when a target is removed we don't
  leave old firmware behind.
- Remove CDC drivers, which do not work on Windows 8 or later and are
  not needed for most use cases.
- Edit various text items for clarity
- Add to the CDC inf LUX.  Fixes #529

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/602)


mlyle:
Not yet tested.

mlyle:

tracernz:
I think we should just not attempt to install the CDC driver on Windows 8 (but still unzip it somewhere), and not include it on Windows 10 (since it is unnecessary). Not including it at all on Windows <10 reduces functionality. We still have a lot of Windows 7 users.

mlyle:
But do we expect any of them to want CDC?  We can give them instructions on how to do it themselves.

Win 7 x64 requires signed drivers too, btw.

tracernz:
We cut off the possibility of BLHeli passthrough by removing CDC support (unless we re-implement BlHeliSuite ourselves).

tracernz:
Also prevents using ComUsbBridge for passthrough connection to OSDs etc.

mlyle:
So I looked and I was surprised-- 64 flight boards on Windows 7.  

The question is-- what proportion of them will still be Windows 7 when we ship next (some attrition can be expected), what proportion of those are on 32 bit Windows (so they could benefit from this), do not already have the Tau drivers installed, and want to use CDC.  I bet that boils down to almost nothing.

We need to figure out CDC on Windows 10 because this is the future.  We also may have to do some "new" kind of driver thing for tunneling serial through UAVO.

mlyle:
Also note I left the .inf in our tree (and updated it for LUX), and I have no objection to documenting using it.

My concern is that this has made numerous people nervous, and they've attributed other problems to their "driver problem" and have fiddled with the installer.  If we're going to make this dependent on Windows version it's a lot of testing for a relatively small, ephemeral benefit-- it's something most people don't use.

tracernz:
So let's remove the installer but still leave the option that copies the driver inf file though?

> We need to figure out CDC on Windows 10 because this is the future. We also may have to do some "new" kind of driver thing for tunneling serial through UAVO.

I didn't find any good open-source solutions that would allow GCS to provide a "soft/virtual" serial port on Windows? Easy on Linux/OS X though yes.

mlyle:
> I didn't find any good open-source solutions that would allow GCS to provide a "soft/virtual" serial port on Windows? Easy on Linux/OS X though yes.

http://stackoverflow.com/questions/1605721/faking-an-rs232-serial-port

> So let's remove the installer but still leave the option that copies the driver inf file though?

How about I just document it with a link to the blob?  That's as easy as telling people where in the destination tree we copy it.

tracernz:
> http://stackoverflow.com/questions/1605721/faking-an-rs232-serial-port

Same problem with no signed drivers.

> How about I just document it with a link to the blob? That's as easy as telling people where in the destination tree we copy it.

How many browsers like to download .inf files without lots of scary warnings? :P

mlyle:
>  Same problem with no signed drivers.

Some of the links are naughty signed drivers ;)

Some docs proposed here:
https://github.com/d-ronin/dRonin/wiki/USB

> How many browsers like to download .inf files without lots of scary warnings? :P

Don't know, but I know just our normal installer sets of Windows SmartScreen with scary warnings :P

tracernz:
Well not as bad as the warning Ubuntu used to generate with the deb :-P, it used to outright tell you that the package WOULD damage your computer :P.

mlyle:
Just spent some time trying to explain the CDC stuff to a user on IRC.  It's not fun / difficult to support.

I've tested this, both from clean installs and upgrades.  It nukes the old fw_coptercontrol file, etc, properly.

PR #457, opened 2016-01-16
Autotune improvements

droidicus:
This is a PR to improve autotune performance. Changes include the following:
- Change kalman inputs to use '4_u_t' instead of just 'u' (as suggested in the original autotune derivation by korken)
- Reduce the R-matrix expected measurement noise to more closely match the measured value on a real system
- Added a matlab script to help analyze autotune measurement progress from a flash log

edit by MPL:

Required flights and observations:
- [x] Flight F1 brushless; compared vs. baseline:
- [x] Flight F1 small quad; compared vs baseline:
- [ ] Flight F4 small quad; compared vs baseline:
- [x] Flight F4 large quad; compared vs baseline:
- [ ] Sure would be nice to tune one of these generic 250s we have showing up in great numbers

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/457)

edit by MPL: fixes #112, fixes #170

Issue #228, opened 2015-12-10
Zero/low throttle stabilization. a.k.a borisb 'air mode'


fujin:
Discuss and potentially implement stabilization at low/zero throttle to aid with hang-time manoeuvres.

```
19:58 <icee> fujin: it's basically 3 things:
19:59 <icee> 1) the actuators are nonlinear, so you may need more aggressive PIDs at low throttle settings (unless you do actuator linearization which we can do)
19:59 <icee> 2) choosing to enable stabilization when throttle=0
19:59 <icee> 3) doing clipping management of some kind so the commands make sense
```

References https://github.com/d-ronin/dRonin/issues/182 https://github.com/d-ronin/dRonin/issues/229

dustin:
Cons:  Makes landings harder/less safe.
Pros:  Don't need it.

mlyle:
I think we can do something better without the downsides, as mentioned in the other issue.

dustin:
@mlyle Can you summarize here?  I'm a little dubious, but people seem to want it (I mean, it sounds OK when I'm not landing, at least).

I'm also a little curious as to how the history has unfolded.  People built flight controllers and wrote stabilization loops.  They added specific code to disable stabilization at zero throttle.  Cleanflight added a feature to override that and continue to allow stabilization at zero throttle.  Betaflight forked cleanflight and created that feature again.

It might be worth having a short writeup that describes why each of those decisions were made (i.e. what went wrong that required that change) along with whatever is done here and how it shows that we're not just repeating past mistakes.

mlyle:
@dustin : So, basically, we need logic to do smart things when commands clip.

If you have a throttle setting of 5%, and there's a pitch command of magnitude 10%, and say a yaw command of magnitude 1%, here's what happens:

The firmware spins the front left rotor at 14%.  It spins the front right rotor at 16%.

It wants to spin the rear left rotor at -4% and the rear right motor at -6%.  This doesn't happen and they spin at minimum speed.   So when there's clipping (either against full throttle or no throttle) the desired command doesn't get achieved.

In this case, the thrust delivered is (.14 + .16) / 4 = 7.5%.  This is more than the user asked for, and a reason when there's big oscillations or big commands multirotors tend to climb (along with actuator nonlinearity and the fact actuators may speed up faster than they slow down).

The pitch command is (.14+.16)/4 = 7.5% instead of the desired 10%.

The yaw command is (.02)/4 = 0.5% or half of what the user asked for.

[...]

PR #608, opened 2016-02-03
Actuator clipping management

### LINKED ISSUE #229: Actuator Clipping Management (ghettomixer v2+)
Discuss ghettomixer, settle on sane controls implementation @mlyle should probably sign off on

References https://github.com/d-ronin/dRonin/issues/182

mlyle:
This is the beginning, no "air mode" or anything yet.  Just wanted to get this in, and in place.

Adds a clipping alarm.  Removes the previous "anti-negative-clip" mechanism as it was unsafe.  

Unfortunately this will break a small part of #151 / #457 because of code relocation.

Fixes #229 
Connects to #228

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/608)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -13,7 +13,7 @@
  * @file       actuator.c
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2015
- * @author     dRonin, http://dronin.org Copyright (C) 2015
+ * @author     dRonin, http://dronin.org Copyright (C) 2015-2016
  * @brief      Actuator module. Drives the actuators (servos, motors etc).
  *
  * @see        The GNU Public License (GPL) Version 3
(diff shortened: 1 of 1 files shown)


mluessi:
Nice. I wonder what the best way to handle this is. It seems to me that knowledge of the mixer input and matrix would be helpful, so actuator output could be managed to give the desired response without clipping.

mlyle:
@mluessi IMO this does the sane thing-- if the commands are just bumped off the top it sacrifices throttle.  If the commands also exceed total range, it scales, which has the effect of mostly preserving the extreme / large command and slightly sacrificing other axes.

> It seems to me that knowledge of the mixer input and matrix would be helpful, so actuator output could be managed to give the desired response without clipping.

You've got to give up on something.  And it has to be the next layer up / an actual control loop that does it, so that integrators don't wind, etc.  That's a future item to work on (actuator clip feedback).

Next step after this merges will be to handle the negative-clipping case, low-power rate loop integrator windup cases, and to decay throttle before going to force-all-motors-to-neutral (air mode).  We also really need at least a minimal actuator curve fit, but Droidicus's autotune thing has to merge before that.

Also need space freed on sparky before this one can merge, so all in all a lot of dependencies to manage.

droidicus:
Feel free to break #457, it is fixable.

mlyle:
jenkins, test this please

fujin:
yay, even a clipping alarm! excited to test this when I'm back in NZ.

mlyle:
Interesting effect when playing with this at zero-throttle.  Moving the sticks causes an "ACTUATOR" alarm, because anything clips.

mlyle:
Apologies for world's worst video.... unknown cause of that.

Here's some clipping management (and setting off the clip alarm) during aggressive rolling.

https://youtu.be/-1gIK2GcjP8?t=1m57s

dustin:
This looks pretty good so far.  :lgtm:

---

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/608#-:-K9nF0TD_yY-MD8J0ldp:-215887419)_

PR #403, opened 2016-01-07
Add Sbus Non Inverted Mode

jihlein:
Fixes #390.

This PR implements a non-inverted Sbus interface on the following targets: AQ32, Naze32, and Sparky.  The mode is called Sbus Non Inverted on the GCS setup screens.  The 3 targets implemented here give an F1, F3, and F4 example.

I've tested the capability to selected both Sbus and Sbus Non Inverted on all applicable ports on the above 3 targets.  Each target was able to boot and execute as expected.  As I don't have any Sbus receivers, I can not test further.  Hopefully @dustin and @Nub can test with SBus receivers.  Testing should include, at a minimum, both manual setup and wizard setup as applicable.  A quick test flight wouldn't hurt either.

As the port drop down boxes were getting long and somewhat disorganized, I alphabetized the selections so the lists don't look as disorganized.

A future PR should implement the Sbus Non Inverted mode on the remaining targets, along with the alphabetized selection list so all targets look uniform.

If there's a better way to implement some of this, let me know.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/403)

PR #618, opened 2016-02-06
Flight - MSP Bridge stack space overflows causing reset.

jihlein:
During bench testing of latest next plus #403, resets were observed when both GPS and MSP modules were enabled.  Cause of reset traced to insufficient MSP bridge stack space, likely due to new comp GPS calculations.  With MSP bridge stack size increased to 768 bytes, reset no longer observed, and free stack space of ~80 bytes indicated.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/618)


mlyle:
jenkins test this please

mlyle:
Thanks for the report and figuring this out!

PR #619, opened 2016-02-06
python: remove dumb regression

mlyle:
Prevented non-serial use of API.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/619)


mlyle:
Verified I can do stuff with the API (serial, TCP, file) and it's happy.

PR #623, opened 2016-02-07
makefile: manage flight-side recursion

mlyle:
better manage the all_flight target

avoids double-building flight targets and racing with ourselves.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/623)


tracernz:
Builds all of the things correctly.

PR #625, opened 2016-02-07
Config: Use UAVO field descriptions for widget tooltips

tracernz:
When widgets don't already have a tooltip defined.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/625)


tracernz:
Added HTML escaping to this and to the existing UAVO browser tooltips after discussion with @mlyle.

PR #633, opened 2016-02-09
AH throttle_desired integral calculation error

droney:
The current altitude hold loop has two active paths which allow control to be passed to the throttle_desired pid_apply_antiwindup() function during flight. The first is the loop timeout specified by dt_ms and the second is changes to the AltitudeHoldDesired object. The latter occurs every 20 ms as defined by the ManualControl task rate. Unfortunately, when the AltitudeHoldeDesired object is updated, regardless of whether the throttle changed or not, the pid_apply_antiwindup is called at some time interval less than dt_ms. Therfore a velocity error is computed for this shortened time interval yet pid_apply_antiwindup is still passed the dt_ms time value, resulting in a throttle_desired PID integral calculation error.

I am opening this up for the experts to weigh in on this proposed change where the ManualControl and AltitudeHold task be performed independent of each other, where altitudeHoldTask will sample AltitudeHoldDesired only when the dt_ms timer expires. This proposed change eliminates the integral calculation error associated with the incorrect velocity error calculated off cycle from dt_ms. I further suggest this change allows the AH task to run at a more reasonable loop rate (20ms or longer), thereby freeing up cpu resources.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/633)

### DIFF
--- a/flight/Modules/AltitudeHold/altitudehold.c
+++ b/flight/Modules/AltitudeHold/altitudehold.c
@@ -168,6 +168,7 @@ static void altitudeHoldTask(void *parameters)
 			timeout = engaged ? dt_ms : 100;
 
 			continue;
+
 		} else if (ev.obj == AltitudeHoldSettingsHandle()) {
 			AltitudeHoldSettingsGet(&altitudeHoldSettings);
 
--- a/flight/Modules/AltitudeHold/altitudehold.c
+++ b/flight/Modules/AltitudeHold/altitudehold.c
@@ -134,7 +134,6 @@ static void altitudeHoldTask(void *parameters)
 	struct pid velocity_pid;
 
 	// Listen for object updates.
-	AltitudeHoldDesiredConnectQueue(queue);
 	AltitudeHoldSettingsConnectQueue(queue);
 	FlightStatusConnectQueue(queue);
 
@@ -145,7 +144,7 @@ static void altitudeHoldTask(void *parameters)
 	AlarmsSet(SYSTEMALARMS_ALARM_ALTITUDEHOLD, SYSTEMALARMS_ALARM_OK);
 
 	// Main task loop
-	const uint32_t dt_ms = 5;
+	const uint32_t dt_ms = 20;
 	const float dt_s = dt_ms * 0.001f;
 	uint32_t timeout = dt_ms;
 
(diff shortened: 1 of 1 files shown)


mlyle:
jenkins, test this please

mlyle:
Hey, thanks for noticing this!

This is a good change; of course, it effectively changes the control coefficients/default tuning.  While before there was a subtle dependency on the other control 

We might want to also change it to use the actual measured dT.

Have you flown this?

Can you please add the project standard copyright notice to the top of the file?

I am likely to merge #569 first-- which will probably conflict with this.  We can offer help (if you need it) on how to rebase this once that happens.

droney:
mlyle - what do you mean 'change it to use the actual measured dT'? The loop rate is based on the specified dt_ms parameter.

I have flown this on a 680 heavy quad and noticed improved attitude compensation, especially on slow pitch/roll attitude changes. I have also noticed higher Velocity Kp and Ki settings are possible while maintaining smooth control. Lastly, altitude adjustments while in AH mode sound better (smoother and uniform).

Regarding copyright... are you referring to a dronin copyright, since the taulabs copyright already exists?

As I am new to git, please provide instructions for re-basing should you choose to merge 569 first.

mlyle:
> mlyle - what do you mean 'change it to use the actual measured dT'? The loop rate is based on the specified dt_ms parameter.

Ideally we could measure the time between successive invocations of the loop-- e.g. the actual task time.   This is noisy because RTOS jitter is included, but it handles cases where the task is delayed or runs prematurely better.   e.g. here is an example from attitude.c:

```
        // Compute the dT using the cpu clock
        dT = PIOS_DELAY_DiffuS(timeval) / 1000000.0f;
        timeval = PIOS_DELAY_GetRaw();
```

Yes.  We're fully diverged / separate from Tau Labs.  The line to add above or below the Tau Labs license line is:

```
 * @author     dRonin, http://dRonin.org/, Copyright (C) 2016
```

and is found in LicenseTemplate.txt.

It looks like we got lucky and don't need a rebase after 569 landed.

droney:
mlyle - looks like I screwed up the diff posting and the only thing showing up is a blank line. I am ready to resubmit with the copyright and measure dT, but I want to make sure all changes are visible. What is the procedure to do this?

mlyle:
@droney -- it looks like so far you have 2 commits.  One of which is a blank line, and one is your logic changes.

What you will want to do is to make additional change to the code on your branch, then git commit it, then git push to "add it on" to this pull request.

I'd also suggest you consider a short first line for your commit message, and then a blank line, and any other additional detail you'd like to add... e.g.

```
flight: althold: use measured dT in control loop

any additional detail you want here ...
```

Because there are tools we use that expect git messages to be in that form (they treat everything up to the blank line as the "short" commit message when we're seeing them in lists).

I know the learning curve with git is a bit high, but we'll have you sorted in no time :)

mlyle:
@droney-- thanks for the PR.  I'm going to rebase this for you and re-submit because of conflicts.  Then I think it can get tested and merged.

mlyle:
Superseded by #707

PR #639, opened 2016-02-10
failsafe: Prevent arming in a different fashion

### LINKED ISSUE #629: Flight: Failsafe flight mode causes config error
Bad usability bug introduced in #473. The SystemHealth widget doesn't tell you why either. I don't see a quick fix unfortunately.

mlyle:
Fixes #629

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/639)

### DIFF
--- a/flight/Libraries/sanitycheck.c
+++ b/flight/Libraries/sanitycheck.c
@@ -264,6 +264,13 @@ static int32_t check_safe_to_arm()
 			case FLIGHTSTATUS_FLIGHTMODE_STABILIZED3:
 			case FLIGHTSTATUS_FLIGHTMODE_ALTITUDEHOLD:
 				break;
+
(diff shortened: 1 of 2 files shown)


flight/Libraries/sanitycheck.c
```diff
@@ -264,6 +264,13 @@ static int32_t check_safe_to_arm()
 			case FLIGHTSTATUS_FLIGHTMODE_STABILIZED3:
 			case FLIGHTSTATUS_FLIGHTMODE_ALTITUDEHOLD:
 				break;
+
+			case FLIGHTSTATUS_FLIGHTMODE_FAILSAFE:
```
tracernz, line 272: should note in the doxygen docs for the function (or make some), or rename it. As is there it is implied that it's safe to arm while in failsafe mode.

Issue #472, opened 2016-01-18
Corrupt flash log files


droidicus:
Occasionally I will get a log file from the the flash download that is not parsable from the Matlab LogConvert.m script. The problem always occurs at the end of the file. Attached is a log that is demonstrating this issue. Attached is a zip containing a log file, and the matlab script that demonstrates this.

[autotuneLogsCorrupt.zip](https://github.com/d-ronin/dRonin/files/95006/autotuneLogsCorrupt.zip)

The output generated by Matlab is as follows:
```
>> LogConvert('ShortFlyNoiseDefaults.drlog' , true)

***Tau Labs log parser***

warning: Git hashes do not match. Incorrect file importer.
warning: called from
    LogConvert at line 2399 column 2
warning: implicit conversion from numeric to char
wrongSyncByte instances:             0
wrongMessageByte instances:          0

Completed bytes:        36 of    221085
Est. time remaining, 00h:00m:07s
warning: implicit conversion from numeric to char
wrongSyncByte instances:             0
wrongMessageByte instances:          0

Completed bytes:     50055 of    221085
Est. time remaining, 00h:00m:03s
warning: implicit conversion from numeric to char
wrongSyncByte instances:             0
wrongMessageByte instances:          0

Completed bytes:    100078 of    221085
Est. time remaining, 00h:00m:03s
warning: implicit conversion from numeric to char
wrongSyncByte instances:             0
wrongMessageByte instances:          0

Completed bytes:    150079 of    221085
Est. time remaining, 00h:00m:02s
warning: implicit conversion from numeric to char
wrongSyncByte instances:             0
wrongMessageByte instances:          0

Completed bytes:    200081 of    221085
Est. time remaining, 00h:00m:01s
warning: implicit conversion from numeric to char
wrongSyncByte instances:            20
wrongMessageByte instances:          0

Completed bytes:    221088 of    221085
Est. time remaining, -1h:59m:60s
error: LogConvert: A(I): index out of bounds; value 221086 out of bound 221085
error: called from
    LogConvert at line 5219 column 20
```

If line 5219 of LogConvert.m is commented out, then the file will successfully parse.

This file was downloaded from a Sparky2.

mlyle:
Matlab should be able to deal with this-- it's just truncation (trying to read past the end / get the last record which isn't fully there.)

mlyle:
Just to note, the "fixes #..." needs to be in the top comment (edited in after is OK) of the PR for autoclose to work.

[...]

PR #651, opened 2016-02-11
Matlab, add buffer length check

droidicus:
In case of truncated file, skip decoding of objects that have been truncated.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/651)


droidicus:
Fixes #472

mlyle:
You've tested thoroughly?

droidicus:
I tested it with 5 log files that are truncated (fail with current next), and 10 more "normal" log files. Seems to be robust.

mlyle:
GOod enough for me.

Changeset looks good; cleaned up somewhat.

PR #652, opened 2016-02-11
gcs: config tab selection: behavior improvements

mlyle:
1. Remember the "current tab" so that we "get back to" where the user was last on connection.  Inherited from an unmerged @guilhermito changeset.
2. Hide the OSD config tab if we don't have that board capability.

Tested on brain and sparky2 several cycles.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/652)

### DIFF
--- a/ground/gcs/src/plugins/config/configgadgetwidget.cpp
+++ b/ground/gcs/src/plugins/config/configgadgetwidget.cpp
@@ -230,6 +230,7 @@ void ConfigGadgetWidget::onAutopilotDisconnect() {
     icon->addFile(":/configgadget/images/hardware_selected.png", QSize(), QIcon::Selected, QIcon::Off);
     QWidget *qwd = new DefaultHwSettingsWidget(this, false);
     ftw->insertTab(ConfigGadgetWidget::hardware, qwd, *icon, QString("Hardware"));
+    lastTabIndex = ftw->currentIndex();
     ftw->setCurrentIndex(ConfigGadgetWidget::hardware);
 
     emit autopilotDisconnected();
@@ -280,7 +281,7 @@ void ConfigGadgetWidget::onAutopilotConnect() {
     icon->addFile(":/configgadget/images/ins_selected.png", QSize(), QIcon::Selected, QIcon::Off);
     qwd = new ConfigAttitudeWidget(this);
     ftw->insertTab(ConfigGadgetWidget::sensors, qwd, *icon, QString("Attitude"));
-
+    ftw->setCurrentIndex(lastTabIndex);
     emit autopilotConnected();
 }
 
--- a/ground/gcs/src/plugins/config/configgadgetwidget.cpp
+++ b/ground/gcs/src/plugins/config/configgadgetwidget.cpp
@@ -4,7 +4,7 @@
  * @file       configgadgetwidget.cpp
  * @author     E. Lafargue & The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2014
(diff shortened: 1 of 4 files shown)


tracernz:
What happens if the last tab index doesn't exist anymore (e.g. OSD)?

mlyle:
It'll take you back to OSD.  ;)

mlyle:
(It still exists; just hidden).

mlyle:
(I thought about adding conditionals for this case, but.. meh.  If you plug in a brain, and you plug in something else having left it on the OSD pane, you can illicitly get back to the OSD, shame shame :P)

mluessi:
Great idea. I think it would also be useful to hide the camera stabilization and txpid icons if the modules are not enabled, WDYT?

DTFUHF:
I like this change, very useful.

I also agree with @mluessi it would cut down on clutter some.

mlyle:
@mluessi I'd like to eliminate the page full of checkboxes, so that would mean moving to turning camera stab and txpid on and off on their respective pages.

(The bridges and GPS could be enabled just by configuring ports appropriately)

Battery, airspeed, vibration, geofence, Hott Telemetry, Picoc have their own tabs that we could set to turn their module on and off.  Perhaps camerastab should be moved to a 'Modules' subpage.

Things that don't fit into this model:
- Logging -- probably should get its own module subtab
- Path planner/path followers (should get their own 'stabilization' pane.  Arguably we could enable flight mode modules when the flight mode is on a switch position).
- Alt hold -- @ekaszubski is going to be factoring this into stabilization, so no more module there.

ekaszubski:
I'm a big fan of this in general. I also really like the idea of enabling flight mode modules when the corresponding flight modes are attached to a switch position.

tracernz:
As part of that we'd need to handle running out of memory better, as it would become much more difficult to see what's going on.

ground/gcs/src/plugins/config/configgadgetwidget.cpp
```diff
@@ -260,12 +277,14 @@ void ConfigGadgetWidget::onAutopilotConnect() {
...
-        else {
+        } else {
             qwd = board->getBoardConfiguration();
             if (qwd == NULL) {
                 qwd = new DefaultHwSettingsWidget(this, true);
-            } else {
+            }
+
+            if (!board->queryCapabilities(Core::IBoardType::BOARD_CAPABILITIES_OSD)) {
+                hasOSD = false;
```
tracernz, line 287: Can we do `if (index == ConfigGadgetWidget::osd) index = ConfigGadgetWidget::hardware;` here?

mlyle, line 287: it makes me sad but OK.

tracernz, line 287: It's just ugly to go to a hidden tab is all. :(

mlyle:
Updated / haven't built or tried it myself yet though

ground/gcs/src/plugins/config/configgadgetwidget.cpp
```diff
@@ -260,12 +277,18 @@ void ConfigGadgetWidget::onAutopilotConnect() {
...
+
+                if (index == ConfigGadgetWidget::osd) {
```
tracernz: Ah, I put you wrong here sorry. Needs to be:

``` c
                if (lastTabIndex == ConfigGadgetWidget::osd) {
                    lastTabIndex = ConfigGadgetWidget::hardware;
                }
```

tracernz:
Tested okay with the modification posted above @ L289. Once fixed, LGTM.

mlyle:
OK, different version of the change blindly made ;)

PR #653, opened 2016-02-11
Allow more autonomous flight modes for helicp

ekaszubski:
- Various non-critical helicp-specific fixes
  - GeoFence will go collective-neutral instead of full-negative, which will result in an only slightly less dramatic fall-out-of-the-sky crash; probably shouldn't use it on any heli you are fond of
- Various small fixes to enable helicopters to use more autonomous flight modes including:
  - PositionHold
  - ReturnToHome
  - PathPlanner

Testing progress
- [ ] Test for regressions on multis
- [x] Bench-test PositionHold on 480-size heli with GPS
  - behavior appears sane
- [x] Bench-test ReturnToHome on 480-size heli with GPS
  - difficult to verify behavior
- [x] Bench-test PathPlanner on 480-size heli with GPS
  - difficult to verify behavior
- [x] Flight-test PositionHold on 480-size heli with GPS
  - success; seemed a bit aggressive but definitely held position
  - thrust axis was in manual in PH; need to verify whether this is correct behavior
- [x] Flight-test loiter on 480-size heli with GPS
  - seems to work as expected
  - probably need to tweak some gains for position + velocity
- ~~Flight-test ReturnToHome on 480-size heli with GPS~~
- ~~Flight-test PathPlanner on 480-size heli with GPS~~

480-size testing platform with GPS (in foreground) for reference:
20151211_131353-small

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/653)

### DIFF
--- a/flight/Modules/ManualControl/geofence_control.c
+++ b/flight/Modules/ManualControl/geofence_control.c
@@ -42,6 +42,7 @@
 #include "geofencesettings.h"
 #include "stabilizationdesired.h"
 #include "systemalarms.h"
+#include "systemsettings.h"
 
 //! Initialize the geofence controller
 int32_t geofence_control_initialize()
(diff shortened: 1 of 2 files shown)


mlyle:
:) We probably don't need to pass the whole set of tests here to merge.

Hey, dumb question, and I think I asked you before-- what should we do with throttle on heli on failsafe?  It really should throttle down, at least eventually, right?

ekaszubski:
Just due to the fact that most helis are gonna get trashed in any typical failsafe scenario, it's hard to say what exactly should happen. My first reaction when I'm about to crash is to cut throttle, and you can really reduce crash damage if the motor isn't spinning when you hit the ground; worst case, you can still autorotate.

Ideally, heli failsafe would eventually implement some basic autonomous auto-rotation logic that cuts throttle and looks at acceleration + baro to try to maintain a slow rate of descent via slight downward collective while keeping the heli level. That'll be a fun flight mode to test :P

tracernz:
Yep, first sign of impending doom = cut throttle and start an auto, then you have time to decide what to do next :P.

mlyle:
P.S. I don't suggest you let RTH actually touch down in your heli.  It doesn't know when it's touched, and will play with cyclic on the ground which is probably not good.

mluessi:
Code LGTM

ekaszubski:
> :) We probably don't need to pass the whole set of tests here to merge.

Agreed, I'm just trying to cover all the bases here and also report on how well stuff is actually working on helis

> P.S. I don't suggest you let RTH actually touch down in your heli. It doesn't know when it's touched, and will play with cyclic on the ground which is probably not good.

Agreed, thanks for the reminder :)

Also, just to clarify, the current failsafe behavior for helicp is to cut throttle and set collective to neutral; I think this is what we should be doing until we get a more advanced failsafe/recovery system in place.

mlyle:
You've flown many of these and can check the boxes, right?

ekaszubski:
@mlyle Tried flying today but my sparky2 crapped out (luckily it was on the ground in between flights), looks like a short between 3.3v and GND somewhere. The checklist is up-to-date.

mlyle:
Ah, sorry to hear that :/

And I thought you had flight-tested PH and said it did kinda OK.

ekaszubski:
I flew AH for #569 and bench tested PH, RTH, and PathFollower for this PR. PH looked sane on the bench but it was hard to confirm whether the rest were behaving correctly. Presumably they are.

ekaszubski:
Swapped out my FC today with another sparky2, aiming to flight-test PH this afternoon.

ekaszubski:
Just flew another successful AH on the new FC on my 480 (only other AH test was on my 450). Swapping batteries and trying AH + INS, then PH.

ekaszubski:
Just flew PH, can confirm that it works okay (the gains seem a bit high but it was definitely staying around the same spot). One thing I need to confirm from the code is whether PH is supposed to imply AH (I was expecting that it would); during my testing, the thrust axis was definitely _not_ under autonomous control in PH.

mlyle:
It's a setting that defaults off-- VtolPathFollowerSettings / AllowAltControl or similar.

mlyle:
P.S. Very cool that it works :)

xfce:
Do you plan to develop such a attitude flight mode, when the heliCP flipped 180 degrees, I means propeller bellow, FC can give the negative-pitch to keep the cp stable fly

mlyle:
@xfce Yes, 3D modes are coming-- both for reversible motors and controllable pitch.  First just throttle in that range, but eventually bistable attitude mode, etc.

It's just a question of when.

xfce:
@mlyle yes, I am waiting for your bistable attitude mode, helicp can fly in attitude mode while propeller always below :)

ekaszubski:
@xfce I've written and flown a bistable attitude + global thrust mode that could lock onto an inverted level orientation and automatically project thrust from the global frame into the vehicle's frame allowing it to hover upright or inverted without manually changing the direction of the collective input. This was back when I still used OpenPilot. The code is trivial to implement (as a new flight mode) for helicp.

However, I've been focusing on improving support in general for helicp (including configuration in GCS as well as things like autotune and autonomous thrust control) and am just now getting to the point where I would consider re-implementing bistable attitude and global thrust.

As @mlyle said, the question is not if better 3D support will happen, it's a question of when. The next items on my list are #581 and moving altitude hold functionality into the Stabilization module. After that, I'm all for resuming work on 3D flight modes.

ekaszubski:
> It's a setting that defaults off-- VtolPathFollowerSettings / AllowAltControl or similar.

Yep, realized this afterwards.

I think this PR is safe to merge; I can't imagine loiter is going to be particularly awful, and althold has already been vetted.

ekaszubski:
Just flew another test of AH and PH; everything seems to be working pretty well.

PH kinda orbits the desired location but I was able to completely let go of the transmitter for around 2 minutes while the heli flew autonomously ("PositionHold video 1" below).

[AltitudeHold video](https://youtu.be/8GufBK9QYsA) (with bonus landing while still in AH)
[PositionHold video 1](https://youtu.be/i2Eik5NZ4Yk?t=38s) (flew in PH without any pilot correction for about 1m 40s)
[PositionHold video 2](https://youtu.be/wZwYXE4QO2M?t=57s) (with bonus emergency landing due to low battery)

One thing I've noticed during all of this is that the rate mode seems to wander a ton with attitude filtering set to INSOutdoor vs Complementary. Need to test with complementary to verify that it isn't something else that's changed.

@mlyle @mluessi @tracernz IMO this is definitely ready for merge :)

flight/Modules/ManualControl/geofence_control.c
```diff
@@ -74,30 +75,28 @@ int32_t geofence_control_select(bool reset_controller)
...
+		stabilization_desired.StabilizationMode[STABILIZATIONDESIRED_STABILIZATIONMODE_ROLL] = STABILIZATIONDESIRED_STABILIZATIONMODE_DISABLED;
+		stabilization_desired.StabilizationMode[STABILIZATIONDESIRED_STABILIZATIONMODE_PITCH] = STABILIZATIONDESIRED_STABILIZATIONMODE_DISABLED;
+		stabilization_desired.StabilizationMode[STABILIZATIONDESIRED_STABILIZATIONMODE_YAW] = STABILIZATIONDESIRED_STABILIZATIONMODE_DISABLED;
 	} else {
-		/* Pick default values that will roughly cause a plane to circle down and */
-		/* a quad to fall straight down */
-		stabilization_desired.Thrust = -1;
-		stabilization_desired.Roll = -10;
-		stabilization_desired.Pitch = 0;
-		stabilization_desired.Yaw = -5;
```
mluessi, line 97: It seems like we have lost the "circle down to ground" feature. On planes, this is probably a good idea.

mluessi, line 97: Or are we going to handle this in the failsafe flight mode?

ekaszubski, line 97: This is being deferred to failsafe flight mode, which currently implements the "circle down to ground" feature: https://github.com/d-ronin/dRonin/blob/next/flight/Modules/Stabilization/stabilization.c#L388

mlyle:
The INS kinda sucks still-- suggest 'complementary' for attitude stuff.  Sounds like a great test report.  Circling/toilet-bowling usually indicates magnetometer calibration error.

PR #656, opened 2016-02-12
python: change starting line

mlyle:
Use /usr/bin/env  ... which in turn searches for python in the path.
Nicer on Windows and elsewhere.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/656)


mluessi:
Nice!

PR #659, opened 2016-02-13
GCS:Config: Add voltage ratio box

### LINKED ISSUE #658: GCS:Modules: Allow specifying Vbat ratio rather than mV/V
^As per. Should probably keep the mV/V box as well.

e.g. 11.0 rather than 90.9 mV/V for 10k/1k divider

tracernz:
Connected to voltage factor box so changing one will update the other.

divided

Also changed step size for the voltage factor spinbox from 0.00001 to 0.01 mV/V for sanity purposes.

Fixes #658.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/659)


tracernz:
Added another commit which puts spaces between values and units in all the spinboxes on the battery tab (a pet hate of mine is no space there :-P).

henn1001:
neat :)

mluessi:
LGTM

PR #666, opened 2016-02-15
simulation: eliminate simvisualization

mlyle:
It speaks uavtalk over TCP; if we want to do pretty things that's
sufficient.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/666)


tracernz:
Can you explain more, I don't really understand. Did you mean insufficient?

mlyle:
This is a mechanism to stream out simple vehicle state over udp to a visualization interface.  We don't have anything that speaks that protocol, and the sim also speaks uavtalk like flight boards. This duplicate mechanism should be removed.

tracernz:
Gotcha.

flight/Modules/SimVisualization/simvisualization.c
```diff
@@ -1,133 +0,0 @@
...
-		uav_data.q[1] = simData.q2;
-		uav_data.q[2] = simData.q3;
-		uav_data.q[3] = simData.q4;
-		uav_data.NED[0] = simData.Position[0];
-		uav_data.NED[1] = simData.Position[1];
-		uav_data.NED[2] = simData.Position[2];
-		uav_data.camera_roll = camera.Roll * 30;
-		uav_data.camera_pitch = camera.Pitch * 45;
-		sendto(s, (struct sockaddr *) &uav_data, sizeof(uav_data), 0, (struct sockaddr *) &server, sizeof(server));
-		usleep(100000);
```
mlyle, line 124: P.S. if we decide to retain this, at the least the usleep 0.1sec should be removed, since that will do weird/undefined things with the rest of sim...  (not using chibios primitives)

PR #669, opened 2016-02-17
GCS: Do not automatically apply UAV settings with import

### LINKED ISSUE #668: GCS: Importing UAV settings sets all UAVOs
When importing an .uav file in the GCS, all objects present in the file are set, even if they are not selected to be saved to flash. This should not happen, as settings are changed without informing the user and attempting to fly without first restarting the flight controller (which will cause the correct settings to be read from flash) can cause a crash/unintended behaviour etc.

jpbarraca:
This will import all settings to a temporary object, but nothing will be modified in the FC without user intervention. The apply button will make settings temporarily active (apply to ram), while the save button will save settings to the persistent memory (flash)

Fixes #668

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/669)

### DIFF
--- a/ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
+++ b/ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
@@ -186,17 +186,14 @@ void ImportSummaryDialog::doTheApplySaving(int op)
         QCheckBox *box = dynamic_cast<QCheckBox*>(ui->importSummaryList->cellWidget(i,0));
         if (box->isChecked()) {
             UAVObject* importedObj = importedObjects->getObject(uavObjectName);
+            UAVObject* boardObj = boardObjManager->getObject(uavObjectName);
 
-            if(op & UAVSettingsAction::apply) {
-                UAVObject* boardObj = boardObjManager->getObject(uavObjectName);
+            quint8* data = new quint8[importedObj->getNumBytes()];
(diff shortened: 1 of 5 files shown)


mlyle:
jenkins, test this please

mluessi:
@jpbarraca very nice. I just tested this, and it seems like the settings are only saved when clicking Apply and then Save. I think the save button should do both, apply the settings and save them to flash.

jpbarraca:
@mluessi The save button will put the elements to flash but it will not apply them. In order to maintain a consistent experience this can be easily modified.

Update: Done

mlyle:
I think we should remove the apply button and just have one button do both,
like we changed the main ui to do.

jpbarraca:
@mlyle Ok. Fixed.

mlyle:

mlyle:
jenkins, test this please

mlyle:
jenkins, test this please

mlyle:

mlyle:
I broke jenkins playing with androidgcs :/  sorry

jenkins, build artifacts please (let's try again)

mlyle:
I like this.  Can someone easily try and merge (I have no flight boards here?) else I'll be back in a few days.

pug398:
I like it also. It works now as I always thought it should :)

mlyle:
Looks great from what I can see here.

Issue #514, opened 2016-01-23
Add stick scaling and expo for Throttle


ekaszubski:
- curve1 and curve2 should be used for control linearization and not for "control feel"
- Thus, we should allow users to set stick scaling and expo for Throttle just like we allow it for R/P/Y

ekaszubski:
To clarify, this effectively depends on #515, so we'd be enabling scaling/expo on the Power/Thrust axis which will cover both multirotors (via throttle) and helicopters (via collective).

mlyle:
Relates to #671

PR #671, opened 2016-02-17
Throttle curve dilutes AH/PH altitude control

droney:
During altitude hold and position hold, throttle_desired is used to
maintain the current altitude. However, the throttle_desired value is
desensitized when the throttle curve is applied in actuator.c. In an
attempt to overcome this desensitizes, one can crank up the attitude
compensation as well as the Kp/Kd terms which make uav setup different
between similar crafts with and without throttle curves. Furthermore, when
the user exceeds their throttle curve knees, due to controlled altitude
change with the stick or under strong winds conditions, the altitude change
can be drastic and unexpected due to the higher AH settings.

This change simply disables the throttle curve during AH and PH
flightmodes. Not sure if the proposed change is the best solution, so
interested in any feedback.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/671)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -440,6 +440,15 @@ float process_mixer(const int index, const float curve1, const float curve2,
  */
 static float throt_curve(float const input, float const * curve, uint8_t num_points)
 {
+        uint8_t flightMode;
+        FlightStatusFlightModeGet(&flightMode);
+
+        // don't use the throttle curve during altitude/position hold to avoid desensitizing
+	// the compensation being applied to maintain altitiude
+        if( flightMode == FLIGHTSTATUS_FLIGHTMODE_ALTITUDEHOLD
+                        || flightMode == FLIGHTSTATUS_FLIGHTMODE_POSITIONHOLD )
+                return input;
+
 	return linear_interpolate(input, curve, num_points, 0.0f, 1.0f);
 }
 


mlyle:
Good find.  This is reported (or at least mentioned) in #514.

I'm not sure this is the correct long-term fix (vs. the proposed solution in #514)-- any thoughts?

One other concern I have is with how we set the integral term for thrust when entering PH/AH--- flipping the switch into altitude-hold will no longer be initially thrust-neutral.

mlyle:
@ekaszubski  --- your thoughts?  This could make sense as a short term hack, but we'd have to be sure to remove it.

Jenkins, test this please.

ekaszubski:
@droney Like @mlyle said, I think this is an important discussion to have and I'm glad you're thinking about possible solutions.

> However, the throttle_desired value is desensitized when the throttle curve is applied in actuator.c.

Is this necessarily true? If the throttle curve range is [0,1] then the resulting throttle output range should not be reduced. The only situation I can see where it _would_ be reduced is if the throttle curve range is less than [0,1], and defining a throttle curve with a reduced range seems like a bad thing to do, especially since the min/max throttle signal value can and should be set via the PWM outputs.

So as per #514 and past discussions, I believe that the function of the throttle curve should be to help linearize the system outputs by mapping desired thrust/power to actuator percentages. It should not be used to set control feel, except where a change in feel is simply a byproduct of more linear outputs. If this premise is holds, then the throttle curve should always have its range on [0,1] (regardless of whether it is a straight line or a curve) and applying the throttle curve for AH should actually result in _more_ desirable behavior than just passing the raw throttle value through as the vehicle response will be more linear.

Also as mentioned in #514 I believe the better overall strategy is to add an expo function for desired thrust that is applied only in the ManualControl module (just like the expo for R/P/Y). This will ensure that any non-linearities in thrust response will be a function of user config or vehicle setup and will not be exacerbated by the stabilization modules.

droney:
hello ekaszubski - sorry I was not clear, I am not suggesting changing the range of the throttle, I am merely pointing out the behavior of applying the user throttle curve during AH/PH. The AH algorithm has this linear relationship between current altitude and the velocity error, via PositionKp. A new throttle value is computed from this velocity error. To simplify things, think of it as a change in throttle from the previous iteration of the loop to correct for the positional change from the previous loop. However, when you run this new desired throttle value through the user's throttle curve fit, the resulting change in actuator throttle is actually less than the throttle computed by AH because you are hovering in the flat region of the user's throttle curve (a change in stick produces a _smaller_ change in throttle). This ultimately results in a weaker AH because we are not applying the full correction... only a portion of it. I computed a needed throttle and I applied something less than that to the motors to correct for the error.

Seems to me we are applying a non-linear throttle response where a linear one produces better results. Most people test AH by bringing it to hover and flipping the switch.... yep, it hold reasonably well, cool. However, when you attempt slow roll or pitch movements during AH, it becomes real obvious altitude is not as maintained well in the presence of a throttle curve. After removing the throttle curve, you can easily hear and see the snappier responses to altitude errors as opposed to sluggish responses with the throttle curve enabled.

I am wondering if what you are suggesting in #514 is a bit different from the user defined throttle curve being applied during AH. #514 would employ customer functions to produce the desired linear response. The user's throttle curve does just the opposite in this case by producing a non-linear response to altitude errors, especially when outside disturbance such as wind results in larger corrections and pushes you beyond the knee of your throttle curve.
sorry I was not clear, I am not suggesting changing the range of the throttle, I am merely pointing out the behavior of applying the user throttle curve during AH/PH. The AH algorithm has this linear relationship between current altitude and the velocity error, via PositionKp. A new throttle value is computed from this velocity error. To simplify things, think of it as a change in throttle from the previous iteration of the loop to correct for the positional change from the previous loop. However, when you run this new desired throttle value through the user's throttle curve fit, the resulting change in actuator throttle is actually less than the throttle computed by AH because you are hovering in the flat region of the user's throttle curve (a change in stick produces a _smaller_ change in throttle). This ultimately results in a weaker AH because we are not applying the full correction... only a portion of it. I computed a needed throttle and I applied something less than that to the motors to correct for the error.

Seems to me we are applying a non-linear throttle response where a linear one produces better results. Most people test AH by bringing it to hover and flipping the switch.... yep, it hold reasonably well, cool. However, when you attempt slow roll or pitch movements during AH, it becomes real obvious altitude is not as maintained well in the presence of a throttle curve. After removing the throttle curve, you can easily hear and see the snappier responses to altitude errors as opposed to sluggish responses with the throttle curve enabled.

I am wondering if what you are suggesting in #514 is a bit different from the user defined throttle curve being applied during AH. #514 would employ customer functions to produce the desired linear response. The user's throttle curve does just the opposite in this case by producing a non-linear response to altitude errors, especially when outside disturbance such as wind results in larger corrections and pushes you beyond the knee of your throttle curve.

droney:
shoot... wrong button. Reopening.

droney:
mlyle - regarding your concern for setting the integral term and thrust neutrality, today that term is cleared when AH is engaged and thrust and consequently the integral term is adjusted based on current position. There is also the deadband present in AH, so as long as AH is engaged/disengaged while in the deadband region, is there an issue?

mlyle:
I am away travelling so I can't follow all of this now.  Definitely throttle adjustments in the curve for control feel could create variable gain that the ah logic could have trouble fighting.

My concern with the throttle curve thing is that stuff seems to be structured on the assumption you will engage AH when close to hover throttle and then move the stick to the center.

If a high-powered quad hovers at 18% real power and the user has used a curve to map that to 50% stick movement, this logic will result in setting the integral term to 50% throttle and cause a big initial climb.

ekaszubski:
So I talked to @mlyle about this some more and it seems like there needs to be some redesign of the throttle curve since people are using it for control feel but as this PR mentions, that can have bad results for autonomous modes.

However, as @mlyle mentioned, this PR looks like it will introduce a new issue if someone has a throttle curve set up that scales the throttle output _down_ relative to the throttle inputs (e.g. 50% throttle stick input is mapped to 18% throttle actuator output). In that situation the current AH logic will use a _larger_ value than intended when the mode is first initiated:
- assume we're in a hovering state in normal attitude mode; throttle stick is at 50% and throttle actuator output is 18% as per the throttle curve
- StabilizationDesired.Thrust is being set to 0.5 by [transmitter_control.c](https://github.com/d-ronin/dRonin/blob/next/flight/Modules/ManualControl/transmitter_control.c#L1080)
- when AH is first enabled it grabs StabilizationDesired.Thrust, which was 0.5, and sets its velocity PID iAccumulator to this value [[ref](https://github.com/d-ronin/dRonin/blob/next/flight/Modules/AltitudeHold/altitudehold.c#L167)]
- assuming the vehicle is in a hover at the correct altitude, all error values are zero and a value of 0.5 is passed out of the velocity PID (from iAccumulator) to the Stabilization module via StabilizationDesired.Thrust [[ref](https://github.com/d-ronin/dRonin/blob/next/flight/Modules/AltitudeHold/altitudehold.c#L251)]
- as is, the Stabilization module just passes Thrust through to the Actuator module via ActuatorDesired.Thrust [[ref](https://github.com/d-ronin/dRonin/blob/next/flight/Modules/Stabilization/stabilization.c#L281)]
- the Actuator module grabs ActuatorDesired.Thrust, sees that we're in AH/PH, skips the throttle curve, and returns 0.5 (instead of 0.18 returned by the throttle curve) for the throttle output [[ref](https://github.com/d-ronin/dRonin/pull/671/files#diff-6756853a8b52a94d64d1f12e4377cb31R448)]
- the vehicle shoots up into the sky

Thus, the solution to this problem is not as simple as just disabling the throttle curve in actuator, as this can have unexpected results.

mlyle:
Mapping the integral term through the throttle curve would work, though, though this is a little tricky and duplicative.

mlyle:
I've thought about this... I think we should move the throttle curve implementation up to stabilization.  We have motorinputoutputcurvefit which is more generally useful for linearization / gain scheduling down at the actuator level (looks at actual motor powers, not just throttle).

Then when we move althold into stabilization, too, solving this will be straightforward.

@ekaszubski @droney thoughts?

ekaszubski:
@mlyle Sounds like a good option. Will put more thought into it after #581 and althold -> stabilization.

mlyle:
Deferring-- sigh, really need to fix this.

mlyle:
This is implicitly superseded by removal of the throttle curve functionality.

PR #672, opened 2016-02-17
STM32F4 USART: Add single wire mode

mluessi:
Support single wire mode for UART on STM32F4. This is e.g. useful for S.Port telemetry, but it requires an external bi-directional inverter as the F4 doesn't have built-in inverters.

Status: bench tested to be working.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/672)

### DIFF
--- a/flight/PiOS/STM32F4xx/pios_usart.c
+++ b/flight/PiOS/STM32F4xx/pios_usart.c
@@ -219,6 +219,12 @@ int32_t PIOS_USART_Init(uintptr_t * usart_id, const struct pios_usart_cfg * cfg)
 	if (usart_dev->cfg->tx.gpio != 0)
 		GPIO_Init(usart_dev->cfg->tx.gpio, (GPIO_InitTypeDef *)&usart_dev->cfg->tx.init);
 
+	/* Enable single wire mode if requested */
+	if (usart_dev->cfg->single_wire == true)
+		USART_HalfDuplexCmd(usart_dev->cfg->regs, ENABLE);
+	else
+		USART_HalfDuplexCmd(usart_dev->cfg->regs, DISABLE);
+
 	/* Configure the USART */
 	USART_Init(usart_dev->cfg->regs, (USART_InitTypeDef *)&usart_dev->cfg->init);
 


tracernz:
I know we support this on F3, but do we have it on F4.

mluessi:
@tracernz well, this adds support for single wire on F4 ;). Basically, F4 supports single wire, whereas F3 supports single wire + inverters. Using single wire on F4 has the advantage that only the Tx pin is needed, so you can use the Rx pin for something else. However, for it to work with s.port, you also need a bi-directional single wire inverter.

tracernz:
Yes, but it's good for more than just S.PORT ;). F3 can do it on the Rx pin too which is good for serial RX + telemetry, e.g. Jeti. Wish the F1 & F4 had the F3 peripherals. :( What I meant to ask above is do we have it on F1, doh. :P

ekaszubski:
@mluessi This is great! We still need to modify something else (maybe hwshared) to actually give the user the option to use half-duplex for S.Port telemetry or for other purposes, right?

tracernz:
@ekaszubski The Sparky already used it so check board_hw_defs.c to see where it's used.

mluessi:
@ekaszubski yes, as @tracernz the uart config from sparky can be used (basically, just remove the rx pin and set `single_wire` to `true`). But, you still need an external bi-directional inverter and I'm not sure if there is an off the shelf part exists that does this. @tracernz any ideas?

ekaszubski:
@mluessi @tracernz So this doesn't look like a user-configurable option based on what is being said here. I currently use a homemade (proto-board) inverter for FrSky S.Port telemetry that assumes separate tx and rx lines on the flexi port, but if I (or upstream) modified the usart config for S.Port on sparky2 to enable single_wire, it looks like I'd be forced to either change my inverter hardware or maintain a branch of the code that disables single_wire mode. Is this correct? If so, I think we should consider adding an option to set single_wire mode in GCS (similar to the way we allow DSM mode to be configured in GCS).

In the end, I think using single_wire would be the better long-term solution, but it would be unfortunate to force that mode onto people who are currently using two-wire mode on the FC side.

mluessi:
@ekaszubski the single wire mode is just an option and none of the targets use it at the moment. We definitely don't want to make this the default as it would mean existing setups no longer work (as you said). Further, it would only make sense to make it a user configurable option for existing targets if there is an easy way to build the bi-directional inverter.

The reason I added this is because I'm working on new hardware that will have the bi-directional inverter built-in.

ekaszubski:
@mluessi Ah, so single_wire=true just means "this usart _can_ run in single wire mode", it doesn't mean "make this usart run in single wire mode"?

Are there any details available on the hardware you're developing? I'm looking to replace one of my sparky2s that burned out recently and am considering switching to a new FC.

mluessi:
@ekaszubski, no single_wire=true means that the usart is in single wire mode. But by default it is single_wire=false, so this is not a problem, i.e., for existing targets nothing changes and the usart will be in two wire mode.

I can't release too much about the new hardware, other than it will be similar to brain (built-in OSD) but better ;).

ekaszubski:
@mluessi Got it.

As a heli pilot, if the board has at least 5 servo headers (throttle/power from esc + 4 servos) and can be powered with 7.4v (for HV servos), I'd consider it.

Not sure how far along you are in the design process but one major issue I have with sparky2 is that the main/flexi ports have their VCC pin connected to VCC_UNREG so those ports can't be directly used to power a GPS or other peripherals with a 5v input limit; having main/flexi ports with their VCC pin connected to onboard regulated voltage (3.3 or 5v) or having the onboard regulated voltage exposed via a header would be a huge plus for me.

mluessi:
@ekaszubski it will support up to 8 PWM outputs, 6 without losing any other functionality. The best option is usually to run a separate 5V BEC to power the flight controller, GPS, etc and run the HV servos directly from a 2S battery. Most flight controllers don't have a regulated 5V supply, just 3.3V, but GPS etc usually need 5V (as they use another regulator to generate 5V internally).

ekaszubski:
@mluessi So adding the extra weight and complexity of a separate 2s power system to run my servos is really really undesirable when I'm already running huge 6s (1550 mAh for 450/480-class and 5000 mAh for 500-class) batteries and since many heli ESCs already integrate an adjustable BEC (5v/6v/7.4v) specifically designed to power the servos and FC.

This also implies that I would have to build a separate servo hub board with at least 4 servo headers just to get power and ground to the servos, not to mention all the extra wiring to get the PWM signals from the FC to the servos. I could understand it for a gas/nitro/kerosene heli but not for a full-electric heli. On the other hand, I would definitely consider getting an external 5v reg for GPS, and even the FC, but if the FC can't also at least act as a hub for HV servos, then that's definitely a deal-breaker for me :/

BTW the GPS I'm using has a 3.3v reg onboard but it's apparently totally happy using the 3.3v supply from sparky2's dedicated receiver header (though I understand this will not always work depending on the regulator and whether the board is smart enough to bypass the reg if the supply is already 3.3v).

PR #677, opened 2016-02-22
GCS:Uploader: Usability tweaks

tracernz:
Just some small tweaks I've wanted to make. The UI really needs some major work, but this is some small steps to make what we have better until that happens.
- Button visibility changed
- OSD capability added
- Progress bar only shown when used
- Partition browser sizing fixed
- General small tweaks

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/677)

### DIFF
--- a/ground/gcs/src/plugins/uploader/tl_dfu.cpp
+++ b/ground/gcs/src/plugins/uploader/tl_dfu.cpp
@@ -211,7 +211,7 @@ void DFUObject::run()
 bool DFUObject::DownloadPartition(QByteArray *fw, qint32 const & numberOfBytes, dfu_partition_label const & partition)
 {
     EnterDFU();
-    emit operationProgress(QString("%0 partition download").arg(partitionStringFromLabel(partition)), -1);
+    emit operationProgress(QString("Downloading %0 partition...").arg(partitionStringFromLabel(partition)), -1);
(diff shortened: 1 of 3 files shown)


mlyle:
Can't test yet.  But if this does what I think you'd do--- prevent hiding buttons-- thank God.

mluessi:
Very nice. I just tested this. Not having the button "jump around", i.e., always showing the same buttons at the same locations but disabling them is a huge usability improvement IMO.

An idea I had the other day: How about automatically selecting the correct firmware file when the board is connected? So one could just do "Rescue, Connect Board, Flash". Naturally, the "Open File" button should still be there in case one wants to open a bl_ file etc.

tracernz:
@hshred, yep, I have a branch somewhere fixing the firmware file location stuff but osx is being a problem.

tracernz:
Oops, @mluessi ^^

mluessi:
@tracernz I didn't just mean the file location (which btw is broken in the last release), I meant when the board is connected, the uploader could automatically select the file, so one wouldn't have to click on "Open File".

tracernz:
@mluessi, I understand. Finding the files is the pre-req and then it would be trivial to auto-load. Could probably go in the same  PR. ;-)

mlyle:
Confirmed working by @mluessi, code looks good.  Thanks much!

PR #683, opened 2016-02-23
flight/receiver activity: double activity detection range

### LINKED ISSUE #68: TX wizard: make robust to small movements/jitter

mlyle:
Make things less likely to grab the wrong channel as an interim measure
for Tanto release.

Fixes #68

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/683)


tracernz:
We could potentially do away with this UAVO if we re-write the input wizard to be a bit better and look at the channel delta itself. Anyway, LGTM for now.

PR #689, opened 2016-02-24
gcs: edit file dialogs to default to home dir

### LINKED ISSUE #684: GCS: OSX Default file location is unwritable.
File -> Export UAV Settings tries to put it in Machintosh HD, which is unwritable. Should put it somewhere better.

Also Tools -> Download Log.

mlyle:
fixes #684

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/689)


tracernz:
Might be better on OSX but it's worse elsewhere because now it overrides the last used path.

mlyle:
Yah... good point.

Unfortunately there's not really a great way to fix this, then.

PR #691, opened 2016-02-25
flight/manualcontrol: separate failsafe-disarm timer

### LINKED ISSUE #424: Consider seperate adjustable failsafe-timer.
To continue the discussion that i accidentally started in #357, it should be considered to implement a separate failsafe timer which disarms the craft after a predefined time after connection loss.

The current implementation is fine for a setup in which idle throttle means = motors off. In case of a failsafe event the motors will cut the power and after 30 seconds (default) the craft will disarm.

But i think the new conscious in the scene is to let the motors always spin when armed. See also -> #406 
With this change though, the parameter ArmingTimeout seems to be not sufficient enough to bring the craft, in case of a failsafe situation, into a safe state. The motors will just keep spinning at neutral speed until the fc disarms after 30 seconds by default (which is extremely long).

To quote the old OP wiki: 
"By consensus it is advised that for multi rotors, the safest condition is one where the throttle is cut to zero.  While the results may seem catastrophic for the vehicle (and often are), it is agreed by most that this is the safest state that can be applied to the widest set of operational circumstances.  It is better to let the copter fall to the ground, than to have it drift at hover throttle into a crowd of unsuspecting observers who generally don't realise just how dangerous a full set of spinning propellers can be.  Spinning props under a non zero throttle condition which come in to contact with grass, branches, or other objects can also cause ESCs to burn up, short out, and conceivably cause the LiPo battery to catch fire in an unknown location."

Also if something like this #228 is considered, then there will be a dilemma. You should to cut the motors immediately in case of a failsafe, but you also don´t want to accidentally disarm your craft in midair just because your throttle was too low for x seconds. This contradicts the current implementation with the ArmingTimeout parameter.

mlyle:
Defaults to off / just using the low throttle mechanism

No UI yet.  Fixes #424, but a new post-tanto issue should be opened for changing the
default and creating UI.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/691)

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -715,9 +715,21 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
 		bool low_throttle = cmd->Throttle < 0;
 
 		// Check for arming timeout if transmitter invalid or throttle is low and checked
-		if (!valid || (low_throttle && check_throttle)) {
-			if ((settings->ArmedTimeout != 0) && (timeDifferenceMs(armedDisarmStart, lastSysTime) > settings->ArmedTimeout))
+		if (!valid) {
+			/* This is a separate armed timeout for invalid input only.
+			 * It defaults to 0 for now (in which case we still check the below
+			 * low-throttle-or-invalid-input timeout */
+			if ((settings->InvalidArmedTimeout != 0) && (timeDifferenceMs(armedDisarmStart, lastSysTime) > settings->InvalidArmedTimeout)) {
 				arm_state = ARM_STATE_DISARMED;
+				break;
+			}
+		}
+
+		if (!valid || (low_throttle && check_throttle)) {
+			if ((settings->ArmedTimeout != 0) && (timeDifferenceMs(armedDisarmStart, lastSysTime) > settings->ArmedTimeout)) {
+				arm_state = ARM_STATE_DISARMED;
+				break;
+			}
 		} else {
 			armedDisarmStart = lastSysTime;
  		}
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -719,7 +719,7 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
 			/* This is a separate armed timeout for invalid input only.
 			 * It defaults to 0 for now (in which case we still check the below
 			 * low-throttle-or-invalid-input timeout */
-			if ((settings->InvalidArmedTimeout != 0) && (timeDifferenceMs(armedDisarmStart, lastSysTime) > settings->InvalidArmedTimeout)) {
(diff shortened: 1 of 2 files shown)


mlyle:
Tested with frsky on Lux; numerous failsafes.  Low throttle timeout and failsafe timeout are now distinct.  If no failsafe timeout configured, low throttle timeout is properly used.

dustin:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

_[shared/uavobjectdefinition/manualcontrolsettings.xml, line 138 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/691#-KBfNFQi-kYcJh5QbKoV:-KBfNFQjUTrIWQE9HI6N:1502627689) ([raw file](https://github.com/d-ronin/dronin/blob/92fd429ffdb506fef4b8d5371079f1f20d7a99b9/shared/uavobjectdefinition/manualcontrolsettings.xml#L138)):_
This name doesn't seem intuitive to me.  I don't think of failsafe while armed as being "invalid."

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/691)_

mlyle:
How do you feel about this?  I'm trying to be consistent with the code in transmitter control that calls the flag 'valid' for valid RX.

tracernz:
Commands while in failsafe are invalid... :/

mlyle:
I don't understand what you're saying here, @tracernz -- who you're agreeing or disagreeing with.

basically, if the receiver input is invalid, the flag valid is not set in transmitter_control.  This field is then the time from last positive throttle to disarm if the receiver is invalid.  Note it stacks with zero throttle, so if the invalid threshold is 5 seconds, and the low throttle threshold is 30.. and you have 6 seconds of low throttle and then _any_ invalid RX (out of range input, lack of input) it will disarm.

tracernz:
@mlyle That sounds good to me. Will review tomorrow.

dustin:
I can understand this as an implementation detail, but if I were wanting to
set the disarm timeout on failsafe, I wouldn't associate it with an invalid
command.

In fact, I'm just now wondering if this "invalid command" thing is why my
150 disarmed Saturday.  Is it possible the built-in OpenLRS decided to
consider something out of range and disarmed for that reason?

tracernz:
Code looks sane, :lgtm: 

---

Reviewed 2 of 2 files at r2.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/691#-:-KBkXZy4ob0s-hy2YGVZ:-1699641535)_

mlyle:
@dustin -- did it disarm or failsafe?

out of allowed range will definitely make us failsafe.

mlyle:
Anyways, this is effectively disabled by default.  I think it should merge.

dustin:
Yes, the 150 situation is different.

henn1001:
I like! :)

PR #692, opened 2016-02-25
GCS:Welcome: Exclude merge commits from GH activity

tracernz:
SNR improved significantly IMO.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/692)


tracernz:
Without:
screenshot from 2016-02-25 20 56 58

With:
screenshot from 2016-02-25 20 57 11

tracernz:
Added another commit that compacts stuff down like this to fit more:
screenshot from 2016-02-25 21 18 38
Can remove if desired.

Issue #147, opened 2015-12-05
Bootloader doesn't check success of partition erase operation


tracernz:
https://github.com/d-ronin/dRonin/blob/7a7c24987db33c93bb43d5aff802c1982e7e7e05/flight/targets/bl/common/bl_xfer.c#L252

PR #695, opened 2016-02-25
bl: fix partition erase

henn1001:
bug i stumbled upon today.
Would obviously mean that the bootloader version has to be bumped and updated for a new release.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/695)

### DIFF
--- a/flight/targets/aq32/board-info/board-info.mk
+++ b/flight/targets/aq32/board-info/board-info.mk
@@ -1,6 +1,6 @@
 BOARD_TYPE          := 0x94
 BOARD_REVISION      := 0x01
-BOOTLOADER_VERSION  := 0x84
+BOOTLOADER_VERSION  := 0x85
 HW_TYPE             := 0x00
 
 MCU                 := cortex-m4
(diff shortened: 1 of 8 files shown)


mlyle:
Good find.  Not something most people do, but we should bump bootloader vers too.  Are you willing to do that in this PR?

jenkins, test this please.

henn1001:
Sure thing - i´ll read into it how to do it.

mlyle:
It's BOOTLOADER_VERSION in board-info.mk

henn1001:
Done!
We should also consider including other changes to the bl when this is merged before a new release.

mlyle:
jenkins, test this please

tracernz:
Yep, worth considering if there's anything else we'd like to do to BL.

mlyle:
Yup.  Though this isn't enough of a change that we should advise users to upgrade the bootloader, IMO.  Just something nice to have when someone upgrades anyways.

tracernz:
We should address https://github.com/d-ronin/dRonin/issues/147 if we're bumping the version too.

PR #696, opened 2016-02-25
Remove unused variable in GCS, rename input type get/set functions

### LINKED ISSUE #694: GCS: Setup wizard: Reboot is always requested
due to this: https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/boards_brainfpv/brain.cpp#L180-L181

because this sends void instead of 0: https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/setupwizard/pages/inputpage.cpp#L84

the port number itself seem superfluous, as the board type's function has the ability to scan through all available ports (as this specific hardware target does)

DTFUHF:
fixes #694

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/696)

### DIFF
--- a/ground/gcs/src/plugins/boards_brainfpv/brain.cpp
+++ b/ground/gcs/src/plugins/boards_brainfpv/brain.cpp
@@ -124,14 +124,10 @@ bool Brain::isInputConfigurationSupported()
 /**
  * Configure the board to use a receiver input type on a port number
  * @param type the type of receiver to use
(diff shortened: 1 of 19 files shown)


DTFUHF:
tested on Brain

mlyle:
Tested both variants on lux.  The right things happen.  Thank you very much!

PR #711, opened 2016-02-27
stabilization settings: turn up weak leveling rate

mlyle:
We suggest using weakleveling a lot to learn rate, but it needs to be
more effective to do so.  The current coefficients do almost nothing--
.1 corresponds to taking out 63% of the pitch/roll in 10 seconds, or 30
seconds to remove 95% of the bank.

New value of .33 takes out 63% of the pitch/roll in 3 seconds, or 9
seconds to get 95% of the way to level.

Put another way, at 45 degree roll, it'll want to initially level at 15
deg/sec.  With 150 deg/sec default full-stick deflection rates, it'll
take 10% deflection to hold that attitude.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/711)

### DIFF
--- a/shared/uavobjectdefinition/stabilizationsettings.xml
+++ b/shared/uavobjectdefinition/stabilizationsettings.xml
@@ -38,8 +38,8 @@
 	<field name="MaxAxisLock" units="deg" type="uint8" elements="1" defaultvalue="15"/>
 	<field name="MaxAxisLockRate" units="deg/s" type="uint8" elements="1" defaultvalue="2"/>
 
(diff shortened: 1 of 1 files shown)


mlyle:
Flew weak leveling from new defaults .... I had preferred .5 in the past, but chose .33 to be conservative.  It's -very- weak leveling but should be enough to help newbies.  Should be pretty good for planes, too.

dustin:
I've never tried WeakLeveling, but changing defaults seems harmless.
:lgtm:

---

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/711#-:-KBfPcJIqrMuu_TfQs_2:2053020562)_

PR #712, opened 2016-02-27
naze32: configurable interrupt port for mpu6050

### LINKED ISSUE #702: Afromini Amaze fails to be detected properly
Afromini Amaze is coming up with `hse_value == 8000000`.  AFAICT, the code to figure out the `hse_value` is just about the same as baseflight.  Baseflight seems to also detect this as ≠ rev5, but manages to get i2c working.

If I force it to be considered a rev 5, it works fine.

(note:  I also tried next + one-true-mpu and had the same thing, just to be safe)

I'm sure I can fight with the baseflight code a bit more to figure this out.

dustin:
The port moves about among the various naze incarnations and isn't as
simple to detect as "is this a rev 5?"  Afromini Amaze is effectively a
rev 4, but has the MPU on port C.

This allows the user to override that when the default isn't working
correctly.  My Amaze works now.

Closes #702 

dronin_gcs

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/712)

### DIFF
--- a/flight/targets/naze32/fw/pios_board.c
+++ b/flight/targets/naze32/fw/pios_board.c
@@ -141,7 +141,7 @@ static const struct pios_ms5611_cfg pios_ms5611_cfg = {
 #include "pios_mpu6050.h"
 //#define PIOS_MPU6050_I2C_ADDR PIOS_MPU6050_I2C_ADD_A0_HIGH
 #define PIOS_MPU6050_I2C_ADDR PIOS_MPU6050_I2C_ADD_A0_LOW
-static const struct pios_exti_cfg pios_exti_mpu6050_cfg __exti_config = {
+static const struct pios_exti_cfg pios_exti_mpu6050_cfg_pb13 __exti_config = {
 	// MPU_INT output on rev4 hardware (PB13)
 	.vector = PIOS_MPU6050_IRQHandler,
 	.line = EXTI_Line13,
@@ -171,7 +171,7 @@ static const struct pios_exti_cfg pios_exti_mpu6050_cfg __exti_config = {
 	},
 };
 
-static const struct pios_exti_cfg pios_exti_mpu6050_cfg_v5 __exti_config = {
+static const struct pios_exti_cfg pios_exti_mpu6050_cfg_pc13 __exti_config = {
 	// MPU_INT output on rev5 hardware (PC13)
 	.vector = PIOS_MPU6050_IRQHandler,
 	.line = EXTI_Line13,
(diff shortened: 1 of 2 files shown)


dustin:
Tested on:
- Naze < rev 5
- Naze rev5
- Afromini Amaze

DTFUHF:
The work is appreciated but I have to argue against this, unless the C option is made to be the default

-What happens with a fresh flash on afromini v2/v3, with Auto as default? Does it boot, but not work until C is selected? Or does it not boot at all?

-Can anyone with an afrominiv2 verify that mpuint is on PC13? (I have at least two here, but I looked and can't find them :'(

-Do you really want to support Rev4, at any cost, a closed-source board obsoleted by 2 newer generations, last produced 2-3 years ago in far smaller quantity than the newer stuff?

-Naze target is already a bit of a black sheep, that works quite differently from the fully supported targets, and this change ensures that a newbie setting up an afromini/amaze is even more confused than normal

henn1001:
I think the flip32 has the rev4 layout too which is still widely available.

DTFUHF:
Damn clones! OK, this might be the best option then, if clone hardware is to be supported - and short of some trickery with board revisions and/or pin auto-detection

mlyle:
> -What happens with a fresh flash on afromini v2/v3, with Auto as default? Does it boot, but not work until C is selected? Or does it not boot at all?

He said before it booted but didn't update the AI / had sensor alarms.

I like this.  Real autodetection would be better but is hard.  

I suggest that we name this differently, though.  MPU6050IntPin... Auto/C13/B13.  Lets us have more options in the future if we need to (sigh) and it's just a little more clear in UI what's going on to people who understand GPIOs/interrupts, I think.

dustin:
Updated as suggested.

Tested flip32, naze rev5 and Afromini Amaze (which required changing the pin and rebooting).

mlyle:
Thx for doing this :)

If you get a chance, can you add a mention to https://github.com/d-ronin/dRonin/wiki/OpenNaze-Naze32 ?

tracernz:
Uhhh, it shouldn't boot without the interrupt pin working. I don't like this at all, hopefully we can find something better that doesn't require user intervention.

mlyle:
that would be even better.  But at least now it's possible.

DTFUHF:
I'm guessing it passes the WHOAMI check, but then never receives the mpu conversion complete interrupt so subsequently never receives data.

mlyle:
What's on the pin otherwise?  Can we enable both as conversion interrupts? ;)

DTFUHF:
PC13 was used for older BMP085 sensors, which isn't on rev4+ or afrov2+

DTFUHF:
So yeah, could be safe to do that

DTFUHF:
and PB13 is NC on rev4

dustin:
If there's a way to detect, I'm all for that.  Right now, this works the
way it did before, but also can work on my amaze.

Issue #642, opened 2016-02-10
power-up issue with a sparky2


thunderdan:
After setup for my fixed wing craft I observe a problem when I connect my flight battery without connecting USB.  I have no issue when I connect USB first, followed by battery second.  There is a problem if I just connect the battery only.

The problem looks like a glitching of the power rails.  The power LEDs blink abnormally.  The servos glitch and then end up at their extreme positions instead of centered like they should be.  

The battery is attached to an ESC with BEC and the cable then attached to servo output channel 3.

I'm using the dronin release from January 26th 2016 for both the gcs and the firmware onboard the plane.

Things I've ruled out:
  low battery (i've used a fresh 3 cell battery charged to 12.6V)
  incorrect servo connnections (control of surfaces and throttle is correct w/ usb plugged in)

Any suggestions on things to try?

thunderdan:
I added a video of the problem on youtube.  Link is http://youtu.be/LPm7VbA4zF0

thunderdan:
Update:

I erased all settings from board, ran through setup again, and tried again and it worked as expected this time.  Maybe the default initialization of the board has the PWM outputs set to 1000 us and once the firmware/settings load then the netrual values are used (1500 us for my plane). 

I'm not sure anyone should start to work on this unless I can reproduce the problem again.

thunderdan:
I have seen this problem again when I took my plane out to fly today.  I believe it is a power droop issue.  I have two batteries that I have used for about a year.  They are the zippy compact 3 cell, 2200mah, and 25C... and yes, they are charged to capacity (12.6V).  With these batteries, the sparky2 does not complete power-up and goes into a loop (see youtube video).

When I tried a new battery in my collection, venom 3 cell, 1300mah, 20C, there are no problems.  Power-up occurs just fine multiple times in a row.

Here are the things drawing power on my board:
- 3 servos on the pwm output pins
- 1 receiver (frksky d4r-ii) on the receiver port
- GPS (neo 8m based) on the mainport

If initialization took into account planes with servos and didn't drive them to 1000 us before setting them to neutral at 1500 us, it might fix this issue?  I know on quad copters the motors all need to be off (1000 us) but on planes output needs to be set to a midway point neutral (1500 us).

Last but not least, thanks for the great project.

thunderdan:
I confirmed the same problem exists on fw/sw from Tau Labs 2015-11-23 release.

mlyle:
[...]

PR #719, opened 2016-02-29
sparky2: bl: don't drive channels low on reset

mlyle:
Connects to #642

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/719)

### DIFF
--- a/flight/targets/sparky2/bl/pios_board.c
+++ b/flight/targets/sparky2/bl/pios_board.c
@@ -49,19 +49,11 @@ void PIOS_Board_Init()
 	
 	const struct pios_board_info * bdinfo = &pios_board_info_blob;
 
-	// Make sure all the PWM outputs are low
-	const struct pios_tim_channel * channels = pios_servo_cfg.channels;
-	uint8_t num_channels = pios_servo_cfg.num_channels;
-	for (int i = 0; i < num_channels; i++) {
-		GPIO_Init(channels[i].pin.gpio, (GPIO_InitTypeDef*) &channels[i].pin.init);
-	}
-
 #if defined(PIOS_INCLUDE_LED)
 	const struct pios_led_cfg * led_cfg = PIOS_BOARD_HW_DEFS_GetLedCfg(bdinfo->board_rev);
 	PIOS_Assert(led_cfg);
 	PIOS_LED_Init(led_cfg);
 #endif	/* PIOS_INCLUDE_LED */
-
 
 #if defined(PIOS_INCLUDE_SPI)
 	/* Set up the SPI interface to the flash and rfm22b */


mlyle:

mlyle:
@thunderdan Hey, this may improve your issue.  You'll need a new bootloader.  Are you willing to try out a bootloader update I provide to see if it changes things?

mlyle:
http://jar.lyle.org/~mlyle/bu_sparky2.tlfw

Procedure:  Rescue, flash this to board, "boot", wait for it to do its thing / stop blinking.

Then rescue, flash normal firmware to board.

Test IAP reset again to see if happier.

tracernz:
Yeah, I think this is a good idea. Will put it on a scope at some point to see what it does before/after.

thunderdan:
Hi.  I gave this a shot tonight with the two batteries I have for my plane.  With one battery, the start-up completes and the servos are centered.  With the other battery, the problem remains.  Start-up fails and it goes into a loop with the servos driven to their extreme position (which i think is the zero position).  The servos always move after attaching power with either battery.

Thanks @mlyle for the boot loader build.  This project is fun.

pug398:
@thunderdan Sounds like the first battery cannot source enough current to actuate all the servos and causes a brownout. Do the servos start from a random position and jump to center or are they centered and jump to an extreme when first powered? A servo's zero position is in the middle and is driven there by a 1500us pulse. 1000 or 2000 would be at each extreme. In your dealing with this problem did you try latest TL next which sets pwm default to off instead of 1000us?

thunderdan:
Yes I agree that is the situation.

Other suggestions to fix the issue were to add capacitance near the Sparky2 or a separate BECs for flight controller and servos.

I'll build the latest Tau Labs next branch tonight and give it a try.

It's great to see such an active community after open pilot.

-- Dan Morgan --

pug398:
The TL is just a test. Flash firmware and then in partition browser>settings right click and erase settings, then boot. Power up and note servos. Then setup vehicle type. Power up and note servos. Continue setup until you encounter issue again or you may get same issue immediately. Just gathering more info.

thunderdan:
Build is TL next from 20160302

Updated bu_sparky2.tlfw, fw_sparky2.tlfw, and erased settings

applied battery power with usb power still attached
- servos moved but they aren't at their extremes and board completed power-up
  disconnect usb and applied battery power (bat1)
- servos moved to extremes and board completed power-up
  usb still disconnected, applied battery power (bat2)
- servos moved to extremes and board completed power-up

Configured FC
applied battery power with usb power still attached
- servos moved from extremes to center and board completed power-up
- it was quick so power-up may complete then be driving the servos to correct position
  disconnect usb and applied battery power (bat1)
- servo move a little then move back to center
  usb still disconnected, applied battery power (bat2)
- power-up reset loop with servos bouncing

Thoughts:  In my fixed wing setup with 3 cheap servos powered through sparky2 the FC is sensitive to power dips.  I need to change my wiring to seperate power for FC and servos or add capacitance for FC.  Changes to BL and FW can't remove the pulse on the PWM lines that cause my servos to move during first power-up.

I have one other experience that really points to a power problem.  Check out this mid flight power-loss.  This is a very short flight where the battery should normally still have enough power to glide in.  Instead the FC resets.
https://youtu.be/Fh1Vij385bw

pug398:
I think separating power for fc and servos is a good idea. After configuring board it looks like you got correct behavior with a good battery. With a configured board a small jump (depending on initial position) and then moving to center can be expected. dRonin should have the same response when configured. Make sure the configured min/max endpoints don't push the mechanical limits of the system.

thunderdan:
I think this one can be closed too since #642 is closed.  The bootloader improvements were good and the final solution to my power brown-outs was adding more capacitance.   Thanks again for the help everyone.

mlyle:
A decision needs to be made whether to merge this-- this is a PR that you tested to eliminate the power-up glitch.  It seems you and @pug398 like it, but @tracernz is probably waiting to scope it before he pushes the merge button.

PR #720, opened 2016-02-29
IAP reset: Disable PWM in many cases during reset wait

mlyle:
Relates to #642.

When we go to deliberate reset, if the output channel is not inverted, generate all-low PWM.  (Inversion check is there to try and deal better if someone is doing something odd, e.g. P-fet high side switching of brushed motors).

This will let some ESCs disarm during the interval and not be fooled by short pulses.  It will also ensure the channel is low when reset makes us "let go" of it.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/720)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -693,12 +693,7 @@ static bool set_channel(uint8_t mixer_channel, float value)
 		return true;
 	}
 	case ACTUATORSETTINGS_CHANNELTYPE_PWM:
-#if defined(PIOS_INCLUDE_HPWM)
-		// The HPWM method will convert from us to the appropriate settings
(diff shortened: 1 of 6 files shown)


pug398:

mlyle:
Use caution with this-- props off for initial testing, etc.

mlyle:
Also note it doesn't contain 719.  I talked to simonk and it will not be sufficient to disarm simonk ESCs, but it will be enough for blheli and some digital servos.

pug398:
So I tested this on F1, F3 and F4. F4 with #719. I didn't see any adverse effect with the output pulled low briefly and then going high and staying there. I believe it may have disarmed my sk esc because they give a short beep when output goes low and then continuously beep while high. Worked fine also with blheli/oneshot esc.

pug398:

mlyle:
This has rotted / isn't getting merged.  Closing.

PR #723, opened 2016-03-01
Flight: Storm32bgc

### LINKED ISSUE #288: POItracking to brushless gimbal via MAVLink
Existing code (Potentially removed) is aimed at direct servo control, Propose enhanced POITracking control via MAVLink or STorM32 protocol directly to STorM32 brushless gimbal.

Project should re-enable and modify existing position tracking / following code with the STorM32 filling the role of the SparkyBGC with a readily available, low cost and opensource project.

Similar basic functionality is currently available for the Pixhawk

http://copter.ardupilot.com/wiki/common-storm32-gimbal/

https://github.com/TauLabs/TauLabs/wiki/User-Guide:-POI-tracking

https://vimeo.com/127901407

http://www.olliw.eu/storm32bgc-wiki/Main_Page

jihlein:
Simple control of Storm32bgc.  Enables sending cameraDesired.Declination and  cameraDesired.Bearing to gimbal controller over serial link.

A UAVObject for the controller's ACK response was added to make it easy to verify proper communication with Storm32bgc is occurring.  A value of 1 indicates successful ACK, all other values indicate an error of some sort occurred.

Implemented for F4 and F3 targets only.

Tested on AQ32, Quanton, and Sparky targets.  Could only test  cameraDesired.Declination since I only have a 2 axis gimbal (pitch/roll) setup at the moment.  No reason I can see why cameraDesired.Bearing would not work.

Tests were also conducted using the GCSRCVR function since all my receivers have homes in airframes at the moment.

edit by mpl: fixes #288

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/723)


tgreer:
Subscribing as relevant to my interests... Awesome work @jihlein

flight/targets/aq32/fw/Makefile
```diff
@@ -66,6 +66,7 @@ OPTMODULES += PicoC
 OPTMODULES += UAVOFrSKYSPortBridge
 OPTMODULES += Geofence
 OPTMODULES += Logging
+OPTMODULES += Storm32bgc
```
tracernz: Should be camel case "Storm32Bgc"

tracernz:
There isn't another UAVO you can use? A whole UAVO is quite heavy to hold a single true/false value.

tracernz:
We already have an alarms infrastructure which other modules use.

jihlein:
So for example in systemalarms.xml add:

```
    <field name="Storm32BgcAck" units="" type="uint8" elements="1"/>
```

tracernz:
You could, or you could add it to the existing alarm field, then it's easy to handle in GCS with nice UI for each error. See https://github.com/d-ronin/dRonin/tree/next/ground/gcs/src/plugins/systemhealth/html
The downside of this is that it will take an extra few bytes RAM on F1, and it would be nice to generalise the alarms to all BGCs we might support.

``` xml
        <field name="Alarm" units="" type="enum" options="Uninitialised,OK,Warning,Error,Critical" defaultvalue="Uninitialised">
            <elementnames>
                <elementname>OutOfMemory</elementname>
                <elementname>CPUOverload</elementname>
                <elementname>StackOverflow</elementname>
                <elementname>SystemConfiguration</elementname>
                <elementname>EventSystem</elementname>
                <elementname>Telemetry</elementname>
                <elementname>ManualControl</elementname>
                <elementname>Actuator</elementname>
                <elementname>Attitude</elementname>
                <elementname>Sensors</elementname>
                <elementname>Stabilization</elementname>
                <elementname>Geofence</elementname>
                <elementname>PathFollower</elementname>
                <elementname>PathPlanner</elementname>
                <elementname>Battery</elementname>
                <elementname>FlightTime</elementname>
                <elementname>I2C</elementname>
                <elementname>GPS</elementname>
                <elementname>AltitudeHold</elementname>
                <elementname>BootFault</elementname>
                <elementname>TempBaro</elementname>
                <elementname>GyroBias</elementname>
                <elementname>ADC</elementname>
                <elementname>Storm32Bgc</elementname>
            </elementnames>
        </field>
```

To set: https://github.com/d-ronin/dRonin/blob/next/flight/Modules/Stabilization/stabilization.c#L236

jihlein:
Couple of things:

1)Wasn't planning on adding to F1 targets.  Resources are too limited, particularly on Naze32 wrt to serial ports.

2)After more thought, I'm not sure the gimbal criticality is at the same level as the sensors or other truly flight critical systems in the alarm list.  All I'm really looking for from the gimbal is a successful ack to the command(s).

3)I didn't look to see the true impact of the dedicated UAVObject, but I like the way it operates.  If the gimbal module is not enabled and has no port selection, the object remains uninitialized.  If its enabled with a port, object is initialized and updated with the ACK response status.  Seems to give a clear path of troubleshooting steps to the user who is experiencing problems.

4)I also want to expand on what data I monitor from the gimbal, but this was a good first implementation to get things started.

Guess I'm arguing to keep the dedicated UAVObject........

mlyle:
If the object is going to be extended to have versions, etc, in the future, this makes sense.  We don't want to clutter up the namespace of objects with a billion objects, and they are kinda high overhead.  But being able to read stuff out about gimbal status could be nice.

jihlein:
So a possible alarm implementation would be:
1. If module not active or port not assigned, set alarm to uninitialised (actually that's the default).  Alarm will stay in this state until module is active and port assigned.
2. Once active and port assigned, if no response from the gimbal controller, set alarm to error.
3. If gimbal controller acks command correctly, set alarm to ok.
4. Set alarm to warning for any other gimbal controller response.

I think this gives the best of both worlds, does not need a new UAVObject, and clearly illustrates if the gimbal module is active and if the gimbal is responding without errors.

One concern:  would having the gimbal in a warning state or error state preclude the system from arming?  Wrt to gimbal, I don't think that's desirable.

I'm not sure how to set up anything wrt the GCS user interface.......

mlyle:
Warning prevents arming, error doesn't.

I'd reverse your two cases there.  Error for unexpected response (who knows
what's going on?), warning for gimbal missing..

jihlein:
Actually, it's error that prevents arming.......

So it's working with alarms as I described above.  I haven't pushed the code yet.  I noticed we have an empty box on the wca panel.  Could that be labeled gimbal, and be greyed over when the module is inactive, green when okay, yellow when in warning state, and red when in error state?  That would be neat, but I have no idea how to mess with the GCS code.  Any pointers on what needs modifications, anyone want to take over that part?  Should I push the current changes first?

mlyle:
haha, yes, somehow i reversed that :P

mlyle:
and yes, current changes first.  I think just a box needs to be added to the SVG graphic.

tracernz:
I can do the GCS changes and PR your branch if you like @jihlein. Note: I also have a branch that auto-generates the boxes from the UAVO rather than using an SVG so this will be easier one day.

jihlein:
@tracernz, that would be helpful, thanks.

mlyle:
Has this been tested all-together?

Also I see that the little "Gimbal" label is a little offcenter/probably has a different text-align, but I can look at fixing that separately.

jihlein:
I tested it all together last night.  I did not push the latest updates though.

It works, but there are a few cleanups necessary.  The alignment of the gimbal text and the red error "X" is off and needs to be adjusted some.  They display fine in Inkscape, but not the GCS for some reason.  Also I'm not sure the pop up messages are working, need to revisit those.  

Last, I want to change the port labels from Storm32bgc to Storm32Bgc (camel case change), and slightly restructure what gimbal return msgs are classified as warning, critical, and error.  There is no critical error right now, I'll need to add the pop up msg for that.  I'm leaning toward:

ALARM_OK - Response with correct CRC and no error code.
ALARM_WARNING - Response with correct CRC and an error code.
ALARM_CRITICAL - Response with incorrect CRC
ALARM_ERROR - No response

mlyle:
OK, let us know when you push what you want reviewed and its test status please.

(I think missing gimbal should be WARNING, still, btw)

jihlein:
Will do.  Trying to sort thru the alarms from least to worst severe.  Maybe given the criticality of the gimbal, and the granularity of the error detection we can do, we could simplify it to ALARM_OK (green) for a proper ACK and no error code, ALARM_WARNING (yellow) for everything else.  Or maybe ALARM_CRITICAL (red) for everything else.  It's either working and responding correctly, or not.

mlyle:
I think when it's missing, it may be benign-- warning.  But if something is there and giving unexpected responses-- what does it mean?  the wrong hardware is hooked up?  the gimbal is there and the user expects it to work but the baud rate / etc is wrong?

tracernz:
I think the error is misaligned on most, or all, of the alarms, not just Gimbal. The whole image is scaled really tiny, maybe integer rather than float positioning is happening somewhere in the code. :/

jihlein:
Rebased, tested, and latest pushed.  Ready for review.

Alarms are set as follows.  Gimbal module uninitialized:

image

Gimbal module initialized, gimbal not powered:

image

Gimbal module initialized, gimbal powered, responding normally:

image

If there is a fault, the gimbal alarm will turn red (critical), untested but assumed okay.

mlyle:
:lgtm:

---

Reviewed 18 of 43 files at r1, 10 of 11 files at r2, 2 of 7 files at r4, 19 of 19 files at r6.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/723#-:-KC3Md9_g7Y6DYrTP3yI:1417088721)_

tgreer:
@jihlein do you have any documentation on configuring and setting this up we could stick in the wiki?

jihlein:
@tgreer, I can pull something together for basic operation.  We can expand to POI later.

jihlein:
@tgreer, I took a bunch of screen shots to detail the basic setup.  Is there a folder for wiki images?

mlyle:
@jihlein it's dumb, but you can check out the wiki git repo and then push to it

tracernz:
Can you just drag them onto the github editor? That's how you do it for github comments.

jihlein:
Don't know, never edited a wiki before.  I cloned the wiki, added my images, but when I go to commit them, it shows a lot of deleted files in addition to my as of yet non versioned images.  Do I want to check all boxes for the commit, or just the non versioned images?  Don't want to mess this up.....

jihlein:
Tried to update the wiki, I don't have the proper permissions to push to the wiki, so I can't upload the figures.

mlyle:
Hmm. :/ I thought the wiki was open for push permission.  How are you trying to push and what happens?

jihlein:
This is a push attempt thru Tortise GIT:

image

PR #726, opened 2016-03-01
actuator: don't LPS forever in failsafe

### LINKED ISSUE #724: Report: Hangtime stays engaged in failsafe
The hangtime is correct for low throttle while valid signal but failsafe kicks in LPS for ~25 seconds when hangtime is set to 2 seconds. This may be expected behavior just not what I was expecting :)

mlyle:
fixes #724

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/726)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -298,7 +298,7 @@ static void actuator_task(void* parameters)
 			throttle_source = desired.Thrust;
 		}
 
-		bool stabilize_now = throttle_source >= 0.0f;
+		bool stabilize_now = throttle_source > 0.0f;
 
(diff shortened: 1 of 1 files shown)


fujin:
Gosh darn off by one. :+1:

mlyle:
Off by .00000000000000000000000000000000000034 here. ;)

The main problem is the 0 value is now used during failsafe.

pug398:
Just for info with LPS+Failsafe, failsafe would kick in when ArmedTimeout expired so it was somewhat adjustable after all :)

tracernz:
Discussed on IRC, seems reasonable. We can consider a different value for failsafe in future but that's outside the scope of this PR.

PR #732, opened 2016-03-01
gcs: autotune: improve workflow when applying.

### LINKED ISSUE #717: Confirmation on clicking "Apply Computed Values" on Autotune Setup page
It would be nice if you would get a confirmation that the action was actually executed.

mlyle:
When you apply autotune, a helpful dialog now shows up, and takes you to
the stabilization page to tweak/save.

Fixes #717

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/732)

### DIFF
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -54,7 +54,6 @@
 #include "extensionsystem/pluginmanager.h"
 #include "coreplugin/iboardtype.h"
 
-
 //#define CONF_ATUNE_DEBUG
 #ifdef CONF_ATUNE_DEBUG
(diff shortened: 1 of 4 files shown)


mlyle:
image

tracernz:
This is a good change for Tanto. We should look to improve things so this is unnecessary in post-Tanto though I think.

mlyle:
Yes.

What I was thinking-- we'd get down to one pane of autotune, with the instructions, the module enable checkbox, and a 'review results' button.  The review results button can pop up the results in a window, and let you select which ones to use (e.g. basic only? yaw?  outer kI?), whether to save, and whether to share, in one operation.

tracernz:
Looks good, works well.

PR #737, opened 2016-03-03
PiOS:HAL: Standardise panic codes between boards

tracernz:
As per title. Should be much more user friendly as all boards will be the same. Requires test.

TODO:
- [x] Add documentation to wiki

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/737)

### DIFF
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -189,7 +189,7 @@ uintptr_t pios_com_debug_id;
  * @param[in] led_id The LED to blink
  * @param[in] code Number of blinks to do in a row
  */
-void PIOS_HAL_Panic(uint32_t led_id, int32_t code) {
+void PIOS_HAL_Panic(uint32_t led_id, enum pios_hal_panic code) {
 	while (1) {
 		for (int32_t i = 0; i < code; i++) {
 			PIOS_WDG_Clear();
--- a/flight/PiOS/inc/pios_hal.h
+++ b/flight/PiOS/inc/pios_hal.h
@@ -17,6 +17,23 @@
 #include <pios_openlrs_priv.h>
 #endif /* PIOS_INCLUDE_RFM22B */
 
+/**
+ * @ brief LED panic codes for hardware failure
+ */
+enum pios_hal_panic {
+	// start at 2, so that panic codes are obvious to user
+	PIOS_HAL_PANIC_IMU  = 2,
+	PIOS_HAL_PANIC_MAG,
+	PIOS_HAL_PANIC_BARO,
+	PIOS_HAL_PANIC_FLASH,
+	PIOS_HAL_PANIC_FILESYS,
+	PIOS_HAL_PANIC_I2C_INT,
+	PIOS_HAL_PANIC_I2C_EXT,
+	PIOS_HAL_PANIC_SPI,
+	PIOS_HAL_PANIC_CAN,
+	PIOS_HAL_PANIC_ADC,
+};
+
 /* One slot per selectable receiver group.
  *  eg. PWM, PPM, GCS, SPEKTRUM1, SPEKTRUM2, SBUS
  * NOTE: No slot in this map for NONE.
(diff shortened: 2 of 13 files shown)


tracernz:
Dox, could be expanded upon if needed.

## Panic Codes

| **LED Flashes** | **Error** |
| --- | --- |
| 2 | IMU initialisation error |
| 3 | Magnetometer initialisation error |
| 4 | Barometer initialisation error |
| 5 | Flash initialisation error |
| 6 | Filesystem initialisation error |
| 7 | Internal I2C bus initialisation error |
| 8 | External I2C bus initialisation error |
| 9 | SPI bus initialisation error |
| 10 | CAN bus initialisation error |
| 11 | ADC initialisation error |

P.S. @kubark42 I know you're interested in this PR.

dustin:
I like this a lot, except for the funny spelling of initialization.

tracernz:
@dustin https://github.com/d-ronin/dRonin/wiki/User-Guide:-Troubleshooting/_history ;)

jihlein:
I would work to remove the panic states from the I2C initialization.  It's confusing to users when they select external mag for instance, and the board stops responding because the external mag is disconnected, not powered, etc., and the I2C bus faults.  If they've done this inadvertently, there is no easy recovery.

PIOS_HAL initialization of I2C ports currently sets the I2C alarm according to the status returned by the I2C init routines.  It's difficult to tell whether an I2C port is internal or external, so I'd treat them both the same way.  This is the way AQ32 and Sparky currently work, I'm (slowly) working on moving other targets in this direction.  I have Colibri and Quanton almost ready for PRs.

Otherwise this is great!

tracernz:
Perhaps only panic for internal I2C then? I can see your point for external I2C and agree.

dustin:
I had to disable the baro on my sparky to make it work because of the
panic.  I know this should be a problem, but it doesn't boot with the code
in place, and it flies OK with it disabled.

tracernz:
You should fix your broken board. I think masking hardware failures is a bad idea in general.

dustin:
I kind of understand that, but at the same time, I have to choose between a
board that doesn't boot because of a feature I don't care much about, and
being able to use it.  This seems to not be terribly uncommon with sparky2.

jihlein:
Is there an easy way to tell internal vs external?  PIOS_HAL treats them the same from an initialization point of view.

If internal i2c fails to init, I would think the sensors connected to it would "panic", so we'd be covered in that situation.

That said, an external mag should not panic, as that would lock the board up too.

DTFUHF:
Somewhat related, is there a way to dynamically see which sensors are
available to the flight controller? I know there is some stuff in the GCS
board plugins. This kind of feature could be used to allow more hardware
flexibility for already existing targets - and simply throw warnings in GCS
if e.g. an althold flight mode is selected when baro is not present.

tracernz:
> If internal i2c fails to init, I would think the sensors connected to it would "panic", so we'd be covered in that situation.

Yes but it'd be less clear what the problem is, and will always misattribute it to the first sensor that tries to init on that bus. Internal/External I2C is up to the board init code to decide e.g. https://github.com/tracernz/dronin/blob/621be312ea23b21d3bc570d8419d605dec1b4b6d/flight/targets/flyingf3/fw/pios_board.c#L190-L200.

> I kind of understand that, but at the same time, I have to choose between a
> board that doesn't boot because of a feature I don't care much about, and
> being able to use it.  This seems to not be terribly uncommon with sparky2.

Which sensors do we ignore and are all the filters stable without different combinations of those sensors?

dustin:
If I don't physically install a baro, will it still fail to boot?

It seems like it's a nice-to-have feature that is unnecessary if I don't
have flight modes that depend on it.  Gyro might be required to boot, but
if there's a problem talking to an optional component, maybe we can just
not do that.

I don't actually know how bad it could get, though.

tracernz:
In any case, way outside the scope of this PR which simply replaces existing panics.

jihlein:
> Internal/External I2C is up to the board init code to decide

We'll have to make a small change to both AQ32 and Sparky then, they use PIOS_HAL to set the I2C ports, whether internal, external, and flex.  All PIOS_HAL can do is set the I2C alarms, which is inconsistent with how we want this to behave.

There's still one remaining issue to work out.  Internal I2C init does not set any alarms, but external/flex int does.  A board with only internal I2C will display the I2C alarm greyed out, where a board with external/flex I2C setup will display the I2C alarm as green (OK).  Different behavior.  Maybe that's okay, or maybe on successful internal I2C init we set I2C alarm to OK.  Topic needs more discussion.

PR #740, opened 2016-03-03
F1: Fix halt, by not double-initing board, fix safe boot

tracernz:
Introduced here https://github.com/d-ronin/dRonin/blame/f72a42d1311b4a7113fc2395a8f5cb0ea5c9cef9/flight/Bootloaders/CopterControl/main.c#L74

Halt now works reliably for me.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/740)

### DIFF
--- a/flight/targets/bl/f1/main.c
+++ b/flight/targets/bl/f1/main.c
@@ -90,7 +90,6 @@ int main() {
 	GO_dfu = (USB_connected == true) || (User_DFU_request == true);
 
 	if (GO_dfu == true) {
-		PIOS_Board_Init();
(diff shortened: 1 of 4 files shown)


tracernz:
Might be worth holding off on merging this for a bit until we can figure out the flash failure the first time you click the button.

tracernz:
First flash failure fixed: https://github.com/mlyle/dronin/pull/3

tracernz:
I fixed safe boot on F1 in the bootloader since the version is already being bumped here, and also in GCS (in mlyle#3) so it works for older bootloader versions too.

PR #742, opened 2016-03-03
Add support for STM32F446

mluessi:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/742)

### DIFF
--- a/flight/PiOS/Common/Libraries/ChibiOS/os/hal/platforms/STM32/stm32.h
+++ b/flight/PiOS/Common/Libraries/ChibiOS/os/hal/platforms/STM32/stm32.h
@@ -65,7 +65,8 @@
 #include "stm32f37x.h"
 
 #elif defined(STM32F401xx)     || defined(STM32F40_41xxx)  ||               \
-      defined(STM32F427_437xx) || defined(STM32F429_439xx)
+	  defined(STM32F427_437xx) || defined(STM32F429_439xx)  ||              \
+	  defined(STM32F446xx)
 #include "stm32f4xx.h"
 
 #elif defined(STM32L1XX_MD)    || defined(STM32L1XX_MDP)   ||               \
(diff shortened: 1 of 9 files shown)


fujin:
Maybe grab https://github.com/fujin/dronin/commit/d372d2ca9ae0645e7b3b576dacd7428dc989bca2 into this?

mluessi:
@fujin I added it slightly differently as it is the same as for other F4s (see second last commit)

fujin:
@mluessi ah that makes more sense. I see it now. Thanks!

mlyle:
How should these things come in?

I have a feeling that RE support isn't going to be complete for tanto.. Should this stuff merge to tanto and add risk?  Should we merge it immediately after?

dustin:
Is there any risk to having half a new target?

I suppose there's no real risk in having branches hanging out waiting to
become targets, especially since the majority of the development is
happening in one place.

fujin:
I don't feel as though it is risky to bring in the support for stm32f446, it looks to all be guarded off with preprocessor directives appropriately. This isn't RE1 specific, either, a user with a 446 Nucleo or someone else developing stm32f446 FC could use it, no?

mluessi:
My idea was to open these PRs now, as they are only indirectly related to RE1 and can be used in other future targets. The changeset in this PR is fairly small, so I don't think it's a big risk for existing targets.

tracernz:
@dustin They touch code used by other targets though. The risk is pretty minimal in this particular case though. Is there any harm in holding them until after Tanto?

mlyle:
@tracernz, from IRC, he says that maintaining these things in his branches is getting to be a pain.  Meh.  Might as well, but the bar for these to get merged is going to increase as we get closer.

tracernz:
None of these are files we modify regularly, if at all, but okay I guess...?

-e- Except for pios_servo

PR #751, opened 2016-03-05
GCS:Welcome: Cleanup QML, use Qt function to launch URLs

tracernz:
Before we called a C++ function. Also, can now click author in Github feed, and it doesn't spam debug console with clicked URLs.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/751)


tracernz:
Also, now opens the webpage in the foreground rather than under GCS.

mlyle:
Looks good.

PR #753, opened 2016-03-05
BU: Invalidate own metadata, reboot to BL when finished

tracernz:
@mlyle is planning to make GCS check FW metadata during boot to see if upgrade is needed, will tie nicely into this.

Tested on Sparky2. Might be wise to check Sparky1, and CC3D.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/753)

### DIFF
--- a/flight/targets/bu/common/main.c
+++ b/flight/targets/bu/common/main.c
@@ -104,6 +104,12 @@ int main(void) {
 	const uint32_t zero = 0;
 	PIOS_FLASH_start_transaction(fw_partition_id);
 	PIOS_FLASH_write_data(fw_partition_id, 0, (uint8_t *)&zero, sizeof(zero));
+	/* 
+	 * Invalidate FW metadata so GCS can know FW is invalid
+	 * We replace the 3rd and 4th letters of the TlFw sentinel with 0
+	 */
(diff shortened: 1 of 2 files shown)


tracernz:
Better idea, just overwrite the part after TlFw/OpFw. That way we can know that our firmware is onboard but it has been intentionally invalidated since all fields are zero. Alternately make use of the 20-char user field at the end, perhaps something unlikely like "BUcOmPlEtEd". Addresses https://github.com/d-ronin/dRonin/pull/755#issuecomment-192816637

mlyle:
We should use the user field to "remember" the githash during upgrade-in-progress.  E.g. take the whole state machine on the GCS side out of the equation and allowing the process to be resumed.

tracernz:
Another good idea.

tracernz:
Made a hash of this the first time, second attempt. Now just replaces the 'w' in "TlFw" at the start of the metadata with \0. This will work quite nicely with string functions in Qt so we can check it and be certain that it's a completed BU. This allows us to maintain bootloader compatibility with other firmwares that don't use our metadata system.

mlyle:
LGTM.  Thanks, sorry STM made it hard on us :P

PR #755, opened 2016-03-05
GCS:Uploader: Detect mismatched firmware version, offer update, auto-load correct firmware, stop at BL when firmware metadata invalid

tracernz:
- On device metadata is updated after flashing (see TauLabs/TauLabs#923).
- Progress bar resets to 0 at start of flash, previously it would keep 100% from previous flash during the erase phase.
- Stop in bootloader if firmware metadata is invalid, works great with #753, part of the way to @mlyle's upgrade work.
- Firmware CRC format is consistent (base 10), previously base 16 was used in one place.
- Find correct firmware file for board
- Bundle OSX firmware in app bundle
- Auto-load firmware file when bootloader connects
- Detect incompatible firmware version and offer update, automatically enters bootloader if required, auto-loads file and highlights Flash button

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/755)

### DIFF
--- a/ground/gcs/src/plugins/uploader/uploader.pro
+++ b/ground/gcs/src/plugins/uploader/uploader.pro
@@ -14,7 +14,9 @@ HEADERS += uploadergadget.h \
     uploader_global.h \
     fileutils.h \
     bl_messages.h \
-    tl_dfu.h
+    tl_dfu.h \
+    ../../../../../build/ground/gcs/gcsversioninfo.h
+
 SOURCES += uploadergadget.cpp \
     uploadergadgetfactory.cpp \
     uploadergadgetwidget.cpp \
--- a/ground/gcs/src/plugins/uploader/uploader.pro
+++ b/ground/gcs/src/plugins/uploader/uploader.pro
@@ -6,6 +6,8 @@ QT += testlib
 include(uploader_dependencies.pri)
 include(../../libs/glc_lib/glc_lib.pri)
 
+CONFIG(release, debug|release):DEFINES += FIRMWARE_RELEASE_CONFIG
+
 INCLUDEPATH *= ../../libs/glc_lib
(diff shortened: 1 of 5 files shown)


tracernz:
Doing some more work on this to detect firmware incompatibility and deal with it.

tracernz:
Well this PR got a whole lot more interesting

tracernz:
Only tested on Linux. Doubtful if finding firmware file works properly on OSX or Windows yet.

tracernz:
Note: currently the offer to auto-update firmware version is always enabled but my intention is to only have it enabled in release config builds when this is tested, it will be too annoying for devs otherwise IMO. Another option might be a setting to disable it.

mlyle:
> Stop in bootloader if firmware metadata is invalid, works great with #753, part of the way to @mlyle's upgrade work.

Is this going to be negative for our compat with CF, etc, on CC3D?  Would hate for them to advise to erase our loader..  Maybe all 1's could be considered OK?

tracernz:
CF doesn't support anything other than erasing our loader. They removed OPBL support some time back (they wanted the extra 12k of flash).

mlyle:
Ah, OK.

tracernz:
It's a good point to raise though. Needs more thought.

tracernz:
~~One thing we could consider is putting an enum in the firmware metadata which says whether it's firmware, bu, or upgrade helper, and for the latter two GCS can see it run once and then it comes back to the bootloader again it knows to halt and do whatever it's up to in the upgrade process. Backwards compatible by assuming firmware when the metadata is invalid or the enum is missing from metadata, or better still, by making firmware option = 0.~~

tracernz:
See https://github.com/d-ronin/dRonin/pull/753#issuecomment-192884949 RE: stop at BL when metadata invalid.

PR #756, opened 2016-03-05
Limit to the 10 most recent tuned boards.

dustin:
This still has a little bit of scrolling, but isn't like, a lot more
than 10.  Previously, it could be up to 500 results depending on how
many were duplicate.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/756)

### DIFF
--- a/ground/gcs/share/welcome/AutoTunePanel.qml
+++ b/ground/gcs/share/welcome/AutoTunePanel.qml
@@ -36,7 +36,7 @@ Item {
 
     JSONListModel {
         id: jsonModel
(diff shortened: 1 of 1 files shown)


tracernz:
Maybe put something like GCS:Welcome: so it's easier to tell what this is in commit list in future ;). Thanks! :lgtm_strong: 

---

Reviewed 1 of 1 files at r1.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/756#-:-KC4xKvVLLhVCuWK4O6M:1705699538)_

PR #757, opened 2016-03-05
One MPU to rule Naze only, for now

### LINKED ISSUE #298: Support the Naze32 rev6
dRonin should support the new Naze32 revision 6 board, which is now the only official Naze32 you can get from many official Abusemark resellers. It doesn't seem like there's a lot of work required to get this working.

The main difference is the use of the MPU-6500 instead of the MPU-6050, and this new chip is already supported by the BrainFPV MPU-9250 i2c driver.

Tracernz has an example hack to get the Naze32 rev6 working at https://github.com/tracernz/TauLabs/tree/naze-rev6-hax

It also uses the BMP280 barometer, for which jihlein has written a driver that can be found on his TauLabs fork at https://github.com/jihlein/TauLabs/tree/myNext/flight/PiOS  - Common/pios_bmp280.c inc/pios_bmp280.h and inc/pios_bmp280_priv.h

tracernz:
Rebased to next (ouch, took a lot of work). Still needs lots of work but putting this up so it's visible. Mag support not yet working, if you can spot anything obvious for 9250, please let me know.

Works on Naze rev 5 & 6, 4 untested.

edit by mpl: fixes #298 

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/757)


tracernz:

flight/PiOS/Common/pios_mpu.c
```diff
@@ -0,0 +1,1090 @@
...
+
+
+/**
+ * WHOAMI ids of each device, must be same length as pios_mpu_type
+ */
+static const uint8_t pios_mpu_whoami[PIOS_MPU_NUM] = {
+	[PIOS_MPU60X0] = 0x68,
+	[PIOS_MPU6500] = 0x70,
+	[PIOS_MPU9150] = 0x68,
+	[PIOS_MPU9250] = 0x71
```
kubark42: Consider a `,` here.

flight/PiOS/inc/pios_mpu.h
```diff
@@ -0,0 +1,192 @@
...
+};
+
+enum pios_mpu60x0_gyro_filter {
+	PIOS_MPU60X0_GYRO_LOWPASS_256_HZ = 0x00,
+	PIOS_MPU60X0_GYRO_LOWPASS_188_HZ = 0x01,
+	PIOS_MPU60X0_GYRO_LOWPASS_98_HZ  = 0x02,
+	PIOS_MPU60X0_GYRO_LOWPASS_42_HZ  = 0x03,
+	PIOS_MPU60X0_GYRO_LOWPASS_20_HZ  = 0x04,
+	PIOS_MPU60X0_GYRO_LOWPASS_10_HZ  = 0x05,
+	PIOS_MPU60X0_GYRO_LOWPASS_5_HZ   = 0x06
```
kubark42, line 68: Consider `,` here.

flight/PiOS/inc/pios_mpu.h
```diff
@@ -0,0 +1,192 @@
...
+};
+
+enum pios_mpu6500_gyro_filter {
+	PIOS_MPU6500_GYRO_LOWPASS_250_HZ = 0x00,
+	PIOS_MPU6500_GYRO_LOWPASS_184_HZ = 0x01,
+	PIOS_MPU6500_GYRO_LOWPASS_92_HZ  = 0x02,
+	PIOS_MPU6500_GYRO_LOWPASS_41_HZ  = 0x03,
+	PIOS_MPU6500_GYRO_LOWPASS_20_HZ  = 0x04,
+	PIOS_MPU6500_GYRO_LOWPASS_10_HZ  = 0x05,
+	PIOS_MPU6500_GYRO_LOWPASS_5_HZ   = 0x06
```
kubark42, line 78: Consider `,` here, etc...

flight/PiOS/Common/pios_mpu.c
```diff
@@ -0,0 +1,1090 @@
+/**
+ ******************************************************************************
+ * @file       pios_mpu.c
```
kubark42, line 3: I wonder if it would be better to rename this, e.g. to "mpuxxxx". MPU for me has meant [other stuff](https://en.wikipedia.org/wiki/MPU), some of which could feasibly be on an autopilot.

DTFUHF, line 3: Or maybe pios_invensense or pios_invensensemotion, as their new lineup has dropped the mpu prefix but is largely compatible.

flight/PiOS/Common/pios_mpu.c
```diff
@@ -0,0 +1,1090 @@
...
+	[PIOS_MPU6500] = 0x70,
+	[PIOS_MPU9150] = 0x68,
+	[PIOS_MPU9250] = 0x71
+};
+
+/**
+ * I2C addresses to probe for device
+ */
+static const uint8_t pios_mpu_i2c_addr[] = {
+	0x68,
```
kubark42, line 74: What do you think about the approach here, https://github.com/TauLabs/TauLabs/pull/2206/files#diff-db1c8b7b78676a2aada18cdf6f5e4428R62, where the enum spells out the magic number?

flight/PiOS/Common/pios_mpu.c
```diff
@@ -0,0 +1,1090 @@
...
+		gyro_data.y *= gyro_scale;
+		gyro_data.z *= gyro_scale;
+		gyro_data.temperature = temperature;
+
+		PIOS_Queue_Send(mpu_dev->accel_queue, &accel_data, 0);
+		PIOS_Queue_Send(mpu_dev->gyro_queue, &gyro_data, 0);
+
+#ifdef PIOS_INCLUDE_MPU_MAG
+		if (mpu_dev->use_mag) {
+			uint8_t st1 = mpu_rec_buf[IDX_MAG_ST1];
```
kubark42: Not sure if you're interested in doing this, but the MPU-9250 supports external magnetometers (really, any I2C device which follows a simple sensor reading scheme) and so it might be good to architect this so as to support arbitrary magnetometers. We use an HMC5883 attached to our MPU-9250 and have had a lot of good results with it. More realistically, I think the MPU-6500 supports the same kind of I2C driver and it doesn't already have the built-in AK8963.

Now, arbitrary **sensors** would be really cool! But I don't see how the driver could be written that generically since it's anyone's guess what might be attached there.

flight/PiOS/Common/pios_mpu.c
```diff
@@ -0,0 +1,1090 @@
...
+	return 0;
+}
+
+static int32_t PIOS_MPU_CheckWhoAmI(enum pios_mpu_type *detected_device)
+{
+	// mask the reserved bit
+	int32_t retval = PIOS_MPU_ReadReg(PIOS_MPU_WHOAMI);
+	if (retval < 0)
+		return -PIOS_MPU_ERROR_READFAILED;
+	for (int i = 0; i < PIOS_MPU_NUM; i++) {
```
kubark42, line 230: Smart approach. Wish the MS5XXX baro sensors would have had a unique ID register. (Maybe it does, there's a 16-bit reserved register in the PROM. But undocumented.)

flight/PiOS/Common/pios_mpu.c
mlyle, line 74: That would prevent indexing this by type.

flight/PiOS/Common/pios_mpu.c
kubark42, line 3: Yeah, great point. We just had Invensense reps stop by this morning and they have a ton of new and interesting sensors beyond the MPU line.

mluessi:
We should probably defer this. Thoughts?

mlyle:
It really hurts to, but that's what's happening by default if it isn't ready super soon.

(Would really, really like to have at least minimal r6 support).

flight/PiOS/Common/pios_mpu.c
tracernz, line 3: Might rename to pios_mpuxxx0. This supports MPU-XXX0 only so I don't want to name it as if it supports all Invensense families, and I want to follow the existing convention of naming using the model rather than the manufacturer. If it is modified later to support additional device families it's a case of renaming and /s/pios_mpuxxx0/pios_shinynewthingstoo/.

flight/PiOS/Common/pios_mpu.c
tracernz, line 230: Sadly MPU-6000/6050/9150 all share the same WHOAMI, but the newer sensors (6500/9250) are unique.

flight/PiOS/Common/pios_mpu.c
tracernz: A good idea for a future extension/PR, and probably a bit of a refactor of the HMC driver to avoid duplicating all that code? At this stage we have no hardware utilizing the MPU I2C bus in this way.

flight/PiOS/inc/pios_mpu.h
```diff
@@ -0,0 +1,192 @@
...
+ * @brief Which type of MPU was detected?
+ */
+enum pios_mpu_type PIOS_MPU_GetType(void);
+
+#endif /* PIOS_MPU_H */
+
+/** 
+  * @}
+  * @}
+  */
```
tracernz, line 182: Needs newline.

tracernz:
We can use this on Naze only to give r6 support without completing the mag work (9250 works, at least on my local branch, 9150 doesn't). Will tidy and prepare for merge this evening after work.

flight/PiOS/inc/pios_mpu.h
tracernz, line 78: Not going to do it for these ones since they're locked into silicon/won't change.

tracernz:
Okay, rebased on next with latest mag work and tidy-up from my local branch included. Some minor review changes. Works for me on Naze rev. 5 and 6. Tried it with autotune on the bench okay, 316 bytes of stack free with PPM.

tracernz:

Please test if you have a Naze, esp. rev 6.

mlyle:
Flew a dRonin brushed naze R5 (quanum pico) test sequence ...setup from scratch, hangtime, attitude, rate, horizon, autotune.  All 100% happy.

flight/PiOS/Common/pios_mpu.c
```diff
@@ -0,0 +1,1118 @@
...
+	}
+
+	return dev;
+}
+
+static int32_t PIOS_MPU_Validate(struct pios_mpu_dev *dev)
+{
+	if (dev == NULL)
+		return -1;
+	if (dev->magic != PIOS_MPU_DEV_MAGIC)
```
mlyle, line 217: i hate these conventions.  we need to end them.  but not for this PR

(trying to tolerate / error-out sanely when.. someone is passing us something that's not the IMU as the IMU?)

flight/PiOS/inc/pios_mpu.h
mlyle, line 182: Yes it does ;)

mlyle:
Just got my R6 and tested / confirmed this-- it is indeed happy.

PR #758, opened 2016-03-05
Stabilization / Actuator Optimization

mluessi:
Some CPU optimizations along the stabilization / actuator path by only fetching UAVOs if they have been updated. This saves some CPU cycles and also reduces UAVO lock contention, which should reduce jiter in actuator outputs etc.

On Brain with IMU running at 1kHz, this reduces the CPU utilization by about 2% (35% to 33%), for RE1 with IMU at 1.6kHz by about 4% (39% to 35%).

Status: Bench tested, seems to be working.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/758)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -76,6 +76,9 @@
 static struct pios_queue *queue;
 static struct pios_thread *taskHandle;
 
+static bool flightStatusUpdated = true;
+static bool manualControlCommandUpdated = true;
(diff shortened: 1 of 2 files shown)


mlyle:
Main thing we need to check is F1 resource consumption.

mlyle:
Seems OK.  Projected to use ~50 bytes more on F1, which per @pug398's comment should be OK.

PR #760, opened 2016-03-06
F1 Upgrader (larger bootloader variant that runs from main firmware slot)

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/760)


mlyle:
Removed python changes from this PR and force pushed

mlyle:
Ready for review.  This is "good enough" for now, though it is likely there will be a couple last changes.

I have test sequences I need to go through with this before merge-- this has only seen a sliver of test.

mlyle:
OK, I am happy with this- took bootloader, upgrader, and bootupdater through some test sequences on CC3D a couple of times.  Would be helpful for it to merge ASAP since it interacts with other changesets.

tracernz:
Works for me too. Looks good.

PR #762, opened 2016-03-06
WIP: F1 boot cleanup

tracernz:
UNTESTED on CC3D and PipXtreme so far. BU and BL untested, could also do with more cleanup.
- Do startup code in C instead of ASM
- Removes a lot of common code
- Init Naze without RTOS task, saves over 700 bytes of heap(!!)

Still more that can be cleaned up I think. Do we want to defer this until after Tanto, or pick up the risk?

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/762)


tracernz:
This does the board init on the IRQ stack but it looks like that's going to be very tight and I don't want to increase that stack size. Will go back to a similar approach to before with a temporary init stack at the end of RAM, and then changing to IRQ stack + extending heap over the init stack, after board init but before modules start up. The downside of that approach is that you need to be really sure that you're not blowing the init stack because it will clobber heap and we won't know about it unless something goes really bad.

mlyle:
Does this save any heap on CC3D?  I am worried if Naze is tight on heap that CC3D is probably tight too (i thought we were making positive progress there but have not really been following it in extreme detail)

tracernz:
No such luck.

pug398:
As of yesterday morning CC3D had ~2600 bytes without AT enabled and ~500 with AT enabled. Checked both times with sbus enabled so there has been some memory freed up.

mlyle:
Seems naze needs the memory now (?) which i'm confused by, because we autotuned before and I thought hte overall trend was to free up ram.

tracernz:
Depends what drivers you're using I guess.

tracernz:
3392 bytes of stack remaining on Naze32 with default settings/no modules.

mlyle:
```
/var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/PiOS/STM32F10x/startup.c:59:17: error: 'INIT_STACK_SIZE' undeclared here (not in a function)
 char   init_stack[INIT_STACK_SIZE] __attribute__((section(".initstack")));
                 ^
/var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/PiOS/STM32F10x/startup.c:60:16: error: 'IRQ_STACK_SIZE' undeclared here (not in a function)
 char   irq_stack[IRQ_STACK_SIZE] __attribute__((section(".irqstack")));
                ^
make[1]: *** [/var/lib/jenkins/jobs/dron/workspace/Nodes/master/build/bl_cc3d/startup.o] Error 1
make: *** [bl_cc3d_bin] Error 2
```

tracernz:
Yeah, needs ifndefs in startup.c, so it can be overridden in BL/BU.

flight/PiOS/STM32F10x/link_STM32103CB_sections.ld
```diff
@@ -107,25 +103,36 @@ SECTIONS
...
-        . = ALIGN(4);
-        _eheap = . ;
+        
+        /* extend the heap up to the bottom of init stack */
+        . = ORIGIN(SRAM) + LENGTH(SRAM) - (ABSOLUTE(_init_stack_top) - ABSOLUTE(_init_stack_end));
+        _eheap = .;
+    } > SRAM
+
+    /*
+        * Place the Init stack at the bottom of SRAM so that an overflow
```
tracernz, line 127: Oops, wrong comment.

flight/PiOS/STM32F10x/startup.c
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Initial startup code.
+ */
+void
+_main(void)
+{
+	// load the stack base for the current stack before we attempt to branch to any function
+	// that might bounds-check the stack
+	asm volatile ("mov r10, %0" : : "r" (&irq_stack[0]) : );
```
tracernz, line 80: @mlyle Do we need the stack limit in R10 these days? http://infocenter.arm.com/help/topic/com.arm.doc.100165_0201_00_en/ric1414056337922.html Looks like unless we specify -fstack-limit-register=r10 this does nothing?

mlyle, line 80: We don't, but there's no reason to not do this here.

(I'm thinking about building stack enforcement using R10 to replace the rtos-enforcey-stuff we took out)

mlyle:
F1, DIAF.

PR #769, opened 2016-03-08
gcs: collapse buttons together, improve workflow in upgrade

mlyle:
In preparation for deeper work.  Speedy review and merge appreciated.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/769)

### DIFF
--- a/ground/gcs/src/plugins/uploader/uploader.ui
+++ b/ground/gcs/src/plugins/uploader/uploader.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>822</width>
-    <height>703</height>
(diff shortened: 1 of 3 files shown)


tracernz:
Is there a good reason for separate no and ignore buttons? I like the idea of ignoring for the rest of the session when the no or ignore (keep whichever you think is best and ditch the other IMO) is clicked. Simplifying the options to just two seems better if we can.

mlyle:
OK.  It's kinda been nice in testing, but you're right that this is an unnecessary option.

mlyle:
(done, but unbuilt)

tracernz:
Tested on CC3D, seems good.

PR #771, opened 2016-03-08
MSP: Send meaningful RSSI values at the edges.

dustin:
I ended up quite confused when I'd set my RSSI upper limit just a bit
below the highest value and ended up at ~101% or so, which was showing
up with a blinking 0% on MWOSD.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/771)


dustin:
Dictated, but not read.

mlyle:
Looks great, thanks :)

PR #772, opened 2016-03-08
python: importer improvements, XML export support

mlyle:
[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/772)


python/dronin/logfs.py
```diff
@@ -3,7 +3,13 @@
 
 from struct import Struct
 
-config_arena_magic = 0x3bb141cf
+known_magics = {
```
mluessi: Where do these come from?

mlyle: board_hw_defs, fs_magic.  No idea why different targets use different magics, haphazardly.

tracernz: Maybe so if you took your SDCard from one board to a different type it would be erased? I suspect SD predates the magic words though...

tracernz:
Works well on S2. :lgtm: . jenkins, hurry up please

---

Reviewed 1 of 3 files at r1, 4 of 4 files at r2.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/772#-:-KCOjUeP0TyGlBUKzCaV:707659508)_

PR #775, opened 2016-03-10
Use the cloud uavo builder for any subprocess failures

dustin:
In this case, there's no subprocess.Popen at all.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/775)


mlyle:
Ah, I see.

PR #776, opened 2016-03-11
gcs: uploader: significant cleanup, work towards upgrade / export config

mlyle:
- Removes "partition bundle" functionality
- Moves lots of code to be procedural/serial instead of connected by signals.
- Remove "incompatible UAVOBJs" message in favor of @tracernz's new mechanism
- Improvements to "f1 upgrader" second stage bootloader
- Add a new button that exports a settings partition and saves an XML dump of the configuration.

This change is [Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/776)


tracernz:
Partition bundle seems like it could be nice when trying to troubleshoot weird issues, we can just ask the user to do one thing and get a full dump of their memory which we can load up and debug on our own hardware...

mlyle:
Rebased.  Need to get the settings-translation button working, then this PR can stand on its own.

(Also consider for this-- 'boot' on F1 upgrader is not meaningful, so maybe teach it to escape back to outer loader).

mlyle:
I think this has reached a point where it should merge before I go further, to keep changeset size sane and to continue to get feedback.

Up next:  further factoring changes, full-on upgrade state machine, additional parameters sent to cloud service.

mlyle:
Also need to change boot-upgrader (UP) detection logic to be smarter.

mluessi:
@mlyle can you summarize what this PR does? I know it can be deducted from the changes, but I think it's good to have a description. Makes it easier to review too, as it makes clear what the programmer was intending to do.

mlyle:
@mluessi Wrote brief subset of the change messages at the top of this; the commit messages are authoritative / a fair good description of what is happening.

It cleans things up, and it adds the ability to export configuration from the loader.  Further work is needed to stitch this into a cohesive upgrade thingy, which I'll begin after this merges.

tracernz:
Looks okay from here. Merging so work can proceed.

PR #780, opened 2016-03-11
GCS:Autotune: Add experimental yaw and outer Ki

tracernz:
This is a little ugly due to the problematic nature of disabling the experimental yaw values when the box is unchecked... :/ It will save the old yaw gain settings when the box is first checked, and these saved values will be overwritten if the tune is applied (this last part is debatable...). When the box is unchecked yaw gains will go back to the saved settings.

@dustin Adds some extra fields to the tune JSON and consequently bumps the data version.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/780)

### DIFF
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -80,6 +80,7 @@ ConfigAutotuneWidget::ConfigAutotuneWidget(ConfigGadgetWidget *parent) :
     connect(m_autotune->rateNoise, SIGNAL(valueChanged(int)), this, SLOT(recomputeStabilization()));
 
     connect(m_autotune->cbUseYaw, SIGNAL(toggled(bool)), this, SLOT(onYawTuneToggled(bool)));
+    connect(m_autotune->cbUseOuterKi, SIGNAL(toggled(bool)), this, SLOT(recomputeStabilization()));
 
     addUAVObject(ModuleSettings::NAME);
 
@@ -369,7 +370,7 @@ void ConfigAutotuneWidget::recomputeStabilization()
     // critically damped;
     const double zeta_o = 1.3;
     const double kp_o = 1 / 4.0 / (zeta_o * zeta_o) / (1/wn);
-    const double ki_o = 0.75 * kp_o / (2 * M_PI * tau * 10.0);
+    const double ki_o = (m_autotune->cbUseOuterKi->isChecked()) ? (0.75 * kp_o / (2 * M_PI * tau * 10.0)) : 0.0;
 
     // For now just run over roll and pitch
     for (int i = 0; i < 3; i++) {
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -79,6 +79,8 @@ ConfigAutotuneWidget::ConfigAutotuneWidget(ConfigGadgetWidget *parent) :
     connect(m_autotune->rateDamp, SIGNAL(valueChanged(int)), this, SLOT(recomputeStabilization()));
     connect(m_autotune->rateNoise, SIGNAL(valueChanged(int)), this, SLOT(recomputeStabilization()));
 
+    connect(m_autotune->cbUseYaw, SIGNAL(toggled(bool)), this, SLOT(onYawTuneToggled(bool)));
+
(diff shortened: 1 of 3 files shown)


tracernz:
screenshot from 2016-03-11 21 08 22

dustin:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

_[ground/gcs/src/plugins/config/configautotunewidget.cpp, line 413 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/780#-KCZiBAK0dUpX3WlHAWs:-KCZiBALcAnncrfuMTVa:62034966) ([raw file](https://github.com/d-ronin/dronin/blob/be0e6fccd4c536f204aa8b7b9fc9be1e5016ff56/ground/gcs/src/plugins/config/configautotunewidget.cpp#L413)):_
Should we limit this to quads? 

Not even sure if it's applicable to any other configurations.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/780)_

tracernz:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

_[ground/gcs/src/plugins/config/configautotunewidget.cpp, line 413 [r1]](https://reviewable.io:443/reviews/d-ronin/dronin/780#-KCZiBAK0dUpX3WlHAWs:-KCZkotEE-GqXAagjBsc:-1782211982) ([raw file](https://github.com/d-ronin/dronin/blob/be0e6fccd4c536f204aa8b7b9fc9be1e5016ff56/ground/gcs/src/plugins/config/configautotunewidget.cpp#L413)):_
good call, done

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/780)_

mlyle:
I don't think it needs to be limited to quads, so I'd put that back :/

Also, ideally, I think we should grey the checkbox is yaw beta is out of range.

dustin:
It worked very poorly on my tri.  Is there any reason to think it'd work
better on someone else's?

I don't think anyone's tested it on hex or octo configurations.  Should it
work there in theory?

mlyle:
Tricopters are a very special case-- especially with cheap slow servos that
respond very differently / more slowly than the escs.  It's likely to work
well on anything that uses similar systems for yaw actuation to the the
other axes.  It has been successfully used on hexcopters.

dustin:
Then perhaps we just exclude it from tricopters (with an explanation)?

mlyle:
I'm not positive it's always useless on tris.  

Similarly autotune is problematic on helicopters but we still try.  Dunno what best practice would be here.  Warn?

dustin:
If we declare it experimental, perhaps it'd be good to write a small bit
about the experimentation.

tracernz:
Quad stuff removed. If you want additional UI strings please post them. :-P

mlyle:
I think we should add a checkbox for Use experimental outer KI, and then I'd be glad to merge

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -367,14 +369,30 @@ void ConfigAutotuneWidget::recomputeStabilization()
...
+
+        switch (i) {
+        case 0: // Roll
+        case 1: // Pitch
+            ki = a * b * wn * wn * tau * tau_d / beta;
+            kp = tau * tau_d * ((a+b)*wn*wn + 2*a*b*damp*wn) / beta - ki*tau_d;
+            kd = (tau * tau_d * (a*b + wn*wn + (a+b)*2*damp*wn) - 1) / beta - kp * tau_d;
+            break;
+        case 2: // Yaw
+            beta = exp(0.6 * (systemIdentData.Beta[SystemIdent::BETA_PITCH] - systemIdentData.Beta[SystemIdent::BETA_YAW]));
```
theothercliff: Did you intend that to be +, not -?

mlyle, line 391: (This is here because of a deleted question, not a comment on the code)

This is supposed to be -;  it basically looks at the gain difference observed between pitch and yaw, and picks a value inbetween to scale it.  It doesn't scale things by the full difference, because actuators are nonlinear and it's better to have yaw be undertuned a little than couple badly into the pitch and roll axes.

fujin:
Tested OK.

tracernz:
Rebased and added checkbox + warning for outer Ki.

PR #787, opened 2016-03-12
Auto-enable modules based on port configuration

tracernz:
Seems to work well. Can't do S.PORT because there are two different modules for that??? I'm unfamiliar with FrSky stuff so don't know why we need two separate modules or how to decide which one to enable if possible, any suggestions welcomed. Also, cleaned up the FrSky config stuff to not need an ugly define added to lots of files.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/787)


mlyle:
One is sport, one is sensor hub-- so there's separate types and separate modules (you could even run them both at once).

tracernz:
On CC3D, code size without:
- Code size: 110168
- Heap remaining: 2532

With:
- Code size: 110144
- Heap remaining: 2532

> One is sport, one is sensor hub-- so there's separate types and separate modules (you could even run them both at once).

No, I've done sensor hub. But there are two S.PORT modules, UAVOFrSKYSPortBridge and UAVOTaranis.

mlyle:
Oh, sorry-- UAVOTaranis is the pipxtreme one for ground-side use.

tracernz:
S.PORT done. One small problem with this, it uses TaskInfo to indicate which modules are enabled on the modules tab, but F1 doesn't have TaskInfo... :/

tracernz:
Found a mistake in the original commit (had altered the S.PORT checkbox rather than MSP). Fixed the mistake, and added a commit to clear the checkboxes upon board connection. This stops the state being retained from the last connected board when the TaskInfo UAVO is not present on the new board (i.e. F1 boards).

tracernz:

Need some test on this.

mlyle:
Passes smoke test.  Need more detailed test on this, but we'll take it together for the release.

PR #794, opened 2016-03-13
GCS:Config: Refactor default board config, add features

tracernz:
Refactor this to remove a lot of unnecessary stuff checked twice.
- Add reboot and reset defaults buttons
- Support arbitrary field types, not just enum
- Fields now expand to full width (could change this if desirable but I kinda like it)
- Removed apply button (it makes no sense on hardware settings that are only applied during boot)
- Improve hiding OSD tab behaviour (it no longer shows when connected to an unknown board, e.g. simulator)

---

This change is [Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/794)


tracernz:
Testing int and float fields with units.
screenshot from 2016-03-13 19 19 16

mlyle:
Pretty

PR #796, opened 2016-03-14
flight/freertos: neuter vTaskDelay when scheduler not running

mlyle:
F1 targets often init hardware and filesystems when the RTOS isn't running.
If they reach a code block that does a PIOS_Thread_Sleep, then they'll
hang forever... e.g. when erasing a fresh flash settings filesystem onto CC3D.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/796)

### DIFF
--- a/flight/PiOS/Common/Libraries/FreeRTOS/Source/tasks.c
+++ b/flight/PiOS/Common/Libraries/FreeRTOS/Source/tasks.c
@@ -723,6 +723,10 @@ tskTCB * pxNewTCB;
 	portTickType xTimeToWake;
 	portBASE_TYPE xAlreadyYielded, xShouldDelay = pdFALSE;
 
+	if (xSchedulerRunning != pdTRUE) {
+		return;
+	}
+
 		configASSERT( pxPreviousWakeTime );
 		configASSERT( ( xTimeIncrement > 0U ) );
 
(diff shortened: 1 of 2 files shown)


mlyle:
Confirmed to make my cc3d where I erased the settings partition to work nicely.

tracernz:
Kind of reluctant to mess with the FreeRTOS source. It's the kind of thing that will get missed when we upgrade it (which really should happen if we're going to keep it). It also makes the function no longer do what it says it does (a delay). I wonder if there's another way, like ifdef PIOS_INCLUDE_FREETOS around the delay calls? That leaves it up to the developer to add else PIOS_Delay_Waitms if the delay is actually necessary.

mlyle:
The delay is necessary when the RTOS is running, to yield when a flash operation is in progress and not busy-wait.

The delay is an infinite loop when the RTOS is not running (during initialization).

Erasure can be needed in either case.

mlyle:
If you can think of a better way to do this --- to busy wait before RTOS start and to yield afterwards-- it'd be great.  It seems like it'd take a lot of screwing around to plumb a proper flag here and do it at the PiOS layer though.

mlyle:
Closed in favor of #797

PR #807, opened 2016-03-16
watchdog changes-- assert if flag not registered

mlyle:
from taulabs/taulabs#2226

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/807)


tracernz:
This looks good, nice work @kubark42.

PR #809, opened 2016-03-16
Watchdog: Change the order of task vs. watchdog initialization

mlyle:
The tasks run immediately after they are created, which means the watchdog
won't have been registered yet before it the UpdateFlag reset function is
called on it. The result was a crash.

This fixes the order of initialization.

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/809)

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -7,7 +7,7 @@
  *
  * @file       actuator.c
  * @author     dRonin, http://dRonin.org/, Copyright (C) 2015-2016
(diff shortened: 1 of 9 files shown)


mlyle:
Both we (and they) were very sloppy here.

mlyle:
Tests well.

tracernz:
Hahaha. Will merge when Jenkins is done.

kubark42:
Yup, should have tested it before pushing the `PIOS_ASSERT()` change. It seemed so logical at the time...

tracernz:
Stuff happens, no harm done.

PR #811, opened 2016-03-18
gcs: autotune: set minimum damping factor to .85

mlyle:
Values under 0.8 almost always produce nonsensical results.

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/811)


mluessi:
Thanks!

PR #813, opened 2016-03-19
Flight: Update AQ32 and Sparky ext mag implementation.

jihlein:
Single change to make AQ32 and Sparky external mag implementation work the same as Colibri, Quanton, and Revolution.  

Both AQ32 and Sparky targets compile without errors.

Tested on Sparky, works fine.  

Same change made to AQ32 but not tested, I can't find my extra I2C cable for that board.

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/813)


tracernz:
Shouldn't this line be done once in pios_board.c before any I2C is init'ed, and then alarms won't be masked when the board has more than one I2C bus? https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/Common/pios_hal.c#L544

jihlein:
But then if you don't have any I2C ports to initialize, the I2C alarm will show okay for a feature not being used.

I coded it this way because I have some boards that use SPI only for the sensors, and there is no I2C bus to init, so the alarm stays in the uninitialized state.  If I select the external mag, the external I2C bus gets initialized, and the alarm reports accordingly.

In the case of internal/external I2C buses, the internal bus will either set the alarm okay or panic, in which case the external bus initialization will never be reached.  This works as long as the internal bus is initialized first.

Wondering if splitting the I2C alarm into internal/external sections might help clean up the logic.

tracernz:
Okay, after you've inited each I2C port.

``` c
if (AlarmsGet(SYSTEMALARMS_ALARM_I2C) == SYSTEMALARMS_ALARM_UNINITIALISED)
    AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_OK);
```

jihlein:
How about this:

``` c
    case HWSHARED_PORTTYPES_I2C:
#if defined(PIOS_INCLUDE_I2C)
        if (AlarmsGet(SYSTEMALARMS_ALARM_I2C) != SYSTEMALARMS_ALARM_CRITICAL)
            AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_OK);

        if (i2c_id && i2c_cfg) {
            if (PIOS_I2C_Init(i2c_id, i2c_cfg)) {
                PIOS_Assert(0);
                AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
            }
            if (PIOS_I2C_CheckClear(*i2c_id) != 0)
                AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
        }
#endif  /* PIOS_INCLUDE_I2C */
        break;
```

Use cases:
1. Internal I2C fails init: System will enter panic state.  No attempt made to init anything else.
2. Internal I2C successful init: I2C alarm will be set to OK.  The if statement in the external I2C init sequence will execute, but have no effect on the system due to the alarm already being set to okay. External I2C init proceeds and sets I2C alarm to critical if there are failures.
3. No Internal I2C to initialize:  I2C alarm will be uninitialized.  The if statement in the external I2C init sequence will execute, setting the I2C alarm to OK.  External I2C init proceeds and sets I2C alarm to critical if there are failures.
4. 1st External I2C successful init: Initialization completes, I2C alarm stays at OK.
5. 1st External I2C fails init: Initialization completes, but I2C alarm set to critical.  A second call to init another external I2C bus will start with the I2C alarm at critical, and the alarm will remain critical whether or not the 2nd external I2C bus successfully initializes.

BTW - I have no idea why I can't get the inserted code block to display correctly.

tracernz:
I don't see a reason to overwrite anything other than unitialised with okay. Warning or Error might be used in future code and only checking that it's not critical before overwriting with okay is just a bug waiting to happen.

> BTW - I have no idea why I can't get the inserted code block to display correctly.

Fixed, edit it and have a look ;).

jihlein:
@tracernz - Thinking about it some more, I misunderstood your original suggestion.  Actually makes perfect sense.  Corrected in latest push.

tracernz:
Looks nice :D.

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2.

---

_[flight/targets/aq32/fw/pios_board.c, line 361 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/813#-KDS0J0S3EmA2oc8Eg-X:-KDS0J0S3EmA2oc8Eg-Y:345219530) ([raw file](https://github.com/d-ronin/dronin/blob/2b68ad225dc8843b9de0374d3492030ab18ab9d3/flight/targets/aq32/fw/pios_board.c#L361)):_
I think it'd be a good idea to guard this with a check too? Never know what kind of re-ordering might happen in future.
`if (AlarmsGet(SYSTEMALARMS_ALARM_I2C) == SYSTEMALARMS_ALARM_UNINITIALISED)`

---

_[flight/targets/sparky/fw/pios_board.c, line 357 [r2]](https://reviewable.io:443/reviews/d-ronin/dronin/813#-KDS0V2jv9LaT2MD1-1n:-KDS0V2ka8Pj0pckGCUp:160841994) ([raw file](https://github.com/d-ronin/dronin/blob/2b68ad225dc8843b9de0374d3492030ab18ab9d3/flight/targets/sparky/fw/pios_board.c#L357)):_
Same as above.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/d-ronin/dronin/813#-:-KDS0Xc03zyWNGL9ee4H:344932402)_

jihlein:
Cleaned up internal I2C alarm handling on AQ32 and Sparky.  Also added Revolution to keep it consistent.

mlyle:
@jihlein --- this looks good to me.  What test has the latest version seen?

jihlein:
The last round of alarm changes have not been tested. Hope to test on AQ32 tonight.

jihlein:
Alarms are working correctly on AQ32 V1 board with either the internal or external HMC5883 selected.

mlyle:
Yay, LGTM, good enough for merge :)

PR #814, opened 2016-03-19
GCS:SetupWizard: Disable unsupported input types

tracernz:
Fixes #422

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/814)

PR #816, opened 2016-03-19
GCS:Sparky2: Set supported receiver types

tracernz:
Oops, missed this board in #814.

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/816)


mluessi:
Thanks.

BTW: What is the reason that HOTT SUMD/SUMH is not supported on Sparky2? Isn't it just a serial protocol and could easily be supported?

tracernz:
It's currently setup on the RX port which can't support it, but could be added on mainport. See https://github.com/d-ronin/dRonin/commit/ceb42bd5b0777bd382832dfb4510a0e3d38fba73. I just raised #817 to fix this and some other anomalies.

tracernz:
And when I look at that commit... shouldn't he have disabled HoTT telemetry, not HSUM?

Issue #610, opened 2016-02-04
Vibration test sucks.


mlyle:
- The way the configuration works in GCS is extremely fidgety; requires restart of GCS to pick up a new number of spectrogram bins / easy to provoke to crash.
- Only can seem to run it with 64 or fewer bins.  With more bins, data never shows up.
- One UAVO instance per output point.
- With that, what's one left with?  64 bins is not much spectral data. 
- Why even do the FFT flight side to begin with?  There's no buffering / accumulation of the spectral energy after FFT.  There's no magical downsampling.  It boxcar's accelerometer data, and then when it has 64 points, dumps 64 points into the FFT and gets bins out.  Might as well just hand the raw accumulated accel data to GCS to do less retarded things with it.

I nominate for removal.

jpbarraca:
From a user perspective:
- GCS support is bad due to the required restart. However that can be easily fixed.
- It's fidgety but works and is rather useful.
- With a Sparky 2.0 I'm also experiencing loss of data with more than 64 bins.

EDIT: unrelated ~~Do we have the bandwidth in the telemetry channel to report vibration data more often (1-5ms)?~~

jpbarraca:
@mlyle Please check PR #822 as an initial attempt at this issue.

PR #822, opened 2016-03-22
Upgraded Vibration Analysis Module

jpbarraca:
An attempt to address #610

Currently the values are sampled at the FC and sent to GCS in chunks of 16 values per axis.
Spectrogram was enhanced in order to process this data and show it. Optionally GCS can calculate an FFT using the FFTReal library.

Improvements:
- Can potentially higher number of bins (1024 tested)
- Memory consumption is reduced at the FC
- Analysis parameters can be modified on runtime without restarting GCS
- Spectrogram can apply the FFT or not (optional)
- Enable the application of further analysis from the RAW accel values
- Reduced load in the FC

Drawbacks:
- Telemetry bandwidth is increased and may lead to a bursty behaviour.
- Adds an external header (FFTReal with Free License)

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/822)

### DIFF
--- a/flight/Modules/VibrationAnalysis/vibrationanalysis.c
+++ b/flight/Modules/VibrationAnalysis/vibrationanalysis.c
@@ -7,7 +7,7 @@
  *
  * @file       vibrationanalysis.c
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2014
- * @brief      Performs an FFT on the accels to estimation vibration
+ * @brief      Gathers data on the accels to estimate vibration
  *
  * @see        The GNU Public License (GPL) Version 3
  *
@@ -33,13 +33,12 @@
  * Output object: @ref VibrationAnalysisOutput
  *
  * This module executes on a timer trigger. When the module is
- * triggered it will update the data of VibrationAnalysiOutput, based on
- * the output of an FFT running on the accelerometer samples. 
+ * triggered it will update the data of VibrationAnalysiOutput,
+ * with the accumulated accelerometer samples. 
  */
 
 #include "openpilot.h"
 #include "physical_constants.h"
-#include "arm_math.h"
 #include "pios_thread.h"
 #include "pios_queue.h"
 
@@ -52,7 +51,7 @@
 // Private constants
 
 #define MAX_QUEUE_SIZE 2
-#define STACK_SIZE_BYTES (200 + 484 + (13*fft_window_size)*0) // The fft memory requirement grows linearly 
+#define STACK_SIZE_BYTES (200 + 484 + (6*window_size)*0) // The memory requirement grows linearly 
 																				  // with window size. The constant is multiplied
 																				  // by 0 in order to reflect the fact that the
 																				  // malloc'ed memory is not taken from the module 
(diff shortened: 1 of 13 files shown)


mlyle:
This looks really good!

It's getting late in the Tanto cycle, so I don't consider this a candidate for merging for the next release.  But this fixes most of my complaints about vibration analysis.

The main thing I see-- I'm not sure we should be converting back and forth from Q15 instead of just keeping floats.  What do you think?

@droidicus -- this now uses a very similar strategy to your new-autotune approach.

jpbarraca:
I left the Q15 conversions because I do not know if native floats can create problems in other boards (soft float?).
If you know it is safe I can clean it.

mlyle:
Yes-- all the data comes in as floats, all of our F3 and F4 parts have a FPU (and F1 doesn't have memory or code space for this).  So while it'll use slightly more memory, I think avoiding the conversion to fixed point is a good idea.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
```diff
@@ -86,8 +84,6 @@ static struct VibrationAnalysis_data {
 	int16_t *accel_buffer_complex_x_q15;
```
mluessi: The data is no longer complex, so the variables should be renamed.

ground/gcs/src/plugins/scope/scopes3d/ffft/DynArray.h
```diff
@@ -0,0 +1,101 @@
...
+        DynArray.h
+        By Laurent de Soras
+
+--- Legal stuff ---
+
+This program is free software. It comes without any warranty, to
+the extent permitted by applicable law. You can redistribute it
+and/or modify it under the terms of the Do What The Fuck You Want
+To Public License, Version 2, as published by Sam Hocevar. See
+http://sam.zoy.org/wtfpl/COPYING for more details.
```
mluessi, line 12: Is this license GPL compatible? If not, we could use [FFTW](http://www.fftw.org/)

jpbarraca:
@mluessi I'm cleaning up the code as we speak. ~~Handling floats directly and~~ removing traces of complex math.

mluessi:
Regarding Q15/float, I wonder if it would be better to quantize the data to 16bits when sending it to the GCS via UAVO. After all, the sensor is 16bits, so we shouldn't lose much and it would reduce the data rate by 50%.

ground/gcs/src/plugins/scope/scopes3d/ffft/DynArray.h
jpbarraca, line 12: Don't know about hidden consequences of this license, but this seems to be very compatible:
...
TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
1. You just DO WHAT THE FUCK YOU WANT TO.

jpbarraca:
@mluessi Given the purpose of this module, that may be acceptable. One drawback of using floats and 16 value chunks is that the UAVObject is rather large and will introduce jitter to other telemetry data.

mlyle:
If we do fixed point 16-bit, it probably shouldn't be Q15; we set up the sensor to be + / - 16G usually, but Q15 can measure only 0-1G; it means Z will spend most of the time clipped and relatively small vibrations can cause clipping in other axes (which in turn will generate odd/even harmonics).

jpbarraca:
Well... there are a few bugs and is not ready for prime time. Will provide a fixed version for the next milestone.

mlyle:
OK.  I appreciate it, whenever you're ready :)

PR #826, opened 2016-03-23
Flight: Add Frsky Sensor Hub + GPS rx configuration.

jihlein:
Here's something I've been playing with on my Naze32Pro F3 boards.  It's FrSKY Sensor Hub and GPS RX on the same uart port.

GPS config needs to be saved to the GPS flash, and the GPS baud rate needs to be 9600.

On an F3 board, I now have FrSKY Sensor Hub telem, GPS, external mag, and MSP running.  With the gyro rate at 500 Hz, I'm at 50% CPU load with a little over 4K of heap remaining.  That's also with vtol path follower, altitude hold, and autotune modules enabled.  My mini tri is now fully NAV capable.

If this is a worthwhile change for the general population, I'll add the mode to the target  files.

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/826)

### DIFF
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -469,20 +469,47 @@ void PIOS_HAL_ConfigurePort(HwSharedPortTypesOptions port_type,
 
 	case HWSHARED_PORTTYPES_FRSKYSENSORHUB:
 #if defined(PIOS_INCLUDE_FRSKY_SENSOR_HUB)
-		usart_port_params.init.USART_BaudRate            = 57600;
+		usart_port_params.init.USART_BaudRate            = 9600;
 		usart_port_params.init.USART_WordLength          = USART_WordLength_8b;
 		usart_port_params.init.USART_Parity              = USART_Parity_No;
 		usart_port_params.init.USART_StopBits            = USART_StopBits_1;
 		usart_port_params.init.USART_HardwareFlowControl = USART_HardwareFlowControl_None;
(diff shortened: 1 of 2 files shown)


mluessi:
This is neat, but I'm a bit worried that it will be hard to use for normal users, especially since the GPS has to be configured using uCenter and the configuration has to be stored in the GPS (some models, like the Neo M8Q, don't even support that). Thoughts?

mlyle:
It's a pain but I've done it on the GPS+Mavlink setting we have.

In general, I like the ability to share unidirectional ports-- it's challenging to have enough (e.g. on my plane-- mavlink for OSD + GPS main, FrSky telemetry flexi, Magnetometer I2C, PPM in RCVR, telemetry RFM22 uses all of my ports/connectivity.

But then the unfortunate thing is we get to a lot of single-purpose code and code duplication for all of the combinations of unidirectional things which is just gonna increase with time.  It'd sure be nice if we could solve this in a generic way.

jihlein:
Yes it's a little hard to setup, but it's neat to have all this functionality out of 2 uart ports. It is sort of a power user thing, but if documented well in the wiki, should be okay.

mlyle:
I think it's my (weak) preference to not merge this in favor of something more general eventually.  I worry about the amount of mode combinations we could end up with things like this, and with how hard it is to support people setting up GPS themselves.

@jihlein are you OK with this closing unmerged?

jihlein:
@mlyle - sure, no problem.  I have it on my branch with the Naze32Pro and thought I'd see if it was useful for the larger user community.

Issue #821, opened 2016-03-21
Known issues with upgrade meta-issue


mlyle:
After #801 lands, known items for cleanup.
- [x] Missing ID complaint during import --- fix for this in #833 
- [x] Failed to connect device: ____ popup https://github.com/d-ronin/dRonin/pull/801#issuecomment-200162563  fixed in #836
- [x] Blank flight controllers need to be handled better - fix for this in #851
- [x] Can't import settings for not-running modules - fix for this in #838
- [x] Completion screen should look better -- fix for this in #837 
- [x] Issues with window height / focus, though these are problematic to address -- fix for this in #837 too
- [x] Verification of availability of cloud service, etc, in the beginning - fix for this in #851
- [x] Better text on first dialog asking about upgrade - fix for this in #851
- [x] Need to not upgrade pipxtreme this way, or bug user in the same way about Naze32-- fixed in #834 
- [x] Add git information for Tau Labs and BrainFPV Tau Labs fork to cloud service to make it a smoother experience for users coming from these platforms. -- done by @dustin

tracernz:
Should it be a modal dialog? I see it sometimes ends up behind the main window (e.g. if an error occurs).

mlyle:
It is setmodal.  There are odd things with modalness and our main window.  (if you have more than one modal thing, Qt acts weird, etc).

mlyle:
P.S.  This is also why most of the errors/subsidary things have been moved into the main dialog.

PR #836, opened 2016-03-25
gcs: connmgr: only pop up error dialog on manual connect

mlyle:
Connects to #821

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/836)


tracernz:
Is there a reason manual connect should be any different? It seems the me that this is just masking an underlying issue which applies to both manual and automatic connections?

mlyle:
@tracernz: It's pretty frequent to get jittery-enumeration and the failure message e.g. I see this 1 out of 5 times pulling a USB cable on my mac laptop, which I assume is from losing and gaining electrical contact.

There's similar things during reset sequences, too, which may be our fault in the firmware somehow, or may be the fault of the individual hardware -- some people tend to get enumerate, disappear, enumerate sequences on Windows.

mlyle:
tested by @droidicus -- not regressed, at least.

mlyle:
@pug398 tested a branch with this and likes the results

mluessi:
Thanks!

PR #838, opened 2016-03-25
flight: initialize settings objects even if module not running

mlyle:
Exceptions: txpid and autotune on F1

This way you can import or export or access settings even when the module isn't running.

Connects to #821

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/838)


mluessi:
LGTM, this will make updating much smoother.

mlyle:
On Revolution, confirmed I was able to import flightbattery, txpid, vtolpathfollower, and camerastab settings.

mluessi:
Thanks!

PR #846, opened 2016-03-28
gcs: set saving TxPIDSettings not mandatory

### LINKED ISSUE #842: Enabling TxPid module in GCS raises a failure to save the txpid settings object
(Not present on HW).

With #838 this is fixed for F3/F4 targets, but it's still relevant to F1-- we need to cope with that object not saving.

mlyle:
fixes #842


mlyle:
Tested and confirmed to work.

Issue #849, opened 2016-03-28
Upgrader cloud-side improvements


mlyle:
- [ ] Should actually migrate objects to new version
- [ ] Should take into account differences in defaults
- [ ] Should add infrastructure to adapt objects, change object names, etc, as part of upgrade
- [ ] Should annotate objects with descriptive data on the migration results
- [ ] GCS should display the descriptive data on import.

Issue #848, opened 2016-03-28
Upgrader improvements for developer branches


mlyle:
- [x] Build system should figure out common ancestor with "upstream" if possible
- [x] This should be stored into firmware descriptors
- [x] In turn, this should be used as "alternates" to migrate to/from.

mlyle:
Closely related: #849

mlyle:
was fixed in autotown, too. so all's good.

PR #850, opened 2016-03-28
upgrade: Fallback to GITHASH_OF_LAST_RESORT

dustin:
This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/850)

### DIFF
--- a/python/dronin/uavo_collection.py
+++ b/python/dronin/uavo_collection.py
@@ -12,6 +12,8 @@ import uavo
 import operator
 import os.path as op
 
+GITHASH_OF_LAST_RESORT = 'Release-20160120.3'
+
 class UAVOCollection(dict):
     def __init__(self):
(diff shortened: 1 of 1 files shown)


mluessi:
Is there a way to automatically grab the tag of the latest release? It will be easy to forget adjusting this line with every release.

mlyle:
See #848 and #849 ... yes, in summary, but it's not easy

mlyle:
Confirmed this to do the right things-- fallback behavior and detection of missing rev.

dustin:
Yes, we can figure out an appropriate tag dynamically.  We may even have
that information loaded, but not in a ready form.

I think the longer term goal here is to have the caller roughly know the
applicable history, so this is mostly a "last resort" for existing versions
that don't have deeper knowledge.

PR #853, opened 2016-03-29
build: windows: bash_profile adjustments

### LINKED ISSUE #657: Windows build environment needs to tolerate other build tools better
This has been a support headache, and I've encountered it myself now that I have anaconda on my Windows development machine.  
- Should prepend to paths in bash_profile
- Should more thoroughly search for tools that may be missing (need to find new things introduced in QT5.5, etc)
- More thought in general

mlyle:
1. prepend paths instead of postpending, allowing the dev tools we
   find to take precedence over other tools that may be in the path.
2. search for qtwebprocess instead of qmake, in case qmake is
   already in the path (e.g. the Anaconda python distribution has a
   qmake bin).

This is not a cure-all but makes things better.  Fixes #657

---

This change is [Reviewable](https://reviewable.io/reviews/d-ronin/dronin/853)

### DIFF
--- a/make/winx86/bash_profile
+++ b/make/winx86/bash_profile
@@ -19,11 +19,11 @@ test_and_add_to_path ()
   fi
 
   if [ -f "$1/$2" ] ; then
-    PATH=$PATH\:$1
+    PATH=$1\:$PATH
   fi
 }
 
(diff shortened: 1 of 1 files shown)


mlyle:
When this merges, I want to adjust the Windows install directions to use Anaconda instead of the pyhton dist.

Issue #817, opened 2016-03-19
Verify Input Configurations and Setup Wizard on All Boards


tracernz:
There are oddly unsupported input types on some boards (e.g. no HoTT on Sparky2, S.BUS non-inverted not marked as supported in most board plugins), and some of the connection diagrams are out-of-date I think.

tracernz:
Let's defer the remainder of this work.

Issue #716, opened 2016-02-29
DSM binding on inverter targets (Sparky2, Quanton, Brain, CC3D, Revo) sucks


mlyle:
- Some targets silently squash binding pulses when DSM is on an inverter port (S2, Revo)
- Some targets drive the external-inverter port and busfight (Quanton, Brain, CC3D?)
- Revo erroneously configures DSM to an inverter port by default.

@tracernz -- does this cover it?

tracernz:
@mlyle Yeah I think so.

tracernz:
The Revo is really a separate issue IMO and is a regression rather than a pre-existing bug.

henn1001:
Yeah sorry about the revo - i didn´t quite get the logic behind the module :/

jihlein:
In the FWIW category, an easy fix is to remove DSM support on ports with hardware inverters.  That's the path I took on the AQ32 target, no DSM on uart3.  Need to review targets in question, don't know if it would create other difficulties.

mlyle:
We could do that, but if you have an already bound satellite it's a little limiting.  It may be the better of choices, as people will get confused about thigns not binding.

OTOH binding works so damn poorly :/

tracernz:
Might be cleanest just to remove DSM on those ports, although not ideal if it happens to be a dedicated Rx port..

mlyle:
Won't be further improving this before samsara; deferring.  Sorry :(

PR #617, opened 2016-02-06
Revo: Flexi-Io Uart Support

henn1001:
Probably breaks everything and has still a lot of errors but will update it here. Just wanted put it out here and draw attention to it :) 
Mostly a copy & paste from the brain because it´s using a similar pinout.

I think a lot of people out there are buying more revos, now that they are available for cheap. So this should be quite welcome. 

Fixes #261

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/617)

PR #858, opened 2016-03-30
gcs: revolution: boardplugin-- put back to how it was

mlyle:
Before #617.

Connects to #817.

Connects to #716.

### DIFF
--- a/ground/gcs/src/plugins/boards_openpilot/revolution.cpp
+++ b/ground/gcs/src/plugins/boards_openpilot/revolution.cpp
@@ -166,13 +166,13 @@ bool Revolution::setInputType(enum InputType type)
         settings.MainPort = HwRevolution::MAINPORT_SBUS;
         break;
     case INPUT_TYPE_DSM:
-        settings.MainPort = HwRevolution::MAINPORT_DSM;
+        settings.FlexiPort = HwRevolution::FLEXIPORT_DSM;
         break;
     case INPUT_TYPE_HOTTSUMD:
-        settings.MainPort = HwRevolution::MAINPORT_HOTTSUMD;
+        settings.FlexiPort = HwRevolution::FLEXIPORT_HOTTSUMD;
         break;
     case INPUT_TYPE_HOTTSUMH:
-        settings.MainPort = HwRevolution::MAINPORT_HOTTSUMH;
+        settings.FlexiPort = HwRevolution::FLEXIPORT_HOTTSUMH;
         break;
     default:
         return false;


tracernz:
getInputType should only check things that are set by setInputType, otherwise the wiring diagram could be wrong. Also need to check that the wiring diagram is actually correct, particularly for S.BUS. It may be a good idea to just separate the CC3D and Revo connection diagrams to avoid icky workarounds if they're different.

mlyle:
@tracernz Are you sure?

If someone has DSM set on a (non-standard) port already, and they pick DSM again, I don't think it should think the existing configuration is not DSM and switch the ports on them (even if the wiring diagram is wrong in that case).

mlyle:
Also if we do it just like that, and someone picks DSM,... in the case where we would have shown a bad wiring diagram before, we'd now configure DSM to an additional port and make it not work at all, which doesn't seem good.

mlyle:
From applyHardwareConfiguration in VehicleConfigurationHelper https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/setupwizard/vehicleconfigurationhelper.cpp#L127

```
    Core::IBoardType::InputType newType = m_configSource->getInputType();
    bool success = boardPlugin->setInputType(newType);
```

It looks like right now we'll happily create duplicate settings in these circumstances, etc.

I think we should suppress setting it if it's the same setting we currently have (though there are some risks there-- we preserve the user's configuration but do not necessarily have the right wiring diagram).

tracernz:
Setup wizard, wiring diagram etc. don't support any configuration other than that set by the setup wizard. There code would all need to be extended to support anything else.

tracernz:
For this release I think we should just fix what's there (make getInputType inverse of setInputType) since that's the way it's always supposed to have been. For next release we should look at overhauling the setup wizard and diagram functionality. I'd like the diagram to be more general and available somewhere other than setup wizard.

mlyle:
@tracernz, if getinputtype is the inverse of setinputtype, I don't think it will handle any situation better.  I may not be understanding.

It seems like this is how it works:
1. If the thing is on the "correct" port-- e.g. if someone picks "DSM", and flexi is already configured to DSM--- the current code and what you want is the same.
2. If the thing is not on any port-- the current code and what you want is the same-- it will configure it and tell the user to reboot.
3. a. If the thing is on the "wrong" port-- if someone picks "DSM", and mainport is configured to "DSM."-- both your proposed and old code will end up configuring flexiport to DSM and leaving the main port as DSM (nonfunctional config).   Current code in this PR will not tell the user to reboot, your proposed code will tell the user to reboot.

b. If I add a line to the above based on original config, it will leave mainport configured to DSM and not touch flexiport, and not tell the user to reboot.  If the user looks at the wiring diagram, it will be wrong for mainport DSM.

c. Alternatively, I could change the code to de-configure RX protocols before setting the configured type.   It would work right, but would silently move mainport to flexiport.

tracernz:
None are particularly appealing but that has been the situation longer than you or I have been involved ;-). I have some ideas to make this a lot better which have been sitting in my head for a while, will discuss with you this evening and maybe get working on it.

tracernz:
Let's just do this for now.

PR #867, opened 2016-04-01
hw: TauLinkModule: remove because it violates license

### LINKED ISSUE #865: TLM hardware should be removed because of license compliance issues
Upstream taulabs/taulabs#1998 seems to not be getting fixed.

mlyle:
Fixes #865 .  Unfortunately, TauLinkModule is mostly a schematic-level
clone of PipXtreme and does not properly attribute C. Moss and is listed
under an incompatible license.

Also adds OpLink Mini schematic, so that we have a useful reference (of what the prevailing board out there actually is.


tracernz:
Do we want to put the license text in the OPLink folder?

mlyle:
Unfortunately, I don't know _exactly_ what the license is.  It's marked CC-BY-NC-SA, and what we're doing complies with any version of it.  But the version number of the license is not marked upstream.

tracernz:
Okay. LGTM

PR #868, opened 2016-04-02
uploader: disable "multiple devices in bootloader state" logic

mlyle:


mlyle:

mlyle:
Worked for GearheadRed2 on IRC.

mlyle:
I think this should probably merge; the benefits of it / the situations it protects against are less severe than the problems it causes.

PR #876, opened 2016-04-03
flight: dsm: change resolution autodetect behavior

mlyle:
In cleanflight/cleanflight#1985, @SpektrumFPV provided different logic
from what we have now for resolution autodetect.  Use that in the most
simple case.  If it works, great, else fall back to our kludgey autodetect.


flight/PiOS/Common/pios_dsm.c
```diff
@@ -182,33 +182,37 @@ enum dsm_resolution PIOS_DSM_DetectResolution(uint8_t *packet)
...
-
-		if (bit_10 && !bit_11)
-			return DSM_10BIT;
-		
-		if (bit_11 && !bit_10)
-			return DSM_11BIT;
-		
-		return DSM_UNKNOWN;		
+	// If transmitter information byte != 0, master satellite
+	if (packet[1] != 0x00) {
```
tracernz: don't even need the if, the default case in the switch will achieve the same.

tracernz:
Also from Spektrum:

> 1. Choosing between 1024 or 2048 resolution shouldn't be an option. It's confusing for people that dont have a good understanding of the receiver bind types and usually requires some research for people to figure out what they should select. The options should be either DSM2 or DSMX remote.
> 2. Based on above option, binding startup pulses will be either 5 for DSM2 or 9 for DSMX. With these two bind types, receivers will provide their bind type and resolution in the first byte of a packet. This detection and autoset have been implemented in the changes I've submitted in the pull request.

We should nail down this bind pulses thing and label the appropriate ones as DSMX and DSM2.

tracernz:
Oh, and 

> Data scaling modified to match intended PWM channel value. This will provide consistency in channel output values between spektrum receivers using serial, PPM, or PWM to send channel data to the flight controller. This should also an increase resolution and will cause an increase in travel (People currently using 150% travel on cleanflight with a spektrum transmitter should reduce to 125%).

mlyle:
I like our current approach of not scaling/normalizing, even though it's somehow confusing to end-users.

The bind pulse thing is problematic because of the free/spurious pulses we get at powerup.

mlyle:
Forcepush'd based on your if observation

tracernz:
> I like our current approach of not scaling/normalizing, even though it's somehow confusing to end-users.

Yeah, ultimately we normalize it on [-1,1] anyway. Probably not useful due to our wizard.

> The bind pulse thing is problematic because of the free/spurious pulses we get at powerup.

Future work to improve this, it's too messy/unfriendly ATM.

SpektrumRC:
@tracernz 
when I said 

> Data scaling modified to match intended PWM channel value. This will provide consistency in channel output values between spektrum receivers using serial, PPM, or PWM to send channel data to the flight controller. This should also an increase resolution and will cause an increase in travel (People currently using 150% travel on cleanflight with a spektrum transmitter should reduce to 125%).

I'd like to point out that the increased resolution depends on how the channel data is stored. Currently, cleanflight just bitshifts 2048 data once to the right to make it 1024. This cuts it in half. The change in the pull request I made scales this data by the appropriate decimal, but I realized it stores it in an int so decimal values just get rounded anyway, so changing travel from 150% to 125% with that change would actually slightly decrease resolution. I'm not sure how your channel data is stored but just thought I'd correct myself on that remark.

tracernz:
@SpektrumFPV Thanks for the note. We just output the raw channel value from the low-level driver. It is then scaled on the interval [-1.0f,1.0f] in a higher level module so very little precision should be lost (scaling is done in single-precision floating point).

mlyle:
Worked for me on Naze32.

PR #877, opened 2016-04-03
CircQueue: Check success of malloc

tracernz:
Maybe will help us fail more gracefully when running out of heap (ideally I want us to be able to alarm/fault-out rather than crashing, but I don't think this PR alone is enough).


tracernz:
Tagged with Tanto for consideration.

PR #882, opened 2016-04-03
GCS:Modules: Don't enable module tabs unless modules running

### LINKED ISSUE #871: GCS: Some module settings tabs always enabled
Since #838. Tagging into Tanto for now.

tracernz:
Fixes #871. Tanto or no? :-P


mlyle:
Are we sure we want it at all?  It's kinda nice you can change the settings without it running (on hardware that supports it).

tracernz:
At the moment it's very inconsistent so only 3 tabs (IIRC) are enabled all the time. If we made it consistent (all the objects present -> all tabs enabled) I'd be okay with that but as-is I think it will only be more confusing.

mlyle:
OK :P

PR #866, opened 2016-04-01
gcs: move to Qt5.6

mlyle:
5.6.x is long term support and we can avoid tracking Qt so closely in
the future.

Fixes #729
Fixes #730

Tested just on OS X so far.  Still WIP.

PR #884, opened 2016-04-03
make $(error) when qt sdk is non existant or out of date

nathantsoi:
not sure if this is the right solution, but the problem is moving from master -> next, qt was upgraded in the make files, so the paths were wrong. hopefully folks wont need to dig through the makefile to figure this out in the future

the problems w/ this is that this line guards all commands except qt_sdk_install when the qt version is out of date: https://github.com/d-ronin/dRonin/compare/next...nathantsoi:make-err-on-qt-err?expand=1#diff-d848f181cd782d5d32ac4749ce706bf1R31

idk if more commands should be allowed, or this should just be a warning?

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -22,7 +22,19 @@ ifdef OPENOCD_FTDI
 endif
 
 # Set up QT toolchain
-QT_SDK_DIR := $(TOOLS_DIR)/Qt5.5.1
+QT_VERSION := 5.5.1
+QT_SDK_DIR := $(TOOLS_DIR)/Qt$(QT_VERSION)
+
+# On any platform except windows
+ifndef WINDOWS
+# Check for a current QT SDK dir, abort without
+ifeq ($(wildcard $(QT_SDK_DIR)),)
+ifneq ($(MAKECMDGOALS), qt_sdk_install)
+	$(error "QT SDK not found, please run `qt_sdk_install`")
+endif
+endif
+endif
+
 
 ifdef LINUX
   ifdef AMD64


mlyle:
@nathantsoi -- thanks for the PR :)

So, a few of pieces of feedback--  first, it's not necessary to nuke old versions of QT from the tool directory.  It's helpful when changing between development branches that depend on different QT versions.

Second, this logic needs to be disabled for Windows, because on most Windows installations a system-level QT install is used-- mostly because of the very poor symbolic link support on Windows for having multiple trees.

Third, I'm not certain we should factor the QT version out of the URLs like that, because often other parts of the URL change (e.g. mingw491->mingw492 going to 5.5) when Qt upstream changes, and it's easier to just paste the URLs in.  It conflicts with a pending PR #866.

Would you consider altering the first bit to be conditional on non-Windows and eliminating the rest?

mlyle:
@dronin-ci P.S. I hate you.

jenkins, test this please

nathantsoi:
something like this https://github.com/d-ronin/dRonin/pull/884/commits/ec539df0753f47f7bb100fdac0cdc0194a10d138 @mlyle?

mlyle:
Looks good, but I think you might still want to remove the QT_SDK_EXISTING_DIR stuff.   I'm cool either way, though.

I'll merge this in a week or two after we ship our new release.

Hope you're having fun flying next :)

nathantsoi:
ahh, right! fixed and force pushed for a cleaner merge

mlyle:
jenkins, build this please

mlyle:
jenkins, test this please

PR #891, opened 2016-04-04
telemetry: remove redundant telemetry updateSettings()

mlyle:
This got introduced by an errant rebase during telemetry refactor.

updateSettings is supposed to happen in the tx task, to avoid blowing watchdog during module initialization when bluetooth is configured.


mlyle:
Tested this out.  Should merge.

PR #892, opened 2016-04-05
Flight: Msp send INS filtered altitude

### LINKED ISSUE #845: MSP: sends baroaltitude directly instead of INS filtered altitude
(Uses baroaltitude instead of positionactual...  which is probably necessary on small targets, but has poor results on larger targets).

Relates to #443 (pressures displayed are altitude MSL instead of altitude relative to home).

henn1001:
Compiles but not tested yet.

Fixes #845


mlyle:
lgtm

PR #893, opened 2016-04-05
New vibrationAnalysis module and Spectrogram with FFT

### LINKED ISSUE #610: Vibration test sucks.

jpbarraca:
New take at #610 

VibrationAnalysis Module
- module will only acquire and send data (no FFT)
- using 16 bits to reduce memory footprint. Floats scaled without clipping (considering +/-16g accel)
- module will reallocate buffers according to the number of samples required or allocates a single buffer (using an ifdef)
- reconfiguration will only take place at the start of an acquisition period so that jitter is minimised
- supports up to 1024 bins and can potentially support much more. Be careful of the bandwidth required.

GCS Spectrogram
- calculates FFT from data (optional according to settings)
- updates window_width dynamically
- scales FFT magnitude so that the values presented are similar to the real G. This is an aesthetical feature.
- applies a Hann window to the data captured

VibrationAnalysisOutpudData object
- is multi instance and will grow as needed (but will not shrink), or can be single instance reducing memory footprint.
- each instance carries 16 values for each axis

edit by mpl: fixes #610

### DIFF
--- a/flight/Modules/VibrationAnalysis/vibrationanalysis.c
+++ b/flight/Modules/VibrationAnalysis/vibrationanalysis.c
@@ -205,7 +205,7 @@ static int32_t VibrationAnalysisStart(void)
         uint16_t existing_instances = VibrationAnalysisOutputGetNumInstances();
         if(existing_instances < instances) {
             //Create missing instances
-            for (int i=existing_instances; i<instances; i++) {
+            for (int i = existing_instances; i < instances; i++) {
                 uint16_t ret = VibrationAnalysisOutputCreateInstance();
                 if (ret == 0) {
                     // This fails when it's a metaobject. Not a very helpful test.
@@ -361,7 +361,7 @@ static void VibrationAnalysisTask(void *parameters)
     {
 
         // Only check settings once every 100ms and not in the middle of a buffer accumulation to reduce artifacts
-        if (PIOS_Thread_Systime() - lastSettingsUpdateTime > SETTINGS_THROTTLING_MS && sample_count == 0 && instance_number==0) {
+        if (PIOS_Thread_Systime() - lastSettingsUpdateTime > SETTINGS_THROTTLING_MS && sample_count == 0 && instance_number == 0) {
             //First check if the analysis is active
             VibrationAnalysisSettingsTestingStatusGet(&runAnalysisFlag);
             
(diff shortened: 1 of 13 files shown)


mlyle:
@jpbarraca Very nice!

We're going to be busy for the next week shipping Tanto but then I look forward to checking this out.

mlyle:
@droidicus this is tangentially related to some things you've been looking at too.

jpbarraca:
@mlyle Sure. No need to rush with this.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
```diff
@@ -83,17 +84,54 @@ static struct VibrationAnalysis_data {
...
 } *vtd;
 
 
 // Private functions
 static void VibrationAnalysisTask(void *parameters);
 
+/*
+*   Releases any memory dinamically allocated
+*   Because this module can have a big footprint this will ensure we can
+*   reduce it when something fails.
```
tracernz, line 109: FYI: PIOS_free is a NOP so this actually doesn't free anything. It's still nice to put the code in though in-case we change that some day.

jpbarraca, line 109: But in this case every time settings change a new buffer is allocated.

The options I see:
- Pre-allocate the maximum size (1024)
- Disable dynamic allocation: will imply a reboot when changing values
- Try to use the stack only: probably will present issues with larger number of sample buffers
- Implement PIOS_free

tracernz, line 109: Allocating maximum size at boot or only allocating current setting at boot are the methods used by other code. Implementing a functional free, and dynamic allocation in general, is problematic.

mlyle, line 109: Another option is to always just allocate buffers of the size that fit in one instances and to update individual instances as we get enough samples.

The only reason why we buffer more than that currently is so we can adjust for magnitude and remove DC bias-- but this could happen on the GCS side instead where memory is cheap.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
```diff
@@ -52,7 +51,15 @@
...
-#define STACK_SIZE_BYTES (200 + 484 + (13*fft_window_size)*0) // The fft memory requirement grows linearly 
+
+/*
+    Stack usage reported by gcc:
+    VibrationAnalysisCleanup     16
+    VibrationAnalysisInitialize  24
+    VibrationAnalysisStart       48
+    VibrationAnalysisTask        192
+*/
+#define STACK_SIZE_BYTES (192 + 48 + 16 + (2*3*window_size)*0) // The memory requirement grows linearly
```
mlyle: I don't think this is the right way to calculate the stack size, because ... not all of these functions run at once.  Also, they call into other library functions that use stack. 

Probably best practice is to pick a small value, like 480, and then when the module is running, ensure in taskstats that at least 100 bytes is free.

I have some code to do static stack-depth analysis but it's pretty challenging to use, because we use function pointers so much.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
jpbarraca, line 109: That could work, but it could also introduce some artifacts. I have this running on the simulator an can compare the two approaches.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
jpbarraca: Of course is not correct... 
I will look into this. Probably first using the code in the simulator target, and then in a FC.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
mlyle, line 109: It should be OK if each of the instances is updated in series (with all of its values filled) and the GCS only processes when the last one shows up.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
mlyle: OK.  Note the simulator doesn't use the declared stack size.

flight/Modules/VibrationAnalysis/vibrationanalysis.c
```diff
@@ -82,6 +83,7 @@ static bool module_enabled = false;
 static struct VibrationAnalysis_data {
 	uint16_t accels_sum_count;
 	uint16_t window_size;
+    uint16_t buffers_size;
```
tracernz, line 86: Use tabs in flight code :)

flight/Modules/VibrationAnalysis/vibrationanalysis.c
```diff
@@ -116,7 +118,8 @@ static void VibrationAnalysisCleanup(void) {
         PIOS_Thread_Delete(taskHandle);
         taskHandle = NULL;
     }
-    
+
+#ifdef PIOS_FREE_IMPLEMENTED
```
tracernz, line 122: This isn't strictly necessary, but it's still okay, you can call free even though it does nothing. ;-)

mlyle:
I've confirmed this works very nicely.

mlyle:
Thanks very much @jpbarraca for this change.

tracernz:
Good stuff @jpbarraca.

PR #896, opened 2016-04-06
gcs: eliminate uavobject to widget relation debug spew

### LINKED ISSUE #802: GCS: UAVObject to widget relation on QLabel tries to write UAVO
Maybe need a readonly field on the UAVO relation that stops it trying to do this. Generates heaps of these when we update input form:
`17:58:33[DEBUG]setFieldFromWidget widget to uavobject relation not implemented for widget:  "inputChannelForm" -> "channelCurrent" of class: QLabel`

mlyle:
fixes #802


tracernz:
Looks good, LGTM.

PR #900, opened 2016-04-07
GCS: Solves Logging crash when existing. Fixes #746

### LINKED ISSUE #746: GCS: Quitting when logging causes a crash
`File closed
QThread: Destroyed while thread is still running
Dump file was written to: /var/folders/kb/g2m5j1f576l3z7lk3ypyq6hr0000gn/T/dr_mpl_connectfaster_20160304_1a3e402c/73C00019-BA51-4767-82D0-12D5869BD092.dmp`

or sometimes it doesn't crash:

`File closed
QThread: Destroyed while thread is still running
QWaitCondition::wakeAll(): mutex lock failure: Invalid argument
QWaitCondition::wakeAll(): cv broadcast failure: Invalid argument
QWaitCondition::wakeAll(): mutex unlock failure: Invalid argument`

jpbarraca:
Let the thread exit and wait for it during shutdown.

edit by mpl: fixes #746 -- github is picky that this has to be in this top comment block

### DIFF
--- a/ground/gcs/src/plugins/logging/loggingplugin.cpp
+++ b/ground/gcs/src/plugins/logging/loggingplugin.cpp
@@ -484,7 +484,7 @@ void LoggingPlugin::extensionsInitialized()
 
 void LoggingPlugin::shutdown()
 {
-    if(state == LOGGING) {
+    if (state == LOGGING) {
         stopLogging();
 
         loggingThread->wait();
--- a/ground/gcs/src/plugins/logging/loggingplugin.cpp
+++ b/ground/gcs/src/plugins/logging/loggingplugin.cpp
@@ -490,7 +490,7 @@ void LoggingPlugin::shutdown()
         loggingThread->wait();
     }
 
-    if (loggingThread == NULL) {
+    if (loggingThread != NULL) {
         delete loggingThread;
         loggingThread = NULL;
     }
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/logging/loggingplugin.cpp
```diff
@@ -485,7 +484,16 @@ void LoggingPlugin::extensionsInitialized()
...
 void LoggingPlugin::shutdown()
 {
-    // Do nothing
+    if(state == LOGGING) {
+        stopLogging();
+
+        loggingThread->wait();
+    }
+
+    if (loggingThread == NULL) {
```
mlyle: This should be !=, right?

mlyle:
Thanks for fixing this :)

ground/gcs/src/plugins/logging/loggingplugin.cpp
tracernz: Should maybe enclose the previous if too?

ground/gcs/src/plugins/logging/loggingplugin.cpp
```diff
@@ -485,7 +484,16 @@ void LoggingPlugin::extensionsInitialized()
 
 void LoggingPlugin::shutdown()
 {
-    // Do nothing
+    if(state == LOGGING) {
```
tracernz: Missing space between if and (.

ground/gcs/src/plugins/logging/loggingplugin.cpp
jpbarraca: damn... I will correct it asap :/

mlyle:
Works great.

PR #907, opened 2016-04-09
GCS:SystemHealth: Tell the user if they select autotune while disarmed

### LINKED ISSUE #906: sysconfig error for autotune mode doesn't give error explanation.
On tanto beta2 (git hash 409c43c8)

Enable autotune module and mode.

Switch into autotune mode while disarmed.  Sysconfig gets a red X as expected. But clicking that gives no details about why system config is in error.  

It does give a nice error now if I disable the module but leave the mode available - so that's nice, I don't remember getting that nice of an error in the past.

tracernz:
Fixes #906 

Strictly speaking this case isn't actually a configuration error... but this is a good minimal fix.

### DIFF
--- a/ground/gcs/src/plugins/systemhealth/html/SystemConfiguration-Error-UnsafeToArm.html
+++ b/ground/gcs/src/plugins/systemhealth/html/SystemConfiguration-Error-UnsafeToArm.html
@@ -1,0 +1,12 @@
+<html>
+  <head>
+    <title></title>
+    <meta content="">
+    <style></style>
+  </head>
+  <body style="color: black">
+    <h1>System Configuration: Error</h1>
+    <h2>Error Code: Unsafe To Arm</h2>
+    <p>The flight board cannot arm in the currently selected flight mode. Arming while in <strong>A]autotune</strong> or autonomous modes is not allowed (you must first takeoff and then switch modes). Please select another flight mode before attempting to arm.</p>
+  </body>
+</html>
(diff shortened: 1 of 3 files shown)


tracernz:
Can we make a "User" field in SystemAlarms to trap things like this, rather than abusing "SystemConfig"?

ground/gcs/src/plugins/systemhealth/html/SystemConfiguration-Error-UnsafeToArm.html
```diff
@@ -0,0 +1,12 @@
+<html>
+  <head>
+    <title></title>
+    <meta content="">
+    <style></style>
+  </head>
+  <body style="color: black">
+    <h1>System Configuration: Error</h1>
+    <h2>Error Code: Unsafe To Arm</h2>
+    <p>The flight board cannot arm in the currently selected flight mode. Arming while in <strong>A]autotune</strong> or autonomous modes is not allowed (you must first takeoff and then switch modes). Please select another flight mode before attempting to arm.</p>
```
mlyle: Is the bracket here a typo?

tracernz: Yep, was supposed to be a backspace.

tracernz: Thanks for spotting. Managed to fix it on my phone :-D.

PR #911, opened 2016-04-11
build: windows: alter bash_profile

mlyle:
based on discussion with @ufoDziner --- interoperability with Anaconda
which installs QT make tools for 4.x.x.

### DIFF
--- a/make/winx86/bash_profile
+++ b/make/winx86/bash_profile
@@ -14,21 +14,21 @@ QT_MINGWVER="mingw492_32"
 
 test_and_add_to_path ()
 {
-  if which $2 >/dev/null 2>&1 ; then
-    return
-  fi
+#  if which $2 >/dev/null 2>&1 ; then
+#    return
+#  fi
 
   if [ -f "$1/$2" ] ; then
     PATH=$1\:$PATH
   fi
 }
 
-test_and_add_to_path "$QT_BASEDIR/$QT_VERSION/$QT_MINGWVER/bin" "QtWebProcess.exe"
-test_and_add_to_path "$QT_BASEDIR/Tools/$QT_MINGWVER/bin" "gcc.exe"
 test_and_add_to_path "/C/Python27" "python.exe"
 test_and_add_to_path "/C/OpenOCD/0.4.0/bin/" "openocd.exe"
 test_and_add_to_path "/C/Program Files/NSIS/Unicode" "makensis.exe"
 test_and_add_to_path "/C/Program Files (x86)/NSIS/Unicode" "makensis.exe"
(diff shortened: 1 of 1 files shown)


make/winx86/bash_profile
```diff
@@ -14,21 +14,21 @@ QT_MINGWVER="mingw492_32"
...
 }
 
-test_and_add_to_path "$QT_BASEDIR/$QT_VERSION/$QT_MINGWVER/bin" "QtWebProcess.exe"
-test_and_add_to_path "$QT_BASEDIR/Tools/$QT_MINGWVER/bin" "gcc.exe"
 test_and_add_to_path "/C/Python27" "python.exe"
 test_and_add_to_path "/C/OpenOCD/0.4.0/bin/" "openocd.exe"
 test_and_add_to_path "/C/Program Files/NSIS/Unicode" "makensis.exe"
 test_and_add_to_path "/C/Program Files (x86)/NSIS/Unicode" "makensis.exe"
+test_and_add_to_path "$QT_BASEDIR/$QT_VERSION/$QT_MINGWVER/bin" "QtWebProcess.exe"
+test_and_add_to_path "$QT_BASEDIR/Tools/$QT_MINGWVER/bin" "gcc.exe"
```
tracernz, line 35: What does re-ordering these do?

-e- To clarify, I'm fine with it staying in the PR regardless. Just curious if there is something subtle going on here.

make/winx86/bash_profile
```diff
@@ -14,21 +14,21 @@ QT_MINGWVER="mingw492_32"
 
 test_and_add_to_path ()
 {
-  if which $2 >/dev/null 2>&1 ; then
-    return
-  fi
+#  if which $2 >/dev/null 2>&1 ; then
+#    return
+#  fi
```
tracernz, line 23: Delete rather than commenting IMO, or alternately add a comment explaining why they're commented out.

make/winx86/bash_profile
mlyle, line 35: The paths prepend.  We definitely want to have the true QT stuff at the beginning of path... because sometimes stuff installs its own qmake etc :P

make/winx86/bash_profile
mlyle, line 23: added clarifying comment

make/winx86/bash_profile
tracernz, line 35: Yeah, but none of the other things we prepend should have the Qt stuff. It's okay anyway just to be extra sure.

mlyle, line 35: But they do-- the reason why this was opened is because Anaconda (python distribution) includes qmake 4.xx tools.

tracernz, line 35: Right, that's what I was asking  earlier :-P. I thought it must have just added them to PATH ;-).

PR #918, opened 2016-04-12
flight: don't busfight on DSM binding

### LINKED ISSUE #716: DSM binding on inverter targets (Sparky2, Quanton, Brain, CC3D, Revo) sucks

mlyle:
When there's an inverter/xor present, it's essential that we not drive
its output.  Fixes #716, makes Quanton, Brain, CC3D consistent with S2
and revo.

### DIFF
--- a/flight/targets/brain/fw/pios_board.c
+++ b/flight/targets/brain/fw/pios_board.c
@@ -385,7 +385,7 @@ void PIOS_Board_Init(void) {
 			NULL,                                // pwm_cfg
 			PIOS_LED_ALARM,                      // led_id
 			&pios_mainport_dsm_aux_cfg,          // dsm_cfg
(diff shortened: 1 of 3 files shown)


tracernz:
Also needed on Naze rev 6 (assuming it supports binding on that port).

mlyle:
Yes, once implemented there it should follow the pattern.  (Specifically, how Sparky2 does it with revision testing, I think).

tracernz:
The problem already exists whether we actively drive the XOR input or not right?

tracernz:
Fixed in #920.

Issue #924, opened 2016-04-13
Quanton: support USART6 on rcvrport


mlyle:
More serial is always good, right?

Unfortunately the CLK line for it is used for I2C; else we could have used it to get a pseudo-SPI out.

jihlein:
PWM IN2 would function as USART6 TX, PWM IN3 would function as USART6 RX.

I would suggest adding 3 more selections to the RCVR port selection:

PPM+Serial
PPM+Serial+ADC
Serial

If desired, serial could also be added with the PWM output function, but the permutations of pin use start to get a little silly in my opinion.

USART6 port selection would be the typical cast of characters.

mlyle:
I agree re: permutations.

Any strong reason to not have PPM+Serial+ADC only?

jihlein:
Not really.  I already have the three selections mentioned above ready for testing, but can easily pare them down if that's the desired path.

mlyle:
Awesome :)  Whatever you think is best.  Thanks for just fixing everything always ;)

tracernz:
Done.

PR #926, opened 2016-04-13
Flight: Add Usart6 to Quanton

jihlein:
Addition of Usart6 to Quanton.  New rcvrport modes:

PPM+Serial
PPM+Serial+ADC
Serial

Usart6 TX is PWM In2, Usart6 RX is PWM In3.

Setup DSM satellite on Usart6 RX, PPM+Serial selected.  Was able to successfully complete the input wizard.  Satellite functioned as expected.  Changed to PPM+Serial+ADC, and satellite still worked as expected.  Same with Serial selected.  Selecting PPM caused satellite to become unresponsive.

This is the extent of the testing I can do with the hardware I have readily available.  If someone could test PPM and ADC function, that would be a more complete test.

Addresses #924.

### DIFF
--- a/flight/targets/quanton/board-info/board_hw_defs.c
+++ b/flight/targets/quanton/board-info/board_hw_defs.c
@@ -799,6 +799,19 @@ static const struct pios_dsm_cfg pios_usart5_dsm_aux_cfg = {
 	},
 };
 
+static const struct pios_dsm_cfg pios_rcvrserial_dsm_aux_cfg = {
+	.bind = {
+		.gpio = GPIOC,
+		.init = {
+			.GPIO_Pin   = GPIO_Pin_7,
+			.GPIO_Speed = GPIO_Speed_2MHz,
+			.GPIO_Mode  = GPIO_Mode_OUT,
+			.GPIO_OType = GPIO_OType_PP,
+			.GPIO_PuPd  = GPIO_PuPd_NOPULL
+		},
+	},
+};
+
 #endif	/* PIOS_INCLUDE_DSM */
 
 #if defined(PIOS_INCLUDE_SBUS)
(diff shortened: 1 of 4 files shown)


mlyle:
Awesome :)

Just wanted to confirm also-- you don't see any way to get a SPI or a USART's clock out of Quanton, do you?

mlyle:
Diff looks good to me, need to test.

shared/uavobjectdefinition/hwquanton.xml
```diff
@@ -159,4 +186,4 @@
 		<telemetryflight acked="true" updatemode="onchange" period="0"/>
 		<logging updatemode="manual" period="0"/>
 	</object>
-</xml>
+</xml>
```
tracernz: Don't remove newline.

flight/targets/quanton/fw/pios_board.c
```diff
@@ -775,4 +802,4 @@ void PIOS_Board_Init(void) {
 
 /**
  * @}
- */
+ */
```
tracernz: Don't remove newline.

flight/targets/quanton/board-info/board_hw_defs.c
```diff
@@ -2027,4 +2075,4 @@ void PIOS_ADC_DMA_irq_handler(void)
 /**
  * @}
  * @}
- */
+ */
```
tracernz: Don't remove newline.

tracernz:
What are we doing with regard to GCS/Setup Wizard?

tracernz:
P.S. An explanation for linebreaks at the end of files: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline :)

-e- Also http://stackoverflow.com/a/72409

mlyle:
@tracernz -- is initially nothing OK?  It's kinda a specialty configuration for people who really need more USARTs (there's a mailing list discussion pointing out if you use mag, GPS telemetry, DSM, and openlog you run out of USARTs)

jihlein:
Guess my editor is removing the newline at the end of the file.  Going to check default settings.

tracernz:
@mlyle, I guess so, but the RcvrPort name is likely to be confusing?

tracernz:
I'd almost be tempted to follow the rest of the Quanton ports and call it UART6 rather than RcvrSerial and in the options in the RcvrPort replace Serial with UART6. We could make a proper board config form that disables the UART6 box when the appropriate RcvrPort options aren't selected.

mlyle:
RcvrPortSerial or RcvrSerial would be in line with what we've done for other targets; though admittedly Quanton is weird / doesn't make it look so much like a port.

tracernz:
Yeah but totally out of line for this one so it's likely to cause a lot of confusion for Quanton users.

tracernz:
Another option if we're going to change the UAVO, is to go with what's marked on the silkscreen and call it InPort/InSerial rather than RcvrPort/RcvrSerial. That's what's done with the UARTs. I think that could be the least confusing option.

mlyle:
That sounds best.  @jihlein, what do you think?  InPort / InPortSerial?  

(Though this will cause the setting to go back to the default when upgrading, but that's OK for now, relatively few Quanton users..)

jihlein:
IMHO, uart6 on the Quanton looks exactly like uart1 on the Naze32, and the two targets should treat them the same.  We have a port, that when used in it's intended application, is for receiving 8 PWM inputs from a receiver.  It's a rcvrport.  It has an option to use a PPM input, leaving the remaining 7 rcvrport pins free for other uses.  So at the moment, uart6 was implemented the same way on Quanton as uart1 is on Naze32, from both a fc target and GCS perspective.  This way of thinking ignores the silkscreens on the target boards, and I'm okay with that, as we are using the pins in ways not originally envisioned by the hardware designers.

That said, it's only a name, and if InPort / InportSerial is desired, I'm cool with that.  We might want to revisit the Naze32 uart1 implementation to keep things similar, particularly from a GCS point of view.  But that's the subject of another PR.......

Changes like this need to be documented in the wiki, because regardless of if it's called RcvrPort / InPort / InPortSerial, the fact that uart6 TX is PWM IN2 and uart6 RX PWM IN3 needs to made clear to the user.

jihlein:
@mlyle - Forgot this - USART6_CK is PC8, which is PWM IN4.

And if you really want to mess around with the pin definitions: 
SPI3_MISO / PB4   / PWM 1,
SPI3_MOSI / PB5   / PWM 2, 
SPI3_SCK   / PB3   / PWM IN6, 
SPI3_NSS   / PA15 / PWM IN5

USART6 is the easiest path. :)

mlyle:
Ahhhh.  I missed that USART6_CK alternate.  That's very nice.

mlyle:
I think I like @tracernz's suggestion for inport/inportserial.

tracernz:
Yep, I'm also in favour on InPort, InPortSerial. We should stick to silkscreen markings in general to make things easier for users.

jihlein:
rcvrport renamed to inport, rcvrportserial renamed to inport serial as discussed.  Tested as above with DSM satellite.  Still works as expected.

Rebased and pushed.

mlyle:
Thanks so much!  Exciting stuff.

PR #932, opened 2016-04-19
flight: Validated UAVO initialization in Modules. Fixes #879

### LINKED ISSUE #879: Modules Call UAVO Functions Without Checking UAVO Init Suceeded
e.g. https://github.com/d-ronin/dRonin/blob/next/flight/Modules/Attitude/attitude.c#L233

One more reason why we crash rather than failing with an alarm when heap fills up.

jpbarraca:
Added checks to all UAVO initialisations in all modules.
Also added similar checks to some mallocs

When the error is detected (HEAP is full and memory cannot be allocated), the module will not execute and will stay disabled.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -127,24 +127,35 @@ int32_t ActuatorStart()
 int32_t ActuatorInitialize()
 {
 	// Register for notification of changes to ActuatorSettings
-	ActuatorSettingsInitialize();
+	if (ActuatorSettingsInitialize()  == -1) {
+		return -1;
+	}
 	ActuatorSettingsConnectCallbackCtx(UAVObjCbSetFlag, &settings_updated);
 
 	// Register for notification of changes to MixerSettings
-	MixerSettingsInitialize();
+	if (MixerSettingsInitialize()  == -1) {
+		return -1;
+	}
 	MixerSettingsConnectCallbackCtx(UAVObjCbSetFlag, &settings_updated);
 
 	// Listen for ActuatorDesired updates (Primary input to this module)
-	ActuatorDesiredInitialize();
+	if (ActuatorDesiredInitialize()  == -1) {
+		return -1;
+	}
+
 	queue = PIOS_Queue_Create(MAX_QUEUE_SIZE, sizeof(UAVObjEvent));
 	ActuatorDesiredConnectQueue(queue);
 
 	// Primary output of this module
-	ActuatorCommandInitialize();
+	if (ActuatorCommandInitialize() == -1) {
+		return -1;
+	}
 
 #if defined(MIXERSTATUS_DIAGNOSTICS)
 	// UAVO only used for inspecting the internal status of the mixer during debug
-	MixerStatusInitialize();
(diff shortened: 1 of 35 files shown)


mlyle:
I'm a little concerned about this PR.  The "always running" modules will still be started (since MODULE_INITIALISE_ALL is not really capable of checking error codes) with portions of their memory uninitialized if they get an allocation failure.  I'm also concerned it will not help, because these allocations are not likely to fail (it's probably task stacks that make us run out of memory, and all the Start functions are called after the Init ones).

jpbarraca:
@mlyle Should the "always running modules" be initialised and started before the others? Or should we trigger an alarm when some memory allocation failed? If we had a dependency chart for all modules they could be started one by one starting with the more vital ones.

This PR will not solve all memory allocation issues, just add some protection mechanisms to when an allocation fails.

mlyle:
@jpbarraca I think it would be difficult to make the always-running-modules conditional.

My concerns with this boil down to two things:
1. All the initialize routines are called; and then all the start ones are run.  However, if any of the allocation calls failed, there will be uninitialized queues in some of the tasks that still run.  This means that the code doesn't help, and will cause an assertion failure in this case.
2. I don't think these allocations will fail.  UAVOs are small compared to task stacks, and it is the Start functions that allocate the task stacks.  So we probably don't run out of memory until the end.

tracernz:
If the UAVO inits fail, the queues (which come later) will also fail.

jpbarraca:
@tracernz What about initialising and starting one module at a time, avoiding starting a task if a module failed the init? I have a sparky2 that seems to be running fine with this [approach](https://github.com/jpbarraca/dRonin/commit/1a8ccd4022a30938f1fe71eb0cdeefb8e701b083)  (it boots and modules start as previously but cannot do a test flight right now).

tracernz:
I don't know if that's a good idea. Modules could rely on another module being init'ed before starting.

tracernz:
Possibly could start core modules (i.e. sensors, stabilization, attitude etc.) first, basically the ones with watchdog to ensure it doesn't cause a reboot when they fail to start.

mlyle:
@tracernz I think that's the compromise we want.

I think I've come around to your argument about this being no worse from an undefined behavior standpoint.  What do you think about merging, and then we can create a new macro/subsection for "always-on modules" and have linker scripts put them first.

tracernz:
Okay, so "Core" modules, and "Optional" modules. That separation is already there in the build infrastructure, just not in the init/start code so it's a good idea I think.

jpbarraca:
@mlyle I will extend the change to all targets besides Sparky2 and ~~update~~ propose another PR

mlyle:
@jpbarraca -- i'm not sure I understand.  We were talking about merging and then changing the way the startup code does module initialization.

jpbarraca:
@mlyle Sorry, my mistake. Go ahead with the merge.

mlyle:
@jpbarraca thanks much for your contributions as always.  I'll write the module-startup-order stuff if it's OK-- I think I have a neat trick in mind to do it with minimal code change.

jpbarraca:
@mlyle Sure.

PR #941, opened 2016-04-26
flight: f3/f4 fpu: enable flush to zero mode

### LINKED ISSUE #931: Should consider using flush-to-zero mode in FPU
Currently we don't use flush to zero mode, which means we can get into weird modes handling subnormals that may be very slow / performance may not be deterministic.

The smallest normal value is about 1e-38.  IEEE compliant mode represents numbers smaller than this as subnormals, which uses software (very slow) handling by default to handle them with reduced precision.  Flush-to-zero mode instead replaces these numbers by 0 when they come up.

I don't think there's any quantities that we expect to handle that are smaller than 1e-38.  But various variables may incidentally end up there, and result in very slow FP handling.

mlyle:
Fixes #931 

Required for deterministic handling of very small numbers / underflow.

### DIFF
--- a/flight/PiOS/STM32F30x/startup.c
+++ b/flight/PiOS/STM32F30x/startup.c
@@ -76,6 +76,8 @@ _main(void)
 
 	/* configure default FPU state */
 	FPU->FPDSCR |= FPU_FPDSCR_DN_Msk;	/* enable Default NaN */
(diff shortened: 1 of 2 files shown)


tracernz:
LGTM

mlyle:
Flown on Sparky2, autotune, leveling, acro.

Verified the INS initialized, but did not fly a navigation flight.

PR #944, opened 2016-04-26
flight: avoid getting uavos unnecessarily

### LINKED ISSUE #471: Request for comment: change code pattern
Very frequently with UAVOs, we do something like:

StabilizationDesiredGet(&stabDesired);
stabDesired.Throttle=...
stabDesired.Whatever=...
StabilizationDesiredSet(&stabDesired);

I think if we intend to set all fields, we should default or bracket initialize things instead of the initial get.  This way we're not tainted by past states, we don't contend the giant uavo lock, and we don't copy data as much.

Thoughts?

mlyle:
Largely fixes #471.

Avoids copying lots of data and removes some pressure on UAVO lock.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -831,14 +831,12 @@ static int32_t setNavigationRaw()
 		getNED(&gpsPosition, NED);
 
 		NEDPositionData nedPosition;
-		NEDPositionGet(&nedPosition);
 		nedPosition.North = NED[0];
 		nedPosition.East = NED[1];
 		nedPosition.Down = NED[2];
 		NEDPositionSet(&nedPosition);
 
 		PositionActualData positionActual;
-		PositionActualGet(&positionActual);
 		positionActual.North = NED[0];
 		positionActual.East = NED[1];
 		positionActual.Down = cfvert.position_z;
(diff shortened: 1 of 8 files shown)


mlyle:
Flight tested, including autotune, on Sparky2.

Flight tested, including autotune, on Naze32.

flight/Modules/Attitude/attitude.c
```diff
@@ -1410,7 +1406,6 @@ static void updateNedAccel()
 	accel_ned[2] += GRAVITY;
 	
 	NedAccelData accelData;
-	NedAccelGet(&accelData);
```
tracernz, line 1413: This one is set by GroundPathFollower as well, https://github.com/d-ronin/dRonin/blob/080150a5b271fcfb3fe36aa1e20f92fc0d956aca/flight/Modules/GroundPathFollower/groundpathfollower.c#L484

flight/Modules/ManualControl/failsafe_control.c
```diff
@@ -78,7 +78,6 @@ int32_t failsafe_control_select(bool reset_controller)
 	SystemSettingsAirframeTypeGet(&airframe_type);
 
 	StabilizationDesiredData stabilization_desired;
-	StabilizationDesiredGet(&stabilization_desired);
```
tracernz, line 81: This one is set in various places.

flight/Modules/Attitude/attitude.c
mlyle, line 1413: Yes, but we fully set all the values.

tracernz, line 1413: Yeah, I'm thinking of the UAVO fields changing in future though, it'll be quite fragile.

tracernz:
Discussed on IRC. Setting UAVOs in multiple places like this is problematic in general and StabDesired is already in plan for future work. As long as we are writing all the fields what is done here will be okay/better than it was before.

Issue #664, opened 2016-02-14
GCS:Libs: Upgrade to Eigen 3.3


tracernz:
When released (currently at beta http://eigen.tuxfamily.org). Previous versions use things that are deprecated in C++11, and cause compilation to fail with GCC 5.3 for @fujin.

fujin:
confirmed

jpahka:
Because of this in order to build current next, I need to add

QMAKE_CXXFLAGS += -Wno-error=deprecated-declarations

to the end of file ...ground/gcs/gcs.pri as per icee's suggestion.

mlyle:
Related: TauLabs/TauLabs#2240

Hopefully Eigen 3.3 will ship soon.

fujin:
you can cherry pick https://github.com/fujin/dronin/commit/34a79bff57af77b915dbea14242cafb5e940831dif it helps

tracernz:
Hopefully not far away https://listengine.tuxfamily.org/lists.tuxfamily.org/eigen/2016/04/msg00017.html

mlyle:
Eigen 3.3 beta-2 was released on July 26, 2016, so maybe there's some hope...

mlyle:
Hopefully soon.

mlyle:
Eigen 3.3 expected to ship next week.

tracernz:
[Wowowowow](http://eigen.tuxfamily.org/index.php?title=News:Eigen_3.3_released!) :open_mouth:

PR #945, opened 2016-04-27
GCS:Config: Work around Eigen bug with GCC5

tracernz:
Workaround for #664


mlyle:
Sigh.  Sucks to merge this when the next eigen may be imminent... but then again it may not.

PR #946, opened 2016-04-27
Flight: Fix F3 version of pios_internal_adc.c

jihlein:
I've been working on implementing a servo position feedback using an ADC channel.  I'm reading on ADC channel from the actuator task.  I discovered that I could get stable readings from F4 targets, but F3 targets would not read reliably.

After studying pios_internal_adc.c for F3 targets, I noticed that the ADC clock is set (clock divisor set to 16), then both the master and slave (if applicable) ADCs are calibrated.  So far so good.  Then the clock divisor is changed to 256.  I don't think the ADC clock should be changed after calibration is done.  

After removing the code code that changes the divisor after calibration, the F3 targets now return stable readings.

I'm not sure what the reasoning behind the change from divide by 16 to divide by 256, but that seems to be the cause of the issue I was seeing.

Tested on Sparky 1, on both ADC0 and ADC1.

### DIFF
--- a/flight/PiOS/STM32F30x/pios_internal_adc.c
+++ b/flight/PiOS/STM32F30x/pios_internal_adc.c
@@ -232,11 +232,6 @@ static void PIOS_INTERNAL_ADC_Converter_Config(uint32_t internal_adc_id)
 			;
 	}
 
-	if (adc_dev->cfg->adc_dev_master == ADC1 || adc_dev->cfg->adc_dev_master == ADC2)
-		RCC_ADCCLKConfig(RCC_ADC12PLLCLK_Div256);
-	else
-		RCC_ADCCLKConfig(RCC_ADC34PLLCLK_Div256);
-
 	/* Do common ADC init */
 	ADC_CommonInitTypeDef ADC_CommonInitStructure;
 	ADC_CommonStructInit(&ADC_CommonInitStructure);
--- a/flight/PiOS/STM32F30x/pios_internal_adc.c
+++ b/flight/PiOS/STM32F30x/pios_internal_adc.c
@@ -212,9 +212,9 @@ static void PIOS_INTERNAL_ADC_Converter_Config(uint32_t internal_adc_id)
 		ADC_DeInit(adc_dev->cfg->adc_dev_slave);
 
 	if (adc_dev->cfg->adc_dev_master == ADC1 || adc_dev->cfg->adc_dev_master == ADC2 )
-		RCC_ADCCLKConfig(RCC_ADC12PLLCLK_Div16);
+		RCC_ADCCLKConfig(RCC_ADC12PLLCLK_Div32);
 	else
-		RCC_ADCCLKConfig(RCC_ADC34PLLCLK_Div16);
+		RCC_ADCCLKConfig(RCC_ADC34PLLCLK_Div32);
 
 	ADC_VoltageRegulatorCmd(adc_dev->cfg->adc_dev_master, ENABLE);
 	PIOS_DELAY_WaituS(10);


mlyle:
Very nice troubleshooting.  Have you confirmed that the CPU usage remains sane (since I assume we're getting 16x as many conversions)?

(Note that I have not inspected our ADC code at all)

tracernz:
Should probably change the Div16 to Div256 to keep the final sample rate the same as before.

jihlein:
I'm looking into why, but setting to Div256 causes the target to enter a bootloop.  Div64 works.  I'll get some throughput numbers a little later.

jihlein:
I think Div256 takes to long to calibrate, and it resets the watchdog timer before it completes.  This is true even with a watchdog reset right before the ADC initialization.  I'm going to review the reference manual to see if I can confirm this theory, as I have no way to measure it.

Here's some throughput numbers:

Div256    Won't Boot
Div128    45%, Conversions are very noisy
Div64      45%, Conversions are noisy
Div32      47%, Conversions are clean
Div16      51%, Conversions are clean

Test setup is simply 2 10K 5% resistors across 3.3 volts.  2048 is the expected return value, I get a pretty clean 2054 with Div32 and Div16.

With the testing so far, I'm thinking Div32 should be used.  We could experiment with ADC sample time/cycle time and see if Div64 or Div128 results could be improved, but for 2% or so, I'm somewhat inclined to just go with Div32.

mlyle:
Div32 sounds great.  Thanks for the testing and figuring it out!

mlyle:
Thanks much!

jihlein:
Well unfortunately there is still something odd with the F3 ADC driver.  The accuracy of the conversion is affected by CPU load.  At an MPU rate of 500Hz, both the Sparky 1 and Naze32Pro boards have accurate conversions, I'm reading ~2048 bits with the 10K/10K resistor divider network.  But increase the MPU rate to 1000 Hz, and the conversions are no longer accurate.  The Sparky 1 is getting to about 80% loading, the Naze32Pro about 64%.  Naze32Pro loading is lower due to all sensors being on SPI, there is no active I2C.

The AQ32 with the F4 always returns ~2048 bits, regardless of the MPU rate/loading.

I'm going to start digging into the reference manuals/datasheets/code some more and see if I can fiind something.

tracernz:
Look into sample time, this is how long the ADC holds the sample with it's internal cap while it measures IIRC.

jihlein:
The problem is with the number of ADC pins that are defined, along with the adc clock value, sample time, and number of samples, the nexessary conversions simply can't be completed at 1000 Hz.  That's why the values look corrupted at higher MPU rates.

I've set the ADC clock divider at 16, the sample time at 61 cycles, and reduced the oversampling from 32 to 12.  With these numbers, on Sparky the 3 ADC channels can be converted at ~1.7 kHz, on Naze32Pro the 4 ADC channels can be converted at ~1.3 kHz.

Both these boards now return the same ADC conversion values at 500 Hz and 1000 Hz MPU settings.  Sparky is showing yellow on the CPU usage, but I don't think it's recommended to use Sparky at 1000 Hz.

I'll try to get another PR put together as quick as I can, but as I'm leaving on 2 weeks of business travel, it might be a little while.....

mlyle:
I don't really like how downsampling and this whole fifo thing works at all.  It would be nice if we just returned the most frequent downsampled value.  We're not reading audio or something where we want to reconstruct a waveform, we just want to have an idea of the plant's present value.

jihlein:
One other thing - for the typical use of battery and current monitoring, which are low rate, the existing implementation is fine.  It's only a problem when you want to do things at higher rates, like reading analog inputs in the actuator loop with the MPU rate at 1 kHz, or a larger number of ADC pins.  We could leave the driver as is until we can come up with something better, and I can keep "high speed" version of the driver on my private branch were it solves the problems I was having reading servo positions at 1 kHz.

mlyle:
Yah, I'm proposing maybe we fix this in general, to just provide a recent downsampled/filtered value, and not require you to sample at 12KHz or whatever to get 1KHz of points (seems like overkill; adds significant delay)

tracernz:
The current implementation is setup to slow the DMA transfer half/fully complete interrupts down so we don't burn CPU constantly on something we only need every few seconds.

PR #949, opened 2016-04-28
targets: remove colibri target and board plugin

mlyle:
Things that would be nice to do, but are not in scope for this PR, and
optional for release (so not even worth a subsidiary issue):
- WOuld be nice to make quanton board plugin know about existing colibri
  board id and USB id-- we have no mechanism for multiple IDs in GCS, so
  it requires some duplication.
- If so, need to teach upgrade to replace the loader, too, with one that
  can support a full quanton image.

Only 2 users of Colibri ever on dR; and I'm not sure they stayed around.


tracernz:
I'm okay with this. If it ever does come back it needs to be as part of the quanton target.

tracernz:
Just checked to make sure the board plugin form etc. aren't useful for Quanton, they definitely aren't. Some of the things in this target's configuration settings are kinda lulz. LGTM

https://github.com/d-ronin/dRonin/blob/next/shared/uavobjectdefinition/hwcolibri.xml#L5 ?
https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/boards_tbs/colibriconfiguration.cpp#L107 ???

mlyle:
GCS tested and Quanton/Sparky2 targets tried (not on Colibri hardware).   No regressions seen on non-Colibri HW.  Looks OK to me

tracernz:
@mluessi You okay with this?

mluessi:
Good with me. Clicking the button..

PR #950, opened 2016-04-28
logging: minor changes

mlyle:
- Uncrustify the uavtalk code.
- Remove an unnecessary mutex.

UAVTalk is a mess and requires thorough study to be maximally fast.  Should probably defer that until we feel better about telemetry in general.


mlyle:
Logged some on Sparky2.  Seemed OK.

PR #957, opened 2016-04-30
Config:Autotune: Use convergence check

### LINKED ISSUE #391: Autotune: Use convergence check rather than hard-coded iteration count
See TauLabs/TauLabs#2004

It is quite clear that we regularly fail to converge and worst of all it's not obvious to the user unless they can recognise the bogus solution. Warn the user/invalidate the solution when convergence fails.

tracernz:
Add a convergence check to autotune, rather than a hardcoded iteration count/hope for the best approach. The user will be notified and the apply button disabled if convergence fails, sharing is still enabled (seems useful to be able to share failed tunes?). Convergence criteria is very arbitrary/guestimate but it seems to work well.

@dustin Two new fields in the JSON (converged, and iterations), and data version bumped.

Fixes #391.

### DIFF
--- a/ground/gcs/src/plugins/config/autotune.ui
+++ b/ground/gcs/src/plugins/config/autotune.ui
@@ -266,8 +266,8 @@ p, li { white-space: pre-wrap; }
            <rect>
             <x>0</x>
             <y>0</y>
-            <width>796</width>
-            <height>755</height>
+            <width>794</width>
+            <height>960</height>
            </rect>
           </property>
           <property name="sizePolicy">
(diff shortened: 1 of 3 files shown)


tracernz:
Fail:
screenshot from 2016-04-30 15-13-07

Great Success:
screenshot from 2016-04-30 15-15-18

mlyle:
Beautiful.

jhitesma:
Seems like it's going to be a little more complex than that on windows:

> 
> jhitesma@backroom MINGW64 /d/dronin/python ((bd06bf9...))
> $ conda install pyqtgraph
> Fetching package metadata .......
> Solving package specifications: ..........
> 
> # All requested packages already installed.
> # packages in environment at C:\Anaconda2:
> #
> pyqtgraph                 0.9.10                   py27_1
> 
> jhitesma@backroom MINGW64 /d/dronin/python ((bd06bf9...))
> $ ./logview.py
> Traceback (most recent call last):
>   File "./logview.py", line 3, in <module>
>     import pyqtgraph as pg
> ImportError: No module named pyqtgraph
> 
> jhitesma@backroom MINGW64 /d/dronin/python ((bd06bf9...))
> $
> 
Though I did already have non-anaconda Python installed so I had to install anaconda first...wonder if the old python left something behind causing issues.

PR #958, opened 2016-04-30
PiOS:Servo: Auto-select timer resolution, use hardware one-pulse for SyncPWM

### LINKED ISSUE #219: Auto-select PWM timer frequency

tracernz:
Timer prescaler, and consequently period are now automatically calculated based on the maximum pulse length configured within the timer bank. Hardware one-pulse timer mode is now used for SyncPWM which makes it more robust and allows us to safely use aggressive timer frequencies (Sparky1 was doing 72 MHz with Oneshot125 during my testing). The hardware one-pulse mode aligns the trailing edges of the pulses, although only per-timer at the moment. Could study further the possibility of using triggers to sync all SyncPWM timers. Small change to F3 timer clock selection, we no longer assume that APB1 is SysClk/2.

Oneshot42 is now also supported in GCS setup wizard.

Tested so far:
- Sparky1: Oneshot42, Oneshot125, 1000-2000us PWM in normal and SyncPWM mode, high/brushed frequencies on outputs 1, 2, 3 and 4.

A lot more test required, esp. F1 and F4.

Likely to use more stack in actuator, haven't checked.

TODO:
- [x] Ignore unused channels in Max pulse length code
- [x] Do something safe with unused channels
- [x] Open blocker to make upgrader deal with ActuatorSetting appropriately re: unused channels with 2000 us max pulse
- [x] Test zero pulse length on a flight board

Test required:
- [x] Analog servo
- [x] Digital servo
- [x] 400 Hz ESC
- [x] Multiple F4 targets

Fixes #219


flight/PiOS/Common/pios_servo.c
```diff
@@ -101,137 +97,126 @@ int32_t PIOS_Servo_Init(const struct pios_servo_cfg * cfg)
...
+		timer_banks[bank].max_pulse = MAX(timer_banks[bank].max_pulse, channel_max[i]);
 	}
 
+	// configure timers/banks
 	TIM_TimeBaseInitTypeDef TIM_TimeBaseStructure = servo_cfg->tim_base_init;
 	TIM_TimeBaseStructure.TIM_ClockDivision = TIM_CKD_DIV1;
 	TIM_TimeBaseStructure.TIM_CounterMode = TIM_CounterMode_Up;
+	for (int i = 0; i < banks; i++) {
+		/* Calculate the maximum clock frequency for the timer */
+		volatile uint32_t max_tim_clock = max_timer_clock(timer_banks[i].timer);
```
tracernz: Oops, volatile left over from debug.

flight/PiOS/Common/pios_servo.c
```diff
@@ -101,137 +97,126 @@ int32_t PIOS_Servo_Init(const struct pios_servo_cfg * cfg)
...
-		for (uint8_t j = 0; (j < i) && new; j++) {
-			new = new && (chan->timer != servo_cfg->channels[j].timer);
+		// output rate of 0 means SyncPWM
+		if (out_rate[i] == 0) {
+			/* 
+			 * Ensure a dead time of 2% + 10 us, e.g. 15us for 250us
+			 * long pulses, 50 us for 2000us long pulses
+			 */
+			volatile float period = 1.02f * timer_banks[i].max_pulse + 10.0f;
+			volatile float num_ticks = period / 1e6f * max_tim_clock;
```
tracernz: Oops, volatile left over from debug.

flight/PiOS/Common/pios_servo.c
```diff
@@ -101,137 +97,126 @@ int32_t PIOS_Servo_Init(const struct pios_servo_cfg * cfg)
...
+			TIM_SelectOnePulseMode(timer_banks[i].timer, TIM_OPMode_Single);
+			// Need to invert output polarity for one-pulse mode
+			uint16_t inverted_polarity = (servo_cfg->tim_oc_init.TIM_OCPolarity == TIM_OCPolarity_Low) ?
+					TIM_OCPolarity_High : TIM_OCPolarity_Low;
+			TIM_OC1PolarityConfig(timer_banks[i].timer, inverted_polarity);
+			TIM_OC2PolarityConfig(timer_banks[i].timer, inverted_polarity);
+			TIM_OC3PolarityConfig(timer_banks[i].timer, inverted_polarity);
+			TIM_OC4PolarityConfig(timer_banks[i].timer, inverted_polarity);
+		} else {
+			volatile float num_ticks = (float)max_tim_clock / (float)out_rate[i];
```
tracernz: Oops, volatile left over from debug.

tracernz:
Not sure if setupwizard does the right thing. Also, should take the Oneshot42 GUI work from #442. Could do Multishot too if wanted, looks like it's just 5-25 us PWM despite marketing materials to the contrary so probably no point having Multishot.

mlyle:
```
/var/lib/jenkins/jobs/dron/workspace@2/Nodes/master/flight/Modules/Actuator/actuator.c: In function ‘actuator_task’:
/var/lib/jenkins/jobs/dron/workspace@2/Nodes/master/flight/Modules/Actuator/actuator.c:466:3: error: implicit declaration of function ‘PIOS_Servo_Update’ [-Werror=implicit-function-declaration]
   PIOS_Servo_Update();
   ^
/var/lib/jenkins/jobs/dron/workspace@2/Nodes/master/flight/Modules/Actuator/actuator.c: In function ‘actuator_set_servo_mode’:
/var/lib/jenkins/jobs/dron/workspace@2/Nodes/master/flight/Modules/Actuator/actuator.c:727:4: error: passing argument 2 of ‘PIOS_Servo_SetMode’ makes pointer from integer without a cast [-Werror]
    ACTUATORSETTINGS_TIMERUPDATEFREQ_NUMELEM, actuatorSettings.ChannelMax);
```

mlyle:
I'm super excited about this :)

tracernz:
Heh, forgot about simulator. :P

tracernz:
Force pushed sim fixes (to avoid breaking build for bisect etc.).

mlyle:
```
In file included from /var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/Modules/Actuator/actuator.c:44:0:
/var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/Modules/Actuator/actuator.c:876:58: error: ‘PIOS_SERVO_MAX_BANKS’ undeclared here (not in a function)
 DONT_BUILD_IF(ACTUATORSETTINGS_TIMERUPDATEFREQ_NUMELEM > PIOS_SERVO_MAX_BANKS, TooManyServoBanks);
                                                          ^
/var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/PiOS.posix/openpilot.h:47:70: note: in definition of macro ‘DONT_BUILD_IF’
 #define DONT_BUILD_IF(COND,MSG) typedef char static_assertion_##MSG[(COND)?-1:1]
```

tracernz:
Okay, pulled in Oneshot42 + F3 clock fix from #442, and fixed the remaining sim issue. Should be done now.

@mluessi Can you please test this and make sure it fixes your timer/clock issue?

tracernz:
For docs or something:

| Protocol | Min Pulse | Max Pulse | Suggested Update Rate |
| --- | --- | --- | --- |
| Regular PWM | 1000 | 2000 | 400 |
| Oneshot42 | 42 | 83 | SyncPWM |
| Oneshot125 | 125 | 250 | SyncPWM |
| Multishot | 5 | 25 | SyncPWM |

flight/PiOS/Common/pios_servo.c
```diff
@@ -101,188 +97,152 @@ int32_t PIOS_Servo_Init(const struct pios_servo_cfg * cfg)
 		return -1;
 	}
 	memset(output_channel_resolution, 0, servo_cfg->num_channels * sizeof(typeof(output_channel_resolution)));
```
mluessi: Where is the memory for `output_channel_resolution` allocated?

flight/PiOS/Common/pios_servo.c
```diff
@@ -101,188 +97,152 @@ int32_t PIOS_Servo_Init(const struct pios_servo_cfg * cfg)
...
+	// configure timers/banks
 	TIM_TimeBaseInitTypeDef TIM_TimeBaseStructure = servo_cfg->tim_base_init;
 	TIM_TimeBaseStructure.TIM_ClockDivision = TIM_CKD_DIV1;
 	TIM_TimeBaseStructure.TIM_CounterMode = TIM_CounterMode_Up;
+	for (int i = 0; i < banks; i++) {
+		/* Calculate the maximum clock frequency for the timer */
+		volatile uint32_t max_tim_clock = max_timer_clock(timer_banks[i].timer);
+		// check if valid timer
+		if (!max_tim_clock)
+			return;
```
mluessi, line 167: Would it make sense to use `continue` to just skip the timer? OTOH, this really shouldn't happen unless there is a bug in the hardware configuration, so maybe a assert would be better?

mluessi:
Awesome! After discovering the timer bug the other day, I though about how to best implement this and came up with something very similar. But you probably did it better / cleaner than I would have ;).

mluessi:
Are we currently ignoring unused channels? I.e., channels that are not assigned to any functions?

What happens if a user configures a tricopter and configures channels 1.3 for OneShot125 but leaves channel 4 with the default (as it is not used)? Will in this case the 2000us from the unused channel cause a very low timer resolution?

tracernz:
> Are we currently ignoring unused channels? I.e., channels that are not assigned to any functions?
> 
> What happens if a user configures a tricopter and configures channels 1.3 for OneShot125 but leaves channel 4 with the default (as it is not used)? Will in this case the 2000us from the unused channel cause a very low timer resolution?

Good question. Currently not doing anything to mitigate that scenario. The setup wizard configures all channels but we need to work when the user configures themselves too.

tracernz:
It's not straightforward to solve this. We can construct an array of enabled(true/false) from get_mixer_type(https://github.com/d-ronin/dRonin/blob/next/flight/Modules/Actuator/actuator.c#L819) and pass it into pios_servo but the problem is what happens to unused channels? Currently they are driven at their minimum pulse but obviously this will no longer work if our timer isn't able to run that long. I think the only sane option left is to set the GPIO config of un-assigned channels as IPU.

mlyle:
IPU is bad if they're on a fet gate for brushed.

tracernz:
In which case they should have a strong pulldown resistor on the gate, otherwise they're already in trouble during boot etc.

mlyle:
actuator_task stack analysis:

```
F1, with this code:
>$ actuator_task                       504      272       10

F1, without:
>$ actuator_task                       504      272       10

F4, with this code:
>$ actuator_task                       480      272        7

F4, without:
>$ actuator_task                       424      280        7
```

So all's good.  The F3/F4 actuator stack is ridiculously big.  And there's no change to F1.

mlyle:
Re: unused channels-- they should default to 0 min/max/neutral, I think.  Wizard should fix them up to real values.

mlyle:
@tracernz -- can you remove those volatiles that don't belong and work on getting things to their final form?  Then I can flight test brushed F1 at minimum.

tracernz:
@mlyle Yeah, I should get some time this evening.

tracernz:
Is there a reason MixerType and MixerVector aren't just declared as a single 10 element field? It seems like a lot of code is awkwardly working around the fact that they can't be treated as arrays.

mlyle:
Flies well on F1 16666Hz brushed (naze32/Quanum Pico).

tracernz:
It's just the issue with unused channels to resolve. It's not as straightforward as I would have liked. I'll think on it and read some code this evening and see if I can come up with a good/clean solution.

tracernz:
- Actuators now default to 0 pulse length (min/neutral/max). Untested on real flight board!
- GCS can now autoset channel banks
  screenshot from 2016-05-04 21-34-27

@mlyle Do you have some recommended defaults for brushed? I'd like to add that as an option if possible.

flight/PiOS/Common/pios_servo.c
tracernz, line 167: Maybe assert would be appropriate. The existing code used return for all fatal cases.

mluessi:
@tracernz I think this goes in the right direction. Maybe we shouldn't call it "Channel Type", as Type means something different below.

I like the 0 max pulse default. We will need to figure out how to handle this for people upgrading. This would probably be a good case for some upgrader Python magic to set unused channels to max 0 when translating settings.

tracernz:
@mluessi I'm trying to think of a name without type in it and drawing blanks :P.

Re: upgrader, icee and I had that same discussion on IRC earlier. Already added an action in the PR description to open a blocker for it. 👍

mluessi:
@tracernz sounds good. How about let auto set also set the PWM frequency, so users could just set OneShot125 (or whatever) and be done? The Auto Set buttons should be on top if you agree.

tracernz:
It already does that, so putting it on top might be a good idea too., thanks. :+1: I'm thinking about separating the GCS part into another PR so the flight side stuff can go in sooner.

mlyle:
I think the auto-set should be a pull-down that makes it clear it's an action.    I'm going to look for an example of an icon.  Or it could say 'Change...' to make it clear it's not a special "Auto set" type that's selected.

Brushed 16666 0/3/60

mluessi:
Also, it would be nice if the pull-down autoset listed the frequencies. E.g. "Analog Servo (1000us - 2000us, 50Hz)", "Digital Servo (1000us - 2000us, 400Hz)", etc

mlyle:
I'm thinking like, a hamburger-type button with a little arrow pointing down indicating it spawns a menu.  I know it's a pain, but ... combo-box type pulldowns are not usually command interfaces.

mlyle:
Somehow this has accumulated conflict.

I kinda recommend we take the UI stuff out and land it in a separate PR.  lets us spend more time to get it right and doesn't complicate the testing of this one.

tracernz:
Yeah, mentioned that but haven't had a chance to do it yet. I bet the conflict is in the actuator module.

tracernz:
Yep, was changes to Actuator init. GCS auto-set split to another branch.

tracernz:
GCS auto-set PR will have to wait for this to merge since it depends on some changes in this PR.

tracernz:
Tested with MKS DS95i servo at 50 Hz, 330 Hz, 400 Hz, and 900 Hz. Worked great on all banks except 3. Something odd I will need to troubleshoot tomorrow.

mlyle:
Standing by to test this when you're happy with it.

tracernz:
Fixed a really dumb bug in the initial bank setup. Refactored timer clock functions a little to minimise common code. Squashed the pios_servo commits so it's not left in a broken state in any commit. @mlyle This is now ready for testing.

ufoDziner:
Installed on a BrainFPV board in a 220mm 5" quad with 1500kv motors on 6s.  The ESCs did not make the idle arming noise upon initial plug in.  I double checked the throttle and arming switch positions, then verified the correct model was chosen.  I then flipped the arming switch to see if somehow it got reversed.  Immediately it armed but the motors didn't idle.  So, I flipped the switch to disarm so I could remove the battery.  As soon as I flipped the switch, the quad launched at what seemed to be full throttle doing spirals, and eventually spiraled into the side of the hill.  1 motor was running full throttle still (prop missing) the others were off.  I unplugged the battery and brought it back.  The UAV file can be found here: https://dl.dropboxusercontent.com/u/59924814/bad_958.uav

Edit:  The arming switch had no effect throughout the event.

tracernz:
APB clock defines were incorrect for all F4 targets causing pulses to be twice as long as they should be (since this code accounts for the 2x multiplier), and frequency to be halved for non-SyncPWM. Fixed these and tested on BrainFPV. Looks like the defines are only used in pios_servo so no other effects anticipated.

tracernz:
Have scoped with the APB clock fix on Brain, Quanton, Sparky2 and Revo. Additionally tested Brain with @ufoDziner's settings above and it is now okay.

tracernz:
@fujin found a bug when not all banks are used. Fixed and force-pushed.

ufoDziner:
Tested on 2 different quads, both running BrainFPV boards and oneshot 125.  We had no issues at all.

ufoDziner:

mlyle:
Is this expected to be safe on CC3D, where a common timer is used for PWM decode and generation of some outputs?

tracernz:
Shouldn't be different to current situation re: SyncPWM (totally incompatible). For regular PWM, would need to check, I think current input code handles timer overflow?

ufoDziner:
Flew a 3rd additional quad (BrainFPV, oneshot 125) today using the artifacts with no issues.

tracernz:
CC3D changes merged in.

ufoDziner:
Compiled last night with 1018.  Loaded on 1 cc3d and 2 BrainFPV boards.

cc3d autotuned with very low Tau (https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAwJ2lnAsM).  Flew very, very well.  Handling was great, I don't really have anything bad to say.

Autotuned one Brain with it (https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAoNTm2wsM), also had very low Tau, and it too flew very well. 

The other Brain had a tune carried over (ran out of time).  It had no issues either.

mlyle:
Great. :).  Opened the blocker bug, let's merge this so we can get on to RE1

ufoDziner:
Tested Multishot.  No negative issues with it, nor did there appear to be any benefit.  Slider resolution may need to be increased for further support.

PR #960, opened 2016-05-01
GCS:Stabilization: Move hangtime to checkable groupbox

tracernz:
Unchecked sets duration to 0/disables, checked sets it to default value of 2.5s.

hangtime


tracernz:
Oops, forgot the ui file, force-pushed.

tracernz:
Also uses UAVO field limits now rather than limits on individual UI elements. 
Benefits:
- UI will display higher values if they're set manually in the UAVO, but they will be highlighted (before the UI would lie and say the value was at the limit rather than the true value)
- UI will still stop the user pushing the value past the limits
- UAVO browser still doesn't care about the limits
- Will apply to any future UI elements that use the same fields

PR #962, opened 2016-05-03
BrainFPV RE1 Support

mluessi:
Adds support for the BrainFPV RE1 (Racing Edition 1) flight controller. RE1 is an evolution of Brain and has also has a built-in OSD with a lot of other enhancements. Some details: STM32F446 running at 180MHz, gyro / accel: Bosch BMI160 running at 1.6kHz, 128Mbit of flash for logging etc. RE1 also contains additional custom hardware that allows it to control RGB LEDs (WS2812B) over SPI and also an IR emitter ("transponder") for lap timing when racing. The LEDs and the emitter can be controlled using the OSD menu.

Contents of this PR:
- RE1 support files, including the driver for the RE1FPGA (custom logic chip).
- Improvements to the OSD. In contrast to Brain, which uses 2 buffers and 2 SPIs to render the OSD, RE1 uses a single buffer (2 bits/pixel) and the QUADSPI to output the OSD data. The horizontal timing is handled externally, which simplifies the STM32 code quite a bit. This required changes to the drawing routines and a new driver (`pios_video_quadspi.c`). It also needed a different encoding of fonts and images, so scripts to generate the font headers from PNGs have been added.
- Some additional OSD settings (x-scaling, side-by-side 3D support). This is not used by Brain as the hardware can't support it.
- Change to the PWM resolution. #958 will make this unnecessary.

I will open a 2nd PR for a better looking board plugin (similar to Brain), OSD config UI changes, and also the wiring diagrams.


tracernz:
Can't wait to get stuck into this one. :D

flight/targets/brainre1/fw/ir_protocols.c
```diff
@@ -0,0 +1,155 @@
...
+* using the configuration:
+*    Width         = 16
+*    Poly          = 0x1021
+*    Xor_In        = 0x0000
+*    ReflectIn     = False
+*    Xor_Out       = 0x0000
+*    ReflectOut    = False
+*    Algorithm     = bit-by-bit-fast
+*****************************************************************************/
+uint16_t crc_ilap_trackmate(uint16_t crc, const uint8_t *data, uint8_t data_len)
```
tracernz, line 60: This should go into a lib and be called crc16_ccitt or something, this poly is used by other things (offhand Jeti Ex.Bus).

mluessi:
Braking this up into multiple PRs for easier review.

mluessi:
See #965

PR #963, opened 2016-05-03
PIOS: Code to generate IR transponder packets

mluessi:
Use by the IR "transponder" on BrainFPV RE1


flight/PiOS/Common/pios_ir_transponder.c
```diff
@@ -0,0 +1,144 @@
...
+ */
+static void ir_invert_bits(uint8_t * data, uint8_t data_len)
+{
+	for (int i=0; i<data_len; i++) {
+		data[i] = ~data[i];
+	}
+}
+
+
+/* Based on code generated by pycrc v0.9, https://pycrc.org
```
mlyle: just to echo tnz's comment-- can this go in pios_crc?  pios_crc16's polynominal looks like it might be 0x1021.

mluessi: Do you mean the `CRC_Table16` in `pios_crc.c`? I think the polynomial isn't the same. The CRC table for `crc-16-ccitt` is different: https://gist.github.com/mluessi/f8ecbcd305d3eb1b2b8c04fc29cd8734

mlyle: Ah yah, it's not.  Still-- can you move this into that file?  (which -- the other instances in there need a style cleanup / possibly to be changed to use nibble tables instead of these giant byte-oriented tables).

mluessi: I had kept it separate as the CRC implementation may not be suitable for other things, i.e., for serial comms etc a table based implantation may be better but here speed doesn't matter.

mluessi: But I'm happy to move it if you think it should be in the other file.

mlyle: Yah, I'd like to keep it together.  If someone wants a CCITT CRC it'd be nice for them to not reinvent the wheel.

Incidentally we already have an implementation of 0x1021 in flight/Libraries/rscode, but I want to rip that library out, so... :P

mluessi: Moved

mlyle:
Parfait.  Thank you!

flight/PiOS/Common/pios_crc.c
```diff
@@ -170,6 +170,45 @@ uint16_t PIOS_CRC16_updateCRC(uint16_t crc, const uint8_t* data, int32_t length)
 	return _crc;
 }
 
+/* Based on code generated by pycrc v0.9, https://pycrc.org
```
tracernz, line 173: Nice. :)

PR #967, opened 2016-05-05
Startup high priority modules first

### LINKED ISSUE #966: Continuation of #932 -- startup ordering
- "required" modules should init and start first.
- Achieve this with a new macro for "required" modules, and adjusting to start required mods first
- This can probably be safely ignored for simulator

This continues the work started in #932 to handle memory allocation failures gracefully.

mlyle:
Fixes #966 
Continuation of #932 

Have verified that it puts high priority modules first in the modules list.  Need to confirm startup on flight targets still.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -160,7 +160,7 @@ int32_t ActuatorInitialize()
 
 	return 0;
 }
-MODULE_INITCALL(ActuatorInitialize, ActuatorStart);
+MODULE_HIPRI_INITCALL(ActuatorInitialize, ActuatorStart);
 
 static float get_curve2_source(ActuatorDesiredData *desired, SystemSettingsAirframeTypeOptions airframe_type, MixerSettingsCurve2SourceOptions source)
 {
--- a/flight/Modules/Attitude/acro/attitude.c
+++ b/flight/Modules/Attitude/acro/attitude.c
@@ -163,7 +163,7 @@ int32_t AttitudeInitialize(void)
 	return 0;
 }
 
-MODULE_INITCALL(AttitudeInitialize, AttitudeStart)
+MODULE_HIPRI_INITCALL(AttitudeInitialize, AttitudeStart)
 
 /**
  * Module thread, should not return.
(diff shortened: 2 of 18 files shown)


tracernz:
Just to confirm this has all modules that trigger watchdog as HiPri?

mlyle:
I believe it has all modules that start up unconditionally on hipri (i ignored pipxtreme because of its low module count).  Will confirm on wdg specifically.

mlyle:
```
Michaels-MBP:Modules mlyle$ grep -r WDG . | grep UpdateFlag | cut -d : -f 1 | uniq -c
   1 ./Actuator/actuator.c
   1 ./Attitude/acro/attitude.c
   1 ./Attitude/attitude.c
   1 ./Autotune/autotune.c
   1 ./ManualControl/manualcontrol.c
   5 ./RadioComBridge/RadioComBridge.c
   2 ./Sensors/sensors.c
   1 ./Sensors/simulated/sensors.c
   1 ./Stabilization/stabilization.c

vs.

Michaels-MBP:Modules mlyle$ grep -r HIPRI . | cut -d : -f 1 | uniq -c
   1 ./Actuator/actuator.c
   1 ./Attitude/acro/attitude.c
   1 ./Attitude/attitude.c
   1 ./FirmwareIAP/firmwareiap.c
   1 ./ManualControl/manualcontrol.c
   1 ./Sensors/sensors.c
   1 ./Stabilization/stabilization.c
   1 ./System/systemmod.c
   1 ./Telemetry/telemetry.c
```

So everything on flight except autotune.  No radiocombridge because that's pipx/oplink.

tracernz:
Reason I ask is because I suspect watchdog could be the reboot cause if one of these fails to start? I'll look at the init code and see if the watchdog flags are registered before the tasks are created.

tracernz:
Yep, the watchdog flags are registered before the task is created/stack alloc'ed. I'd like to re-order that and not register the flag if the task alloc fails. My goal is to show the user a critical alarm in GCS if we run out of memory rather than rebooting. This is one of our biggest usability issues for F1. The things we need:
- UAVOs
- Telemetry
- Stuff that would cause us to reboot if an alloc fails

We can eliminate most of the last I think.

tracernz:
Although re-ordering that could be undesirable too. :/

mluessi:
I like this approach :)

mlyle:
Tested revo, quanton, cc3d-- seems to work nicely.

mluessi:
Thanks @mlyle

PR #973, opened 2016-05-05
flight: improvements to object persistence

### LINKED ISSUE #810: Saving UAVO shortly after boot occasionally results in reboot on S1
Somewhat rare, but definitely occasionally happens.

mlyle:
- Don't support an object persistence "selection" anymore.  Just save
  the specified object.
- Don't load back objects immediately after saving them to "double
  check".
- Yield for 25ms after any persistence op so other threads can run.

Hoped that this fixes #810

### DIFF
--- a/flight/Modules/System/systemmod.c
+++ b/flight/Modules/System/systemmod.c
@@ -367,55 +367,25 @@ static void objectUpdatedCb(UAVObjEvent * ev, void *ctx, void *obj_data, int len
 		}
 
 		if (objper.Operation == OBJECTPERSISTENCE_OPERATION_LOAD) {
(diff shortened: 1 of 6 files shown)


tracernz:
I'm a bit unclear on this, do we have any way to know if the operation failed now? e.g. ack/nack? It looks like that function doesn't return anything on failure?

tracernz:
Okay, now I see retval is actually written to the ObjectPersistance UAVO... okay..

mlyle:
Yes-- GCS watches for the state to change back to success or failure.

As a bonus, we get a callback for our own state change :P

mlyle:
Ran 10 test cycles without an undesired reset on sparky1, which is where i saw a problem.  I'd call this good.

PR #980, opened 2016-05-07
Significant logging improvements

### LINKED ISSUE #976: Logging performance and correctness changes
Right now, the serial pios_com stuff uses the fifobuf intelligently and stuffs characters in at interrupt-time.  So txstart doesn't block, and in turn attempts to send data don't block.

Unfortunately, streamfs doesn't work like this-- txstart blocks for the entire time writing happens.

As a result, logging cannot be refactored to just write the object in the callback subsystem and to eliminate the logging task (would work great for serial / openlager, but not onboard flash where writes could tie up highpriority tasks), and the buffers in the streamfs pios_com layer are not effectively used for flash.
- streamfs com should be refactored to have a tx task and tx wakeup semaphore, and to grab chunks pseudo-intelligently like the usart subsystem in openlager does.
- logging should be refactored to be task-free and callback oriented.

The big benefit from this is improved performance, eliminated skew and time jitter (where we double or omit objects, and get wrong update times), and reduced memory usage on systems that are logging to serial. (No add'l task).

mlyle:
- Better support for serial loggers
- Cleanup of a couple of issues in pios_com
- Simplification of log download from flash
- Support for logging subsystem in the simulator
- Using callback infrastructure for logger for performance
- Eliminating some data copy
- Eliminating some work done on non-flash log targets
- Factoring initialization of SPI flash logging better so it doesn't use resources when logging isn't enabled.
- Unfortunately, elimination of streamfs unit test

Tested talking at 2470000 bits per second talking to openlager dev hardware.  Tested logging in simulator. 

Fixes #976


mlyle:
Confirmed to log at 250Hz "FULLBORE" on Revolution to internal flash.  249-251 instances of all high rate objects in all seconds of a giant logfile.

tracernz:
Why PIOS_HAVE instead of the conventional PIOS_INCLUDE?

mlyle:
@tracernz corrected per your request

flight/Modules/Logging/logging.c
```diff
@@ -602,42 +627,47 @@ static void register_default_profile()
...
 	send_data((uint8_t *)LOG_HEADER, strlen(LOG_HEADER));
 
 	// Commit tag name
-	info_str = (char*)(bdinfo->fw_base + bdinfo->fw_size + 14);
-	send_data((uint8_t*)info_str, strlen(info_str));
+	// XXX all of thse should use the fw_version_info structure instead of
+	// doing offset math, but we need to factor out the fw_version_info
+	// struct from the templates first.
+	info_str = (char*)(fw_blob + 14);
+	send_data((uint8_t*)info_str, strnlen(info_str, 26));
```
mluessi, line 650: Good call on using `strnlen` it was crashing for me the other day when `fw_blob` was all `FF`. Was on my list of things to fix :).

tracernz:
Looks good, tests well.

mlyle:
Thanks much!

flight/Modules/Logging/logging.c
mlyle, line 650: It bit me in the simulator. :)

PR #986, opened 2016-05-10
androidgcs: sort objects in uav browser

mlyle:


mlyle:
Confirmed to work nicely.

PR #987, opened 2016-05-10
Tools: Allow environment to ignore missing toolchain

tracernz:
Good for e.g. my CI server that just builds doxygen docs, and the general ability to avoid this awfulness (albeit necessary to reduce support requests).


tracernz:
Out of curiosity, was this done with the new squash merge feature? :stuck_out_tongue:

PR #990, opened 2016-05-13
PiOS: Removed unused LPS25H driver

tracernz:
Looks like it came from Brain but it's not used anywhere? @mluessi ??


mluessi:
Yes, could be removed. It was on a earlier version of Brain prototypes. On the other hand, why remove it? It doesn't hurt to have it around and maybe some wants to use it or a successor (e.g., LPS25HB, which is quite similar driver wise).

tracernz:
It can be brought back in future if needed. Unused code suffers bitrot and tends to be worse than no code after a while. We have removed it as a matter of course so far in dRonin.

mlyle:
For things we expect to be functional or mostly-functional, maybe we need an attic repository... just because it's hard to search the past.  If one of us wanted to do something with LPS25 we could peek there and make sure we had no past driver/work for it.

mluessi:
Removing it is fine with me, searching the PR history will make it pretty easy to bring it back. The driver also hasn't been tested a lot, so I wouldn't want anyone to think it really works ;).

PR #993, opened 2016-05-15
GCS: Module tab cleanup

tracernz:
- Clarify auto/manual modules much more
- Move UAVO->Widget relations to UI file
- ConfigTaskWidget (UAVO->widget relations) now supports arbitrary mapping of checkbox state to enum options
- Move GPS to own tab and add constellations + auto-setup
- General improvements

screenshot from 2016-05-15 18-18-58

### DIFF
--- a/ground/gcs/src/plugins/config/config.pro
+++ b/ground/gcs/src/plugins/config/config.pro
@@ -47,7 +47,8 @@ HEADERS += calibration.h \
     configmodulewidget.h \
     configosdwidget.h \
     expocurve.h \
-    autotuneshareform.h
+    autotuneshareform.h \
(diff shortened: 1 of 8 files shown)


tracernz:
Ahh, there is a mixture of True/False and TRUE/FALSE throughout all the UAVOs... sigh.. :(

tracernz:
Changed TaskInfo to TRUE/FALSE since there are a lot of settings objects using uppercase.

tracernz:
Added a logging tab, should be handy with OpenLager coming soon. Should probably extend Logging so the custom option can be customised.

screenshot from 2016-05-15 20-21-53

mlyle:
Beautiful.

PR #994, opened 2016-05-15
eliminate fifo_buf, move to circqueue

mlyle:
Expected to be more efficient and to allow even more efficiency goin' forward.

Needs a thorough read-through and test sequence before proceeding.
- [x] Review, make sure tx data goes to tx and vice versa :P  All directions, queues, correct in pios_com
- [x] Review, make sure new circqueue calls added are correct
- [x] Review, double check use of contig vs. avail throughout the code
- [x] Test in sim
- [x] Test on hardware


tracernz:
Should move removal of BMA180 driver into another PR so it's easier to search if anyone ever wants to bring it back?

mlyle:
We could, but this would depend on it.  I tend to prefer searching logs
(and would hope we wouldn't squash and merge something like this) rather
than merges or PRs.

Either way the driver won't work afterwards because of removal of fifobuf.

tracernz:
Unfortunately Github search doesn't search commit logs. :-(

mlyle:
I search git log ;)   but #996 opened.

mlyle:
rebased

mluessi:
Code looks good (didn't try to understand every detail though ;))

mlyle:
Tested on Quanton.  1% CPU improvement (73% CPU use for fullbore logging at mpurate 500Hz) .  Logfile valid, no missing crud anywhere.  USB, serial telemetry OK.

mlyle:
I'm happy with this now.

mlyle:
Lots more test on this-- all's good g.

tracernz:
Seems to work well. Code looks good although I haven't reviewed in detail.

PR #1000, opened 2016-05-17
GCS:Lux: Rename ports to match silk, fix GUI

tracernz:


mlyle:
If we're going to do this, can we please rename the fields in the uavo to uart2/uart3 at the same time?

tracernz:
Yep, gladly, all ports match the silk now. UART1 tx config added too (although unused ATM).

tracernz:
Okay, now we always do PPM and can use UART1 for lots of things. Completely untested.

mlyle:
I guess with the past refactor, the "always do PPM" behavior got lost.  Very nice.

mlyle:
Reviewed and it looks good to me.  My proto Lux still detects receiver.  Thanks so much for improving this target.

PR #1009, opened 2016-05-18
FIX: RSSI ADC channel

### LINKED ISSUE #999: ADC RSSI only works for ADC0 and ADC1
For some reason, analog RSSI only works for the first two ADC inputs. Tested on Brain and RE1. The same pin / ADC works in the battery module. 

Not sure what the reason is.

mluessi:
Fixes #999 

Tested on Brain.


tracernz:
:-)

PR #1011, opened 2016-05-20
gcs: hidapi: from upstream: hid open mode shared

mlyle:
From https://github.com/signal11/hidapi/commit/b5b2e1779b6cd2edda3066bbbf0921a2d6b1c3c

This may explain some anomalies experienced during upgrade on Windows;
we saw various open failures because of Windows holding hid stuff open.
It may also fix the cc3d "wrong driver for composite device" problem
that users sporadically experience.

It does make running multiple GCS at once potentially more dangerous.


mlyle:

mlyle:

mlyle:
Tested this on multiple targets on Windows.  It's happy.

I can't prove it fixes some of the rarer issues we run across on Windows, but there's a solid reason to expect it does and it's a change that was made upstream, so...

PR #1014, opened 2016-05-27
actuator: don't stabilize when disarmed

### LINKED ISSUE #1008: Actuator alarms, part II: not disabled enough of the time
- Seen in version (release version or githash): release 20160409
- On flight controller (if applicable): any

Issue details:
If the RX has >0 throttle, you can get an actuator alarm even when disarmed that in turn can prevent completing setup wizard.  Noticed by tolan on IRC

mlyle:
Not that we really did anyways, but this prevents actuator alarms when
disarmed which were screwing some people up for setup wizard stuff.

Fixes #1008.


tracernz:
LGTM

PR #1015, opened 2016-05-28
LICENSE: update to mention stdperiph.

### LINKED ISSUE #928: STM32 Libraries not GPL compatible
Need to add appropriate exception to our license to allow linking with STM32 libraries.  Need to make best-effort to clear with existing contributors.  (But should be OK).  Need to be careful going forward-- it'll be difficult to bring in any third-party GPL code.

mlyle:
Fixes #928.


tracernz:
Is there any legal basis for this? (I'm pretty clueless about that side of things)

mlyle:
@tracernz -- the goal here is just to make it explicit that if people contribute code, they're OK with that code being combined with stdperiph, which is otherwise in a grey area with respect to the GPL.

mluessi:
Thanks!

PR #1020, opened 2016-05-30
Change Outer Loop sliders to 200 maximum

purpleears:
Outer Loop text boxes were set to a maximum of 200, but the sliders were set to only 100 causing the sider to sometimes max out. Values of 100+ aren't uncommon. Changed sliders to 200 as well.

### DIFF
--- a/ground/gcs/src/plugins/config/stabilization.ui
+++ b/ground/gcs/src/plugins/config/stabilization.ui
@@ -9831,7 +9831,7 @@ border-radius: 5;</string>
                     <string notr="true"/>
                    </property>
                    <property name="maximum">
-                    <number>100</number>
+                    <number>200</number>
                    </property>
                    <property name="sliderPosition">
                     <number>50</number>
@@ -9913,7 +9913,7 @@ border-radius: 5;</string>
                     <string notr="true"/>
                    </property>
                    <property name="maximum">
-                    <number>100</number>
+                    <number>200</number>
                    </property>
                    <property name="sliderPosition">
                     <number>50</number>
@@ -9995,7 +9995,7 @@ border-radius: 5;</string>
                     <string notr="true"/>
                    </property>
                    <property name="maximum">
-                    <number>100</number>
+                    <number>200</number>
                    </property>
                    <property name="sliderPosition">
                     <number>50</number>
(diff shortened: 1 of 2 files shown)


fujin:
jenkins, test this please

fujin:
jenkins, build artifacts

fujin:

tracernz:
The UAVO limits are set to 10 (with scaling of 0.1 on these widgets so 100) https://github.com/d-ronin/dRonin/blob/next/shared/uavobjectdefinition/stabilizationsettings.xml#L17-L19. The widget limits are managed by ConfigTaskWidget UAVO relation code when "haslimits" is set to "yes" so changing the slider limits in the UI file has limited effect.

The UAVO relation for the first field changed in your PR:

``` xml
                    <property name="objrelation" stdset="0">
                     <stringlist>
                      <string>objname:StabilizationSettings</string>
                      <string>fieldname:RollPI</string>
                      <string>element:Kp</string>
                      <string>scale:0.1</string>
                      <string>haslimits:yes</string>
                      <string>buttongroup:2,10</string>
                     </stringlist>
                    </property>
```

Sadly UAVO-to-widget relations are not well documented. I recently started doing this but I hate writing docs :-P https://github.com/d-ronin/dRonin/wiki/Development-UAVObject-to-Widget-Relations

This is a description of the limits functionality from the original design doc:

> The framework tries to limit the possible user interaction to keep the values inside the limits on the widgets that allow this (QSlider, QDoubleSpinBox…). **With this kind of widgets if the value is externally forced to a value outside the limits the framework will adapt the widgets limits to that value but will change its style sheet to signal the out of limits status. The original stylesheet and limits are used again once the value goes inside limits again.**
> 
> With widgets which are “impossible” to limit the user interaction with, the framework still changes their stylesheet to signal the value going outside the limits.

TL;DR I think the UAVO field limits need to be changed to 0:20 as well if we consider this safe.

-e- The original docs are actually quite good, esp. toward the end. Not sure why I worked all this out the hard way :/. https://github.com/d-ronin/dRonin/blob/next/ground/docs/UAVObj%20relations%20Framework%20document.doc

tracernz:
Tests okay, looks good. Thanks.

PR #1030, opened 2016-06-10
DO NOT MERGE. Proof of concept 16 channel output.

vladkvit:
This fork shows the minimum changes required to get GCS and Flight firmware to have full 16 channel output on Quanton board. It could work on several other boards with minimal changes.

This fork is not recommended for merging because it increases the size of UAVObects, which will cause problems with low-end hardware (CC3D).

I'm making this into a pull request so that the fork shows up in search as a reference starting point, just in case someone looks into this in the future.


mlyle:
jenkins, test this please

tracernz:
Closing, perhaps for the future after F1 is gone.

mlyle:
Have a local copy in case resources become plentiful on F1

PR #1033, opened 2016-06-13
taskinfo/modules: make case/naming consistent for frsky stuff

mlyle:


tracernz:
Did you shorten the name to SBridge rather than SensorHubBridge intentionally? I think it will be more confusing with SBridge and SPortBridge.

mlyle:
Could change the other way, but it's 4 places to change it that way vs. 2 this way (they were mixed).

mluessi:
I would prefer `SensorHubBridge`, as `SBridge` could also mean `SPortBridge`..

jihlein:
Agree with mluessi, SensorHubBridge is more descriptive.

mlyle:
Changed it the other way.

PR #1036, opened 2016-06-14
GCS:Input: Add warnings for bad reprojection settings

### LINKED ISSUE #961: UI needs to enforce that choices for reprojection are OK
as described in https://github.com/d-ronin/dRonin/pull/956#issuecomment-216092075

tracernz:
untitled

Fixes #961 
Although it doesn't enforce, just warns.

### DIFF
--- a/ground/gcs/src/plugins/config/configinputwidget.cpp
+++ b/ground/gcs/src/plugins/config/configinputwidget.cpp
@@ -326,17 +326,13 @@ ConfigInputWidget::ConfigInputWidget(QWidget *parent) : ConfigTaskWidget(parent)
                         ManualControlSettings::CHANNELGROUPS_ACCESSORY2 <<
                         ManualControlSettings::CHANNELGROUPS_ARMING;
 
-    for (int i = 1; i <= 6; i++) {
-        QComboBox *child = this->findChild<QComboBox *>(QString("fmsModePos%1").arg(i));
-        if (child)
-            connect(child, SIGNAL(currentTextChanged(QString)), this, SLOT(checkFlightMode(QString)));
-    }
-    const QStringList axes({"Roll", "Pitch", "Yaw"});
+    // check reprojection settings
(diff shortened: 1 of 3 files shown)


tracernz:
Oops, changed default tab, force-pushed.

ground/gcs/src/plugins/config/configinputwidget.cpp
```diff
@@ -1784,3 +1774,64 @@ void ConfigInputWidget::checkHangtimeConfig()
 
     m_config->lblHangTimeWarning->setVisible(warn);
 }
+
+void ConfigInputWidget::clearMessages(QWidget *widget, const QString type)
```
tracernz, line 1778: Should probs move these message functions to ConfigTaskWidget so they can be used elsewhere, and add Doxygen docs.

ground/gcs/src/plugins/config/configinputwidget.cpp
```diff
@@ -1784,3 +1774,64 @@ void ConfigInputWidget::checkHangtimeConfig()
...
+}
+
+void ConfigInputWidget::addMessage(QWidget *widget, const QString type, const QString msg)
+{
+    QLabel *lbl = findChild<QLabel *>("lblMessage" + type);
+    if (!lbl) {
+        lbl = new QLabel(this);
+        lbl->setObjectName("lblMessage" + type);
+        lbl->setWordWrap(true);
+        widget->layout()->addWidget(lbl);
```
tracernz, line 1794: Should probably not assume the widget has a layout, rather add one if it doesn't before using it.

mluessi:
Looks good. Did a quick test. Maybe it would make sense to highlight it more to alert the user of the error. I think many users will overlook/ignore the messages.

mlyle:
LGTM

mlyle:
Can improve usability later-- it's at least there.

PR #1037, opened 2016-06-14
audible alarms support but not integrated with gcs

droney:
This new module provides simple tones for warnings, errors, and critical
events using one DAC output of the arm processor. This module has not been
integrated with gcs (because I don't know how), however three variables
should be made available to the user through GCS to make module more
useful. 'DAC_Channel' is needed to select DAC channels 1 or 2.
'alarm_ignores' is used to specify event types not to output an audible
alarm, such as SYSTEMEVENT or ACTUATOR. 'audible_alarms_when_disarmed' is
used to enable or disable audible alarms when system is not armed.

In addition to integrating with gcs, compile time include selection will
be required for specifying the appropriate 'stm32f*_dac.h' file, depending
on the FC board being compiled. Currently this is hard coded as
stm32f4xx_dac.h.

Note: code has only been testing on a brain FC.

Note2: added module flight/Module/Audio/audible_alarms.c using github UI, but it's not showing up here, so attached as a zip file as well.
[audible_alarms.c.zip](https://github.com/d-ronin/dRonin/files/314949/audible_alarms.c.zip)

### DIFF
--- a/flight/Libraries/alarms.c
+++ b/flight/Libraries/alarms.c
@@ -396,6 +396,40 @@ int32_t AlarmString(SystemAlarmsData *alarm, char *buf, size_t buflen, bool blin
 }
 
 /**
+ * Return severity of top active alarm type user is not ignoring
+ */
+uint8_t getTopAlarmSeverity(SystemAlarmsData *alarm, uint32_t ignores)
+{
+        uint8_t severity = SYSTEMALARMS_ALARM_OK;
+
+        // walk through all alarm types and check if active
+        for (int i = 0; i < SYSTEMALARMS_ALARM_NUMELEM; i++) {
+                if (((alarm->Alarm[i] == SYSTEMALARMS_ALARM_WARNING) ||
+                     (alarm->Alarm[i] == SYSTEMALARMS_ALARM_ERROR) ||
+                     (alarm->Alarm[i] == SYSTEMALARMS_ALARM_CRITICAL))) {
+
+                        if (!alarm_names[i][0]) {
+                                // zero-length alarm names indicate the alarm is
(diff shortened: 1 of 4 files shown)


mluessi:
@droney it seems like some things are missing (e.g. there is no actual module in in this PR). Did you forget to add the files to git?

droney:
Hello Martin – I attempted to ‘create’ a new file with the PR, but I did not see it included with the other two changed files in the list. I did attach a zip file with the new module to the PR, so you should be able to get to that. If that does not work, please instruct me on how to upload the new file so it gets attached.

Thanks!

mluessi:
@droney just use `git add <filename>` and then commit the changes (`git commit -a -m "added file"`) then push the updated branch here and the new files will appear.

droney:
OK did these steps but file still not showing up. Something else I need to do? git push is failing with a, 
Updates were rejected because the remote contains work that you do not have locally.' error.

droney:
ok... previously did not push the Makefile change since this module not yet integrated with GCS. This I believe was causing the push to fail.

droney:
forgot to mention... you may need to bump the stack size if you have issue. I don;t have a profiler, so stack size of 512 a bit of a swag.

tracernz:
Looks like all the files are now in the PR. :) 

One comment I'd make is that I'd prefer the code that interacts directly with STM32 hardware (DAC, Timer, GPIO) to move into PiOS and/or use/extend the existing PiOS drivers where appropriate. Modules should be hardware agnostic.

droney:
tracernz - excellent point.... I will explore that approach

mlyle:
Superseded #1515

PR #1041, opened 2016-06-16
OSD: hide fields / scales for which no sensor is present

### LINKED ISSUE #1040: default disable compass, gps, barometer-related options while sensors are not present
- Seen in version (release version or githash): all
- On flight controller (if applicable): brain o.g., brain re1, all?

Issue details:
When the brain o.g. has a barometer disabled (in code), it's still possible to enable the climb related fields in the OSD, and they give bogus values (I'll get screenshots)

When the board is attached to gcs, you also see the infinite climbing in the GCS PFD at a pretty high rate.

Similarly in the brain re1 (which has no baro, or magnetometer) when the osd defaults (*at one stage) were/are used, the barometer related fields have insane values (+2~300m/s climb rate) and the compass spins wildly.

The gps fields may be able to be configured similarly but this is harder with GPS/mag requiring 5v from battery.

Is one of the other modules emitting bogus data when a lower layer sensor (baro/mag?) task is running?

mluessi:
Hides fields for which no sensors are present. Also some more fundamental changes: 
- Only initialize BaroAltitude UAVO is a barometer is present
- Only initialize Magnetometer and MagBias UAVOs if a magnetometer is present
- Only initialize GPS UAVOs if GPS module is running

Status: bench tested.

Fixes #1040

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -449,7 +449,7 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool 
 		magData.z = 0;
 
 		// Wait for a mag reading if a magnetometer was registered
-		if ((PIOS_SENSORS_GetQueue(PIOS_SENSOR_MAG) != NULL) && (MagnetometerHandle() != NULL)) {
+		if (MagnetometerHandle() != NULL) {
 			if (!secondary && PIOS_Queue_Receive(magQueue, &ev, 20) != true) {
 				return -1;
 			}
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -449,7 +449,7 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool 
 		magData.z = 0;
 
 		// Wait for a mag reading if a magnetometer was registered
-		if (PIOS_SENSORS_GetQueue(PIOS_SENSOR_MAG) != NULL) {
+		if ((PIOS_SENSORS_GetQueue(PIOS_SENSOR_MAG) != NULL) && (MagnetometerHandle() != NULL)) {
 			if (!secondary && PIOS_Queue_Receive(magQueue, &ev, 20) != true) {
 				return -1;
 			}
@@ -631,7 +631,7 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool 
 	}
 
 	float mag_err[3];
-	if (secondary || PIOS_Queue_Receive(magQueue, &ev, 0) == true)
+	if (MagnetometerHandle() && (secondary || PIOS_Queue_Receive(magQueue, &ev, 0) == true))
 	{
 		MagnetometerData mag;
 		MagnetometerGet(&mag);
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -1438,6 +1438,11 @@ static void check_home_location()
 	if (homeLocation.Set == HOMELOCATION_SET_TRUE)
 		return;
 
+	// We need GPS
+	if (!GPSPositionHandle()) {
+		return;
+	}
+
 	GPSPositionData gps;
 	GPSPositionGet(&gps);
 	GPSTimeData gpsTime;
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -495,9 +495,11 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool 
 
 		complementary_filter_state.arming_count = 0;
 
-		float baro;
-		BaroAltitudeAltitudeGet(&baro);
-		cfvert_reset(&cfvert, baro, attitudeSettings.VertPositionTau);
+		if (BaroAltitudeHandle()) {
+			float baro;
+			BaroAltitudeAltitudeGet(&baro);
+			cfvert_reset(&cfvert, baro, attitudeSettings.VertPositionTau);
+		}
 
 		return 0;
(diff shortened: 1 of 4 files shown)


fujin:

flight/Modules/Attitude/attitude.c
```diff
@@ -449,7 +449,7 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool
 		magData.z = 0;
 
 		// Wait for a mag reading if a magnetometer was registered
-		if (PIOS_SENSORS_GetQueue(PIOS_SENSOR_MAG) != NULL) {
+		if ((PIOS_SENSORS_GetQueue(PIOS_SENSOR_MAG) != NULL) && (MagnetometerHandle() != NULL)) {
```
mlyle: This seems like a pretty odd case; shouldn't we, perhaps, assert if a magnetometer exists but there's no magnetometer UAVO?

mlyle: Or is this queue getting created in either case, which seems broken somewhere else...

tracernz: The queues are created in the sensor drivers and then hooked into PIOS_Sensors after successful sensor init.. so shouldn't happen.

mluessi:
@mlyle @tracernz thanks. Removed the redundant magnetometer check and bench tested on Brain and RE1.

mlyle:
This seems backwards-- I think we were talking about moving it the other way.

Basically-- if we look to the future, I think hardware will have these UAVO-- magnetometer, etc, no matter what.  Whether the queue exists or not will come down to whether the driver is attached.

This way targets will be able to support optional sensors well (and outboard mag/baro).

mluessi:
@mlyle can you explain why? To me it seems logical that hardware that doesn't have a magnetometer also doesn't have the corresponding UAVO instantiated. This also saves some memory on F1 targets.

mluessi:
Re: Offboard sensors, this will already work. If e.g. the GPS module creates and updates a mag queue, the Sensors module will initialize the Magnetometer UAVOs.

mlyle:
Ah-- I missed the move to conditional initialization.

I don't know-- this seems a little out of step with how we've done things in the past-- in that:
1. Many uavos are initialized by any module that would read or write them-- in multiple places.  These ones aren't, but if anyone sprinkles in the other place they'll create a regression.
2. Targets used to (and some out-of-tree may still) generate a function initializing all UAVOs linked into the target (or before that, a section in flash with all the uavo inits).  Simulation still does this.  ( `UAVObjectsInitializeAll()`

That being said, it may be the best approach.  Maybe we mark up the UAVO xml to indicate that it should only be initialized by sensors for this reason?

Also--- how does brainre1 deal with altitude hold?

mluessi:
@mlyle note that it doesn't hurt if all UAVOs are initialized on some targets. Previously, it was a bit odd, e.g., the GPS UAVOs would always be present (except on F1) and there was no simple way of checking whether they are actually "active". So things like the logging module and telemetry would unnecessarily register callbacks wasting memory. I went through the code to make sure `if (GPSPositionHandle()) {..}` etc is used when accessing the UAVO, and it already was in place almost everywhere.

mluessi:
PS: RE1 doesn't do well on altitude hold as it doesn't have a barometer ;)

mlyle:
> PS: RE1 doesn't do well on altitude hold as it doesn't have a barometer ;)

Sure..  But is the module linked in?

If it's not linked in... then when you have a later outboard sensor it won't work.

If it's linked in, what happens when you select it?

mluessi:
It's linked in.. the module doesn't access `BaroAltitude` directly, so I assume it will run but do something crazy. How about raising an error in the module it `BaroAltitude` is missing?

mlyle:
> note that it doesn't hurt if all UAVOs are initialized on some targets. 

Won't it?

```
                // Wait for a mag reading if a magnetometer was registered
        if (MagnetometerHandle() != NULL) {
            if (!secondary && PIOS_Queue_Receive(magQueue, &ev, 20) != true) {                  
```

Won't that end up blocking for 20ms if magnetometer is initialized but there's no driver?  That's why I favored checking the queue instead of the other check.

mluessi:
I guess I can change the OSD code to check for the queue existence, but I thinks it's backwards.

mluessi:
@mlyle how about going back to checking both queue and UAVO existence? If only the queue is checked and UAVOs are initialized only if a sensor is present, it could cause a race condition (i.e. it's possible that the queue exists, but the Sensors module hasn't yet initialized the UAVO). Alternatively, we can also make sure that no targets initialize sensor UAVOs. AFAIK, only the simulation target does and we shouldn't really care about out-of-tree targets.

I like the idea of only having UAVOs for which sensors exist, it e.g. allows also the GCS do check for sensor existence at run time, which currently isn't possible (would be useful for e.g. disabling OSD fields in the GUI).

mlyle:
I'm not sure I like UAVO presence to be the determining factor.  But it sounds like the correct approach for now-- to use both.  And we can add some appropriate SENSORS call to know what's up in the future.

tracernz:
The queues are only attached if the sensor driver has successfully initialised the sensor, otherwise the queue call returns NULL. For the OSDs purposes though I agree that checking both sensor queue and UAVO existence makes sense, since you are calling UAVO functions that depend on the UAVO having been initialised. This is the call to make: https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/Common/pios_sensors.c#L65

mluessi:
@tracernz as a part of this PR, the Sensors module will only initialize UAVOs for optional sensors (Baro, Mag) if the queue is present; this means things like the OSD won't have to check for existence of the queue, just the UAVO.

mluessi:
Superseded by #1049

PR #1042, opened 2016-06-16
GCS: Add new OSD options

mluessi:
Adds new OSD options introduced by RE1. Namely: x-scaling and configuration of the 3D OSD mode.

3d


mlyle:
Looks great

tracernz:
Is it feasible to hide fields which aren't available? Perhaps an IsOsdCapabilityPresent(enum OsdCapabilities cap) call to the board plugin or something?

PR #1043, opened 2016-06-16
simulation: improvements

mlyle:
- Cross compilation hack for pi
- Lots of refactor to coalesce most of pios.posix into pios.


mlyle:
Simulator is happy with these chances.  Bench-tested naze OK.

(Only change that is at all incorporated into other targets is reordering of headers in pios.h)

mlyle:
Quick hop on quanum pico was OK.  OK to merge IMO.

PR #1044, opened 2016-06-17
autotune usability

### LINKED ISSUE #149: Autotune forum share button placement maybe not ideal
The lower right button is "close".

Usually the lower right button is what you want to do by default, "next" or whatever.  So this is counterintuitive.

mlyle:
Fixes #149

Todo:
- [x] Fixup 'done' vs 'finish' -- platform specific -- in final autotune dialog
- [x] Look at on Windows; consider pixmaps
- [x] Consider turning failure to converge red
- [x] Test many variations-- applying yaw, not, etc.
- [x] Test whole lifecycle on real hardware
- [x] Column width / maybe make dialog a little bigger
- [x] Disable button when not connected to autopilot

Done in this PR:
- detect new tunes, and invite user to apply them
- replace second autotune tab with pseudo-wizard
- simplify directions
- make it more likely users share
- do sharing async / in background
- don't pump values for sharing through UI objects


mlyle:

mlyle:
7uzgefl
xw9wudd
as7z0vi
t1imu0t
jtaeuyv
mldf1da

ufoDziner:
Like the direction overall.  Checked on Win 10.
There is no "back" or "previous" button near the "Next" button, only an arrow near the logo.
Does Yaw still need a checkbox?
A button to reset the sliders to defaults would be handy.
Text says "finish button", button is labeled "Finish".
Help question mark doesn't seem to work on the pages.

1044ss

tracernz:
This looks like a nice improvement!

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -685,7 +685,7 @@ void AutotuneSlidersPage::compute()
 
     // handle non-convergence case.  Takes precedence over all else.
     if (!converged) {
-        lblWarnings->setText(tr("Tune didn't converge!  Check noise and damping sliders."));
+        lblWarnings->setText(tr("<font color=\"red\">Error:</font> Tune didn't converge!  Check noise and damping sliders."));
```
tracernz: `<span style="color: red">` is way nicer, and way less deprecated. :-P

mlyle:
Kinda cheesy, but added an OSX watermark from @peabody124's autotune writeup:

image

mlyle:
@ufoDziner --- thanks for feedback.

> There is no "back" or "previous" button near the "Next" button, only an arrow near the logo.

Unfortunately I don't think I can fix this while letting Qt decide which look and feel to use for each platform.  Sadly this is "right" for Windows 10-- I agree it's wonky.

> Does Yaw still need a checkbox?

At minimum, we need to be able to turn it off for when yaw beta is bad.  And I hear this has caused overheating of some motors (just once or twice).

> A button to reset the sliders to defaults would be handy.

There's a reset parameters button in your screenshot...

> Text says "finish button", button is labeled "Finish".

Fixed up.

> Help question mark doesn't seem to work on the pages.

Argh need to figure out how to turn this off.

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -722,56 +337,400 @@ QJsonDocument ConfigAutotuneWidget::getResultsJson()
...
+        av->kp[2] = av->kp[0] * scale;
+        av->ki[2] = 0.8 * av->ki[0] * scale;
+        av->kd[2] = 0.8 * av->kd[0] * scale;
+    } else {
+        av->kp[2] = -1;  av->ki[2] = -1;  av->kd[2] = -1;
+    }
+
+    // handle non-convergence case.  Takes precedence over all else.
+    if (!converged) {
+        lblWarnings->setText(tr("<font color=\"red\">Error:</font> Tune didn't converge!  Check noise and damping sliders."));
```
tracernz: `<span style="color: red">` is way nicer, and way less deprecated. :-P

-e- Not even just deprecated, but obsolete:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font

ufoDziner:
@mlyle FWIW, I had no idea that the "Reset Parameters" button is actually a button.  It looks like a label to me.

tracernz:
I think it might be wise to merge this and test + fix any minor issues in-place given the current position in the release cycle, what do you think @mlyle?

mlyle:
@tracernz I need to make it through a couple small test cycles of the full use-- enable module, fly tune, connect back/detect tune, apply, adjust tune.. which I don't think anyone else has done.  I have some hope of doing that late today, and then I think it can merge.

mlyle:

mlyle:
Tested a fair bit on Windows today.  With these changes I'm happy / ready to merge.

ground/gcs/src/plugins/config/autotunefinalpage.ui
```diff
@@ -17,14 +17,30 @@
    <item>
     <widget class="QLabel" name="lblCongrats">
      <property name="text">
-      <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Congratulations! You've reached the end of the Autotune process.&lt;/p&gt;&lt;p&gt;By clicking the "Finish" button, the values calculated in the previous step will be saved to the board. &lt;/p&gt;&lt;p&gt;In addition, if you select the sharing checkbox, the vehicle data will be shared with the dRonin Cloud Service.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+      <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Congratulations! You've reached the end of the Autotune process.&lt;/p&gt;&lt;p&gt;By clicking the &quot;Finish&quot; button, the values calculated in the previous step will be saved to the board. &lt;/p&gt;&lt;p&gt;In addition, if you select the sharing checkbox, the vehicle data will be shared with the dRonin Cloud Service.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
mlyle, line 20: Qt changed this on its own, but I confirmed it did still do the right thing on OSX.

PR #1046, opened 2016-06-19
naze32: adjust for initcall change

mlyle:
Naze32 was missed as part of the #967 (due to its linker script being in
a different place) and was failign to start up.  Now is OK.


tracernz:
Should maybe move the linker script?

mlyle:
@tracernz Ideally, but this is the minimal fix and is confirmed to work.

PR #1050, opened 2016-06-21
i-BUS: Increase channel count to 10

tracernz:
For their 10 channel radios e.g. FS-i10.

Not tested at all yet.

To merge:
- [x] Confirm this works with my protocol generator
- [x] Test on FlySky hardware

### DIFF
--- a/flight/PiOS/Common/pios_ibus.c
+++ b/flight/PiOS/Common/pios_ibus.c
@@ -33,9 +33,9 @@
 #ifdef PIOS_INCLUDE_IBUS
 
 // private
-#define PIOS_IBUS_CHANNELS 8
-// 1 sync byte, 1 unknown byte, 8x channels (uint16_t), 12 unknown bytes, 2 crc bytes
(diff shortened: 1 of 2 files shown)


tracernz:

mlyle:
jenkins, test this please

mlyle:

tomasa73:
Tested the build, and it works with 10ch, and iBus. Haven´t done any testflight yet. Still the TX wizard doesn´t get through. When I get to the screen when I do the stick calibration (move everything), no respone is seen on the screen. If I then press save, and set the input and channel manually and the do the manual calibration it works.

tracernz:
Thanks for the test! What are the approximate minimum/maximum/neutral values you see when calibrating manually? I will generate some i-BUS packets with the same range and see if I can track down the issue.

tomasa73:
Throttle 1007/1985
Roll 1008/1500/1998
Pitch 1004/1500/2000
Yaw 1000/1500/1992

PR #1052, opened 2016-06-21
pios: fix the build

mlyle:
The pios sim and srxl changes interacted.  With either alone it was
fine; with both we couldn't build.

### DIFF
--- a/flight/PiOS/Common/pios_srxl.c
+++ b/flight/PiOS/Common/pios_srxl.c
@@ -143,6 +143,21 @@ const struct pios_rcvr_driver pios_srxl_rcvr_driver = {
 
 /* Implementation */
 
+/* byte-ordering macros */
+#define ntohl(v) (				\
+	(((v) & 0xFF000000) >> 24) |		\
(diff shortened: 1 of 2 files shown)


mlyle:
(I will be self-merging this once it gets through Jenkins if there are no objections)

tracernz:
Hmm, I would like to make these macros available to all code if possible. They're useful, and should be used in place of other byte-swapping stuff elsewhere too (to standardise). Is `#ifndef ntohl` etc. feasible?

mlyle:
@tracernz then it becomes dependent on other headers, and some platforms they're extern, etc.  Yes, it would be nice to do in a generalized way, but I'm trying to fix the build.

tracernz:
Okay.
P.S. `#include <pios_usart_priv.h>` doesn't look like it was needed in pios_srxl.h, could be for debug stuff I had early on, or it could be needed only in pios_srxl.c. :)

mlyle:
;) yah, i tried a couple of things, ... it'd be nice to make all this cleaner later but i'm willing to incur debt for things workin'

PR #1054, opened 2016-06-22
Flight: Add triflight control algorithms

jihlein:
Implements the triflight control algorithms for improved tricopter tail servo and motor control.

These algorithms were originally implemented in CleanFlight, creating a fork call TriFlight.  The implementation has been modified to fit the Dronin paradigm, and also to make use of the F3/F4 FPU.

As far as testing goes, all functions have been tested on my private branch using the Naze32Pro target.  Initial setup, tail tuning (both servo setup on the bench and torque tuning in the air), and flight all worked as expected with a servo modified for position feedback.  I've also tested the virtual servo model version (no servo feedback) on the AQ32 target with good results.  Finally, it was bench tested on a Naze32 and seemed to run okay.  The Naze32 was not flight tested.  There are a couple of flight parameters I'm still working to refine.

I've also flown a quad and single copter with the triflight branch, with the triflight functions disabled.  No ill effects were observed.

There's more algorithm work to be done, but this is a good point to create the first PR, both from a code integration and functional standpoint.


flight/Modules/Actuator/actuator.c
```diff
@@ -83,13 +86,18 @@ static bool manualControlCommandUpdated = true;
 // set true to ensure they're fetched on first run
 static volatile bool actuator_settings_updated = true;
 static volatile bool mixer_settings_updated = true;
```
mlyle, line 89: We may want to make all of this stuff conditional on not-F1, because RAM is tight.

flight/Modules/Actuator/triflight.c
```diff
@@ -0,0 +1,1018 @@
...
+
+	return correction;
+}
+
+///////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////
+
+#if !defined(ARCH_POSIX) && !defined(ARCH_WIN32)
```
mlyle, line 581: Probably should not exclude this from the simulator.

flight/Modules/Actuator/triflight.c
```diff
@@ -0,0 +1,1018 @@
...
+{
+	static uint32_t buzzer_timestamp;
+	static uint32_t duration;
+
+	if (triflightStatus->BuzzerChannel == 255)
+		return;
+
+	if (buzzer.state == BUZZER_ON)
+	{
+		if (PIOS_DELAY_GetuSSince(buzzer_timestamp) >= duration)
```
mlyle: a separate set of buzzer infrastructure for this isn't great-- we need to figure out a way to unify all the buzzer things.

flight/Modules/Actuator/triflight.c
```diff
@@ -0,0 +1,1018 @@
...
+	else if (!(force < yaw_force_curve[TRI_YAW_FORCE_CURVE_SIZE - 1])) // No force that high
+	{
+		return TRI_TAIL_SERVO_ANGLE_MID + TRI_TAIL_SERVO_ANGLE_MAX;
+	}
+
+	// Binary search: yaw_force_curve[lower] <= force, yaw_force_curve[higher] > force
+	uint8_t lower  = 0;
+	uint8_t higher = TRI_YAW_FORCE_CURVE_SIZE - 1;
+
+	while (higher > lower + 1)
```
mlyle, line 145: reminder to self: i need to carefully read this, because formulating binary search on things like this is errorprone.

mlyle, line 145: (maybe should just go through it in order, actually, to simplify the code, and to eliminate the two conditionals at top).

shared/uavobjectdefinition/triflightsettings.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="MotorDecelDelayMs" units="mSec" type="float" elements="1" defaultvalue="100.0"/>
+        <field name="MotorThrustFactor" units="" type="float" elements="1" defaultvalue="13.8"/>
+        <field name="ServoMaxAngle" units="degrees" type="float" elements="1" defaultvalue="40.0"/>
+        <field name="ServoSpeed" units="dps" type="float" elements="1" defaultvalue="300.0"/>
+		<field name="ServoFdbkPin" units="" type="enum" elements="1" options="ADC0,ADC1,ADC2,ADC3,ADC4,ADC5,ADC6,ADC7,ADC8,NONE" defaultvalue="NONE"/>
+        <field name="AdcServoFdbkMin" units="" type="float" elements="1" defaultvalue="2040"/>
+		<field name="AdcServoFdbkMid" units="" type="float" elements="1" defaultvalue="2048"/>
+		<field name="AdcServoFdbkMax" units="" type="float" elements="1" defaultvalue="2056"/>
+        <field name="VirtualServo" units="" type="uint8" elements="1" defaultvalue="1"/>
+        <field name="UseTriFlight" units="" type="uint8" elements="1" defaultvalue="0"/>
```
mlyle: bool/enum

mlyle:
So two questions.
1.  Is the ADC measurement of angle / feedback really such a win?
2.  Doesn't look like we actually _control_ based on the feedback, just measure it for the purpose of thrust corrections.  Is this true?  If we have the ADC measurement of angle, would we be better off with a PI controller trying to hit the angle?

mlyle:
Overall, this looks good (indenting, above issues aside).  I'm looking forward to its inclusion-- thanks much for the submission.  Probably will take us a bit of time to shake it all out.

shared/uavobjectdefinition/triflightsettings.xml
jihlein: Are you referring to line 9?  All other xml files that use the ADC pins leave the units as "" and the type as "enum".  See flightbatterysettings.xml lines 5 and 6 for example.

flight/Modules/Actuator/triflight.c
jihlein, line 581: Including this part in the sim causes the sim build to fail, due to the calls to the PIOS_DELAY_XXXX functions, as they seem to be undefined in the simulator.  Also since the ADC read in the feedback servo module fails for the same reason, there is no ADC data for this part to act on, so it's of little to no use in the sim environment.

flight/Modules/Actuator/actuator.c
jihlein, line 89: I considered that too.  It does seem to work on F1, but I did not check how much RAM was left.

shared/uavobjectdefinition/triflightsettings.xml
mlyle: UseTriFlight-- this is intended to be a bool, right?  Then it should be treated like other bools (enum, DISABLE/ENABLE values) in UAVO

flight/Modules/Actuator/triflight.c
mlyle, line 581: Note ARCH_POSIX includes upcoming FlyingPi, etc.

flight/Modules/Actuator/actuator.c
mlyle, line 89: My main concern is that it might not let autotune still work on some F1 configurations (that are not even using triflight).

mluessi:
This is awesome. I'm not very happy with the name "triflight"; it makes sense that the original fork was called triflight, as it was a cleanflight fork, but for us it is just a tricopter with a specialized control algorithm. It would be nice if the GCS presented these options when a tricopter airframe is selected.

jihlein:
Addressed the review comments to date, did some debug on the tail servo calibration.  Any better ideas for a naming convention?

jihlein:
Development complete, rebased to current next.  Flight tested on DTFc.  Working toward testing on other targets.

LitterBugs:
Flight tested on RCExplorer F3FC Racing FC.

jihlein:
Overcome by events.

Issue #1048, opened 2016-06-20
Recent next: USB detect after FC reboot on OSX doesn't work


mlyle:
Has to be a fairly recent regression.

mlyle:
Or it's particular to some flight boards-- seen on Lux.

mlyle:
Sparky2 does: 

```
"Telemetry connected"
"Telemetry disconnected"
Automatically opening device
ConnectionManager::devChanged autoconnected USB device
[RawHID] Error writing to device
```

mlyle:
Looking at several targets-- the behavior is varied.
1. Sometimes usbmonitor sees the target go away and come back.  This is a race with our polling.  If usbmonitor sees this, all goes well.
2. Sometimes without that, ConnectionManager decides to try reconnect on its own.  Sometimes this works, sometimes RawHID gets an error writing to device which puts a stop to this.
3. Sometimes connectionmanager never tries.

mlyle:
Fixed in #1081

PR #1056, opened 2016-06-24
gcs: usb: changes to make reboot button better

mlyle:
This is not 100% reliable, but well over 90%.  There's a remaining
failure category (when GCS never notices the departure) that's not
addressed.

Mitigates #1048 on OSX.  @tracernz, feedback would be very helpful
to see if this helps you.  Tested with sparky2 and revo and lux.


tracernz:
Nope, doesn't help.

mlyle:
Really weird that these races always go off differently for you :/

tracernz:
Yeah, adding some debug code in hidapi which changed timing slightly seemed to help it. The whole thing is very fragile. :(

PR #1057, opened 2016-06-24
Add headfree support

mlyle:
Tested in simulation, but.. coordinate systems are weird / i need to really try this FML.


mlyle:
Behavior in sim: https://www.youtube.com/watch?v=LwERr0j6Jyg

mlyle:
Just flight tested it on Lux.  Works perfectly.

mluessi:
Awesome :). Code looks good to me. Do we need to add GCS warnings like we did for the camera angle mode?

mlyle:
@tracernz already put in the correct warnings for headfree, and this connects to them. :D

Issue #1055, opened 2016-06-24
Flight: MPU6000 bottom orientations appear to be incorrect


jihlein:
- Seen in version (release version or githash): Latest Next
- On flight controller (if applicable): Any with MPU6000 mounted upside dowm

Issue details: The inverted MPU6000 orientations don't work correctly.  The sign inversions appear incorrect, and possibly the axis assignments.  I corrected the PIOS_MPU60X0_BOTTOM_90DEG orientation by removing the 2 inversions, and then it works properly.  The rest of the "BOTTOM" orientations need review.  Actually all the assignments need reviewing, as there's some axis swapping going on too that I don't understand.

mlyle:
Surely at minimum you need one inversion to put something on the bottom-- doesn't Z accelerometer switch sign, and Z gyro at minimum?

jihlein:
Yes, of course.  Look at the 90 top orientation vs 90 bottom orientaion, you will see the z axis data is inverted from top to bottom orientation, and that is okay.  It was the x and y that were inverted incorrectly.  If you experiment with the bottom 90 degree orientation, you will see short term the ADI go one way, then long term the other, indicating the gyro and accels are out of phase.  I suspect there are others with the same issue, but I don't have time to go thru them all.  I plan to when I get back from vacation, or someone can take a look if they want to earlier.

Date: Thu, 23 Jun 2016 20:19:52 -0700

mlyle:
What I think I'd really like to see, in general (and it's a bit tough to do) is to remove the gyro/accel sensor orientation stuff, and to add them to the board orientation rotation.  No need to do redundant work when it can be a single rotation.

mlyle:
(But it's quite all right for us to focus on fixing the rotations first).

jihlein:
Corrected with #1058.  Closed.

PR #1058, opened 2016-06-24
Flight: Move MPU6000 Targets to pios_mpu.h/pios_mpu.c

jihlein:
On the PIOS_MPU60X0_BOTTOM_XXXDEG orientations, the gyros and accels where not aligned correctly, causing the complementary filter to produce incorrect attitude estimates.

This patch was bench tested on a Quanton, proving all 8 orientations had the gyros and accels aligned correctly.  It was also specifically tested on an RcExplorer F3FC off tree target in the PIOS_MPU60X0_BOTTOM_90DEG orientation where i had initially discovered the problem, also without issue.

Corrects issue #1055


mlyle:
@jihlein -- excellent work-- even if we don't use this in the mainline targets it's really useful to clear this kind of technical debt that could bite us in the future.

Unfortunately, we have closely related code in pios_mpu, pios_bmi160, pios_hmc5883/5983/5983_i2c, mpu6050/9150/9250/9250_spi.  It sure would be cool if we could factor these transformations common somehow.

tracernz:
Rather than this, check the rotations in pios_mpu.c and swap relevant targets to it (CC3D, revo, ?) and delete this driver IMO.

mlyle:
@tracernz is the mag behavior corrected in pios_mpu?

tracernz:
@mlyle Not sure, haven't looked at it for a long time (need to get back to it sometime). MPU6xxx don't have mag anyway.

tracernz:
Bottom 90 and bottom 270 are opposite to this PR in pios_mpu.c, depends which axis you define as the reference for the rotation I guess.

jihlein:
So regarding Bottom90 and Bottom270, which axis do we want to rotate around?  Our definition of Y as in pios_mpu6000.c, or X as it pios_mpu.c?

mlyle:
Well, isn't it the "same axis"-- just a question of whether you rotate and flip, or flip and then rotate?

jihlein:
Here's how I look at it:
1. To go from top0 to bottom0, there is a 180 degree rotation around X
2. To go from top180 to bottom180, there is a 180 degree rotation around X
3. To go from top90 to bottom90, there is a 180 degree rotation around X
4. To go from top270 to bottom270, there is a 180 degree rotation around X

This way, the same sensor "edge" is forward, whether oriented top of bottom.  Is that correct?  Or is all that matters is that we are consistent?  This subject always gives me headache....  :) :)

jihlein:
"Rather than this, check the rotations in pios_mpu.c and swap relevant targets to it (CC3D, revo, ?) and delete this driver IMO."

Done.  All targets tested to compile, but the following targets need testing on hardware to verify MPU operation:
- [x] AQ32
- [ ] CC3D
- [x] Quanton
- [ ] Revolution

jihlein:
Verified AQ32 target working correctly on all 8 orientations with pios_mpu.h and pios_mpu.c.

jihlein:
Verified Quanton target working correctly on all 8 orientations with pios_mpu.h and pios_mpu.c.

When time permits, hopefully someone can test CC3D and Revolution targets.

PR #1061, opened 2016-06-25
Add CRC16 to logfs

mluessi:
Adds CRC16 to the UAV Objects written to flash. For backwards compatibility, the CRC is written after the object and for now we also accept `0xFFFF` as a valid CRC (as this is the value the flash would have when the CRC isn't written).

Status: Tested on F4 (RE1) checking that objects with and without CRC are saved / loaded correctly.

### DIFF
--- a/flight/PiOS/Common/pios_flashfs_logfs.c
+++ b/flight/PiOS/Common/pios_flashfs_logfs.c
@@ -957,6 +957,8 @@ out_exit:
  * @retval -3 if object not found in filesystem
  * @retval -4 if object size in filesystem does not exactly match buffer size
  * @retval -5 if reading the object data from flash fails
+ * @retval -6 if reading the CRC from flash fails
+ * @retval -7 if CRC is incorrect
  */
 int32_t PIOS_FLASHFS_ObjLoad(uintptr_t fs_id, uint32_t obj_id, uint16_t obj_inst_id, uint8_t *obj_data, uint16_t obj_size)
 {
--- a/flight/PiOS/Common/pios_flashfs_logfs.c
+++ b/flight/PiOS/Common/pios_flashfs_logfs.c
@@ -29,6 +29,7 @@
 
 #include "pios_flash.h"		     /* PIOS_FLASH_* */
 #include "pios_flashfs_logfs_priv.h" /* Internal API */
+#include "pios_crc.h"
 
 #include <stdbool.h>
 #include <stddef.h>		/* NULL */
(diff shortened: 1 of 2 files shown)


tracernz:
Have you checked an object up to the UAVO size limit? I can't quite recall where the limit was but I think it was somewhere in the filesystem side so it may have reduced with this.

-e- https://github.com/TauLabs/TauLabs/pull/1849#issuecomment-135981132

mluessi:
@tracernz I did but for some reason I overlooked that the enum in the header is 4 bytes and assumed the header size was 9 bytes and therefore thought it would fit. For this to work, we would need to change the max size to 242. I added the corresponding commit. We currently don't have an UAVO that is that large and I think adding the check is probably worth it.

mluessi:
I was just thinking: An alternative approach to store the CRC could be to get rid of the size field in the header and store the CRC there instead. The firmware already knows the size based on the object ID, so it's redundant. For backwards compatibility we would accept object for which the CRC corresponds to the size known based on the object ID.

tracernz:
I guess the only value in the size is to protect against the unlikely case of an ID collision.

mluessi:
Yeah, I assume so. But having the CRC there would offer similar protection.

flight/PiOS/Common/pios_flashfs_logfs.c
```diff
@@ -990,6 +1003,26 @@ int32_t PIOS_FLASHFS_ObjLoad(uintptr_t fs_id, uint32_t obj_id, uint16_t obj_inst
...
 		}
+
+		/* Check the CRC */
+		uint16_t crc_fromfs;
+		if (PIOS_FLASH_read_data(logfs->partition_id,
+						slot_addr + sizeof(slot_hdr) + obj_size,
+						(uint8_t *)&crc_fromfs,
+						2) != 0) {
+			/* Failed to read CRC */
+			rc = -6;
```
tracernz, line 1016: Should cover this error in Doxygen docs.

flight/PiOS/Common/pios_flashfs_logfs.c
```diff
@@ -990,6 +1003,26 @@ int32_t PIOS_FLASHFS_ObjLoad(uintptr_t fs_id, uint32_t obj_id, uint16_t obj_inst
...
+			goto out_end_trans;
+		}
+
+		/* For now, we also accept 0xFFFF for backwards compatability */
+		if (crc_fromfs != 0xFFFF) {
+			uint16_t crc = PIOS_CRC16_updateCRC(0, obj_data, obj_size);
+			if (crc_fromfs != crc) {
+				rc = -7;
```
tracernz, line 1024: Doxygen.

mluessi:
@tracernz Updated the doc string.

PR #1066, opened 2016-06-28
GCS: Make board plugins semi-portable

tracernz:
Opened for visibility. Puts hardware UAVOs inside board plugins. Will allow third parties (e.g. @jihlein) to distribute standalone board packages that work with dRonin GCS.

TODO:
- [x] Check addons directory for firmware images (uploader/upgrader)
- [x] Use githash of available firmware image rather than GCS githash for upgrader
- [ ] Manage plugin version/compatibility (maybe it's possible to automatically bump plugin versions (in .pluginspec) when commits that change public interfaces are made?)
- [ ] Packaging for board plugins

### DIFF
--- a/ground/gcs/src/plugins/boards_dtf/dtfc.cpp
+++ b/ground/gcs/src/plugins/boards_dtf/dtfc.cpp
@@ -37,7 +37,6 @@
 #include "uavobjectutil/uavobjectutilmanager.h"
 #include <extensionsystem/pluginmanager.h>
 
-#include "hwdtfc.h"
 #include "dtfcconfiguration.h"
 
 /**
@@ -116,9 +115,16 @@ QPixmap Dtfc::getBoardPicture()
     return QPixmap(":/dtf/images/dtfc.png");
 }
 
-QString Dtfc::getHwUAVO()
+QString Dtfc::getHwUavoName()
 {
     return "HwDtfc";
+}
+
+HwDtfc *Dtfc::getHwUavo()
+{
+    ExtensionSystem::PluginManager *pm = ExtensionSystem::PluginManager::instance();
+    UAVObjectManager *uavoManager = pm->getObject<UAVObjectManager>();
+    return qobject_cast<HwDtfc *>(uavoManager->getObject(HwDtfc::NAME));
 }
 
 //! Determine if this board supports configuring the receiver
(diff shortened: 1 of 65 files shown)


mluessi:
I assume this is to support out-of-tree boards. We had discussed this a while ago (in the Tau Labs days) and the board plugin is only one thing that's needed. The more difficult issue is the UAVO needed for the board, i.e., the GCS would have to know about and out-of-tree UAVO for the hw settings. It also means the UAVO hash would be different.

tracernz:
The UAVO is the bulk of the work in this PR ;-).

tracernz:
Rebased, took care of two of the bullet points. Currently the add-on firmware path is the last searched, in future we might want to make it first which would allow us to update single board images in bugfixes rather than an entire package (currently we're not setup to be able to do this anyway).

tracernz:
Remaining work can come in a future PR.

ground/gcs/src/plugins/boards_dtf/dtfc.cpp
```diff
@@ -116,11 +115,18 @@ QPixmap Dtfc::getBoardPicture()
     return QPixmap(":/dtf/images/dtfc.png");
 }
 
-QString Dtfc::getHwUAVO()
+QString Dtfc::getHwUavoName()
 {
     return "HwDtfc";
 }
 
+HwDtfc *Dtfc::getHwUavo()
```
mlyle: This one should use the template like the other one, right?

mlyle:
I reviewed and this looks really good.  I can help with further packaging.

I do think this PR needs one more minor thing-- package/winx86/dronin.nsi should blow away the addons directory IMO.

ground/gcs/src/plugins/boards_dtf/dtfc.cpp
tracernz: Yes, thanks!

tracernz:
Fixed DTFc, and rebased (so I could test DTFc without bootlooping). Can address the Windows installer thing in a future PR, but don't want to hold this one up while I figure out how NSIS works, test on Windows etc.

mlyle:
OK.  It's just one rmdir line (model after the plugins one), and it's a nasty piece of technical debt if it doesn't get done (because incompatible plugins will make it crash)

mlyle:
Will merge once jenky is happy.

PR #1067, opened 2016-06-28
VTXConfig Module

mluessi:
Module to configure video transmitters. Currently supported are Team Blacksheep (TBS) Unify Pro 5G8 transmitters. The newer "TBS Unify Pro 5G8 HV" should mostly work as well, but the protocol is a bit different. I will get a HV transmitter tomorrow (6/29) to do further tests.

On Brain and RE1 the video transmitter can be configured through the OSD menu. I didn't add the module to the other targets as I'm not sure how useful it is without and OSD.

Status: Bench tested.

To test: Connect audio cable to Tx pin of serial port.


mluessi:
Good news first: I tested this using a second Unify Pro and it works as expected. The bad news is that it doesn't work with the Unify Pro HV and I somehow managed to kill the video output of the TBS Core Pro I have, so I can't use the logic analyzer to look at the data.

This should be ready for review.

mluessi:
OK, turns out I didn't kill the Core Pro, just the cable was wrong. After updating the Core Pro to the latest beta firmware (1.51) it started communicating with the HV video transmitter. The HV video transmitter no longer ignores the CRC so I had to add it to the messages sent by the flight controller. 

Also added support for showing the Tx power in the OSD. If the VTX is started in "Pit Mode" (low power mode only supported by the Unify Pro HV) the flight controller won't attempt to set the transmit power.

fujin:

ufoDziner:
Checked on OG brain.  Doesn't detect the the VTx with it plugged into the Main Port.  Works just fine with it in the Flex port.

mluessi:
@ufoDziner thanks for the test. I just tested it on Brain it is working on all the serial ports (Main, Flx, RxPort) for me. Can you test again to make sure it was not a wiring problem. Sometimes it can take a few seconds for it to connect as the module tries different baud rates until it finds one that works (the Unify Pro VTX doesn't use a quartz oscillator for the UART, so the deviation can be quite large, one model I have works reliably at 5kbps instead 4.8kbps).

ufoDziner:
@mluessi I'm not sure what to check from a wiring perspective.  I'm literally taking the the plug out of the Main port and plugging it into the Flex port, and vice-versa.  When I do this I swap the modules on the Hardware tab as well so that the SRXL and VTX control modules match the locations on the board, then power cycle the entire system.  The Rx works at both locations, so I'm confident that the FC is working properly.

I only have one wire from the VTx going into the plug (aligned with the Tx pin).  Am I supposed have a ground wire split off the VTx gound as well?  Thanks

mluessi:
@ufoDziner there needs to be some ground connection between Brain and VTX, not necessarily from the same port. As you said, it should work when switching from Flx to Main, but other than either a wiring problem or a defect on the Brain you have I don't have an explanation why it would work on one Brain but not on the other.

mlyle:
We had something else recently not work for @ufoDziner on main didn't we?

ufoDziner:
@mluessi I will test on another board, and add a ground wire to the port in a separate test.

@mlyle No, it worked actually.  It was SRXL, and was stated by a LP developer that it wouldn't work on the Main port.  Both @tracernz and I tested it successfully.

mluessi:
@ufoDziner I also had a problem using it with an RE1 I have. Made the receiving code more robust by matching directly for the length field and trying more different baurates. Make sure to try the latest version, it works for me now.

mluessi:

ufoDziner:
@mluessi Successfully tested with no wiring changes on both ports.  Great work!

fujin:
Confirmed working OK w/ TBS UNIFY PRO HV. Channel, band and power select all work fine. Great work. :+1:

fujin:
Ah, `Save and Exit` doesn't seem to work. Apply does?

flight/Modules/OnScreenDisplay/osd_menu.c
```diff
@@ -1671,6 +1816,195 @@ void sticklimits_menu(void)
...
+		}
+	}
+
+	y_pos += MENU_LINE_SPACING;
+	write_string("Save and Exit", MENU_LINE_X, y_pos, 0, 0, TEXT_VA_TOP, TEXT_HA_LEFT, 0, MENU_FONT);
+	if (current_state == FSM_STATE_VTX_SAVEEXIT) {
+		draw_selected_icon(MENU_LINE_X - 4, y_pos + 4);
+		if (current_event == FSM_EVENT_RIGHT) {
+			VTXSettingsSet(&settings);
+			UAVObjSave(VTXSettingsHandle(), 0);
```
fujin, line 1996: do we need to `draw_selected_icon(MENU_LINE_X - 4, y_pos + 4);` here to re-draw (exit)?

mluessi:
@fujin thanks, I had noticed that as well. The necessary state transition  was missing from the FSM. It works now (it did apply/save the changes before, just not exit).

mluessi:

flight/Modules/VTXConfig/VTXConfig.c
```diff
@@ -0,0 +1,317 @@
...
+	[VTXSETTINGS_POWER_500] = 500,
+	[VTXSETTINGS_POWER_800] = 800,
+};
+
+// ****************
+// Private variables
+
+static uint32_t vtxConfigPort;
+static bool module_enabled = false;
+static bool settings_updated = false;
```
mlyle: volatile

flight/Modules/VTXConfig/VTXConfig.c
```diff
@@ -0,0 +1,317 @@
...
+							// somethign went wrong, we need to re-try again
+							update_vtx = true;
+						}
+					}
+				}
+			}
+		}
+		else {
+			// Cycle through different baud rates. The spec says 4.8kbps, but it can deviate
+			if (baud_rate < TBS_MAX_BAUD) {
```
mlyle, line 307: Sure woulda been nice to bitbang this through the DAC instead... maybe next time around

flight/Modules/VTXConfig/VTXConfig.c
mluessi: fixed

flight/Modules/VTXConfig/VTXConfig.c
mluessi, line 307: For sure. I figured it would be faster / more reliable to get it working with an UART first but we should look into using a software based UART using the GPIO pin. The baudrate is very low, so it should be feasible.

mluessi:
Awesome, glad this made it into the release :)

PR #1068, opened 2016-06-28
RE1: Remove AltitudeHold

mluessi:
Don't compile-in this module as RE1 doesn't have a barometer and probably does something crazy if AH is enabled.

### DIFF
--- a/flight/targets/brainre1/fw/Makefile
+++ b/flight/targets/brainre1/fw/Makefile
@@ -43,7 +43,6 @@ MODULES += Telemetry
 
 OPTMODULES += GPS
 OPTMODULES += PathPlanner
-OPTMODULES += AltitudeHold
 OPTMODULES += VtolPathFollower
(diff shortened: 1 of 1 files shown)


flight/targets/brainre1/fw/Makefile
```diff
@@ -43,7 +43,7 @@ MODULES += Telemetry
 
 OPTMODULES += GPS
 OPTMODULES += PathPlanner
-OPTMODULES += AltitudeHold
+#OPTMODULES += AltitudeHold
```
fujin: I think we prefer to delete this rather than comment it (so that grep, etc is more effective going forward)

mluessi: I left it because I plan to enable it again for the next release. But I can delete it as well

mluessi:
@fujin fixed

PR #1069, opened 2016-06-28
RE1: Add FlashID

mluessi:
Adds function to read OPT (one-time programmable) data from SPI flash (works for some Spansion devices, not sure about others). The function is used to read a 16-byte random number from Flash, which we will use for flight controller registration. The STM32 ID could also be used for this but it does not seem to be random enough.

### DIFF
--- a/flight/PiOS/Common/pios_flash_jedec.c
+++ b/flight/PiOS/Common/pios_flash_jedec.c
@@ -56,6 +56,8 @@
 #define JEDEC_STATUS_SEC             0x40
 #define JEDEC_STATUS_SRP0            0x80
 
+#define JEDED_READ_OPT_DATA          0x4B
+
 enum pios_jedec_dev_magic {
 	PIOS_JEDEC_DEV_MAGIC = 0xcb55aa55,
 };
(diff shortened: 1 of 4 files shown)


flight/PiOS/Common/pios_flash_jedec.c
```diff
@@ -459,6 +461,49 @@ static int32_t PIOS_Flash_Jedec_ReadData(uintptr_t chip_id, uint32_t chip_offset
...
+/**
+ * @brief Read one-time progammable data from a location in flash memory
+ * @param[in] chip_id the opaque handle for the chip that this operation should be applied to
+ * @param[in] chip_offset Address within flash to write to
+ * @param[in] data Pointer to data to write from flash
+ * @param[in] len Length of data to write (max 256 bytes)
+ * @return Zero if success or error code
+ * @retval -1 Unable to claim SPI bus
+ */
+int32_t PIOS_Flash_Jedec_ReadOPTData(uintptr_t chip_id, uint32_t chip_offset, uint8_t *data, uint16_t len)
```
tracernz: OPT or OTP?

mluessi:
@tracernz typo fixed

PR #1072, opened 2016-06-29
sparky: work around PPM suck by forcing 1MHz

### LINKED ISSUE #1071: Actuator auto-resolution changes break PPM in on Sparky1
Thanks to poor timer pin assignments

mlyle:
(only when appropriate).

Fixes #1071

### DIFF
--- a/flight/targets/sparky/board-info/board_hw_defs.c
+++ b/flight/targets/sparky/board-info/board_hw_defs.c
@@ -985,6 +985,22 @@ struct pios_servo_cfg pios_servo_cfg = {
 	.num_channels = NELEMENTS(pios_tim_servoport_v02_pins),
 };
 
+struct pios_servo_cfg pios_servo_slow_cfg = {
+	.tim_oc_init = {
+		.TIM_OCMode = TIM_OCMode_PWM1,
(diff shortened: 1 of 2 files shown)


mlyle:
Confirmed on Sparky1-- Make PPM great again.

mluessi:
Code looks good. I wonder if a better solution would be to have an option to only use the first 5 PWM outputs; they use different timers and could run faster than the timer used for PPM. The downside would be that the user would need to configure this and so it becomes less user friendly.

tracernz:
So to be clear this has always been an issue when using output channels 6 & 10 with oneshot?

mlyle:
@tracernz when moving them to a rate other than 1MHz or using oneshot, yes..  With the new autodetect code, they are always set to a rate other than 1MHz.

tracernz:
Should ignore the bank in pios_servo if max_pulse for the bank is 0. And the time base would have already been set by the call in pios_board.c (with the fixed / 1 MHz config in board_hw_defs.c).

mlyle:
That would be an alternate approach, though it will still fail in some cases this will work.

PR #1073, opened 2016-06-29
pios_servo: also use "minimum" values to calc res

### LINKED ISSUE #1070: Actuator auto-resolution calculation assumes that servos are not reversed
And that ChannelMax is the true maximum value.

mlyle:
This is necessary to be robust to reversed axes (servos)

Fixes #1070


mlyle:
Confirmed to work on sparky1 with reversed output axes.

mlyle:
Auditing / thinking about this, this is far less critical – affects oneshot to digital servos etc but is really an edge case on ESCs.

flight/PiOS/Common/pios_servo.c
```diff
@@ -153,6 +153,7 @@ void PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_
 
 		timer_banks[bank].timer = chan->timer;
 		timer_banks[bank].max_pulse = MAX(timer_banks[bank].max_pulse, channel_max[i]);
+		timer_banks[bank].max_pulse = MAX(timer_banks[bank].max_pulse, channel_min[i]);
 	}
```
mluessi: Shouldn't it be 

		timer_banks[bank].max_pulse = MIN(timer_banks[bank].max_pulse, channel_max[i]);
		timer_banks[bank].max_pulse = MAX(timer_banks[bank].max_pulse, channel_min[i]);

flight/Modules/Actuator/actuator.c
```diff
@@ -735,7 +735,9 @@ static bool set_channel(uint8_t mixer_channel, float value)
 static void actuator_set_servo_mode(void)
 {
 	PIOS_Servo_SetMode(actuatorSettings.TimerUpdateFreq,
-			ACTUATORSETTINGS_TIMERUPDATEFREQ_NUMELEM, actuatorSettings.ChannelMax);
+			ACTUATORSETTINGS_TIMERUPDATEFREQ_NUMELEM,
+			actuatorSettings.ChannelMax,
+			actuatorSettings.ChannelMin);
 }
```
mluessi, line 741: This is nitpicking, but why not just use `MAX(actuatorSettings.ChannelMin, actuatorSettings.ChannelMax)` here instead of passing an additional variable?

mlyle, line 741: Each is an array of 10 values; so we could find the max of each element and store it in temporary storage...

mluessi, line 741: Oh, LOL, obviously I haven't had enough coffee..

mluessi:
Looks good, thanks!

PR #1076, opened 2016-07-01
GCS: Make RE1 Vsync threshold configurable

mluessi:
Exposes the threshold in the Hardware configuration tab and adds instruction for how to adjust it. The default seems to work well for most cameras, but for some cameras it may be necessary to adjust it.


tracernz:
It looks like you've changed a lot of unrelated stuff in the UI? Looking at the diff, probably better to put the LED limit in the UAVO so the UI will handle higher values set from the UAVO browser correctly.

mluessi:
@tracernz somehow QTCreator had made a bunch of seemingly unrelated changes. Made it cleaner.

PR #1089, opened 2016-07-06
GCS:Stabilization: Add camera tilt to advanced settings

### LINKED ISSUE #1085: Make StabilizationSettings.CameraTilt configurable in GCS
We need to make this configurable in the GCS. Forcing users to go to the UAVO browser to set the angle is not acceptable IMO.

tracernz:
Fixes #1085. Tested on sim_posix, incl. default button.


tracernz:
Added negative tilt, a note about sign to the field description, and actually use the limits in the UI.

mluessi:
@tracernz thanks. Just tested this and it works fine. Merging.

PS: Somehow having the camera tilt setting under stabilization is not natural as it can also be used for other things (e.g. OSD). Maybe for the next release we should think of putting this somewhere else together with the camera lens focal length (which will be needed for augmented reality type things).

PR #1092, opened 2016-07-08
FIX: PPM timer frequency on F4 targets

### LINKED ISSUE #1091: PPM input broken (preview-20160706)
For some reason PPM is no longer working (preview-20160706). So far confirmed on Brain and BrainRE1. Maybe it's related to the recent timer/servo changes?

mluessi:
Tested on Brain, BrainRE1, Sparky2. On Sparky2 the PPM input timer is shared with the brushes motor outputs, so it only works if those outputs are disabled.

Fixes #1091

### DIFF
--- a/flight/PiOS/Common/pios_servo.c
+++ b/flight/PiOS/Common/pios_servo.c
@@ -161,6 +161,10 @@ void PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_
 	TIM_TimeBaseStructure.TIM_ClockDivision = TIM_CKD_DIV1;
 	TIM_TimeBaseStructure.TIM_CounterMode = TIM_CounterMode_Up;
 	for (int i = 0; i < banks_found; i++) {
+		// Skip the bank if no outputs are configured
+		if (timer_banks[i].max_pulse == 0)
+			continue;
+
 		/* Calculate the maximum clock frequency for the timer */
 		uint32_t max_tim_clock = max_timer_clock(timer_banks[i].timer);
 		// check if valid timer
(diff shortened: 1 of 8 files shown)


mluessi:

flight/PiOS/Common/pios_servo.c
```diff
@@ -161,6 +161,10 @@ void PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_
 	TIM_TimeBaseStructure.TIM_ClockDivision = TIM_CKD_DIV1;
 	TIM_TimeBaseStructure.TIM_CounterMode = TIM_CounterMode_Up;
 	for (int i = 0; i < banks_found; i++) {
+		// Skip the bank if no outputs are configured
+		if (timer_banks[i].max_pulse == 0)
+			continue;
+
```
tracernz, line 167: 👍

mlyle, line 167: Because of this, should we also revert the previous sparky change?  (#1072)

Thinking about it, probably not...

What I want to do for quixote is to adapt the CC3D/Sparky things-- instead of a boolean, to specify a **timer** that should be pegged to 1MHz.  And we can make that conditional on the use of PPM/PWM.

mlyle:
Thanks especially for troubleshooting Sparky2.

flight/PiOS/Common/pios_servo.c
tracernz, line 167: 👍

tracernz:
So all STM32s share this frequency doubler thing for timers.. what is done differently in F4 targets?

PR #1093, opened 2016-07-08
gcs: slay the dragons :(

mlyle:
replace with a boring grid, to avoid mockery from the lame.


fujin:
farewell dragons, was nice knowing you. :dragon: :+1:

SystemsGuy:
Can we leave the blond who birthed 'em? :)

tracernz:
Lame 😢

PR #1094, opened 2016-07-09
gcs: map: set minimum zoom based on width

mlyle:
If we're really wide such that all the map tiles won't fit at zoom 2,
use zoom 3.


mlyle:
git command got outta hand.

PR #1096, opened 2016-07-09
logviewer and misc python enhancements

mlyle:
Works; not completely pretty / useful yet, and the code could use cleanup.  Still I think this is a good first cut.


mlyle:
logview

mlyle:
I need to work on build, etc, to get exes out.. but the basics:

```
01:01 < icee> OK in theory people can try that :P
01:03 < icee> in theory with anaconda on windows,almost all the components should 
              be there
01:03 < icee> just should be like conda install pyqtgraph      or something
01:03 < icee> and then be possible to do ./logview.py  from the python directory
```

(Need to try this out myself, too, for correct directions).

mlyle:
Right mouse button plus dragging zooms in and out.  When over an axis, just zooms that axis.  Left mouse button pans.

tracernz:
Got a sample log to play with?

tracernz:
On OSX:
- `brew install pyqt`
- `pip install pyqtgraph`

Probably some other stuff I already had required too.

jhitesma:
Whoops, left this on the wrong PR at first:

Seems like it's going to be a little more complex than that on windows:

> jhitesma@backroom MINGW64 /d/dronin/python ((bd06bf9...))
> $ conda install pyqtgraph
> Fetching package metadata .......
> Solving package specifications: ..........
> 
> # All requested packages already installed.
> 
> # packages in environment at C:\Anaconda2:
> 
> #
> pyqtgraph                 0.9.10                   py27_1
> 
> jhitesma@backroom MINGW64 /d/dronin/python ((bd06bf9...))
> $ ./logview.py
> Traceback (most recent call last):
>   File "./logview.py", line 3, in <module>
>     import pyqtgraph as pg
> ImportError: No module named pyqtgraph
> 
> jhitesma@backroom MINGW64 /d/dronin/python ((bd06bf9...))
> $
> 
> Though I did already have non-anaconda Python installed so I had to install anaconda first...wonder if the old python left something behind causing issues.  

I'll mess with this more tomorrow.

mlyle:
Yes, I suspect you have two python installations, and are not running the anaconda one.  try `which conda`   and then `which python` and compare the paths.

jhitesma:
Yep, was able to confirm a conflict.  Tried on another system that only has anaconda and appears to start fine.  Working on fixing python on the other computer, but now that one is having fan issues.  Good excuse to hook up the openlager and make some logs to test now.

PR #1097, opened 2016-07-09
POSIX: Fix flash file length check

tracernz:
fseek happily seeks past the end of the file so the old check failed (and ftell will happily return the same value even if it's past the end). Exposed by the change in flash size in #980. SEEK_END is not portable but I don't expect that will ever be an issue.

@mlyle Feel free to defer from Samsara if you want.


tracernz:
Sigh, indentation is wrong in flight/targets/simulation/fw/pios_board.c :(

PR #1098, opened 2016-07-10
gcs: rawhid: catch recursion

mlyle:
On OS X, process_pending_events can fire QT timers.  This caused us to
recurse and corrupt data structures.  So bail out in that case.


tracernz:
Seems to help things for me although very difficult to be certain.

Issue #1099, opened 2016-07-10
GCS:Welcome: Bad PNG


tracernz:
One of the PNGs brought in by @mlyle has a bad colour profile, causing a warning when GCS starts.

DTFUHF:
Addressed this issue with #1103 hopefully it works :)

tracernz:
Sorry @DTFUHF, misdiagnosed this and it was actually https://github.com/d-ronin/dRonin/commit/8dfa22edf8fe5de92ef8db8a7ffffef5a8876e84

PR #1103, opened 2016-07-10
DTFc - replace gcs PNGs with websafe sRGB

DTFUHF:
Re-exported these two PNGs for sRGB with no attached colour profile, to address #1099 

Though I didn't see this error in Windows or Linux hopefully this fixes it!


tracernz:
Really sorry, I misdiagnosed this, it was actually https://github.com/d-ronin/dRonin/commit/8dfa22edf8fe5de92ef8db8a7ffffef5a8876e84

PR #1106, opened 2016-07-11
RE1: Add IBus support

mluessi:
Somehow missed to add this to RE1. Would be nice if it could make it into the release.


tracernz:
👍 We always seem to do this with new targets.

PR #1107, opened 2016-07-11
pios: spi, irq,... lots of cleanup

### LINKED ISSUE #983: SPI DMA on F4 dubious
- It only kicks in for transfers over 128 bytes or where there's a callback.
- I think no one uses the callback interface.
- I believe only flash does transfers over 128 bytes.
- Even for this, since we block and wait the CPU is used during the operation.

So the only benefit we get right now is if there's a task switch or interrupt during a large flash write, that the data continues to get clocked out in the background.

However, it consumes DMA channels and resources.

### LINKED ISSUE #1001: Factor lots of code common in pios
Things that should be easy
- [x] pios_ppm
- [x] pios_pwm
- [x] pios_debug
- [x] pios_irq
- [ ] pios_usb_hid_pwr

Things slightly more than trivial but that should probably be done:
- [ ] pios_led

Things that are moderately difficult and should be considered or skipped:
- [x] pios_spi
- [ ] pios_iap
- [x] pios_wdg

mlyle:
- SPI: eliminate disused DMA callback arg
- SPI: remove DMA from F3 and F4.  wasn't used on F3, was no advantage as
  written on F4.
- SPI: check much more state / wait much more on F3... because what we were
  doing just wasn't healthy.
- SPI: completely unify F1, F3 and F4 implementations
- IRQ: unify
- debug: unify
- pwm/ppm: unify
- wdg: unify
- remove some conditional compilation things that are needed in any case
- factor STM32 specific (but not F1/F3/F4 specific) code into separate dir
- Change how makefiles find code to automatically take the "most specific" implementation

Fixes #1001
Fixes #983


mlyle:
Completely untested.

mluessi:
What's the motivation for removing DMA? For things like sending data over SPI to OpenLager, having DMA support will be useful.

mlyle:
As it's written, DMA is never used (transfers don't hit the threshold, no callers use the callback mechanism-- so even if they hit the threshold the CPU would busy-wait polling registers for it to be complete) and there appear to be obvious bugs in it.  There's no implementation on F3, it looks like there's multiple bugs in it on F4.

It prevents the unification of the different pios_spi implementations, uses memory and code space, etc.

Also, it uses up tons of DMA channels that we could instead use to support DMA on pios_com_usart which would be a more immediate benefit to openlager on more targets.

I'm not opposed to putting a DMA implementation back, if we can find a use case that it actually improves and we can build it correctly in a unified fashion across the stm targets.

mluessi:
Makes sense, it may be good to clean it up and then add things like DMA when we actually use it.

In general, I'm not sure if "if we can build it correctly in a unified fashion across the stm targets" should be a requirement, i.e., we don't want to be forced to only use features that are supported on F1, F3, and F4.

mlyle:
When the code is 90%+ common, it should be in one file and have the different bits conditionalized.

mlyle:
Rebased from conflict-- pios_mpu changes

mlyle:
Benchtested Sparky1, Revolution, and CC3D (one of each generation).  Happy.

flight/PiOS/STM32/pios_irq.c
```diff
@@ -49,6 +47,7 @@ static uint32_t prev_primask;
 */
 int32_t PIOS_IRQ_Disable(void)
 {
+	/* XXX TODO: This is racy */
```
tracernz, line 50: Nothing stopping this func. being re-ordered?

mlyle, line 50: It requires an atomic / analysis (because there could be an intervening task switch).  I just spotted it while I was here, but thinking of the correct fix takes some effort.

tracernz:
Looks reasonable on first pass. Not really a fan of vpath though.

mlyle:
I had it explicitly searching the paths as part of the template before, but it was wordier and probably not as quick as letting make do the work.

Main question is whether the concept is OK-- having "general" implementations and then allowing them to be overridden by arch-specific ones without any makefile surgery.  Hopefully this makes a lot more lines of makefile common so more work can be done to factor them out.

tracernz:
The reason vpath could be an issue is because Re1 will require a different pios_ws2812.c than other F4 boards.

mlyle:
If it's completely independent, that's fine--  vpath is defined to operate in the order the paths are specified. 

> If several vpath patterns match the dependency file's name, then make processes each matching vpath directive one by one, searching all the directories mentioned in each directive. make handles multiple vpath directives in the order in which they appear in the makefile; multiple directives with the same pattern are independent of each other.

So,

```
vpath % $(BOARD_ROOT_DIR)/$(BUILD_TYPE)

vpath % $(foreach d,$(ARCH_TYPES) Common,$(PIOS)/$(d))
```

will first look in the fw directory, then PiOS/STM32F4xx, then PiOS/STM32, then PiOS/Common.   So there could be two pios_ws2812.c's, one in brainre1/fw and then one in PiOS/STM32F4xx.  (perhaps in the future we should have brainre1/common, but that's for the future IMO).

In fact, if I had a concern, it'd be that the ease of this will make copying code and making a couple changes for a single target more tempting, because it's easier.

tracernz:
Need to update used DMA channels table on pios_board.h for all affected targets.

mlyle:
OK.  removed the SPI stuff from the DMA tables.  It was missing/lies on F1 anyways.

Brain's DMA table information is a lie, and I think RE1's may be (is QSPI DMA?)  But IMO that can wait as it doesn't relate to this changeset.

mlyle:
Tried to cram some other srtuff in here but pulled it out.

mluessi:
> Brain's DMA table information is a lie, and I think RE1's may be (is QSPI DMA?) But IMO that can wait as it doesn't relate to this changeset.

They both use DMA for the OSD SPI / QSPI, but the table is likely wrong/outdated. I can fix this in a separate PR.

tracernz:
CC3D doesn't boot firmware or upgrader with this. Quanton and Sparky1 booted okay.

mlyle:
Ick.  I will run that down.  I tried CC3D not too long ago, so it must have been a regression in the most recent commits.

tracernz:
Turns out the breakage is actually #1058.

tracernz:
This moves a long way in the right direction, nice work. 👍

PR #1109, opened 2016-07-12
Make: Fix breakpad install

tracernz:
Google moved the repo from SVN to Git. Also made it slightly better.

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -600,25 +600,34 @@ endif
 openssl_clean:
 	$(V1) [ ! -d "$(OPENSSL_DIR)" ] || $(RM) -rf $(OPENSSL_DIR)
 
-
-BREAKPAD_URL := https://google-breakpad.googlecode.com/svn/trunk
-BREAKPAD_REV := 1498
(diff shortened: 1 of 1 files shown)


tracernz:
Might be better to just have binaries in the tree, as recommended by Google.

tracernz:
Turns out this isn't even enough anymore, you need to download and install some Google tools to fetch the sources (including deps) before you can build successfully. I will prepare some dump_syms and minidump_stackwalk binaries and package them up. Hope to generate some symbols when the next release build runs @mlyle.

tracernz:
Linux and OSX binaries
[breakpad.zip](https://github.com/d-ronin/dRonin/files/358979/breakpad.zip) built from https://chromium.googlesource.com/breakpad/breakpad/+/dfd2da797999be085d55ae9430f2a434788652e3

tracernz:
Got full symbol info dumping on OSX with some mucking about. Need to make sure Linux is all good then put together a PR to make this whole thing super simple and Jenkins ready.

PR #1111, opened 2016-07-13
RE1: only allow valid trackmate IDs

mluessi:
It seems like there are some rules for valid IDs. This seems to generate IDs that are valid.

### DIFF
--- a/flight/Modules/OnScreenDisplay/osd_menu.c
+++ b/flight/Modules/OnScreenDisplay/osd_menu.c
@@ -1082,21 +1082,50 @@ const char * ir_protocol_names[HWBRAINRE1_IRPROTOCOL_MAXOPTVAL + 1] = {
 #define MAX_ID_ILAP 9999999
 #define MAX_ID_TRACKMATE 0xfff
 
-// check if this is a valid trackmate id
+// Get the next valid trackmate id
 // It seems like the ID has some weird requirements:
 // 1st nibble is 0, 2nd nibble is not 0, 1, 8, or F
 // 3rd and 4th nibbles are not 0 or F
-bool trackmate_valid(uint16_t trackmate_id)
+uint16_t next_valid_trackmateid(uint16_t trackmate_id, int16_t step)
 {
-	bool valid = ((trackmate_id & 0x0F00) >> 8) != 0x0;
-	valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0x1);
-	valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0x8);
-	valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0xF);
-	valid = valid && (((trackmate_id & 0x00F0) >> 4) != 0x0);
-	valid = valid && (((trackmate_id & 0x00F0) >> 4) != 0xF);
-	valid = valid && ((trackmate_id & 0x000F) != 0x0);
-	valid = valid && ((trackmate_id & 0x000F) != 0xF);
-	return valid;
+	uint8_t nibble;
+	while (1){
+		trackmate_id += step;
+		if (trackmate_id > MAX_ID_TRACKMATE) {
+			if (step > 0) {
+				trackmate_id = 0;
+			}
+			else {
+				trackmate_id = MAX_ID_TRACKMATE;
+			}
+		}
+		// Test 2nd nibble
+		nibble = (trackmate_id & 0x0F00) >> 8;
+		if ((nibble == 0x00) || (nibble == 0x01) || (nibble == 0x08) || (nibble == 0x0F)) {
+			// step through these quickly
+			if (step > 0) {
+				trackmate_id += 256;
+			}
+			else {
+				trackmate_id -= 256;
+			}
+			continue;
+		}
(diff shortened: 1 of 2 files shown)


ground/gcs/src/plugins/boards_brainfpv/brainre1configuration.cpp
```diff
@@ -105,7 +105,23 @@ void BrainRE1Configuration::generateILapID()
...
+    bool valid = false;
+    while (!valid) {
+        trackmate_id = generateRandomNumber(0xFFF);
+        // It seems like the ID has some weird requirements:
+        // 1st nibble is 0, 2nd nibble is not 0, 1, 8, or F
+        // 3rd and 4th nibbles are not 0 or F
+        valid = ((trackmate_id & 0x0F00) >> 8) != 0x0;
+        valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0x1);
+        valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0x8);
+        valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0xF);
```
mlyle, line 118: These reject 3/16 possibilities.

ground/gcs/src/plugins/boards_brainfpv/brainre1configuration.cpp
```diff
@@ -105,7 +105,23 @@ void BrainRE1Configuration::generateILapID()
...
+        valid = valid && (((trackmate_id & 0x00F0) >> 4) != 0x0);
+        valid = valid && (((trackmate_id & 0x00F0) >> 4) != 0xF);
```
mlyle, line 120: These reject 2/16 of remaining possibilities.

ground/gcs/src/plugins/boards_brainfpv/brainre1configuration.cpp
```diff
@@ -105,7 +105,23 @@ void BrainRE1Configuration::generateILapID()
...
+        valid = valid && ((trackmate_id & 0x000F) != 0x0);
+        valid = valid && ((trackmate_id & 0x000F) != 0xF);
```
mlyle, line 122: This reject 2/16 of remaining possibilities.  Chance of pass is:

13/16 \* 14/16 \* 14/16 = about .622.  So we should only expect to have to try a few times at most.    Seems OK to me.

flight/Modules/OnScreenDisplay/osd_menu.c
```diff
@@ -1204,19 +1221,25 @@ void brainre1_menu(void)
...
 		case FSM_STATE_RE1_IR_IDTRACKMATE:
 			{
+				uint16_t id = data.IRIDTrackmate;
 				int step = held_long ? 100 : 1;
 				if (current_event == FSM_EVENT_RIGHT) {
-					data.IRIDTrackmate += step;
-					if (data.IRIDTrackmate > MAX_ID_TRACKMATE) {
-						data.IRIDTrackmate = 0;
-					}
+					do {
```
mlyle: This seems less good.  Since like 0xx and 1xx are bad, we could have to do 512 + 16 iterations which seems to be getting excessive.

ground/gcs/src/plugins/boards_brainfpv/brainre1configuration.cpp
yamahabest: Wouldn't it be better to calculate the duplicate bitwise and's and bit shifts only once?

And putting the last two statements on top would also be better, since there are no bit shifts there.

flight/Modules/OnScreenDisplay/osd_menu.c
```diff
@@ -1080,7 +1080,24 @@ const char * ir_protocol_names[HWBRAINRE1_IRPROTOCOL_MAXOPTVAL + 1] = {
...
+bool trackmate_valid(uint16_t trackmate_id)
+{
+	bool valid = ((trackmate_id & 0x0F00) >> 8) != 0x0;
+	valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0x1);
+	valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0x8);
+	valid = valid && (((trackmate_id & 0x0F00) >> 8) != 0xF);
+	valid = valid && (((trackmate_id & 0x00F0) >> 4) != 0x0);
+	valid = valid && (((trackmate_id & 0x00F0) >> 4) != 0xF);
+	valid = valid && ((trackmate_id & 0x000F) != 0x0);
+	valid = valid && ((trackmate_id & 0x000F) != 0xF);
```
yamahabest: Wouldn't it be better to calculate the duplicate bitwise and's and bit shifts only once?

And putting the last two statements on top would also be better, since there are no bit shifts in there.

mlyle: The compiler should do common subexpression elimination and short circuit evaluation, though the likely case is all of these are true and thus all need evaluation.

mluessi:
@mlyle optimized the OSD menu code. It now will do a max of 16 iterations (they are now also contained in a single function).

mlyle:
LGTM

PR #1117, opened 2016-07-15
flight: osd: properly display reset reason

mlyle:
We immediately returned before which prevented it from working.

### DIFF
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -737,10 +737,12 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
 	int32_t len = AlarmString(&alarm, buf + pos, sizeof(buf) - 1 - pos,
 			blink, &state);
 
-	len += pos;
-
 	if (len > 0) {
-		buf[len] = '\0';
+		pos += len;
+	}
+
+	if (pos > 0) {
+		buf[pos] = '\0';
 		write_string(buf, x, y, xs, ys, va, ha, flags, font);
 	}
 }
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -731,11 +731,13 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
 		buf[strlen(boot_reason)] = '\0';
 		pos = strlen(boot_reason);
 		buf[pos++] = ' ';
-		return;
 	}
 
 	uint8_t state;
-	int32_t len = AlarmString(&alarm, buf, sizeof(buf) - 1, blink, &state);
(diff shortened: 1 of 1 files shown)


flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -731,11 +731,13 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
...
 		buf[pos++] = ' ';
-		return;
 	}
 
 	uint8_t state;
-	int32_t len = AlarmString(&alarm, buf, sizeof(buf) - 1, blink, &state);
+	int32_t len = AlarmString(&alarm, buf + pos, sizeof(buf) - 1 - pos,
+			blink, &state);
+
+	len += pos;
```
tracernz: Would prefer `pos += len`, or `pos += AlarmString(...` instead of len, and then write a null char at pos but I suppose this will do. :-P

tracernz:
Looks good.

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -731,14 +731,18 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
...
 		buf[pos++] = ' ';
-		return;
 	}
 
 	uint8_t state;
-	int32_t len = AlarmString(&alarm, buf, sizeof(buf) - 1, blink, &state);
+	int32_t len = AlarmString(&alarm, buf + pos, sizeof(buf) - 1 - pos,
+			blink, &state);
 
 	if (len > 0) {
```
dustin, line 740: len won't ever be < 0, so it doesn't seem all that useful to have this condition here.

mlyle, line 740: Just in case it ever returns error in the future, I figured.  This is my second variant of this..

tracernz, line 740: Looks like it will be 0 if there are no warning/error/critical alarms? https://github.com/d-ronin/dRonin/blob/080150a5b271fcfb3fe36aa1e20f92fc0d956aca/flight/Libraries/alarms.c#L329-L396

-e- derp, you are suggesting just unconditionally adding len. Fair comment.

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -731,14 +731,18 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
 		buf[strlen(boot_reason)] = '\0';
```
tracernz: strncpy won't put a null char at the end if AlarmBootReason returns a string longer than buf so this isn't super safe.

tracernz, line 731: Should only copy max of sizeof(buf)-1 in the above or this could write past the end of the buffer by 1 byte. And put the null char at the end of the buffer, then pos = strlen(buf).

tracernz:
Looked over this pretty well. The existing code has potential for a future bug (#1120), but I'm happy this is safe for now.

PR #1119, opened 2016-07-15
python-api: parse xml files iteratively

### LINKED ISSUE #1118: Python API uavo dependency management (parents) problematic
@ufoDziner experienced this:

AttributeError: 'NoneType' object has no attribute 'ENUMR_ChannelGroups'".

The approach used to load 'shared' objects first is ineffective with the ChannelGroups shared thing:

```
        for file_name in glob.glob(os.path.join(path, '*.xml')):
            if op.basename(file_name).lower().find('shared') > 0:
                file_names.insert(0, file_name)
            else:
                file_names.append(file_name)
```

mlyle:
This change properly handles parent dependencies between arbitrary
uavos.

Fixes #1118

### DIFF
--- a/python/dronin/uavo.py
+++ b/python/dronin/uavo.py
@@ -196,7 +196,7 @@ def make_class(collection, xml_file, update_globals=True):
 
     from lxml import etree
 
-    tree = etree.parse(xml_file)
+    tree = etree.fromstring(xml_file)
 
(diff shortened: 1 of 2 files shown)


mlyle:
- tested from inside source tree
- tested from outside source tree so it had to get stuff from web service.

PR #1124, opened 2016-07-17
gcs: send next ancestry information to cloud service

mlyle:
The cloud still needs to use this info.

When there are PR builds, or builds from developer branches... the githashes are not preserved and are not meaningful in the long term.  This creates a problem for developer upgrade, because the logfs importer in the cloud doesn't know how to interpret these hashes.

As a result, save 8 more bytes of the firmware descriptor to store information about the last common ancestor between this build and next.  If the cloud cannot interpret the githash in the upgrade, it should fall back to using the common ancestor-- which may not be an exact match for the UAVOs but should be pretty close.

Connects to #848.


mlyle:
@dustin FYI

tracernz:
This introduces a dependency on the naming of remotes on the local machine right? Probably need to document this in developer docs.

mlyle:
It's a dependency, but a non-fatal one.  Developer docs on readme.io recommend this convention.  One thing to check is whether jenkins builds do the right thing for this, since it pulls refs oddly.

mlyle:

dustin:
I'm not entirely sure what I need to do here.

mlyle:
@dustin I should change the importer reference code in our tree first, probably.

But basically, we need to use the provided "ancestor" form entry as an alternate githash for upgrading-- so if we don't know about the provided one, we try the ancestor.  Ancestor is the merge-base between the built branch and upstream/next, so it should make everything happy / be a sane fallback.

mlyle:
Artifacts above list git hash as d161f5b8 and ancestor hash of e81706808a0ef751

mlyle:
Which looks correct (following hyperlinks above).

mlyle:
Couple other notes:
- This will send undefined stuff-- probably nulls or 20202020202020 ... but could be other things.. as the "user data" portion of the tlfw was in the bytes that i'm using for this and I didn't want to burden us with any kind of "magic" ... for old builds.  But that should be OK if we can't find the hash.
- This is a really big deal for me personally to get working.  I have lots of weird vehicles on weird versions with difficult-to-recreate configs, and often I have trouble figuring out the right uav files for the right things (and it's a big hassle to re-export config for every "last" change.   So it will be nice nice nice to stop losing configurations and stop doing all this bookkeeping.

tracernz:
Yeah, this should be pretty nice.

Issue #1123, opened 2016-07-16
BrainFPV in-flight reset


ufoDziner:
- Seen in version (release version or githash): Build 455
- On flight controller (if applicable): BrainFPV

Issue details:
My son and I have experienced in-flight resets of the BrainFPV FC.  Initially I thought that it was a power issue to the FC, but it it happening randomly across 3 different quads.  Build 455 was used to see what the initial startup message is after an in-flight reset.  Screenshot attached.

![brain-reboot](https://cloud.githubusercontent.com/assets/7434287/16896857/dc51fd56-4b54-11e6-81e1-de7dc6e94631.png)

ufoDziner:
BOOT: UNDEFINED appears to be the normal startup message as well.

dustin:
Just to be extra clear, are you able to see this in the UAVO as well?  I
don't know why it'd be wrong, but if the UAVO disagrees, then it's a
different bug.

ufoDziner:
Requested screenshots

![stackremaining1](https://cloud.githubusercontent.com/assets/7434287/16897859/93ae7776-4b74-11e6-9bb7-a17e156f1319.png)
![stackremaining2](https://cloud.githubusercontent.com/assets/7434287/16897858/93ad25ce-4b74-11e6-845e-82035dd80dcb.png)

ufoDziner:
@dustin Where would I check for a uavo?

dustin:
I can't quite remember which one it is (on my phone), but it's system
something and has a boot reason in it.

mlyle:
SystemSettings.RebootCause I think.  Doesn't need a flight -- just is the startup reason.

ufoDziner:
Normal startup screenshot.

![rebootcause1](https://cloud.githubusercontent.com/assets/7434287/16898418/b49d4fb0-4b8e-11e6-9bfa-72fe350feba3.png)

mlyle:
Not awesome :/

fujin:
#1126 should make it so we can see the proper reason now yea? @ufoDziner if you could upgrade to nightly tomorrow :sunglasses:

mlyle:
fw_sparky2, static stack analysis (with my dodgy tools) on next:

```
>$ stabilizationTask                   456      312        7
```

vs tanto:

```
>$ stabilizationTask                   432      304        7
```

So stabilizationtask stack usage has increased by 24 bytes.

[...]

PR #1125, opened 2016-07-17
Bump stacks for Stabilization/VTXConfig on F4

fujin:
Per #1123. I flew this for 4 batteries (20 minutes or so of airtime) and no boot:undefined/reset so far.


ufoDziner:
My son flew 5 packs and I flew 4 on two different quads using build 459 that includes this PR with no issues.  Funny thing is that I broke mine after the 4th flight, so I took out the backup quad that was still on 455 without this PR.  Halfway through the first pack it rebooted.  So, I'm pretty confident that the issue is fixed.

mlyle:
Well, we'll hang out a bit longer looking for unexplained crashes.

One thign that would be pretty clear evidence is if we saw a quad with less than 68 bytes of stack in the stabilization module (since we increased stack by 60 bytes) after this change.

PR #1126, opened 2016-07-17
brain/brainre1: make RCC flags like other targets

mlyle:
This was preventing reset reasons from properly displaying-- because
the reset reason flags were cleared before the alarms library was
initialized.


tracernz:
Looks good. Happy to merge when build succeeds.

ufoDziner:
Getting an actual reason now.  Thanks!

PR #1128, opened 2016-07-17
python: fixups

### LINKED ISSUE #1127: Output #1 set to 0/0/0 on update
- Seen in version (release version or githash): 3e21fdf7
- On flight controller (if applicable): Brain
  Issue details:
  When doing an auto-update from e8170680 to 3e21fdf7, the values of Output #1 are set to 0/0/0 instead of 125/131/250 as they were prior to the update.  Outputs 2,3, and 4 are correct.  Screenshot and uav file attached.

output1_missing

mlyle:
ipython:
- it needs the `__name__` attribute in the namespace to survive
- changing the interactive shell embed import removes a deprecation warning.

uavo.py:
- join attributes when we encounter lists, not just tuples.  fixes #1127

### DIFF
--- a/python/dronin/uavo.py
+++ b/python/dronin/uavo.py
@@ -144,7 +144,7 @@ class UAVTupleClass():
 
             field_value = raw_dict[field_name]
 
-            if isinstance(field_value, tuple):
+            if isinstance(field_value, tuple) or isinstance(field_value, list):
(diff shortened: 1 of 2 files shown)


mlyle:
@dustin  This one would need to get picked up in the cloud, too

PR #1129, opened 2016-07-17
flight: battery: increase stack size

mlyle:
battery is also very low, and stack analysis confirms usage is increased
in this rev.

connects to #1123


fujin:
LGTM :+1:

PR #1134, opened 2016-07-18
python-api: make work on python 3, packaging stuff...

mlyle:


mlyle:
Minimally tested this-- with logs-- but need to try telemetry sessions too.

mlyle:
Now it distributes the bins, too.

python/setup.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+        # Pick your license as you wish (should match "license" above)
+        'License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)',
+
+        # Specify the Python versions you support here. In particular, ensure
+        # that you indicate whether you support Python 2, Python 3 or both.
+        'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.6',
+        'Programming Language :: Python :: 2.7'
```
tracernz: No python 3?

tracernz:
`pip -U install dronin`? :stuck_out_tongue:  :100:

python/setup.py
mlyle: Will fix for future uploads :)

mlyle:
or dronin[all] to bring in dependencies (Except PyQt4).

tracernz:
Do you want anything specific tested with this? I should be able to tell pip to install this from disk right?

mlyle:
it's already pushed to pypi, so you can just do pip install dronin[all]

I'm happy with this, though I have not tested the full suite of capabilities.  In particular I haven't tested serial FC comms.  But there's going to be further active dev and this is a worthwhile step along the way (and appropriately publishes the code of what we have on pypi)

mlyle:
And now fixed up to work with serial and network telemetry well.

PR #1137, opened 2016-07-22
python-api: support taking a list of githashes to fall back

mlyle:
Connects to #848.

### DIFF
--- a/python/dronin-logfsimport
+++ b/python/dronin-logfsimport
@@ -12,10 +12,10 @@ def main():
 
     arg = parser.parse_args()
 
-    githash = arg.githash
+    githashes = arg.githash.split(':')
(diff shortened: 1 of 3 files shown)


python/dronin-logfsimport
```diff
@@ -12,10 +12,10 @@ def main():
```
tracernz, line 12: Should update the help on the line above, esp. with regard to the separator since it's not ','.

(why does Github restrict which lines you can comment on now? 👎 )

mlyle, line 12: I thought about it but I thought the line would get too long... plus this bit of functionality isn't intended for end users-- it's just there for the cloud (and the split was added here so I could try it out).

PR #1138, opened 2016-07-22
GCS: Fix MSVC regressions

tracernz:
Can build with MSVC 2015 Community Edition and Qt5.6.1 with MSVC2015 kit. It even runs. I changed to MSVC 2015 because that is the most easily available community edition.

### DIFF
--- a/ground/gcs/src/libs/utils/winutils.cpp
+++ b/ground/gcs/src/libs/utils/winutils.cpp
@@ -94,7 +94,9 @@ QTCREATOR_UTILS_EXPORT QString winGetDLLVersion(WinDLLVersionType t,
     }
     VS_FIXEDFILEINFO  *versionInfo;
     UINT len = 0;
-    if (!(*verQueryValueW)(data, TEXT("\\"), &versionInfo, &len)) {
+    WCHAR tmp1[] = TEXT("\\");
+    LPWSTR tmp = tmp1;
+    if (!(*verQueryValueW)(data, tmp, &versionInfo, &len)) {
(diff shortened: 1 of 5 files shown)


ground/gcs/src/libs/utils/winutils.cpp
```diff
@@ -94,7 +94,9 @@ QTCREATOR_UTILS_EXPORT QString winGetDLLVersion(WinDLLVersionType t,
     }
     VS_FIXEDFILEINFO  *versionInfo;
     UINT len = 0;
-    if (!(*verQueryValueW)(data, TEXT("\\"), &versionInfo, &len)) {
+    WCHAR tmp1[] = TEXT("\\");
+    LPWSTR tmp = tmp1;
+    if (!(*verQueryValueW)(data, tmp, &versionInfo, &len)) {
```
tracernz, line 99: Hacky hack hack. MSVC really didn't like this for reasons I'm not sure of.

ground/uavobjgenerator/uavobjgenerator.pro
```diff
@@ -7,10 +7,12 @@ macx {
 
 cache()
 
-# Unfortunately this is ineffective on OSX, due to
-# https://bugreports.qt.io/browse/QTBUG-39417
-# Should use it once upstream defect resolved
-QMAKE_CXX=$$(CCACHE_BIN) g++
+*-g++* {
+    # Unfortunately this is ineffective on OSX, due to
```
mlyle: Can remove this comment too if you want (since ccache is now used on OSX post-qt upgrades and other fixes)

tracernz: oh yep 👍

Issue #1140, opened 2016-07-22
CCM-sram: zero the parts that should be zeroed at startup.


mlyle:
F3 and F4

It comes up zero, so it's "OK", but I don't like having faith in things like that.

PR #1141, opened 2016-07-23
RE1: Remove CCM section in Linker Definition

### LINKED ISSUE #1139: BrainRE1 does not init BSS stuff in CCM-SRAM properly!
This has triggered a bug on next with the new manualcontrol stuff, but---- probably triggers lots of undefined behavior elsewhere.

Putting it into Samsara.  We will have to hotfix.  Even though ill effects have not been observed on Samsara, tehy're surely there.

fujin:
References #1140. confirmed @ 565a761

Edit by mpl: Fixes #1139

### DIFF
--- a/flight/targets/brainre1/fw/sections_chibios.ld
+++ b/flight/targets/brainre1/fw/sections_chibios.ld
@@ -139,27 +139,6 @@ SECTIONS
         __main_thread_stack_end__ = .;
     } > ram
 
-    .ccm :
-    {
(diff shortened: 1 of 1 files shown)


mluessi:
Thanks for fixing this and sorry for the messup. Not sure why I didn't delete that section. RE1 needed a separate linker file exactly because the F446 doesn't have CCM.

mlyle:
No worries, stuff happens..  We're going to have to hotfix this though (unknown what other consequences this might have)

PR #1142, opened 2016-07-23
bash_profile: fixup to point to 5.6.0

mlyle:


mlyle:
Looks like htere's other stuff wrong with the move to 5.6.

mlyle:
Will look at this in detail when deploying new build server

tracernz:
Would be nice to be able to switch between MinGW and MSVC build kits easily. Probably load the MinGW one by default from the bash profile.

PR #1146, opened 2016-07-26
flyingf3: remove target

mlyle:
Only 5 have ever been connected.  No idea how much works.. Bye.


tracernz:
We should really collect these old drivers somewhere.

PR #585, opened 2016-01-31
Centralize stabilization, flight switch, and flight mode options in shared UAVO

ekaszubski:
Adding flight modes and stabilization modes is a pain, as it's necessary to ensure several redundant/identical fields in UAVOs are up-to-date. This PR centralizes stabilization and flight mode enums into a shared UAVO which is referenced by the other duplicate UAVOs.

Note: depends on #569 and #581

edit by MPL: fixes #640 

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/585)

PR #1147, opened 2016-07-26
Added shared defs for flight modes

### LINKED ISSUE #640: Flight modes and stabilization modes should be inherited
from a common object, using the UAVO parent mechanism.

mlyle:
Supersedes  / rescues #585 

Fixes #640 

Prepares for work like #581


tracernz:
This is a good improvement.

PR #1152, opened 2016-07-27
Revert Merge pull request #1058 from jihlein/mpuAlignment

tracernz:
Flight: Move MPU6000 Targets to pios_mpu.h/pios_mpu.c (reverted from commit f3e1986f6e45d643f3ed75ec113a5583dc52d983)

See #1149

PR #1155, opened 2016-07-27
Fix pios_mpu on CC3D

### LINKED ISSUE #1149: CC3D Fails To Boot: SPI not Initialised
Regression from #1058. Seems to be something else stopping it too.

mlyle:
In theory: Fixes #1149 --- initializes SPI and doesn't use semaphores during init.

Still need to confirm this is actually OK-- will test in the next couple of hours.

Thank you, @tracernz, for figuring out how we broke everything :D


mlyle:
uh, why didn't you notice this, jenkins?

jenkins, test this please

mlyle:
Doesn't work at this point, reviewing.

mlyle:
OK, this is unfortunate.  For undetermined reasons interrupt delivery seems not to be working at startup-- but is just fine subsequently.

This code neuters the interrupt-validity check on CC3D, where it wasn't getting checked anyways before in the MPU6000 code.  It works fine on CC3D.

On my test CC3D, it seems that 'zero gyros on arm' is _mandatory_-- yaw bias changes significantly after about 2-3 seconds of uptime.

tracernz:
I have an almost identical branch sitting on my PC at home with the interrupt counter. Like yours it doesn't work. :-(

mlyle:
Happily bench tested with Quanton and CC3D.  Interrupt counter works on others-- interrupts must be disabled/masked somehow during this time on CC3D so there's code here to bypass it.

OTOH, we didn't test this before, so it's no _worse_.

mlyle:
@tracernz can you tolerate this approach or should we merge #1152?

tracernz:
I'd rather something that is ifdefed on CC3D so it doesn't affect other targets and can be cleanly removed when CC3D is dropped without imposing test requirements for other targets. This is definitely better than #1152. I would like to do a little more work this evening to understand why the interrupt doesn't work on the CC3D during init if possible.

mlyle:
Root cause is-- bootloader disables interrupts on reset.   FreeRTOS enables them with `cpsie i` instruction in prvPortStartFirstTask.

(I think enabling earlier could be problematic, because all kinds of rtos interrupts could get triggered before they're expectin' it.)

tracernz:
This gets the CC3D booting with #1156 applied on top. MPU never actually gives us any attitude data though.

PR #1156, opened 2016-07-29
PiOS: Fix F1 SPI after #1107

tracernz:
The old F1 SPI driver enabled these clocks, new one doesn't (F3/F4 targets enable pretty much all peripheral clocks in pios_sys, we should probably discuss that more).


mlyle:
LGTM-- @tracernz may merge if he wants (with 1155 & 1157).

PR #1158, opened 2016-07-29
pios_rcvr: handle the case when no receivers are configured

mlyle:
This is the case on RE1 with defaults.

Connects to #1139.


mlyle:
Confirmed to fix issue.  RE1 starts with default config.

mlyle:

tracernz:
Indentation is wrong in this file, but this PR matches existing so I guess it'll be okay.

PR #1162, opened 2016-07-31
PiOS/Lux: Tidy up some things while debugging

tracernz:
Some things I noticed while debugging.

### DIFF
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -221,6 +221,7 @@ uintptr_t pios_com_debug_id;
  * @param[in] code Number of blinks to do in a row
  */
 void PIOS_HAL_Panic(uint32_t led_id, enum pios_hal_panic code) {
+#if defined(PIOS_INCLUDE_LED)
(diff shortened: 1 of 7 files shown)


flight/PiOS/Common/pios_hal.c
```diff
@@ -230,14 +231,14 @@ void PIOS_HAL_Panic(uint32_t led_id, enum pios_hal_panic code) {
 			PIOS_LED_Toggle(led_id);
 			PIOS_DELAY_WaitmS(175);
 		}
-		PIOS_DELAY_WaitmS(200);
+		PIOS_DELAY_WaitmS(175);
```
tracernz, line 234: Lux was rebooting by watchdog after the first set of flashes..

tracernz, line 234: Ahhh, LSI clock (which IWDG uses) is 32 kHz on F4, and 40 kHz on F1/F3, PIOS_WDG assumes they're all 32 kHz according to comments in it. :/ Watchdog is configured for 250 ms, and 200 ms/250 ms \* 40 kHz/32 kHz = 1 hence the watchdog.

PR #1165, opened 2016-08-02
logging: log everything at start, not just settings

mlyle:
This is necessary to avoid ambiguity in initial state--- especiallly with fullbore logging in order to know initial flight modes etc.

I'm working on the log viewer-- in order to notice flight 'events' relevant to analysis, like flight mode switch changes and arming/disarming.  But in logfiles provided by @ufoDziner, I noticed that the initial flight mode and arming event was not there.


tracernz:
You are still going to miss the arming event if the logging starts on arming right?

mlyle:
THere will be at least a flightstats message that says "armed" and the initial flight mode right at the beginning, and the disarming event shows up at the end properly from what I can see (though this is probably racy)

PR #1168, opened 2016-08-03
python logviewer nearing completion

mlyle:
Builds upon / requires #1161

~~Imports pyqtgraph into our codebase since they've not shipped in forever (so we can get their new stuff)~~

Gets the log viewer much more usable, IMO.  
- Now can plot non-float types
- Now properly handle subelement values (by index)
- Detect important flight events

Various other good stuff..  Remaining todo (in subsequent PRs)--
- PyInstaller configuration
- pyqtgraph updates from upstream
- Pip/pypi/pyqt5 improvements (depend on the publically available pyqt5 wheels)
- Allow plots of multiple series.

### DIFF
--- a/python/dronin/logviewer/logviewer.py
+++ b/python/dronin/logviewer/logviewer.py
@@ -46,6 +46,8 @@ def add_plot_area(data_series, dock_name, axis_label, legend=False, **kwargs):
 
     last_plot = dock
 
+    return pw
+
 def plot_vs_time(obj_name, fields):
     if not isinstance(fields, list):
         fields = [fields]
@@ -66,7 +68,7 @@ def plot_vs_time(obj_name, fields):
     win_num += 1
     dock_name = "TimeSeries%d" % (win_num)
 
-    add_plot_area(data_series, dock_name, left_axis_label, legend=legend)
+    return add_plot_area(data_series, dock_name, left_axis_label, legend=legend)
 
 def clear_plots(skip=None):
     containers, docks = area.findAll()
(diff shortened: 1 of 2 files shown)


ebenolson:
Rather than copying in all that code from pyqtgraph, why not use a `requirements.txt` that installs from github?

mlyle:
@ebenolson Thanks-- I was not familiar with this-- that's a neat trick and I'll look into it.  It's a bit unfortunate still in that (I think) we'd require the end-user to have an "unstable" pyqtgraph systemwide or use a virtualenv.

mlyle:
For now, yanked the controversial code ee433d4 and 8f7ab58; the remainder of stuff is useful on its own

tracernz:
Would be really nice to avoid bringing in an entire library if possible.

mlyle:
Yes, that's why I pulled it.

OTOH getting this "right" lets us do everything through pypi (pyqt5 has wheels for installation unlike pyqt4) and makes other people able to actually use this code.

mlyle:
Rebased post #1161

mlyle:
@ebenolson ---- it doesn't look like I can make something in pypi depend upon / pull in a github repository.   It looks like pip/pypi don't listen to/search out requirements.txt and just rely on the info in setup.py.  There's another mechanism, dependency_links, but it A) looks like it's deprecated, and B) not listened to by default.

So it looks like my main options would be either to A) fork pyqtgraph into another repository, and adjust setup.py to publish to pypi in another namespace and under a different package name... (suggested by @tracernz) or B) to import the library here as I've done.  Do you have any suggestions about how to proceed?  Thanks!

mlyle:
Just to clarify, the goal is to make it as easy as `pip install dronin[all]` to get these tools.

ebenolson:
You're right, python packaging is a kind of a mess and what I suggested would require an additional `pip install -r requirements.txt` step. It would be possible to add a setuptools hook to do that automatically with `os.system` but it's a bit hackish. 

Perhaps forking pyqtgraph and putting the new version up on pypi is the best solution - it might even help some other projects, it looks like people have been asking for an update for a while: https://github.com/pyqtgraph/pyqtgraph/issues/238.

mlyle:
@ebenolson Thanks-- I think that's the plan

PR #1172, opened 2016-08-05
pios_posix: support for SPI

mlyle:
Verified to work with MS5611 on flyingpi. 

Also some small cleanups to stm32 pios_spi, of which only ea182ca has some risk (but I think is a good change).

### DIFF
--- a/flight/PiOS/posix/pios_sys.c
+++ b/flight/PiOS/posix/pios_sys.c
@@ -54,26 +54,86 @@
 
 #include "pios_fileout_priv.h"
 #include "pios_com_priv.h"
+#include "pios_spi_posix_priv.h"
+#include "pios_ms5611_priv.h"
+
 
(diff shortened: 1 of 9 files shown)


flight/PiOS/posix/pios_sys.c
```diff
@@ -54,26 +54,86 @@
 
 #include "pios_fileout_priv.h"
 #include "pios_com_priv.h"
+#include "pios_spi_posix_priv.h"
+#include "pios_ms5611_priv.h"
```
mluessi, line 58: It seems odd to have this included here.

mlyle, line 58: Command line processing (and a lot of stuff that would be usually done in pios_board) happens here, so the same includes that pios_board.c would need to init the hardware are required.

tracernz:
Looks good. Note: STM32 changes have been tested on Quanton as per IRC.

PR #1174, opened 2016-08-05
preliminary bmx055 driver support

mlyle:
Depends on #1172 

Some associated bug fixes to pios_delay and pios_spi in posix.

No mag support yet.  No interrupt support.

Thanks to @yamahabest for significant help putting together the headers for the device's register map.

https://www.youtube.com/watch?v=_TO-OOIRIAA

### DIFF
--- a/flight/PiOS/Common/pios_bmx055.c
+++ b/flight/PiOS/Common/pios_bmx055.c
@@ -1,0 +1,589 @@
+/**
+ ******************************************************************************
+ * @file       pios_bmx055.c
+ * @author     dRonin, http://dRonin.org/, Copyright (C) 2016
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS_BMX055
+ * @{
+ * @brief Driver for Bosch BMX055 IMU Sensor
+ *****************************************************************************/
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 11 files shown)


flight/PiOS/Common/pios_bmx055.c
```diff
@@ -0,0 +1,589 @@
...
+	uint8_t c;
+
+	int32_t ret = PIOS_BMX_ReadReg(slave, address, &c);
+
+	if (ret) {
+		return ret;
+	}
+
+	if (c != expect) {
+		printf("BMX: Assertion failed: *(%d:%02x) == %02x (expect %02x)\n",
```
tracernz: Probs shouldn't have this in here, or should make it conditional on SIM, with PIOS_Assert when not compiling for sim. This driver can be used in regular flight code too.

mlyle: So what I was planning was-- to leave this in, in the meantime.  Ultimately to work on logging infrastructure to supplant flight's debug console, and to send it to syslog(3) on simulator (this will ensure it doesn't block on console output too).

I can comment out the printf's or remove entirely in the meantime if you want, but this is placeholder for messages i want to have ultimately.

mlyle:
Rebased and force push'd.

flight/PiOS/Common/pios_bmx055.c
tracernz: What does printf currently do on flight?

mlyle: Right now there's DEBUG_PRINTF defined in every pios_board, which IMO needs to be replaced with something unified and smarter.

mlyle: (I think printf itself will fail on putchar being missing)

mlyle:
OK, moved these to DEBUG_PRINTF to be consistent with other targets.  Works.

tracernz:
I think that's a good solution for now. 👍

PR #1175, opened 2016-08-06
GCS:Modules: Fix logging UI from #1165

tracernz:
Was broked by missing field.


mlyle:
Ah, I forgot about this.  Will merge on Jenkins happiness.

PR #1188, opened 2016-08-09
PID deadband code.

glowtape:
This runs the error in the PID controller through a cubic deadband function and should influence very low rate flight dynamics, specifically lingering PID responses ending harder maneuvers. The widths and slopes of the roll, pitch and yaw deadbands can be configured via two fields in the StabilizationSettings UAVO (DeadbandWidth and DeadbandSlope, each three elements). The width is clamped in code to 0-20°/s, as is the slope to 0-100%. Zero width is a disabled deadband on the axis.

F1 targets have been tentatively excluded. The UAVO additions rack up to 18 bytes, the necessary variables in memory would add up another 88 bytes to that. Up to debate I guess, with the RAM scarcity and all that.

Currently the deadbands, when non-zero width, are applied to following axis modes:
- Rate
- Acro Plus
- Attitude (inner loop)
- Axis lock (final rate desired, not the axis accumulator stuff)
- Horizon

Some values to start with would be 10°/s and 50-60% slope for roll and pitch, and _maybe_ 5°/s and 50% for yaw. Decent starting values will probably emerge during testing and tinkering.

### DIFF
--- a/flight/Libraries/math/pid.c
+++ b/flight/Libraries/math/pid.c
@@ -47,7 +47,7 @@ static float deriv_gamma = 1.0;
  * @param[in] dT  The time step
  * @returns Output the computed controller value
  */
-float pid_apply(struct pid *pid, float err, float dT)
+float pid_apply(struct pid *pid, const float err, float dT)
 {	
 	if (pid->i == 0) {
 		// If Ki is zero, do not change the integrator. We do not reset to zero
@@ -82,7 +82,7 @@ float pid_apply(struct pid *pid, float err, float dT)
  * @Note based on "Feedback Systems" by Astrom and Murray, PID control
  *  chapter.
  */
-float pid_apply_antiwindup(struct pid *pid, float err,
+float pid_apply_antiwindup(struct pid *pid, const float err,
 	float min_bound, float max_bound, float dT)
 {	
 	if (pid->i == 0) {
@@ -136,7 +136,6 @@ float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const f
 	float err = setpoint - measured;
 	float err_d = (deriv_gamma * setpoint - measured);
 
-#ifndef SMALLF1
 	if(deadband && deadband->width > 0)
 	{
 		err = cubic_deadband(err, deadband->width, deadband->slope, deadband->cubic_weight,
@@ -144,7 +143,6 @@ float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const f
 		err_d = cubic_deadband(err_d, deadband->width, deadband->slope, deadband->cubic_weight,
 			deadband->integrated_response);
 	}
-#endif
 
 	if (pid->i == 0) {
 		// If Ki is zero, do not change the integrator. We do not reset to zero
@@ -218,7 +216,6 @@ void pid_configure(struct pid *pid, float p, float i, float d, float iLim)
  */
 void pid_configure_deadband(struct pid_deadband *deadband, float width, float slope)
 {
-#ifndef SMALLF1
 	if(!deadband)
 		return;
 
@@ -240,7 +237,6 @@ void pid_configure_deadband(struct pid_deadband *deadband, float width, float sl
 
 	cubic_deadband_setup(width, slope, &deadband->cubic_weight,
 		&deadband->integrated_response);
-#endif
 }
 
 /**
--- a/flight/Libraries/math/pid.c
+++ b/flight/Libraries/math/pid.c
@@ -47,7 +47,7 @@ static float deriv_gamma = 1.0;
  * @param[in] dT  The time step
  * @returns Output the computed controller value
  */
-float pid_apply(struct pid *pid, const float err, float dT)
+float pid_apply(struct pid *pid, float err, float dT)
 {	
 	if (pid->i == 0) {
 		// If Ki is zero, do not change the integrator. We do not reset to zero
@@ -82,7 +82,7 @@ float pid_apply(struct pid *pid, const float err, float dT)
  * @Note based on "Feedback Systems" by Astrom and Murray, PID control
  *  chapter.
  */
-float pid_apply_antiwindup(struct pid *pid, const float err,
+float pid_apply_antiwindup(struct pid *pid, float err,
 	float min_bound, float max_bound, float dT)
 {	
 	if (pid->i == 0) {
@@ -130,10 +130,22 @@ float pid_apply_antiwindup(struct pid *pid, const float err,
  * This version of apply uses setpoint weighting for the derivative component so the gain
  * on the gyro derivative can be different than the gain on the setpoint derivative
  */
-float pid_apply_setpoint(struct pid *pid, const float setpoint, const float measured, float dT)
+float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const float setpoint,
+	const float measured, float dT)
 {
 	float err = setpoint - measured;
-	
+	float err_d = (deriv_gamma * setpoint - measured);
+
+#ifndef SMALLF1
+	if(deadband && deadband->width > 0)
+	{
+		err = cubic_deadband(err, deadband->width, deadband->slope, deadband->cubic_weight,
+			deadband->integrated_response);
+		err_d = cubic_deadband(err_d, deadband->width, deadband->slope, deadband->cubic_weight,
+			deadband->integrated_response);
+	}
+#endif
+
 	if (pid->i == 0) {
 		// If Ki is zero, do not change the integrator. We do not reset to zero
 		// because sometimes the accumulator term is set externally
@@ -144,8 +156,8 @@ float pid_apply_setpoint(struct pid *pid, const float setpoint, const float meas
 
 	// Calculate DT1 term,
 	float dterm = 0;
-	float diff = ((deriv_gamma * setpoint - measured) - pid->lastErr);
-	pid->lastErr = (deriv_gamma * setpoint - measured);
+	float diff = (err_d - pid->lastErr);
+	pid->lastErr = err_d;
 	if(pid->d && dT)
 	{
 		dterm = pid->lastDer +  dT / ( dT + deriv_tau) * ((diff * pid->d / dT) - pid->lastDer);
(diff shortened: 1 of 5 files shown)


flight/Libraries/math/pid.c
```diff
@@ -199,6 +230,67 @@ void pid_configure(struct pid *pid, float p, float i, float d, float iLim)
 }
 
 /**
+ * Assigns a pid_deadband struct to the PID controller.
+ * @param[out] pid The PID structure to configure
+ * @param[in] deadband The external deadband structure to assign
+ */
+void pid_assign_deadband(struct pid *pid, struct pid_deadband *db)
+{
```
mlyle: Can this be done dynamically (and lazily) in pid_configure_deadband?  (e.g. have it take a struct pid \* instead), and malloc the structure if it is not present.  Then at least F1 would only pay the tax for this feature if deadbands are configured.

(we're already paying the cost of a pointer-- might as well malloc)

flight/Libraries/math/pid.h
```diff
@@ -32,6 +32,13 @@
 #ifndef PID_H
 #define PID_H
 
+struct pid_deadband {
+	float dbW;				// Deadband width in degrees.
+	float dbB;				// Deadband slope in % (0..1).
```
mlyle: Remove `in %`, which implies 100

tracernz:
I would recommend using uavo field limits to limit values unless they mathematically don't make sense/will cause an error. This way the regular ui will enforce the limits, but developers/power-users are able to test outside the limits through the UAVO browser.

flight/Libraries/math/pid.h
tracernz: Would also prefer longer names so it's obvious when reading code that uses these members. Don't want to have to refer back to struct defs all the time to see what the members mean. Also, mixed-case names are frowned upon (see chapter 4 https://www.kernel.org/doc/Documentation/CodingStyle), but are present in some old code. I don't think we should do it for new code.

glowtape:
I checked the function plot, negative width values don't affect the function, but a negative slope causes funny business, so I think we should still check that it's zero or larger. UAVO field limits are already defined however.

flight/Libraries/math/pid.c
glowtape: Hmm, the pointer assignment stuff is there to be able to disable the deadband between mode switches, while still being able to update it with no fuzz via the settings callback (maybe allow it as option in TxPID). I suppose I could put a check around the pid_configure_deadband calls and just not malloc memory, if all deadbands are zero. If there's no deadband, it'll assign NULL to begin with when passing the pointer along.

That said, the assign and remove functions are mostly for readability and unneeded abstraction, we might aswell turn them into macros in stab.c.

If we malloc for saving RAM, maybe we should also cause a system configuration error on F1 when any deadbands are enabled the same time as autotune.

On further note, thanks to some shower thoughts, I guess it makes more sense to pass the deadband via argument to pid_apply_setpoint (which the rate controllers use), that's four bytes on the call stack. Like this we can scrub the pointer from the pid struct. Also, would we ever need deadbands in the other two, anyway? If not, I'll remove the calls and make some codespace for the checks above.

I'll write a commit with those things for review this evening when I'm back for work.

(--edit: Means when no deadbands, we'd down to 4 bytes for the initial pointer to malloc to and 4 bytes on the stack for another argument on a function call. From 88 bytes, i.e. 10x4 for the pointer in the pid struct, and 4x4x3 from the static deadband structs.)

flight/Libraries/math/pid.h
glowtape: Sure, I can live with that.

flight/Libraries/math/pid.c
mlyle: > Hmm, the pointer assignment stuff is there to be able to disable the deadband between mode switches, while still being able to update it with no fuzz via the settings callback (maybe allow it as option in TxPID)

I don't understand this-- if we have a deadband on the rate controller, shouldn't it always be there-- whether we're flying rate or attitude?

> If we malloc for saving RAM, maybe we should also cause a system configuration error on F1 when any deadbands are enabled the same time as autotune.

We won't survive that long.. so not worth putting in the check that way.  Solution to surviving bad config on F1 is more complicated (and not served by adding special purpose code for each bad combination, unfortunately).

glowtape: > I don't understand this-- if we have a deadband on the rate controller, shouldn't it always be there-- whether we're flying rate or attitude?

Say someone is switching between axis modes with and without deadband during uptime, seems less drama to allocate it once and use it conditionally, than free and reallocate it every time. The code for this is simpler. For instance, I've provisionally excluded weak leveling because of its slow acceleration. Or someone may decide to run autotune and then decide to run down the battery in attitude. SettingsCb would just malloc the structure and it then gets used when needed or NULL gets passed in the respective axis modes. For instance, axis lock might possibly be another for exclusion, too (then again, I think you wanted it gone, anyway).

mlyle: We can't free anything.

I'm just saying-- shouldn't the rate controller always have the deadband if configured?  There is no need to "take it out"

It looks like rihgt now you have no deadband on the inner loop for attitude mode.  Why not?  If the deadband is to prevent ringing in the inner loop, doesn't that hold true whether stick input or the attitude controller is driving it?

glowtape: It was an odd way to make sure it gets used or not when it should be, because the pointer to the deadband was inside the pid struct. I remember you said that autotune wouldn't like it, thus the desire to disable it. Check the new rebased PR, I'm passing it as a parameter instead, which I guess is more elegant to begin with.

mlyle: I'm not sure about autotune as far as sane deadbands are concerned-- we should try with it first and see how it goes.

It might even be helpful, because it'll encourage undertuned behavior near the middle, and autotune likes undertuned conditions.

mlyle:
jenkins, test this please

glowtape:
Finally got that stupid CC3D of mine to work. I get 452 free bytes on heap without it compiled in, 440 bytes with deadbands compiled in, and 392 with deadbands set. Autotune enabled and PWM receiver set.

I have to figure out why it's 12 bytes and not 4, between none at all and none configured. There's only a single pointer when they're not set. --edit: Nevermind, additional code being generated. D'oh.

flight/Libraries/math/pid.c
```diff
@@ -47,7 +47,7 @@ static float deriv_gamma = 1.0;
  * @param[in] dT  The time step
  * @returns Output the computed controller value
  */
-float pid_apply(struct pid *pid, const float err, float dT)
+float pid_apply(struct pid *pid, float err, float dT)
```
mlyle: Starting a new review pass.

This function's signature and the next one should not change.

flight/Libraries/math/pid.c
```diff
@@ -82,7 +82,7 @@ float pid_apply(struct pid *pid, const float err, float dT)
  * @Note based on "Feedback Systems" by Astrom and Murray, PID control
  *  chapter.
  */
-float pid_apply_antiwindup(struct pid *pid, const float err,
+float pid_apply_antiwindup(struct pid *pid, float err,
```
mlyle: No signature change necessary.

flight/Libraries/math/pid.c
```diff
@@ -130,10 +130,22 @@ float pid_apply_antiwindup(struct pid *pid, const float err,
...
  */
-float pid_apply_setpoint(struct pid *pid, const float setpoint, const float measured, float dT)
+float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const float setpoint,
+	const float measured, float dT)
 {
 	float err = setpoint - measured;
-	
+	float err_d = (deriv_gamma * setpoint - measured);
+
+#ifndef SMALLF1
```
mlyle: This ifdef shouldn't be here-- it doesn't save RAM on F1-- just a small amount of code space-- but silently doesn't apply any deadband that you _have_ allocated.

flight/Libraries/math/pid.c
```diff
@@ -199,6 +211,39 @@ void pid_configure(struct pid *pid, float p, float i, float d, float iLim)
...
 
 /**
+ * Set the deadband values
+ * @param[out] pid The PID structure to configure
+ * @param[in] deadband Deadband width in degrees per second
+ * @param[in] slope Deadband slope (0..1)
+ */
+void pid_configure_deadband(struct pid_deadband *deadband, float width, float slope)
+{
+#ifndef SMALLF1
```
mlyle: Likewise-- I don't think this should be here.  If we want to prohibit pid deadbands on F1, the caller should not do it, so we can do it in cases that make sense.

But I think our discussion and a lot of the memory optimization has been with the goal of running this on F1.

flight/Libraries/math/pid.h
```diff
@@ -44,12 +51,13 @@ struct pid {
 };
 
 //! Methods to use the pid structures
-float pid_apply(struct pid *pid, const float err, float dT);
-float pid_apply_antiwindup(struct pid *pid, const float err, float min_bound, float max_bound, float dT);
-float pid_apply_setpoint(struct pid *pid, const float setpoint, const float measured, float dT);
+float pid_apply(struct pid *pid, float err, float dT);
```
mlyle: These two signatures don't change.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -119,6 +119,9 @@ bool lowThrottleZeroIntegral;
 float vbar_decay = 0.991f;
 float gyro_alpha = 0.6;
 struct pid pids[PID_MAX];
+#ifndef SMALLF1
```
mlyle: This is the correct place to suppress the deadbands on F1-- but I think we shouldn't for now.  Change this to `NO_CONTROL_DEADBANDS` .

Later, if we want to turn it off on F1, we can do

```
#ifdef SMALLF1
#define NO_CONTROL_DEADBANDS
#endif
```

mlyle:
Review pass done-- not sure the status of the RAM tests above is valid with the deadbands compiled out (or why any additional ram was used with them with it disabled???).

glowtape:
The RAM thing is strange. I'd expected it to be four bytes due to the pointer for the malloc, not sure why it's 12. It's what the SystemStats reported.

Changed what you've asked. The signature thing was a remainder of having deadbands in these functions before, too. IIRC using _const_ gave me compiler errors trying to write to err (didn't want to declare a new variable).

flight/Libraries/math/pid.h
```diff
@@ -32,6 +32,13 @@
 #ifndef PID_H
 #define PID_H
 
+struct pid_deadband {
+	float width;				// Deadband width in degrees.
```
mluessi: deg/s ?

flight/Modules/Stabilization/stabilization.c
```diff
@@ -744,7 +757,7 @@ static void stabilizationTask(void* parameters)
 									rateDesiredAxis[YAW] = pid_apply(&pids[PID_ATT_YAW], axis_lock_accum[YAW], dT);
 									rateDesiredAxis[YAW] = bound_sym(rateDesiredAxis[YAW], settings.MaximumRate[YAW]);
 
-									actuatorDesiredAxis[YAW] = pid_apply_setpoint(&pids[PID_RATE_YAW],  rateDesiredAxis[YAW],  gyro_filtered[YAW], dT);
+									actuatorDesiredAxis[YAW] = pid_apply_setpoint(&pids[PID_RATE_YAW], NULL, rateDesiredAxis[YAW], gyro_filtered[YAW], dT);
```
mluessi, line 760: No deadband for yaw?

glowtape, line 760: AFAIK coordinated flight is plane-only, so I did not apply it there. Same applies to Virtualbar.

That big ass switch statement gets iterated over every axis, so it's being applied in the other sections (like *_RATE, *_ATTITUDE, etc.).

PR #1189, opened 2016-08-09
Coordinated flight PID gets configured twice.

glowtape:
Some older copypasta mistake, I figure.


mlyle:
jenkins, test this please

mlyle:
We should consider removing virtualbar– and maybe coordinatedflight from F1.  (with a systemconfiguration error if they're selected)

PR #1148, opened 2016-07-27
FlyingPi I/O Microcontroller Support

mlyle:
FlyingPi includes an I/O microcontroller-- a STM32F030-- used for generating PWM, demodulating PPM, adding another serial port, etc.  This PR will add that target in minimal form, and adapt PiOS to run in non-RTOS environments again, and add support for STM32F0xx..

PR #1195, opened 2016-08-11
PIOS F0 & sim support, various code cleanup and factoring improvements

mlyle:
This is the first fraction of #1148, separated for easier review.  Will rebase 1148 on top once this lands.


tracernz:
I've had another read through this. Looks pretty good. What's the motivation for reading the LED value? I think you should make the declaration conditional on sim unless there's some purpose for it that extends to STM32 targets (in which case you should implement it for STM32).

mlyle:
Thanks for the read-through.  Will make the declaration conditional.

It's there so when I do actuator updates, I can send to the subsidiary microcontroller the LED status-- flyingpio is in charge of the heartbeat LED.

e.g. https://goo.gl/photos/bhhMX98mh6u6XAqr8 :) (click to play video)

tracernz:
Sounds good. :+1:

mlyle:
jenkins, test this please

PR #1197, opened 2016-08-12
Pulldown TX pins for F3 UARTs.

### LINKED ISSUE #1034: Unverified: Smart Port sensor discovery doesn't work on X4R-SB
From here:  https://forum.dronin.org/forum/d/28-smartport-telemetry-on-lux-flight-controller/12

> I'm running OpenTX 2.1.8 on the transmitter.
> 
> Yeah it seems like the problem is isolated to sensor discovery and impacts the x4r-sb and not the x8r.
> 
> x8r - sensor discovery works, sensor data makes it to transmitter as expected
> 
> x4r-sb - sensor data does make it to the transmitter, but the sensors have to be set up manually, by script, or carried over from a sensor discovery pass when the x8r is connected.
> 
> Sensor discovery does work on the x4r-sb when the lux is running cleanflight, so I assume the x4r-sb hardware is not faulty, but I don't really know anything about how the sensor discovery protocol works so I'm not sure why the x8r would work for sensor discovery while the x4r-sb fails on dronin.

dustin:
This appears to be necessary for single-wire usage with Frsky S.Port
telemetry.

Closes #1034

### DIFF
--- a/flight/targets/dtfc/board-info/board_hw_defs.c
+++ b/flight/targets/dtfc/board-info/board_hw_defs.c
@@ -336,7 +336,7 @@ static const struct pios_usart_cfg pios_uart1_usart_cfg = {
 			.GPIO_Speed = GPIO_Speed_2MHz,
 			.GPIO_Mode  = GPIO_Mode_AF,
 			.GPIO_OType = GPIO_OType_PP,
-			.GPIO_PuPd  = GPIO_PuPd_UP
+			.GPIO_PuPd  = GPIO_PuPd_DOWN
 		},
 		.pin_source = GPIO_PinSource6,
 	},
@@ -371,7 +371,7 @@ static const struct pios_usart_cfg pios_uart2_usart_cfg = {
 			.GPIO_Speed = GPIO_Speed_2MHz,
 			.GPIO_Mode  = GPIO_Mode_AF,
 			.GPIO_OType = GPIO_OType_PP,
-			.GPIO_PuPd  = GPIO_PuPd_UP
+			.GPIO_PuPd  = GPIO_PuPd_DOWN
 		},
 		.pin_source = GPIO_PinSource2,
 	},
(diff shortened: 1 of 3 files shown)


tracernz:
What about F1/F4? Do they also need this?

dustin:
AFAIK, F1 and F4 don't have the magic single-wire stuff.  The wiki page
suggests you need additional hardware in general.

mlyle:
Looks wunnerful.  Thanks for figuring this out :)

tracernz:
Wait, this affects all users of UART, not just FrSky? That's really bad.

dustin:
As I understand it, it only affects single wire serial use.  If there's a
case that needs tx to pull up as well, then that's going to be complicated.
My osd still works, though, so I don't think it's totally wrong.

tracernz:
All UART protocols I've worked with expect the line to be high while no data is transmitted (which is the equivalent of low for an inverted protocol), how else do you detect the start bit? Proper fix would be in here: https://github.com/d-ronin/dRonin/blob/545576598e794261555c61eb6a2fd687c6c343a2/flight/PiOS/STM32F30x/pios_usart.c#L181

mluessi:
@tracernz the protocol uses inverted serial, so it's low when idle and for the start bit it goes high.

mluessi:
When the serial port is in 2-wire mode, the pull up/down resistor shouldn't matter, as the line is always actively driven.

mlyle:
Yes-- that's the assumption-- that when it's driven at 5-20 ohm resistance, a 40k pullup/pulldown doesn't matter (and just wastes power).

Arguably having any pullup/pulldown is bad in two-wire mode, because it wastes some microamperes.. but that's not a real factor IMO.

tracernz:
Ah. This will affect single wire mode when not inverted though right?

mlyle:
Yes, that's true, but I don't believe we have any non-inverted single wire in use otherwise.  Could be wrong.

tracernz:
Ah, I do in a branch, Jeti EX-Bus. ;-) I'll fix it when I eventually get that ready for a PR.

mluessi:
Unless Jeti has a built-in pull-up that is stronger than the STM32 one, which would make sense.

PR #1203, opened 2016-08-13
Packaging: Collect debug symbols from ground and flight

tracernz:
Debug symbols are now always generated when GCS is built in release config on Linux, macOS and Widnows-MSVC. 

The packaging scripts now generate breakpad symbols for only valid files (and choose the correct versions of libraries on OSX) so we don't get a whole lot of junk along with them. Breakpad symbols, flight firmware debug info (for GDB debugging), and GCS debug info (for GDB/LLDB on Linux/Mac, MSVC on Windows) are collected into a separate artifact, `dronin-*-symbols.(tar.xz|zip)`. Breakpad symbols are also packaged on their own, `dronin-*-breakpad.(tar.xz|zip)`. The intent is that the full symbols package will be archived for tagged release builds, and just the breakpad archive for standard artifact builds. My stack-trace processor can then pick up the breakpad artifacts.

@mlyle Can you make the Jenkins artifact builders archive `dronin-*-breakpad.(tar.xz|zip)` please? `dronin-*-symbols.(tar.xz|zip)` would be nice for release builds too.

Tested on my own Linux, Mac, and Windows machines, seems to work well.


tracernz:
Added a commit I needed to build syms in 64-bit git-bash. Makes us use the 32-bit breakpad binaries.

mlyle:

mlyle:

mlyle:

tracernz:
- Windows: worked 🍺 
- OSX: macdeployqt segfaults when debug symbols are in the package (I tried deleting them from the gcs build dir and then running it, it was fine). Will have to do some workaround to move them out before macdeployqt runs. It worked fine with macdeployqt from 5.6.0 so it's a Qt regression 👎 .
- Linux: Still no execute permission on dump_syms, need to investigate.

mlyle:
It's the user/group/everyone perms.

On our build servers, I typically set up my own dir under `mlyle`, and then have a `build` user that gets a symlink.

Looks like your umask is 077? 

For now I fixed with chmod -R 755 tools/breakpad

mlyle:

tracernz:

tracernz:

mlyle:
LGTM.  Thanks for your patience-- 3 build platforms is a pain

PR #1205, opened 2016-08-13
sparky2: include HSUM

mlyle:
No idea why this was originally omitted.

### DIFF
--- a/flight/targets/sparky2/fw/pios_config.h
+++ b/flight/targets/sparky2/fw/pios_config.h
@@ -93,7 +93,7 @@
 /* Supported receiver interfaces */
 #define PIOS_INCLUDE_RCVR
 #define PIOS_INCLUDE_DSM
-//#define PIOS_INCLUDE_HSUM
+#define PIOS_INCLUDE_HSUM
 #define PIOS_INCLUDE_SBUS
 #define PIOS_INCLUDE_PPM
 #define PIOS_INCLUDE_PWM


mlyle:

tracernz:
Report that this doesn't work: https://forum.dronin.org/forum/d/104-sumd-doesn-t-work/11

mluessi:
Strange; I wonder if we have some other regression in `next` for sparky2 and it's unrelated to this PR.

mlyle:

mlyle:

excalibur46:
Always the problem. When choosing SUMD, the RC input is always red while the channel mapping is ok

excalibur46:
The problem is always there. With SUMD, the RC input module is red while the channel mapping is ok

tracernz:
You configured the min/max/neutral to suitable values? (note: S.Bus is not 1000-2000)

mlyle:
@excalibur46 Have you tried saving it in the hardware settings, resetting the board, and then running the input wizard?

excalibur46:
**@mlile** The input wizard doesn't work  because I think the driver HOTT SUMD is not initialized?
No need to connect a receiver to see the problem. If I choose HOTT SUMD on a port (Rcvr, main, flexi) and set a correct mapping, theoretically the RC input window should turn yellow which means the RC input is OK, but there is no RX activity
If I choose another mode, for example PPM and set the mapping, immediately the RC input becomes yellow and green of course if I connect the correct RX
Important detail , SUMD works perfectly with my REVO board.

mlyle:
@excalibur46 I posted on the forum-- that's not how it goes for me with this build.  I get yellow RC, and the "timeout" error code instead of the no driver.  Please make sure you're running bc1b2433 on the flight controller and that you've reset after configuring the port.

tracernz:
This looks good, should be no reason for it to cause issues, merging.

excalibur46:
Ah OK. Trying the last build with merging and it works  :o)

PR #1206, opened 2016-08-14
brainre1: enable protocols that were skipped

mlyle:
openlog, storm32bgc


mlyle:
Was tested by @glowtape on IRC and found to function. (for the openlog one)

fujin:
lgtm :+1:, is there a way to make these uhh.. less-optional so that they aren't skipped again in the future? some kind of compile-time assertion that they are available? It'd be nice to say a module should be running on all f1/f3/f4 barring some exclusions then be able to check against that (especially automatically.)

mlyle:
I think the important thing is when a new target comes in, that the whole range of next since it was forked should have a similar target looked at for change as part of the review process.

tracernz:
Basically, it's due to lazy rebasing :-P.

mlyle:
Hmm.  It doesn't require a rebase to happen-- just for changes to be made across other targets that don't make their way to an out-of-tree one.

mluessi:
Thanks for the fix and sorry for not adding them in the first place ;)

tracernz:
> Hmm. It doesn't require a rebase to happen-- just for changes to be made across other targets that don't make their way to an out-of-tree one.

Hmm, I would always rebase a new target against next to make sure it's not broken before opening a PR. :stuck_out_tongue:

mlyle:
Yes, but rebase or not, changes to other targets adding new features do not trigger a conflict or other notification.  It takes manual review to catch this (or some kind of compile-time assert about whether the feature has been considered for the target).

PR #1207, opened 2016-08-16
simulation: support for various serial protocols

mlyle:
Code complete; needs some test


flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -454,12 +454,17 @@ static void msp_send_altitude(struct msp_bridge *m)
 	float tmp;
 
 #ifdef SMALLF1
-	if (BaroAltitudeHandle() != NULL)
+	if (BaroAltitudeHandle() != NULL) {
```
tracernz, line 457: You don't have to make the code not conform to the style guide. 😜 
Great catch though. 👍

mlyle, line 457: :<

flight/PiOS/Common/pios_hal.c
```diff
@@ -48,6 +47,18 @@
 #include <openlrs.h>
 #endif
 
+#ifdef PIOS_INCLUDE_USART
+#include <pios_usart_priv.h>
```
tracernz, line 51: Just define anonymous structs for this stuff in public headers :-P. We should seriously look at taking the hit and doing that throughout pios so we can avoid including private headers everywhere like this. :wink: I guess you need to actually access members of the USART config/params in here though :(.

flight/PiOS/Common/pios_hal.c
```diff
@@ -452,6 +463,8 @@ static void PIOS_HAL_ConfigureIBus(const struct pios_usart_cfg *usart_ibus_cfg,
  * @param[in] dsm_mode Mode in which to operate DSM driver; encapsulates binding
  * @param[in] sbus_cfg SBUS configuration for this port
  */
+
+#ifdef PIOS_INCLUDE_USART
```
tracernz, line 467: This function is also needed for I2C. :/

flight/PiOS/STM32/inc/pios_sbus_priv.h
```diff
@@ -85,6 +86,8 @@ struct pios_sbus_cfg {
 	BitAction gpio_inv_disable;
 };
 
+#endif /* PIOS_INCLUDE_USART */
```
tracernz: I don't see a matching ifdef for this? Looks like it will terminate the header guard one.

flight/PiOS/inc/pios_hal.h
```diff
@@ -3,20 +3,27 @@
...
-#include <pios_usb_cdc_priv.h>
-#include <pios_usb_hid_priv.h>
+#endif
 
 #if defined(PIOS_INCLUDE_RFM22B)
 #include <pios_rfm22b_priv.h>
 #include <pios_openlrs_priv.h>
 #endif /* PIOS_INCLUDE_RFM22B */
 
+struct pios_dsm_cfg;
```
tracernz, line 16: These should go in the public headers for the relevant drivers IMO.

flight/PiOS/Common/pios_hal.c
mlyle, line 51: I will move the ones I can into public headers.  Where we get into some trouble is when something like PPM is target specific and in stm32/inc or whatever-- but PIOS_HAL needs the type in its function signature.

flight/PiOS/Common/pios_hal.c
mlyle, line 467: Yes.  It's kinda a hack the way stuff is disabled in PIOS_HAL.  PIOS_HAL is just in the simulator right now because we need it for the speed change function that some modules use-- a lot of this is hackery to get it to compile.  I figure this can get cleaned up a little when I work on Pi I2C.

flight/PiOS/STM32/inc/pios_sbus_priv.h
mlyle: I think I need to adjust the comment there.  Will do.

tracernz: This endif lines up with `#ifndef PIOS_SBUS_PRIV_H`

flight/PiOS/STM32/inc/pios_sbus_priv.h
```diff
@@ -102,8 +105,6 @@ extern int32_t PIOS_SBus_Init(uintptr_t *sbus_id,
 			      uintptr_t lower_id);
 
 
-#endif /* PIOS_INCLUDE_SBUS */
-
 #endif /* PIOS_SBUS_PRIV_H */
```
tracernz, line 106: And this lines up with `#ifdef PIOS_INCLUDE_SBUS`

mlyle:
Thanks for the review-- I'll get most of this cleaned up soon.

flight/PiOS/inc/pios_hal.h
mlyle, line 16: I punted on almost all of this.  Only a few were doable, and the ones that were seemed like a bad idea (other than pios_com_driver).

I will start a series of PRs to clean up individual headers in the style of pios_flyingpio.h.  And it's pretty clear that pios_hal is not general enough and needs some reimagining, since it's very hacked here.

flight/PiOS/STM32/inc/pios_sbus_priv.h
mlyle, line 106: fixed.

mlyle:
OK, ... didn't clean up much.  I'm sorry a lot of this still sucks. :(  There's been a lot of net-cleanup lately.

tracernz:
No prob. I don't except this PR to fix the whole of PiOS. 😁

PR #1209, opened 2016-08-17
stabilizationsettings: increase default rates

mlyle:
Basically everyone, even novices, turns this stuff way up.  This turns
it up a little to be somewhat better.  Chose 250 degrees/35% (163 degree
center stick rate) vs. old 150 deg/sec for P/R.  Since people in
attitude modes are affected by yaw, chose 225 degrees/30% expo (158
degree/s center-stick rate) to not be TOO bad.

### DIFF
--- a/shared/uavobjectdefinition/stabilizationsettings.xml
+++ b/shared/uavobjectdefinition/stabilizationsettings.xml
@@ -4,9 +4,9 @@
 	<field name="RollMax" units="degrees" type="uint8" elements="1" defaultvalue="55" limits="%BE:0:180"/>
 	<field name="PitchMax" units="degrees" type="uint8" elements="1" defaultvalue="55" limits="%BE:0:180"/>
 	<field name="YawMax" units="degrees" type="uint8" elements="1" defaultvalue="35" limits="%BE:0:180"/>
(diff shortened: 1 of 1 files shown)


jhitesma:
I approve of this ;)

Just thinking out loud but maybe it would be a good idea to work it into the setup wizard and give a choice between mild/standard/acrobatic options.

mlyle:
jenkins, test this please

tracernz:
This is good. It would be nice to add rate/expo settings to the setup wizard in future too.

mlyle:
@tracernz I've been thinking about putting it on the autotune wizard, since it's kind of the next logical step at that point and we have the user's attention.

tracernz:
I find it scary to fly 150 deg/sec even on the initial flight while auto-tuning! :stuck_out_tongue:

mlyle:
For yaw?  because hte rest is leveling.

tracernz:
Nah, I always take off in rate mode, and the lack of control authority is scary to begin with. 😉

mluessi:
Good idea.

I agree with @tracernz; it would be nice to have some presets in the wizard / stabilization pane.

mluessi:
Thanks @mlyle

PR #1210, opened 2016-08-17
revolution: preliminary support for revolution acro board

mlyle:
- Support the N25Q128 that's on the Revolution Acro board

None of the additional capacity is used-- this is merely to tolerate
this part.  Reported OK by yds on forum here:
https://forum.dronin.org/forum/d/87-readytoflyquads-revolution-acro/25
- Generalize the "missing sensor" infrastructure currently used for external mags
- Generate ERRORs that can be demoted to WARNINGs through a config setting when an expected sensor is missing.


flight/targets/revolution/fw/pios_board.c
```diff
@@ -166,8 +166,11 @@ void PIOS_Board_Init(void) {
 
 #if defined(PIOS_INCLUDE_FLASH)
 	/* Inititialize all flash drivers */
-	if(PIOS_Flash_Jedec_Init(&pios_external_flash_id, pios_spi_telem_flash_id, 1, &flash_m25p_cfg) != 0)
+	if((PIOS_Flash_Jedec_Init(&pios_external_flash_id, pios_spi_telem_flash_id, 1, &flash_m25p_cfg) != 0) &&
+			(PIOS_Flash_Jedec_Init(&pios_external_flash_id, pios_spi_telem_flash_id, 1, &flash_n25q128_cfg) != 0)) {
```
tracernz: This will always execute the second init which (without reading the flash code) will break things for the M25P right? I think you need to nest another if.

tracernz:
I wonder if we can just make PIOS_Flash_Jedec figure out the type and size of memory connected, and error out the init if it's smaller than the board asks for (or if it doesn't find any supported chip)? That would allow us to be flexible for all boards and have less target-specific code. What would be even cooler is teaching the partition manager to auto-expand the last partition but that might be a lot more difficult.

mlyle:
Yes, I was thinking the latter part (auto-expand last).  Unfortunately, there's a lot of different sector sizes and .. there's implications for the correctness of the filesystems on top.

flight/targets/revolution/fw/pios_board.c
mlyle: C has short circuit eval of && and ||-- so if the first term in an and expression is false, the second part isn't evaluated.

tracernz: Ah, of course, it returns zero on success, not true, duh. 👍

mlyle:
I did a little more of the work along the way to this.  That one define (not enabled by default) allows the firmware to work on either Revolution or "Revolution Acro".  Also added a comment explaining "the plan".

mluessi:
Code LGTM. One thing we should think about is what happens if e.g. AltitudeHold is configured as a flight mode and the baro is missing. The warning doesn't prevent arming, so funny things may happen ;).

tracernz:
Yah, we need to make sure that side of it is solid before removing the halts.

mlyle:
OK.  Pushed a change where autonomy-related modules will not start with missing related sensors.  This causes the sanitycheck to fail because the associated module is not running.

mluessi:
LGTM. Does someone have a Revo Acro to test this before merge?

mlyle:
I have tested everything except the logic to prevent arming, since I don't have an RX hooked up.

mlyle:
Not sure why this had conflicts-- it rebased cleanly.

mlyle:
Test 1: altitude hold module works on real revo.
Test 2: System arms with alt-hold set as a flight mode via GCS rcvr.

mlyle:
Test 3: acro revo starts up with sensors and baro/acro error.  Will not arm.
Test 4: with TolerateMissingSensors=true, acro revo starts up with sensors and baro/acro yellow.  Arms.  Cosmetic: Altitude tape in GPS runs away to positive infinity 
Test 5: with "altitudehold" module enabled, module does not start.  Critical "systemconfig-altitudehold" error when altitude hold in flight mode list.

flight/targets/revolution/fw/pios_config.h
```diff
@@ -98,6 +99,8 @@
 #define PIOS_INCLUDE_GPS_UBX_PARSER
 #define PIOS_GPS_SETS_HOMELOCATION
 
+#define PIOS_TOLERATE_MISSING_SENSORS
```
tracernz: Why redefine? 😕

shared/uavobjectdefinition/sensorsettings.xml
```diff
@@ -9,7 +9,13 @@
         <field name="ZGyroTempCoeff" units="(deg/s)/deg" type="float" elementnames="1,T,T2,T3" defaultvalue="0"/>
         <field name="MagBias" units="mGau" type="float" elementnames="X,Y,Z" defaultvalue="0,0,0"/>
         <field name="MagScale" units="gain" type="float" elementnames="X,Y,Z" defaultvalue="1"/>
-        <field name="ZAccelOffset" units="m/s^2" type="float" elements="1" defaultvalue="0"/>
+	<field name="ZAccelOffset" units="m/s^2" type="float" elements="1" defaultvalue="0"/>
```
tracernz: Rest of the file uses spaces...

mlyle:
Review comments fixed

mlyle:
A few additional commits now, involved in supporting the Airbot F4 Lite board.  Seems to work.

mlyle:
I am happy with this on revo acro and airbot f4 lite prototype.  I need to look at real revo next.

mlyle:
altitude hold module and baro correct on real revo.  Some last test sequence on revo acro...

mlyle:
With this last change I'm super pleased-- now you can erase settings etc too.  I think this is good to merge IMO.

tracernz:
Should probably just swap the alarm and heartbeat LEDs if you're going to do that? Then the real revo at least still gets two separate LEDs.

mlyle:
I swapped which LED is used to blink panics -- but the real revo still has an alarm led I believe

PR #1211, opened 2016-08-18
python-logviewer: correct a problem with actuator graph

mlyle:
Indexed-rendering didn't do the right thing and resulted in lots of
extra points, bad performance, and sometimes smearing.


mlyle:
image

Courtesy of @ufoDziner --- a log with a hangtime flip.

PR #1213, opened 2016-08-19
BrainRE1: Integrate optional I2C mag and baro to enable navigation

jihlein:
Enables the BrainRE1 to use external I2C sensors to allow for navigation.

I've checked the operation of the 8 PWM pins, and they work as expected in banks of 4, 6, or 8 outputs depending on the number of Uart ports needed.  With 4 PWM outputs, both Uart4 and Uart6 are available, with 6 PWM outputs Uart6 is available, and with 8 PWM outputs neither Uart4 or Uart6 is available.

External I2C sensors can be HMC5883L mag, and/or BMP280 or MS5611 pressure sensor.

Uart4 and I2C sensor operation has not been verified yet, continuing to test.


flight/targets/brainre1/fw/pios_board.c
```diff
@@ -439,57 +460,79 @@ void PIOS_Board_Init(void) {
...
+	/* Multi-function port (pwm, 1 serial, 2 serial) */
 	uint8_t hw_mp;
 	HwBrainRE1MultiPortGet(&hw_mp);
 
+	/* Configure PWM Outputs */
+#if defined(PIOS_INCLUDE_SERVO) && defined(PIOS_INCLUDE_TIM)
 	switch (hw_mp) {
-		case HWBRAINRE1_MULTIPORT_PWM:
-			// for us, this means PWM output, not input
+		case HWBRAINRE1_MULTIPORT_PWMOUT4:
```
mluessi: Why did you change this? Shouldn't be necessary.

flight/targets/brainre1/fw/pios_board.c
```diff
@@ -439,57 +460,79 @@ void PIOS_Board_Init(void) {
...
+			PIOS_Servo_Init(&pios_servo_4_cfg);
 			break;
 	}
+#endif /* defined(PIOS_INCLUDE_SERVO) && defined(PIOS_INCLUDE_TIM) */
 
-	// Enable / disable RE1 specific hardware inverters and FrSky RSSI support
-	switch (hw_mp) {
-		case HWBRAINRE1_MULTIPORT_FRSKYSENSORHUB:
-			PIOS_RE1FPGA_MPTxPinMode(false, true);
+	if (hw_mp == HWBRAINRE1_MULTIPORT_PWMOUT4) {
```
mluessi: I don't think we should offer this option. UART4 is on the on the first two motor outputs and this makes it very confusing for users..

flight/targets/brainre1/fw/pios_board.c
```diff
@@ -439,57 +460,79 @@ void PIOS_Board_Init(void) {
...
+				NULL,                                // i2c_cfg
+				NULL,                                // ppm_cfg
+				NULL,                                // pwm_cfg
+				PIOS_LED_ALARM,                      // led_id
+				NULL,                                // dsm_cfg
+				0,                                   // dsm_mode
+				NULL);                               // sbus_rcvr_cfg
+	}
+
+	if ((hw_mp == HWBRAINRE1_MULTIPORT_PWMOUT4) || (hw_mp == HWBRAINRE1_MULTIPORT_PWMOUT6)) {
```
mluessi: This is the same UART port that was called "multiport" before. I don't think we should change this.

mluessi:
@jihlein I would prefer if we only exposed the I2C options and not give the option to change the PWM outputs etc. UART4 is on the first two motor outputs and changing this makes the setup very confusing to users. UART6 was already available before (that's what the "Multiport" options configured).

jihlein:
@mluessi - Sure, I can pull that out of the PR and leave it my personal working branch.  I was more interested in number of ports than number of PWM outputs for some things I'm doing when I initially put this together.

jihlein:
Uart4 and associated support code removed.

External HMC5883L confirmed to work properly.  I need to see if I have an MS5611 breakout board somewhere.

mlyle:
@jihlein -- I am changing the way some of this works.. Can you look at #1210 and tell me what you think?

jihlein:
@mlyle:

"Can you look at #1210 and tell me what you think?"

I think this needs to be finished and merged before I go further with this one.......

Looks great.

I did find an MS5611 breakout board for testing, will need to order a BMP280 to complete testing though.  I did test the BMP280 driver a while back on a Dodo FC FWIW.

Is there a preference to BMP280 over MS5611?  Leave both as a choice?

jihlein:
External MS5611 works.

mlyle:
Nice!

mluessi:
Cool. I should have BMP280 board next week or so, so we will be able to test this as well. I think we should have both as an option. I'm planning to have a RE1 add-on board (stacking) with BMP280, but some people may want to use a MS5611.

mluessi:
@jihlein do you have time to rework this so it uses the features introduced in #1210?

jihlein:
@mluessi - yes, in fact I just looked into rebasing it earlier.  Might have something this evening, or worst case the next few days.  Have you tried the BMP280 yet?  Which breakout board did you order?

mluessi:
@jihlein awesome. I designed my own stackable board with a BMP280. The idea is to sell it as an RE1 accessory. Assuming it works, I will send you a prototype.

jihlein:
Rebased to current next to pull in #1210 changes.  Benched tested, external HMC5883 and MS5611 responded correctly, as do the CWA's when the sensors are configured but not installed.

mluessi:
Bench tested BMP280, seems to be working :)

flight/PiOS/Common/pios_bmp280.c
```diff
@@ -0,0 +1,441 @@
...
+		data.pressure = ((float) dev->compensatedPressure) / 256.0f / 1000.0f;
+		data.altitude = 44330.0f * (1.0f - powf(data.pressure / BMP280_P0, (1.0f / 5.255f)));
+
+		if (read_adc_result == 0) {
+			PIOS_Queue_Send(dev->queue, (void*)&data, 0);
+		}
+	}
+}
+
+#endif /* PIOS_INCLUDE_BMP280 */
```
mlyle: Need a newline at the end of file.

Will this work on other BMPx80 variants?

flight/targets/brainre1/fw/pios_config.h
```diff
@@ -149,6 +150,8 @@
 
 #define BRAIN
 
+#define SUPPORTS_EXTERNAL_I2C_SENSOR
+
```
mlyle: Does this define do anything anymore?  Can it simply be removed?

flight/targets/brainre1/fw/pios_board.c
```diff
@@ -520,6 +541,94 @@ void PIOS_Board_Init(void) {
...
+			}
+		}
+		else {
+			PIOS_SENSORS_SetMissing(PIOS_SENSOR_BARO);
+		}
+#endif // PIOS_INCLUDE_BMP280
+#endif // PIOS_INCLUDE_I2C
+	}
+
+	if (((hw_i2c == HWBRAINRE1_I2CPORT_EXTBARO) || (hw_i2c == HWBRAINRE1_I2CPORT_EXTBAROANDEXTMAG)) &&
```
mlyle, line 583: How feasible is it to factor this stuff into PIOS_HAL similar to the ConfigurePort stuff?  Other targets will want this same functionality.

(Might want it to return what external sensors were attempted to be initialized as flags, so that in the future targets can be factored to not use internal versions of the sensors based on this return value-- or so we can factor some of the on-board hardware support to PIOS_HAL).

mlyle:
I've read this code and it looks decent-- is it still "WIP not for merge" / not for serious review, or should we get it into the process of finalizing it?

flight/PiOS/Common/pios_bmp280.c
```diff
@@ -0,0 +1,441 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS_BMP280 BMP280 Functions
+ * @brief Hardware functions to deal with the altitude pressure sensor
+ * @{
+ *
+ * @file       pios_bmp280.c
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
```
mluessi: Was this ever part of TL? If not, I would just put dRonin as author. Also, don't forget the additional dRonin notice (see license template). Also applies to other BMP280 files.

mlyle: It's much easier to take the skeleton of an existing driver which "infects" that driver with Tau copyright...

I need to write a driver truly from scratch (fixing some things, too) for some random hardware that we can start using for that purpose.

flight/PiOS/Common/pios_bmp280.c
```diff
@@ -0,0 +1,441 @@
...
+		varT2 = (((((raw_temperature >> 4) - ((int32_t)dev->digT1)) * ((raw_temperature >> 4) - ((int32_t)dev->digT1))) >> 12) * ((int32_t)dev->digT3)) >> 14;
+		T = varT1 + varT2;
+		dev->compensatedTemperature = (T * 5 + 128) >> 8;
+	}
+
+	int32_t raw_pressure = (int32_t)((((uint32_t)(data[0])) << 12) | (((uint32_t)(data[1])) << 4) | ((uint32_t)data[2] >> 4));
+
+	varP1 = ((int64_t)T) - 128000;
+	varP2 = varP1 * varP1 * (int64_t)dev->digP6;
+	varP2 = varP2 + ((varP1 * (int64_t)dev->digP5) << 17);
```
mluessi, line 293: The code in the datasheet is a bit different and uses only 32 bit ints, not sure if this is critical.

mluessi:
@jihlein code looks good, one thing I noticed is the different implementation of the temp/pressure calculation in the BMP280 driver. Any reason not to use the exact code from the datasheet? Also, the AltitudeHold module should be added to RE1 again.

flight/PiOS/Common/pios_bmp280.c
jihlein: Hmm - Page 22 of my datasheet explicitly calls out 64 bit integers for best possible accuracy.  Document number BST-BMP280-DS001-11, May 5th, 2015.  We looking at different revs possibly?

Other than some variable names, the driver code here follows the Bosch code revision 1.1, also on page 22 of the data sheet.

flight/PiOS/Common/pios_bmp280.c
jihlein: Don't know it will work with others.  I heard rumors the BMP180 and BMP280 are similar, but would need to research it.

flight/targets/brainre1/fw/pios_config.h
jihlein: Yes, it should be removed.

flight/targets/brainre1/fw/pios_board.c
jihlein: Let me think about that.  Would that be the subject of another PR?

flight/PiOS/Common/pios_bmp280.c
jihlein: It was written for an off tree target (Dodo) quite a while ago.  Never released into the Tau tree.  So we could apply only the dRonin copyright info.

flight/targets/brainre1/fw/pios_board.c
mlyle: Doing all of the above stuff definitely doesn't fit in this PR.

But a basic version-- just for RE1-- would be really nice for this PR, so that it's easy to take what you've done and tack it onto Lux & DTFc.  (They need to move off the "acro" attitude stack though.)

jihlein: Do Lux and DTFc even have an external I2C port to make this possible?  I'm not seeing one on DTFc, don't have a schematic for Lux.

jihlein:
I can remove the WIP from the title, I think we are ready to move forward.  I'll work on cleaning up the above comments later today.

flight/PiOS/Common/pios_bmp280.c
jihlein: My datasheet (rev 1.14) is one revision old, but the code revision is still 1.1, so I believe we are okay here.

flight/targets/brainre1/fw/pios_board.c
mlyle: DTFc's USART1 connector for OSD-- PB6/PB7-- have I2C1_SDA and I2C1_SCL as alternates.

jihlein: Ah, implement that as a flex port on DTFc

mlyle: Yes.  Also, there will be closely related functionality going in PIOS_HAL on the serial port side (there's been a lot of talk of an integrated gps/barometer/magnetometer all multiplexed over serial).

flight/PiOS/Common/pios_bmp280.c
mluessi: I was looking at DS Revision 1.14, May 2015. But there is an ever newer version (1.15). IMO, it would make sense to use the code provided in the latest DS. They may have changed it for a reason.

mlyle: https://github.com/BoschSensortec/BMP280_driver/blob/master/bmp280.c#L1414

has the 64 bit math.

jihlein: BST-BMP280-DS001-12 is Revision 1.15.  There is code section 3.11.3 on page 22 which uses 64 bit integers (which I implemented), and in section 8.2 on pages 45/46 there is an alternate 32 bit integer implementation.  The datasheet seems to imply the 64 bit integer implementation is more accurate.

flight/PiOS/Common/pios_bmp280.c
tracernz: But was it based on some code that was copyrighted by Tau? That's the relevant factor. :wink:

jihlein:
With the exception of any refactoring, review comments pushed.

mlyle:
Looks great.  I'll take this and open a PR quick to factor this to HAL.

PR #1215, opened 2016-08-20
flyingpio: rearrange servo pins to match silkscreen

mlyle:
There was a discrepancy between silkscreen and board art that is
errata'd here and the firmware now matches the silkscreen.


mluessi:
Thanks!

PR #1219, opened 2016-08-21
GCS:Uploader: Fix crash with unknown board

tracernz:
Fixes http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgKCXwOEKDA

### DIFF
--- a/ground/gcs/src/plugins/coreplugin/boardmanager.cpp
+++ b/ground/gcs/src/plugins/coreplugin/boardmanager.cpp
@@ -118,6 +118,16 @@ void BoardManager::aboutToRemoveObject(QObject *obj)
         m_boardTypesList.removeAt(m_boardTypesList.indexOf(board));
 }
 
+IBoardType *BoardManager::getBoard(int type)
+{
+    foreach (IBoardType *board, m_boardTypesList) {
+        if (board->getBoardType() == type) {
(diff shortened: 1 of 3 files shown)


ground/gcs/src/plugins/coreplugin/boardmanager.cpp
```diff
@@ -118,6 +118,16 @@ void BoardManager::aboutToRemoveObject(QObject *obj)
         m_boardTypesList.removeAt(m_boardTypesList.indexOf(board));
 }
 
+IBoardType *BoardManager::getBoard(int type)
```
mlyle, line 121: Nice factoring.  I think there's other places that do a similar thing we can fix to use this ultimately.

tracernz:
Tested by deleting a board plugin and then plugging it in.

ground/gcs/src/plugins/coreplugin/boardmanager.cpp
tracernz, line 121: Yep. Gotta run to work though so no time to do it. 👍

tracernz:
Yay, first fix from reading the new style crash reports. :tada:

PR #1238, opened 2016-08-25
logging: support a couple more data rates

mlyle:
Before, it was rather blunt between 100Hz and 250Hz (2.5x difference),
and between 250Hz and 500Hz (2x difference).  Introduce 333Hz and 166Hz
to help out.


mluessi:
Thanks!

PR #1239, opened 2016-08-25
autotune: buffer some more points.

mlyle:
On BrainRE1 (1600Hz) with heavy logging, points spilled for @ufodZiner.
Make the queue just a little deeper on non-F1 targets where we have the
memory to fight this.


mluessi:
Thanks!

PR #1244, opened 2016-08-26
hack: "trend" display for centermark in OSD

mlyle:
```
// XXX right now looks at command and does not consider at all
// "what's happening" -- should weight both [with filtering]
// XXX for now assumes reprojection-- should inverse-reproject if no reproj
// XXX assumes angles are small and doesn't integrate the 3D rotation
// properly
```

Set centermark to "trend"

Set ArtificialHorizonMaxPitch to half the VFOV or 30% of the diagonal FOV of the camera.

### DIFF
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -93,6 +93,7 @@
 #include "magnetometer.h"
 #include "manualcontrolcommand.h"
 #include "modulesettings.h"
+#include "stabilizationdesired.h"
 #include "stabilizationsettings.h"
 #include "stateestimation.h"
 #include "systemalarms.h"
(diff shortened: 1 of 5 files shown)


mlyle:

mlyle:

mlyle:
won't work, but one more:

mluessi:
jenkins, stop being a bitch please

mlyle:

tracernz:
Can you explain what this attempts to do? "trend" by itself doesn't give much context.

mlyle:
This draws a small + in the middle of the screen, plus a larger "plus" for where stabilizationdesired is pointing right now.  Right now it's really dumb just to look at the validity of the concept.

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -1061,6 +1062,95 @@ void showVideoType(int16_t x, int16_t y)
...
+static void plot_trend_centermark(int max_pitch)
+{
+#define INNER_GAP 2
+#define INNER_MARK_LEN 2
+#define OUTER_GAP 4
+#define OUTER_MARK_LEN 3
+
+#define TREND_FORECAST_LEN 0.3f
+
+	write_line_outlined(GRAPHICS_X_MIDDLE - INNER_GAP - INNER_MARK_LEN,
```
mluessi, line 1079: I wonder if we should take the cam angle into account for this as well.

mluessi:
Cool idea; I have a feeling that projecting the attitude actual forward with more weight than the stick input may be more useful, but we will probably have to play with the weighting to see what works well.

flight/Modules/OnScreenDisplay/onscreendisplay.c
tracernz: Yeah, we discussed that earlier on IRC. This is a quick proof-of-concept opened to get build artifacts from Jenkins. ;-)

mlyle:
Yes, but that's harder and takes more filtering.  Also, since race quads have a very very small delay to hit commanded rates, it's not very much of a difference.

mlyle:
https://www.youtube.com/watch?v=mE9o3KYww9k&feature=youtu.be for reference

mlyle:

mlyle:
https://gfycat.com/ShyMammothAdder

mluessi:
I think in the video above it may not look intuitive because the center mark is above the horizon when flying forward (too much camera tilt?). Also, 300ms prediction may be too much. We could make this configurable.

mluessi:
PS: There is also a CPU warning in the video. Any ideas what is causing it?

mlyle:
Yes, logging task.

mlyle:
Maybe abandoning this work for now.

PR #1245, opened 2016-08-26
autotune: properly save vehicle size

### LINKED ISSUE #1184: Quad size drops from saved autotune parameters
- Seen in version (release version or githash): 20160720.1
- On flight controller (if applicable): 

Issue details:
When on the last "Details" page of the autotune submission, many of the fields maintain the information from the prior submission, but the quad size resets to 0.

mlyle:
Fixes #1184 
Also responsible for when we get vehicle size in weight.

Unfortunately, this is on the save not the restore size, so even after the fix is applied people with old saved settings will trip over this for some time.


mluessi:
Thanks!

PR #1253, opened 2016-08-29
revolution/pios_flash: handle variable capacity flash

mlyle:
Be willing to extend last partition.  Useful for clone revos with 128mbit flash chips for bigger logs.

### DIFF
--- a/flight/PiOS/Common/pios_flash.c
+++ b/flight/PiOS/Common/pios_flash.c
@@ -42,6 +42,63 @@ static uint8_t num_partitions;
 #define PIOS_Assert(x) do { } while (!(x))
 #define MIN(x,y) ((x) < (y) ? (x) : (y))
 
+/** Adjust partition tables to extend to the end of a variable-sized flash.
+ * Used on platforms with variable-length flash.  Requires that the partition
+ * table be mutable (not const/in flash).
+ * @param partition_table the partition table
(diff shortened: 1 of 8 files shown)


mlyle:
Boots, reports correct sizes.   Needs additional verification-- that we can indeed use the whole filesystem and can log everything-- will try and do tonight.

mlyle:
Tested-- normal revolution comes up fine.

Tested-- airbot f4 board was able to log a 7MB logfile that I then downloaded and verified.

1KHz logging to NOR flash seems to be no problem.  Getting the logs off is extraordinarily painful and needs help (outside scope of PR).  Downloading an 81 second 1KHz log took about 10 minutes.

flight/PiOS/Common/pios_flash.c
```diff
@@ -42,6 +42,63 @@ static uint8_t num_partitions;
...
+		uint8_t partition_table_len,
+		const struct pios_flash_chip *descriptor,
+		struct pios_flash_sector_range *sectors,
+		uint32_t num_bytes)
+{
+	PIOS_Assert(descriptor->sector_blocks == sectors);
+
+	uint32_t basic_capacity = 0;
+
+	for (int i=0; i < descriptor->num_blocks; i++) {
```
tracernz: count(' ')+=2 😜

Issue #1230, opened 2016-08-25
Simulation altitude goes NaN


dustin:
- Seen in version (release version or githash): Release-20160720-260-g51bfc90
- On flight controller (if applicable): simulator

dustin:
I tried to find a pattern that led to this, but pretty much anything between wiping state and trying to fly led to this:

![image](https://cloud.githubusercontent.com/assets/1779/17952104/669038e4-6a1c-11e6-830e-335618703e5f.png)

I also tried going back to an older version (specifically, the oldest one with the current QT requirement).  It did eventually go NaN on me, but was kind of useless before that, with seemingly random numbers going into attitude actual when armed.

mlyle:
@dustin -- at some point we need to work together to run this down, as I still don't have a repro (have seen it once on flyingpi, but.)

mlyle:
I believe that I understand this now.  It's racy.  Relates to task switching in a signal handler--- you can't really legitimately do that anymore.  THe major failure here is FP state getting restored by the kernel on sigreturn; if there's multiple task switches in the meantime bad things happen.

PR #1258, opened 2016-08-31
attitude: bound dT

mlyle:
It was bounded in the INS, but we can now get smaller values (RE1).
And it wasn't bounded at all.

This is a candidate for what could cause NaN run away for cfvert, found
by light manual audit.

Connects to #1230


mluessi:
LGTM

PR #1261, opened 2016-08-31
gcs: pfd: fix the altitude tape to support descent

### LINKED ISSUE #1260: Primary flight display shows ascent trends but not descents
Green indicator goes "up" next to altitude tape if you're climbing.  But no green "downward" indicator shows up for descent.

mlyle:
fixes #1260


tracernz:
Works. :+1:  Jenkins...

Issue #1270, opened 2016-09-01
Logging needs documentation


mlyle:
- How to set up an openlog / openlager
- Log viewing tool

mlyle:
https://dronin.readme.io/v201610-xx/docs/common-flight-controller-peripherals#openlog

PR #1271, opened 2016-09-01
gcs: usability: clean up modules tab.

mlyle:
- Remove geofence and vibration test tabs.  If you are using these currently,
  you should know what you're doing and use the UAVO browser IMO.
- Move GPS baud rate to GPS tab.
- Add OpenLog baud rate to Logging tab.
- Move "Baudrates" tab to far right and miscellaneous.
- Rearrange Battery tab to be slightly less tall.


mlyle:
This is relevant to #1270-- let people do everything relating to logging in main UI and don't send them to object browser.

mlyle:
image
image
image

PR #1274, opened 2016-09-02
pios_hal: factor in external mag and baro init

mlyle:
I put this together.  Untested so far.  Review of concept would be really appreciated, especially @jihlein @tracernz @mluessi

### DIFF
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -1243,7 +1243,7 @@ static int PIOS_HAL_ConfigureI2C(uint32_t *id,
 		}
 	}
 
-	if (PIOS_I2C_CheckClear(pios_i2c_external_id) != 0) {
+	if (PIOS_I2C_CheckClear(*id) != 0) {
 		AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
 		return -2;
 	}
@@ -1328,7 +1328,7 @@ int PIOS_HAL_ConfigureExternalMag(HwSharedExtMagOptions mag,
 	switch (mag) {
 #ifdef PIOS_INCLUDE_HMC5883
 	case HWSHARED_EXTBARO_MS5611:
-		ret = PIOS_HMC5883_Init(pios_i2c_external_id,
+		ret = PIOS_HMC5883_Init(*i2c_id,
 				&external_hmc5883_cfg);
 
 		if (ret) goto done;
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -230,6 +230,38 @@ uintptr_t pios_com_debug_id;
 #define PIOS_COM_TBSVTXCONFIG_RX_BUF_LEN 32
 #endif
 
+#ifdef PIOS_INCLUDE_HMC5883
+#include "pios_hmc5883_priv.h"
+
+static const struct pios_hmc5883_cfg external_hmc5883_cfg = {
+	.exti_cfg            = NULL,
+	.M_ODR               = PIOS_HMC5883_ODR_75,
+	.Meas_Conf           = PIOS_HMC5883_MEASCONF_NORMAL,
+	.Gain                = PIOS_HMC5883_GAIN_1_9,
+	.Mode                = PIOS_HMC5883_MODE_SINGLE,
+	.Default_Orientation = PIOS_HMC5883_TOP_0DEG,
+};
+#endif
+
+#ifdef PIOS_INCLUDE_BMP280
+#include "pios_bmp280_priv.h"
+
+static const struct pios_bmp280_cfg external_bmp280_cfg = {
+	.oversampling = BMP280_HIGH_RESOLUTION,
+	.temperature_interleaving = 1,
+};
+#endif
+
+#ifdef PIOS_INCLUDE_MS5611
+#include "pios_ms5611_priv.h"
+
+static const struct pios_ms5611_cfg external_ms5611_cfg = {
+	.oversampling             = MS5611_OSR_4096,
+	.temperature_interleaving = 1,
+	.use_0x76_address         = false,
+};
+#endif
+
 /**
  * @brief Flash a blink code.
  * @param[in] led_id The LED to blink
@@ -810,7 +842,7 @@ void PIOS_HAL_ConfigurePort(HwSharedPortTypesOptions port_type,
 			usart_port_params.init.USART_StopBits            = USART_StopBits_1;
 			usart_port_params.init.USART_HardwareFlowControl = USART_HardwareFlowControl_None;
 			usart_port_params.init.USART_Mode                = USART_Mode_Rx;
-			
+
 			PIOS_HAL_ConfigureSRXL(usart_port_cfg, &usart_port_params, com_driver);
 		}
 #endif  /* PIOS_INCLUDE_SRXL */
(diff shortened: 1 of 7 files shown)


flight/PiOS/Common/pios_hal.c
```diff
@@ -1197,3 +1229,138 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
...
+		ret = PIOS_MS5611_Init(&external_ms5611_cfg, *i2c_id);
+		if (ret) goto done;
+
+		ret = PIOS_MS5611_Test();
+
+		if (ret) goto done;
+#endif // PIOS_INCLUDE_MS5611
+
+	default:
+		PIOS_Assert(0);	// Should be unreachable
```
tracernz, line 1304: Do you think it's better to remove the default case and pick it up at compile time (holds as long as you don't pass in some arbitrary int rather than a member of the relevant enum), or wait until it fails at runtime if it is configured (solid against any input, but will only be detected relatively rarely)?

tracernz:
Concept looks good. Can be applied to other targets with external I2C as well... I guess in a future PR?

flight/PiOS/Common/pios_hal.c
mlyle: If the default case is gone, this will fail to compile if you're not compiling in either ms5611 or bmp280... because they'll still be in the uavo enum, but won't be in the switch.

tracernz: You can't put the ifdef inside the case? Would need a break though. That's how most our other code does it IIRC.

mlyle: OK, my intention here is if a target has an option but is missing the associated define, and a user (developer) has thus been able to put it to an invalid combination, to assert blow.  I know this behavior is debatable...

tracernz: Yeah, I don't really like things that only blow up under very specific conditions that we are unlikely to hit during release cycle. I understand the reason here though.

mlyle:
Yes :)  other mag targets, and F3 ... and Naze...

flight/PiOS/Common/pios_hal.c
tracernz: Noticed tonight that clang actually caters for this :stuck_out_tongue: 
http://i.imgur.com/KI9SKWm.png
We should see if gcc-arm supports that -Wswitch-enum :smile: 

-e- It does. Might be too noisy/undesirable though. :(

flight/PiOS/Common/pios_hal.c
```diff
@@ -1197,3 +1229,138 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
...
+	if (baro == HWSHARED_EXTBARO_NONE) {
+		return 0;
+	}
+
+	int ret = PIOS_HAL_ConfigureI2C(i2c_id, i2c_cfg);
+
+	if (ret) goto done;
+
+	switch(baro) {
+#if defined(PIOS_INCLUDE_BMP280)
```
mluessi, line 1278: Shouldn't this be below case? Otherwise if someone sets the option to BMP280 but `PIOS_INCLUDE_BMP280` is not defined it will crash.

flight/PiOS/Common/pios_hal.c
```diff
@@ -1197,3 +1229,138 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
...
+	case HWSHARED_EXTBARO_BMP280:
+		ret = PIOS_BMP280_Init(&external_bmp280_cfg, *i2c_id) ;
+
+		if (ret) goto done;
+
+		ret = PIOS_BMP280_Test();
+
+		if (ret) goto done;
+#endif // PIOS_INCLUDE_BMP280
```
mluessi, line 1289: Falls through to MS5611 if BMP280 init fails

flight/PiOS/Common/pios_hal.c
```diff
@@ -1197,3 +1229,138 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
...
+
+#if defined(PIOS_INCLUDE_MS5611)
+	case HWSHARED_EXTBARO_MS5611:
+		ret = PIOS_MS5611_Init(&external_ms5611_cfg, *i2c_id);
+		if (ret) goto done;
+
+		ret = PIOS_MS5611_Test();
+
+		if (ret) goto done;
+#endif // PIOS_INCLUDE_MS5611
```
mluessi, line 1299: Also, here falls through to assert if init fails

flight/PiOS/Common/pios_hal.c
```diff
@@ -1197,3 +1229,138 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
...
+		return 0;
+	}
+
+	int ret = PIOS_HAL_ConfigureI2C(i2c_id, i2c_cfg);
+
+	if (ret) goto done;
+
+	switch (mag) {
+#ifdef PIOS_INCLUDE_HMC5883
+	case HWSHARED_EXTBARO_MS5611:
```
mluessi: Should not be MS5611 ;)

flight/PiOS/Common/pios_hal.c
```diff
@@ -1197,3 +1229,138 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
...
+			(orientation == HWSHARED_MAGORIENTATION_TOP270DEGCW)    ? PIOS_HMC5883_TOP_270DEG    : 
+			(orientation == HWSHARED_MAGORIENTATION_BOTTOM0DEGCW)   ? PIOS_HMC5883_BOTTOM_0DEG   : 
+			(orientation == HWSHARED_MAGORIENTATION_BOTTOM90DEGCW)  ? PIOS_HMC5883_BOTTOM_90DEG  : 
+			(orientation == HWSHARED_MAGORIENTATION_BOTTOM180DEGCW) ? PIOS_HMC5883_BOTTOM_180DEG : 
+			(orientation == HWSHARED_MAGORIENTATION_BOTTOM270DEGCW) ? PIOS_HMC5883_BOTTOM_270DEG : 
+			external_hmc5883_cfg.Default_Orientation;
+
+		PIOS_HMC5883_SetOrientation(hmc5883_orientation);
+#endif /* PIOS_INCLUDE_HMC5883 */
+
```
mluessi, line 1359: break missing

mluessi:
Nice :). I'm not sure if it is on purpose, but the cases where either BMP280 or MS5611 defs are not set are a bit interesting, i.e., it will hang / crash (resulting in a boot loop) if the user sets it to that option. IMO, we should either require all external baro / mag types to be enabled at compile time or it shouldn't do anything if the wrong baro type is set.

jihlein:
Haven't had time to review yet, but will do so this evening.  

From a system perspective, this works well for those boards that have dedicated external I2C pins, AQ32 and BrainRE1 for example.  Some thought is needed as to how we make this work with targets having one or more flexports where  I2C and Uart functions are shared.

flight/PiOS/Common/pios_hal.c
jihlein: Need to add break; to all case statements?

jihlein:
One other thing we will need to handle is the MS5611 I2C address is different between targets.  Most use 0x77, AQ32 uses 0x76.

mluessi:
@jihlein for the on-bard MS5611on AQ32 the different I2C ID shouldn't matter as we only use the HAL for external baros

jihlein:
If it's only for the external devices, we should be okay then, as long as all MS5611 external hardware uses the same I2C address.

So in the case of an external mag for example, will we keep two different ways to initialize it?  An external mag hooked up to a flex port (think Quanton) is being initialized in other routines than we are currently discussing.  Maybe the flex port sequence is to init flexport I2c as we currently do, then call the HAL routine for mag init with the same I2C descriptor.  The HAL as written will see I2C already initialized, and move forward with initializing the sensor.

I'm thinking we should strive to initializing all mags and baros, whether internal or external, thru these new routines.

mlyle:
@mluessi The intention is that the user would never be able to get into that state- because the child uavo would have a matched set of options to the hardware that's actually enabled on the board.

>  Maybe the flex port sequence is to init flexport I2c as we currently do, then call the HAL routine for mag init with the same I2C descriptor. The HAL as written will see I2C already initialized, and move forward with initializing the sensor.

This is a good point.  Especially if there's any targets with multiple external I2C options.  I think in those cases you'll have to choose to enable the I2C separately.  We could pass the mag / baro fields to PIOS_HAL_ConfigurePort but it's already getting unwieldy.

> I'm thinking we should strive to initializing all mags and baros, whether internal or external, thru these new routines.

@jihlein I need to think about this.  I intended them for external use.

flight/PiOS/Common/pios_hal.c
mlyle: We actively use this already--- that's why we have that DONT_SWITCH_ON_CHILD_ENUMS thing to catch when people are using the wrong thing in a switch statement / not hitting all the cases in the parent enum.

The problem is the separate enabling of these drivers vs. forming the enumeration lists for the targets.  You can't #ifdef on a token-pasted thing, so it's hard to automatically build worthwhile DONT_BUILD_IFs on the child enum (and the namespaces are weird/hard to figure out).

jihlein:
If this is for external use only, and for targets with dedicated I2C only, I'm "almost" inclined to say leave things as they are, as we are only creating more special cases, internal, dedicated external, flexport external, etc.  We can't forget that on other targets, there's an internal mag we don't want to init when using the external.

If we have to enable flexport I2C separately, might as well leave all targets that way.  Less confusion to the end user.  Why do I have to enable I2C separately on Quanton to use an external mag, and not on BrainRE1 or AQ32, for example.

mlyle:
> If this is for external use only, and for targets with dedicated I2C only,

No, it's not for targets with dedicated I2C only, but it may not do the first initialization there.

It makes things nice and simple on 

>  I'm "almost" inclined to say leave things as they are, as we are only creating more special cases, internal, dedicated external, flexport external, etc.

I'm not necessarily saying we shouldn't factor bits of internal initialization to HAL, but I think that should be carefully thought about.

Right now, if we change something with HMC5883 initialization or external alarms we will need to edit every target.  Factoring it to the HAL means we won't.  It's a big net removal of code.

> Why do I have to enable I2C separately on Quanton to use an external mag, and not on BrainRE1 or AQ32, for example.

I agree it's a little unfortunate-- but that's how it is from an end-user perspective in the current codebase in sparky2, aq32, and this brainre1 commit is now different (with the selection of baro type)-- isn't it?  It's because the port can have multiple roles (just like we have with flexi-io on revolution or receiverport on naze).  It'd be nice if uavo could enforce inter-field dependencies to prevent illegal configs, but that sounds tricky.

flight/PiOS/Common/pios_hal.c
mlyle: Oooh.. nice catch :P

flight/PiOS/Common/pios_hal.c
mlyle: thx :)

mlyle:
Also changed it so that we return positive (e.g. 1) if _NONE is set, which will make it easy to do if (... > 0) { init-internal-mag; }

flight/PiOS/Common/pios_hal.c
mlyle: Added break.. it was backwards, it fell through to assert if init succeeded ;)

mlyle:
@jihlein -- are you opposed to this still?  What can be done to make it better so it can merge?  I really don't want to duplicate all of this external-mag-code between targets.

jihlein:
@mlyle - no, if it's understood its for external only, good to go.

mlyle:
@jihlein can you easily (or have you) tested this on your baro setup?

(I know you have build problems but can ask for a jenkins build)

jihlein:

jihlein:
@mlyle - the Jenkins build runs, but can't be tested as is, it's missing some of the GCS changes, I can't select the external baro or external mag.  

edit: I take that back, the proper UAV objects exist on hwBrainRE1, and using the browser, everything works.  The user interface page needs to be updated, that's why the slection boxes are blank.

So partial success........

mlyle:
Ack!  I'd forgotten the hardware page dependency.  I will do that soon.
Is there any chance you could try it using the hardware browser ?

jihlein:
Works via hardware browser, once I thought of that. :)

mlyle:
@jihlein Danke!  Will work on the GCS stuff soon.

mlyle:
@jihlein has gracefully said he may be able to do the UI part of this :D

jihlein:
Sent a PR with the UI modifications, let me know if it arrives correctly.

tracernz:
Looks good @jihlein. :+1:

mlyle:
Thanks @jihlein! :D 

I just need a quick pass of test on the UI part of this myself before I call it good/ready for merge.

mlyle:
I am happy with this. :)

mluessi:
LGTM, also just tested with BMP280 on RE1. Merging.

PR #1277, opened 2016-09-02
small SAFETY fixes to tx_control.

mlyle:


mlyle:
Before, unchecking "take control" from GCS receiver (which resulted in invalid settings would leave the aircraft HALF disarmed.  UAVOs disarmed (flightstatus), blinking disarmed, etc.  But if control became valid again it would be immediately armed.  Fixed with this change.

mlyle:
jenkins, test this please

tracernz:
So you can still hand-over between GCS control and tx control without disarming?

mlyle:
Yes-- if it's valid.  The issue here was if you trigger the "invalid rx" check-- from the wrong return code from the driver or whatever-- it's supposed to immediately disarm.  But it only ... kind-of disarmed.  Motors would stop, etc, but if the rx became valid again it would be immediately armed.

tracernz:
Let's rewrite this whole mess next release like we discussed.

PR #1278, opened 2016-09-02
PiOS:F4:Fix fail in #1276

tracernz:
Epic fail. :cry:


mlyle:
Super fail that i merged it too
I saw one checkmark, and i think jenkins hadn't started building yet... itchy trigger finger

PR #1279, opened 2016-09-02
GCS: New Arming Settings Tab

### LINKED ISSUE #447: GCS:Config: Rework arming settings tab
Try to make it more intuitive and self-explanatory rather than requiring large amounts of text.

Relates to feedback in #446.

### LINKED ISSUE #1235: Expose InavlidRxArmedTimeout in UI / set it in wizard
See #1234

### LINKED ISSUE #1266: GCS: need to make it obvious not to use non-delay arming modes
This is a big source of support overhead and user frustration.  Needs a fix for Quixote.

tracernz:
@mlyle This is what I have to date. Spent the bulk of my time tonight on back-end stuff (in configtaskwidget, the warnings stuff etc.) so the front-end will be nicer and easier.

TODO:
- Make "Your vehicle is armed by" and "throttle must be low" checkboxes actually do something (others are plumbed. Note that the throttle one is read-only for stick arming modes, might be best to just hide it and add "throttle low and" in front of the options).
- Plumb the checkboxes to disable/zero stuff
- Fill the above items from the UAVO values
- Add more warnings

How it looks so far:
Image of non-Yaktocat

Fixes #1266 
Edit by MPL: Fixes #1235
Fixes #447


mluessi:
This looks very nice :)

mlyle:
Oooh I like the pink background thing.  You've fixed everything I could ever object to :D

jhitesma:
Looks like a huge pile of win to me!  It would be nice to have the warning text inside the pink box, (maybe as a tooltip when you mouse over the warning icon?) but even without that it's a huge improvement over what we have now that really helps clarify things.

Treeed:
As someone who had delay disabled accidentally too I must say that this looks really awesome! 
However I would change "for" to something like "delay until aircraft is armed" as making it a full sentence is nice but hard to understand like this. Especially with switcharming "Your vehicle is armed by 'arming switch' for '2000ms' " sounds weird and misleading, especially with the "for" being placed without too much connection to the above. 
Also: When using switch arming would you disable the "disarmed after 2000ms" or set it automatically to zero even if the user previously changed it? I don't think a disarming delay makes much sense on a switch, but someone might still want it. It also would be a little weird to make a warning for it as it is not critical if someone has it enabled.

tracernz:
We currently enforce a disarm delay of 100 ms on switch. It stops a single bad rx frame from disarming you mid-air which is good for PPM, and S.BUS (no checksum).

mlyle:
Yes-- for clarity-- configured value for disarm time doesn't matter with switches.  It's always 100ms.

tracernz:
Oh, and the UI is dynamic, so it won't show you stuff that isn't applicable to your arming mode.

tracernz:
Have done some more work on this in the last couple of days. Mostly backend stuff to make the front end nicer. Most likely not able to finish it until Monday.

mlyle:
WOAH GITHUB TELLS YOU CONFLICTING FILES NOW.

No rush.  I need to finish up the mpl-buzzbuzz branch of which i've done very little

tracernz:
Uh oh, not a conflict! :O

tracernz:
Rebased, hacked on this a lot. It's quite ugly in places... in order to accommodate the current structure of the UAVO. We should restructure the UAVO to better represent the options when we refactor ManualControl (that will allow us to clean up a lot of the C code for switch arming as well as making some things more general with less code).

I think it works okay now, but no doubt there are a couple of issues I haven't identified.

ground/uavobjgenerator/generators/gcs/uavobjectgeneratorgcs.cpp
```diff
@@ -302,10 +302,11 @@ bool UAVObjectGeneratorGCS::process_object(ObjectInfo* info)
         QString varElemName = info->fields[n]->name + "ElemNames";
         finit.append( QString("    QStringList %1;\n").arg(varElemName) );
         QStringList elemNames = info->fields[n]->elementNames;
-        for (int m = 0; m < elemNames.length(); ++m)
+        for (int m = 0; m < elemNames.length(); ++m) {
```
mlyle, line 305: This looks like something I'd do ;)

ground/gcs/src/plugins/uavobjectwidgetutils/smartsavebutton.cpp
```diff
@@ -134,7 +134,7 @@ void smartSaveButton::processOperation(QPushButton * button,bool save)
 
         // We only allow to send/save Settings objects using this method
         if (!obj->isSettings()) {
-            qDebug() << "[smartsavebutton.cpp] Error, tried to apply/save a non-settings object";
+            qDebug() << "[smartsavebutton.cpp] Error, tried to apply/save a non-settings object: " << obj->getName();
```
mlyle, line 137: Excellent

ground/uavobjgenerator/generators/gcs/uavobjectgeneratorgcs.cpp
```diff
@@ -343,22 +344,28 @@ bool UAVObjectGeneratorGCS::process_object(ObjectInfo* info)
...
                           .arg(info->fields[n]->name + "EnumOptions")
                           .arg(info->fields[n]->name + "EnumIndices")
-                          .arg(info->fields[n]->limitValues));
+                          .arg(info->fields[n]->limitValues)
+                          .arg(defaultValuesInit));
         }
         // For all other types
         else {
-            finit.append( QString("    fields.append( new UAVObjectField(QString(\"%1\"), QString(\"%2\"), UAVObjectField::%3, %4, QStringList(), QList<int>(), QString(\"%5\"), FIELD_DESCRIPTIONS[\"%1\"]));\n")
+            const QString defaultValuesInit = info->fields[n]->defaultValues.join(',');
```
mlyle, line 360: This is really cool also :)

ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.h
```diff
@@ -133,8 +152,8 @@ class UAVOBJECTWIDGETUTILS_EXPORT ConfigTaskWidget: public QWidget
 
     void addWidgetToDefaultReloadGroups(QWidget * widget, QList<int> *groups);
 
-    bool addShadowWidget(QWidget * masterWidget, QWidget * shadowWidget,double shadowScale=1,bool shadowIsLimited=false);
-    bool addShadowWidget(QString object,QString field,QWidget * widget,int index=0,double scale=1,bool isLimited=false, QList<int> *defaultReloadGroups=NULL, quint32 instID=0);
+    //bool addShadowWidget(QWidget *masterWidget, QWidget *shadowWidget, double shadowScale = 1, bool shadowIsLimited = false, bool shadowUseUnits = false);
```
mlyle, line 155: Nuke this comment?

ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
```diff
@@ -1520,6 +1518,91 @@ bool ConfigTaskWidget::getCheckedFromOption(QWidget *widget, QString option)
...
+    case PREFIX_MEGA:
+        return "M" + units;
+    case PREFIX_GIGA:
+        return "G" + units;
+    default:
+        return QString();
+    }
+}
+
+bool ConfigTaskWidget::resetWidgetToDefault(QWidget *widget)
```
mlyle, line 1597: 🎉

ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
```diff
@@ -1250,53 +1243,55 @@ QVariant ConfigTaskWidget::getVariantFromWidget(QWidget *widget, double scale)
...
     } else if (QGroupBox *groupBox = qobject_cast<QGroupBox *>(widget)) {
         groupBox->setChecked(getCheckedFromOption(widget, value.toString()));
         return true;
-    } else if (QCheckBox *checkBox=qobject_cast<QCheckBox *>(widget)) {
+    } else if (QCheckBox *checkBox = qobject_cast<QCheckBox *>(widget)) {
         checkBox->setChecked(getCheckedFromOption(widget, value.toString()));
         return true;
-    } else if (QLineEdit *lineEdit=qobject_cast<QLineEdit *>(widget)) {
+    } else if (QLineEdit *lineEdit = qobject_cast<QLineEdit *>(widget)) {
+        // TODO: units, will need to peel off on the other side
```
mlyle, line 1283: Makes me nervous.  Maybe better to not do this one for now?

mlyle (commented):
This looks really good-- I like the changes to the UAVO infrastructure.  I need to systematically test this– and look very closely at the UI widgets for usability– which I will try and do later tonight.

mlyle:
Looking at it in UI.
1. Probably it should expand to full width or be centered.  It looks a little wonky on a big screen.
2. "your vehicle will remain armed indefinitely when receiver is disconnected" -- the throttle timeout still applies.. a better wording would be nice.
3. It looks like left vs. right is reversed in the arming.  Or I confused myself.
4. Calibrate gyros for: should have a space after it or other gap.  Perhaps put "before arming" after the label, so it's clear this will hesitate.

I tested all the combinations I could think of, and all I saw wrong is the control reversal.  The look/feel nits are minor.

PR #1283, opened 2016-09-03
autotune: outer loop more sedate on extreme setups

### LINKED ISSUE #1280: We seem to overtune outer loop on low tau quads
Either need a stop-gap to clamp outer kp/ki or some real improvement, probably.

mlyle:
Also limit traverse speeds a little higher to prevent ourselves from clamping
and adding phase modulation, and water down the outer Ki term a little.

Fixes #1280


tracernz:
Let's do it.

Issue #1284, opened 2016-09-03
opengl init not working on windows software accelerated env


mlyle:
reported by @jhitesma 

![image](https://cloud.githubusercontent.com/assets/90903/18223175/815021aa-7163-11e6-8dde-b8f5f15aa448.png)

tracernz:
Unfortunately can't repro this running under RDP (RDP requires ANGLE/GLES).

mlyle:
No repro, but not ready to close either.

jhitesma:
Any benefit to me trying a more recent build?  I assume nothing has changed but haven't tried to repo with any newer builds in a few weeks.

mlyle:
Nothing has changed.

How much RAM do you have?  I wonder if that is really an out-of-memory error.

jhitesma:
That machine only has 4G, and it's hurting.  I could try with everything else shut down to see if it works then.

mlyle:
Would be helpful to see if it cures it-- then we could easily close this.

jhitesma:
Just tried it with everything else shut down (even chrome's background stuff) and same error.

Watching Taskmanger my memory usage never went past 56% 
So doesn't look like a memory issue. 

The video card on this computer is old enough that there aren't official drivers for it for W10 so I'm not surprised it's having issues related to video.

tracernz:
Playing with different Qt OpenGL backends a bit... GLC_Lib doesn't play nice with any except the software one. Will be good to get rid of it soon (WIP) and then we can probably fix this quite easily.

mlyle:
Still broked-- too bad

tracernz:
#1540 should be interesting.

tracernz:
This was fixed for @jhitesma by https://github.com/d-ronin/dRonin/pull/1540

mlyle:
Not seen in 5ever

PR #1286, opened 2016-09-03
GCS: Link right OpenGL lib on windows

tracernz:
Doesn't make things worse for me (can run with ANGLE under RDP or with NVIDIA drivers and a real screen still). This is the correct lib for MSVC (as opposed to MinGW) according to http://doc.qt.io/qt-5/windows-requirements.html

### DIFF
--- a/ground/gcs/src/libs/glc_lib/glc_lib.pro
+++ b/ground/gcs/src/libs/glc_lib/glc_lib.pro
@@ -490,6 +490,8 @@ OTHER_FILES += \
     qtc_packaging/debian_harmattan/compat \
     qtc_packaging/debian_harmattan/changelog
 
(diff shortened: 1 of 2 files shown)


tracernz:

I'm skeptical but it could help with #1284

mlyle:
@jhitesma if you could try this later to see if it's any different I'd really appreciate it.

jhitesma:
Still get this:

image

mlyle:
@tracernz -- do you still want to merge this even though it's nofix?  I'm fine either way (not sure it makes any functional change)

tracernz:
Not sure it does anything, but it doesn't make it worse and it's recommended it Qt docs... so we probably should merge it?

PR #1292, opened 2016-09-03
Add link for OpenPilot Revolution at RTFQ

russCloak:
RTFQ (ready to fly quads [dot] com) is a pretty reliable place to get the cheapest revo FC around. It seems right to me that it should be where the link goes. 

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ See [dronin.org](http://dronin.org) for more details.
 - [BrainFPV](http://brainfpv.com/)
 - [Lumenier Lux](http://www.getfpv.com/lumenier-lux-flight-controller.html)
 - [DTFUHF DTFc](http://www.dtfuhf.com/)
-- OpenPilot Revolution
+- [OpenPilot Revolution](http://www.readytoflyquads.com/openpilot-cc3d-revolution-acro)
 - Quantec Quanton
 - Tau Labs Sparky
 - [Tau Labs Sparky2](https://github.com/TauLabs/TauLabs/wiki/Sparky2)


mlyle:
I don't think we'd like to add a RTFQ link, for a couple of reasons---
1. They randomly don't fulfill orders and don't answer emails and require paypal claims for resolution.  Several members of our community have experienced this.
2. They sell Quanton clones in violation of the Quanton design's license http://www.readytoflyquads.com/quanton-flight-control
3. They don't even seem to sell a full revolution-- just clones without the full sensor suite and telemetry radio.

mlyle:
(p.s. that acro variant we just started supporting a few days ago and is not supported in any releases)

cpusmith:
@mlyle Totally agree with #1. I don't shop there anymore.

yamahabest:
Haha. reliable?
Placed an order for 200+ USD at RTFQ, never heard again from them. In the end I had to file a claim at PayPal to get my money back.

russCloak:
> Placed an order for 200+ USD at RTFQ, never heard again from them. In the end I had to file a claim at PayPal to get my money back.

Wow, I'm very surprised to hear that - I've never had any issues with them and I know many other people that have similar stories to mine. I actually know one of the reps from there and I've heard about issues in the past but I thought that was long gone now.

mlyle:
Most of the time it's OK.  But from what I've seen, if you keep ordering it'll happen eventually.

yamahabest:
Others could have other experiences, yes, however for me it was my first and last order there.

PR #1295, opened 2016-09-04
GCS: Allow disabling of debug gadget by env var

tracernz:
Setting NO_DEBUG_GADGET allows it to be disabled so QDebug messages will go through to QtCreator or shell. This allows use of QT_MESSAGE_PATTERN to get originating file/line/function etc. Super helpful for debugging!

Example with QT_MESSAGE_PATTERN = `[%{file}:%{line}] %{function}: %{message}`
`[/home/mike/dev/dronin2/ground/gcs/src/plugins/systemhealth/systemhealthgadgetwidget.cpp:128] SystemHealthGadgetWidget::updateAlarms: [SystemHealth] Warning: The SystemHealth SVG does not contain a graphical element for the  "GyroBias"  alarm.`

### DIFF
--- a/ground/gcs/src/plugins/debuggadget/debuggadgetfactory.cpp
+++ b/ground/gcs/src/plugins/debuggadget/debuggadgetfactory.cpp
@@ -29,6 +29,7 @@
 #include "debuggadget.h"
 #include <coreplugin/iuavgadget.h>
 #include <QTextStream>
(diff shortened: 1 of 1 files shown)


mlyle:
Yay.  I've just removed debug gadget manually to accomplish this in the past, but this is better.

PR #1296, opened 2016-09-04
GCS:Debug: Improve debug widget, connect usagestats

tracernz:
Configurable log level in the usage stats settings.

TODO: make sure DEFINES += QT_MESSAGELOGCONTEXT doesn't have a bad impact on release builds.

TODO in future: Make DebugGadget able to filter log messages by severity, source (class?). Then make stuff emit debug messages all the time rather than defined at compile time.

### DIFF
--- a/ground/gcs/gcs.pri
+++ b/ground/gcs/gcs.pri
@@ -109,6 +109,9 @@ DEFINES += QT_NO_CAST_TO_ASCII
 #DEFINES += QT_USE_FAST_OPERATOR_PLUS
 #DEFINES += QT_USE_FAST_CONCATENATION
 
+# include additional information in debug log (file/line/function)
+DEFINES += QT_MESSAGELOGCONTEXT
+
 unix {
     CONFIG(debug, debug|release):OBJECTS_DIR = $${OUT_PWD}/.obj/debug-shared
     CONFIG(release, debug|release):OBJECTS_DIR = $${OUT_PWD}/.obj/release-shared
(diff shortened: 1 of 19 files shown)


mlyle:

tracernz:
Force-pushed without the random line from another changeset :(

mlyle:
```
/var/lib/jenkins/jobs/dronin/workspace@2/Nodes/master/ground/gcs/src/plugins/debuggadget/debuggadgetfactory.cpp: In function ‘void customMessageHandler(QtMsgType, const QMessageLogContext&, const QString&)’:
/var/lib/jenkins/jobs/dronin/workspace@2/Nodes/master/ground/gcs/src/plugins/debuggadget/debuggadgetfactory.cpp:58:116: error: ‘level’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
     DebugEngine::getInstance()->message(level, msg, QString(context.file), context.line, QString(context.function));
```

tracernz:
GCC is so dumb sometimes. 😢 CLang and MSVC both got it right. Interesting that it passed the normal build but not artifacts with GCC... :/

mlyle:
> . Interesting that it passed the normal build but not artifacts with GCC... :/

Optimizer

mlyle:

tracernz:

tracernz:
Should put timestamps on these so they can be sewn into the usage details if desired.

tracernz:
Something not quite right with the QMake files in UsageStats. Will check/correct this evening.

tracernz:

This should be good now.

tracernz:
Made another change. Now install the message handler always when the DebugGadget plugin loads, not just when the widget is created. This allows us to collect debug logs in UsageStats even when the DebugGadget widget isn't in the users workspaces.

PR #1300, opened 2016-09-05
gcs build: change how we enforce qt versions

mlyle:
- Make sure the exactly correct version is used to build GCS
- Only warn if not in tools dir at beginning of make.

This has a number of purposes.  It allows building flight without
exactly-matching Qt version on Raspberry Pi (because getting the right
Qt version on Pi is often a pain).  It also checks the version used
more strongly where mismatches would be ignroed before (e.g. on
Windows).

### DIFF
--- a/ground/gcs/gcs.pro
+++ b/ground/gcs/gcs.pro
@@ -20,4 +20,9 @@ unix:!macx:!isEmpty(copydata) {
 	QMAKE_EXTRA_TARGETS += bin_install
 }
 
+!equals(QT_MAJOR_VERSION, 5) | !equals(QT_MINOR_VERSION, 6) | !equals(QT_PATCH_VERSION, 1) {
+message("Cannot build dRonin GCS with Qt version $${QT_VERSION}.")
+error("Use 5.6.1 (make qt_sdk_install).")
+}
+
 copydata.file = copydata.pro
--- a/ground/gcs/gcs.pro
+++ b/ground/gcs/gcs.pro
@@ -1,9 +1,3 @@
-#version check qt
-contains(QT_VERSION, ^[0-4]\\..*) {
-    message("Cannot build GCS with Qt version $${QT_VERSION}.")
-    error("Cannot build GCS with Qt version $${QT_VERSION}. Use at least Qt 5.0.1!")
-}
-
 cache()
 
 include(gcs.pri)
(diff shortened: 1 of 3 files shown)


mlyle:

mlyle:
jenkins agent was disconnected.
jenkins, test this please, also

make/tools.mk
```diff
@@ -25,14 +25,11 @@ endif
...
-  ifndef WINDOWS
-    # Check for a current QT SDK dir, abort without
-    ifeq ($(wildcard $(QT_SDK_DIR)/*),)
-      ifeq (,$(findstring _install,$(MAKECMDGOALS)))
-        $(error "QT SDK not found, please run `make qt_sdk_install`")
-      endif
+ifndef WINDOWS
+# Check for a current QT SDK dir, abort without
+  ifeq ($(wildcard $(QT_SDK_DIR)/*),)
+    ifeq (,$(findstring _install,$(MAKECMDGOALS)))
```
tracernz, line 31: Could also do for example `ifneq ($(filter-out %_install %_clean, $(MAKECMDGOALS)),)` to allow a (hopefully small) _list_ of exceptions. That would allow you to clean a previous tool installation before attempting a new one.

ground/gcs/gcs.pro
```diff
@@ -20,4 +20,9 @@ unix:!macx:!isEmpty(copydata) {
 	QMAKE_EXTRA_TARGETS += bin_install
 }
 
+!equals(QT_MAJOR_VERSION, 5) | !equals(QT_MINOR_VERSION, 6) | !equals(QT_PATCH_VERSION, 1) {
+message("Cannot build dRonin GCS with Qt version $${QT_VERSION}.")
```
tracernz: I'm not sure we even need to worry about the patch level, but I suppose it doesn't hurt (most of the time). Could just be a warning if the patch level doesn't match for example.

mlyle: I accidentally shipped artifacts before with the "wrong" patch level windows.  I prefer it be fatal for release builds at least.

tracernz: Alright, fair enough. :+1:

make/tools.mk
mlyle: The main purpose of this is so you can build sim-- if you're doing that, you deserve the warning, but it's OK if it proceeds/succeeds IMO.

tracernz:
Note: We also require Qt to build flight (because it depends on the UAVO generator). Are we happy to build it with old Qt? Could it cause problems? Should add a check to uavobjgenerator.pro to ensure at least Qt5 is available IMO.

mlyle:
QT4 blows up fairly theatrically, but I can add an explicit check there too.

mlyle:

tracernz:
This broke my doc builder :cry:. It was able to build uavo generator before with the system Qt install and IGNORE_MISSING_TOOLCHAIN set.

```
+ make -j2 docs
/var/lib/jenkins/workspace/dronin-docs/make/tools.mk:32: "QT SDK not found, please run `make qt_sdk_install`"
Project ERROR: Use QT5 (make qt_sdk_install).
make: *** [uavobjgenerator] Error 2
```

PR #1305, opened 2016-09-06
PlayUAVOSD Initial Support

tracernz:
Early support, boots and runs OSD, communicates with MSP for most basic stuff. No support for fancy stuff like video input switching yet. A few rough edges in the code.

Here it is running dRonin firmware and talking to a Naze32 running Cleanflight (it talks to dRonin too):
https://gfycat.com/GentleCloseDogwoodclubgall

Hardware files from https://github.com/PlayUAV/PlayuavOSD/tree/master/hardware, assumed the license one level up applies as there isn't one in this dir or a mention of one in the files.


flight/targets/playuavosd/fw/Makefile
```diff
@@ -0,0 +1,326 @@
...
+SRC += $(MGRSLIB)/polarst.c
+
+# MSP protocol lib
+SRC += $(MSPLIB)/msp.c
+
+## PIOS Hardware (STM32F4xx)
+include $(PIOS)/STM32F4xx/library_chibios.mk
+# SRC += $(PIOS)/STM32F4xx/pios_video.c
+
+## PIOS Hardware (Common)
```
mlyle, line 129: Should trim most of this down.

flight/targets/playuavosd/fw/board.h
```diff
@@ -0,0 +1,68 @@
...
+#include "pios_chibios_transition_priv.h"
+
+/*
+ * Setup for STMicroelectronics STM32F4 (sparky2)
+ */
+
+/*
+ * Board identifier.
+ */
+#define BOARD_SPARKY2
```
mlyle: :)

ground/gcs/src/plugins/boards_dronin/playuavosd.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+QString PlayUavOsd::boardDescription()
+{
+    return QString("PlayUAVOSD running dRonin firmware");
+}
+
+bool PlayUavOsd::queryCapabilities(BoardCapabilities capability)
+{
+    switch (capability) {
+    case BOARD_CAPABILITIES_OSD:
```
mlyle, line 68: It ought to be upgradable, too, right?

flight/targets/playuavosd/fw/board.h
tracernz: heh, thanks :)

ground/gcs/src/plugins/boards_dronin/playuavosd.cpp
tracernz: Yeah, can't see why not.

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -1618,12 +1621,13 @@ int32_t OnScreenDisplayInitialize(void)
...
-	StateEstimationNavigationFilterGet(&filter);
-	if (filter != STATEESTIMATION_NAVIGATIONFILTER_NONE) {
-		has_nav = true;
-	} else {
-		has_nav = false;
+	has_nav = false;
+	if (StateEstimationHandle()) {
+		uint8_t filter;
+		StateEstimationNavigationFilterGet(&filter);
+		if (filter != STATEESTIMATION_NAVIGATIONFILTER_NONE) {
```
mlyle, line 1628: This test is probably wrong for playuav.

Makefile
```diff
@@ -961,6 +961,15 @@ $(foreach board, $(EF_BOARDS), $(eval $(call EF_TEMPLATE,$(board),$($(board)_fri
...
 $(foreach board, $(UP_BOARDS), $(eval $(call UP_TEMPLATE,$(board),$($(board)_cpuarch),$($(board)_short))))
 
+bu_playuavosd: bu_playuavosd_px4
+
+.PHONY: bu_playuavosd_px4
+bu_playuavosd_px4: BOARD_ROOT_DIR=$(ROOT_DIR)/flight/targets/playuavosd
+bu_playuavosd_px4: OUTDIR=$(BUILD_DIR)/bu_playuavosd
+bu_playuavosd_px4: bu_playuavosd_tlfw
+	$(V0) @echo "PX4_MKFW    $*"
+	$(V1) $(BOARD_ROOT_DIR)/px_mkfw.py --image $(OUTDIR)/bu_playuavosd.padded.bin --board_id 90 > $(OUTDIR)/bu_playuavosd.px4
```
mlyle, line 971: we should package this

mlyle:

flight/Modules/OnScreenDisplay/onscreendisplay.c
tracernz: Yes, but there is also no StateEstimationHandle so it doesn't get that far (although that's a trap for it one day when it might).

Makefile
tracernz: Yes. I wonder if we can easily put something together to go the other way, at least some instructions. I will look into opening a PR against the playuav software to allow it to flash our stuff too

flight/PiOS/inc/pios_max7456_priv.h
```diff
@@ -0,0 +1,223 @@
...
+#define PIOS_MAX7456_VM0_VBE_DISABLE   0b1
+
+#define PIOS_MAX7456_VM0_SRB_MASK      (1 << 1)
+#define PIOS_MAX7456_VM0_SRB_R(val)    ((val >> 1) & 0b1)
+#define PIOS_MAX7456_VM0_SRB_W(regval, val) ((regval & ~PIOS_MAX7456_VM0_SRB_MASK) | ((val & 0b1) << 1))
+#define PIOS_MAX7456_VM0_SRB_RESET     0b1
+#define PIOS_MAX7456_VM0_SRB_CLEAR     0b0
+
+#define PIOS_MAX7456_VM0_VS_R(val)     ((val >> 2) & 0b1)
+#define PIOS_MAX7456_VM0_VS_W(val)     ((val & 0b1) << 2)
```
tracernz: Need to fix some of these up.

mlyle: I wouldn't worry about this too much-- since I need to unify max7456 stuff.

flight/targets/playuavosd/fw/Makefile
tracernz: In future, or in this?

Makefile
tracernz: Packaged.

tracernz:
Comments address, a few further tidy-ups.

flight/targets/playuavosd/fw/Makefile
mlyle: Seems like it'd be preferable to not build receivers, etc-- just lopping out the obvious things.  On Windows, most of the build time is launching processes.

tracernz: Maybe, running a receiver into the OSD is not that crazy of a thing to do though, esp. if you have separate camera operator + pilot :P, and also might be desirable for some firmwares depending on available telemetry messages. ATM they are unused though.

mlyle: Does it even have an appropriate port to do this on?

tracernz: I need to make a spreadsheet of all the pins and all their AFs, then I'll get back to you on that. :stuck_out_tongue:

mlyle:
Looks reasonable.  I still think we should prune the files built, but can be in a subsequent PR.  Another really important thing to have is to have some kind of no-connection flag on OSD.  But screw it, let's merge!@!@

PR #1307, opened 2016-09-07
python-logviewer:  handle empty data series more gracefully

mlyle:
This is important because the default profile doesn't log actuators..
It's better to not crash under these circumstances.  Thanks to Troy
from IRC for the report.


tracernz:
We should probably start unit-testing some of this stuff, and write tests that bring things like this out of the woodwork earlier. :grey_question:

mluessi:
LGTM, thanks

PR #1308, opened 2016-09-07
Package: Add package_flight dep on FW_FILES

tracernz:
a) makes the zip get updated if an FW file changes and it already exists, stops people inadvertantly making packages with old FW (problem for 3rd parties rather than dRonin build machines)
b) stops us unknowingly shipping an incomplete package if something is missing (zip silently ignores missing input files)

### DIFF
--- a/package/Makefile
+++ b/package/Makefile
@@ -136,6 +136,6 @@ ground_package_common: standalone
 package_flight: $(FLIGHTPKGNAME)
 
 $(FLIGHTPKGNAME): $(PACKAGE_DIR) $(FW_FILES)
-	$(ZIPBIN) -j $@ $(FW_FILES)
+	$(ZIPBIN) -j --must-match $@ $(FW_FILES)
 
 include $(WHEREAMI)/Makefile.$(PLATFORM)
--- a/package/Makefile
+++ b/package/Makefile
@@ -135,7 +135,7 @@ ground_package_common: standalone
 .PHONY: package_flight
 package_flight: $(FLIGHTPKGNAME)
 
-$(FLIGHTPKGNAME): $(PACKAGE_DIR)
+$(FLIGHTPKGNAME): $(PACKAGE_DIR) $(FW_FILES)
 	$(ZIPBIN) -j $@ $(FW_FILES)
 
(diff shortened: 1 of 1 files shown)


tracernz:
`/var/lib/jenkins/jobs/dron/workspace@3/Nodes/master/build/bu_brainre1/bu_brainre1.bin.firmwareinfo.c:75:24: error: invalid suffix "xNone" on integer constant
  .next_ancestor_hash = 0xNone,
                        ^
make[1]: *** [/var/lib/jenkins/jobs/dron/workspace@3/Nodes/master/build/bu_brainre1/bu_brainre1.bin.firmwareinfo.o] Error 1
make: *** [bu_brainre1_tlfw] Error 2`

Don't get this on my machines?

mlyle:
that's an intermittent build problem caused when the hash vanishes underneath (e.g. a merge to next).

jenkins, test this please

mlyle:

tracernz:
Alternately, from the zip man page:

> **-MM**
> **--must-match**
> All input patterns must match at least one file and all input files found must be readable. Normally when an input pattern does not match a file the "name not matched" warning is issued and when an input file has been found but later is missing or not readable a missing or not readable warning is issued. In either case zip continues creating the archive, with missing or unreadable new files being skipped and files already in the archive remaining unchanged. After the archive is created, if any files were not readable zip returns the OPEN error code (18 on most systems) instead of the normal success return (0 on most systems). With -MM set, zip exits as soon as an input pattern is not matched (whenever the "name not matched" warning would be issued) or when an input file is not readable. In either case zip exits with an OPEN error and no archive is created.
> This option is useful when a known list of files is to be zipped so any missing or unreadable files will result in an error. It is less useful when used with wildcards, but zip will still exit with an error if any input pattern doesn't match at least one file and if any matched files are unreadable. If you want to create the archive anyway and only need to know if files were skipped, don't use -MM and just check the return code. Also -lf could be useful.

Sounds perfect, and even intended for this use-case.

mlyle:
Maybe both :D

mlyle:

mlyle:

PR #1311, opened 2016-09-08
Flight: Brain - Correct pios_board.h

jihlein:
Moved the line "#endif /\* STM3210E_INS_H_ */" to the end of the file so that if pios_board.h is included in multiple files, a structure redefined error won't be generated for pios_osd_bw_cfg_t.

Builds correctly on next.  Not run on Brain hardware since I don't have that target available.  Minimal risk, but best if someone can try it.


mlyle:
Looks safe enough to merge blind.  Yay for idempotency.

tracernz:
Probably wouldn't be a bad idea to change the name to something that doesn't only make sense for a board from 2010 that never went into production. :laughing:

PR #1313, opened 2016-09-08
Flight: Don't pass header includes to linker

tracernz:
Works around Windows CreateProcess command length limit ([32768](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425%28v=vs.85%29.aspx)) (sigh, another Windows workaround).
A little bit of a hacky way to do it, but effective.


tracernz:

PR #1314, opened 2016-09-08
gcs: output: support 560Hz servos

### LINKED ISSUE #984: GCS:Output: Should support servo rates higher than 400 Hz
For 760 us servos.

mlyle:
Fixes #984


tracernz:
Great, works fine with the conversion to integer: https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/config/configoutputwidget.cpp#L586

mlyle:
Yup, I liked that when implementing 16666.

Issue #487, opened 2016-01-21
Deprecate *SUM, move to SRXL


tracernz:
Does anybody/would anybody use a driver? I wrote one last night but I'm not sure if anybody would use it. It needs a little cleanup and then I can submit a PR if it's wanted.

tracernz:
Supported by JR, Jeti, Multiplex, and possibly some other radios. I have tested framerates between 5 ms and 30 ms.

tracernz:
The diff: https://github.com/d-ronin/dRonin/compare/next...tracernz:srxl?expand=1

mlyle:
How much code size does it use?

tracernz:
next:

```
/home/mike/Dev/dronin/build/fw_sparky/fw_sparky.elf  :
section                   size        addr
startup                    392   134266880
.text                   209312   134267296
_exti                       44   134476608
.initcalluavobj.init         0   134476652
.initcallmodule.init       208   134476652
.ARM.exidx                   8   134476860
.textalign                   4   134476868
.boardinfo                   0   134233984
.stacks                   2048   268435456
.ccm                        56   268437504
.fastheap                 6088   268437560
.data                     1736   536870912
.bss                      8800   536872648
.heap                    30424   536881448
.debug_info             727852           0
.debug_abbrev           139895           0
.debug_aranges           27336           0
.debug_ranges            30776           0
.debug_line             278006           0
.debug_str              228045           0
.comment                   112           0
.ARM.attributes             55           0
.debug_frame             69708           0
.debug_loc              279829           0
.debug_macro             32709           0
Total                  2073443
```

srxl:

```
/home/mike/Dev/dronin/build/fw_sparky/fw_sparky.elf  :
section                   size        addr
startup                    392   134266880
.text                   209936   134267296
_exti                       44   134477232
.initcalluavobj.init         0   134477276
.initcallmodule.init       208   134477276
.ARM.exidx                   8   134477484
.textalign                   4   134477492
.boardinfo                   0   134233984
.stacks                   2048   268435456
.ccm                        56   268437504
.fastheap                 6088   268437560
.data                     1736   536870912
.bss                      8804   536872648
.heap                    30420   536881452
.debug_info             730387           0
.debug_abbrev           140714           0
.debug_aranges           27400           0
.debug_ranges            30880           0
.debug_line             278833           0
.debug_str              228650           0
.comment                   112           0
.ARM.attributes             55           0
.debug_frame             69884           0
[...]

PR #1316, opened 2016-09-08
hsum: eliminate scaling to 1000-2000 range

### LINKED ISSUE #912: Flight: SUMD: Strange stick scaling on inputs
This line: https://github.com/d-ronin/dRonin/blob/080150a5b271fcfb3fe36aa1e20f92fc0d956aca/flight/PiOS/Common/pios_hsum.c#L255

has the effect of scaling SUMD's inputs around a center value of 1500, the protocol is described here on page 4: http://www.deviationtx.com/media/kunena/attachments/98/HoTT-SUMD-Spec-REV01-12062012-pdf.pdf

It's likely this scaling is used for another Graupner protocol, SUMD only requires simple /8 on these values.

mlyle:
Leave it raw, so that we can have maximal resolution.

Fixes #912.

This will screw up upgrade for these users, but A) it's a small population, and B) easily fixed, and C) worth the improvement to the resolution.

### DIFF
--- a/flight/PiOS/Common/pios_hsum.c
+++ b/flight/PiOS/Common/pios_hsum.c
@@ -251,8 +251,7 @@ static int PIOS_HSUM_UnrollChannels(struct pios_hsum_dev *hsum_dev)
 			s += sizeof(uint16_t);
 			/* save the channel value */
 			if (i < PIOS_HSUM_NUM_INPUTS) {
-				/* floating version. channel limits from -100..+100% are mapped to 1000..2000 */
-				state->channel_data[i] = (uint16_t)(word / 6.4f - 375);
+				state->channel_data[i] = word;
 			}
 		} else
 			/* this channel was not received */


dustin:
Is it possible for the upgrade process to intentionally break the input
config?

tracernz:
Or scale it when going from rev < this to rev >= this && type == hsum?

mlyle:
Yes, it's totally possible and I thought about doing it.  There's a very small number of users of hsum though and it's completely outside the range so should be safe-ish, so I'd rather just avoid the complexity.

mlyle:
> Or scale it when going from rev < this to rev >= this?

This is surprisingly hard because it's difficult to compare / order git revisions.

tracernz:
Yes, I have played with that a little and mostly failed. :(

tracernz:
Need to do https://github.com/d-ronin/dRonin/issues/487 :wink: . For that I'd like to build a module for my RX tester/generator that generates valid HSUM, so I can at least confirm it's not totally broken without owning Graupner gear. Unless we can track down somebody with real hardware to test!

PR #1319, opened 2016-09-08
setupwizard: don't configure higher channels

### LINKED ISSUE #1317: VehicleConfigurationHelper::applyActuatorConfig configures *ALL* channels
Not just the ones used.

This is super bad when it conflicts with a PPM resource, e.g. on sparky2.

mlyle:
This may require refactor in the future if we do smarter channel
assignment.  Fixes #1317.

This is important so that we don't cook any servos that are hooked up to non-motor inputs, and so that we preserve timer resources that are used for PPM.

Relates to https://forum.dronin.org/forum/d/139-setup-problem/

### DIFF
--- a/ground/gcs/src/plugins/coreplugin/iboardtype.cpp
+++ b/ground/gcs/src/plugins/coreplugin/iboardtype.cpp
@@ -82,4 +82,15 @@ QString IBoardType::getBoardNameFromID(int id)
 
     return "Unknown";
 }
+
+int IBoardType::getBankFromOutputChannel(int channel)
+{
+    for (int i = 0; i < channelBanks.count(); i++) {
+        foreach (const qint32 chan, channelBanks.at(i)) {
+            if (chan == channel)
+                return i;
+        }
+    }
+    return -1;
+}
 }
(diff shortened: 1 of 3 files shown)


tracernz:
Still makes that big assumption that your motors are on pins 1-# :stuck_out_tongue:

mlyle:
Yes, but the wizard does that too.  I think this is the best option short of a full refactor.

mlyle:
This is insufficient.  Timer bank 3 needs to be unconfigured in some cases, but.. some output maps need timer bank 3.

mlyle:

mlyle:
no good, need to figure out why the bank mapping isn't working.

mlyle:

mlyle:
WORKSFORME -- at least quad cases on sparky and revo.

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -82,4 +82,15 @@ QString IBoardType::getBoardNameFromID(int id)
 
     return "Unknown";
 }
+
+int IBoardType::getBankFromOutputChannel(int channel)
+{
+    for (int i = 0; i < channelBanks.count(); i++) {
```
mluessi: Should use `getChannelBanks()` so it picks up channel bank changes. See #1359

mlyle: OK.

PR #1323, opened 2016-09-09
transmitter_control: use expected enum constant

mlyle:
Has the same value, but we used the wrong one.  Fixes build on OSX (and
is more correct).


mlyle:

Will merge upon success if no complaints.

PR #1318, opened 2016-09-08
pios_servo: fix on CC3D/S1 "fake oneshot" mode

mlyle:
This was inverted before.  Big woops.

PR #1324, opened 2016-09-09
firmwareiap: stop PWM before reset (when appropriate)

### LINKED ISSUE #1216: Safety-Related  Motors spin when connecting battery
- Seen in version (release version or githash): f1c776c9
- On flight controller (if applicable): DTFc

Issue details:
  While plugging in the battery on a few occasions the 2 back motors spun.  @mlyle suggested testing the "reboot" button in the GCS with the battery connected to see if they spin.  I tried this 3 times, and all 3 times at least one motor spun up, like it was armed then disarmed immediately.

mlyle:
Fixes #1216, where @ufoDziner has a reliable repro.

Supersedes #720.

Needs to wait for #1323 / build to be fixed before building artifacts.

Conflicts #1318

### DIFF
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -45,7 +45,9 @@ static const struct pios_servo_cfg *servo_cfg;
 
 //! The counter rate for the channel, used to calculate compare values.
 static uint32_t *output_channel_resolution;  // The clock rate for that timer
-enum channel_mode {UNCONFIGURED = 0, REGULAR_PWM, SYNC_PWM} *output_channel_mode;
+static enum channel_mode {UNCONFIGURED = 0, REGULAR_PWM, REGULAR_INVERTED_PWM, SYNC_PWM} *output_channel_mode;
+
+static bool resetting;
 
 /* Private function prototypes */
 static uint32_t timer_apb_clock(TIM_TypeDef *timer);
@@ -248,7 +250,13 @@ void PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_
 			const struct pios_tim_channel *chan = &servo_cfg->channels[j];
 			if (timer_banks[i].timer == chan->timer) {
 				/* save the frequency for these channels */
-				output_channel_mode[j] = (rate == 0) ? SYNC_PWM : REGULAR_PWM;
+				if (rate == 0) {
+					output_channel_mode[j] = SYNC_PWM;
+				} else if (channel_max[j] >= channel_min[j]) {
+					output_channel_mode[j] = REGULAR_PWM;
+				} else {
+					output_channel_mode[j] = REGULAR_INVERTED_PWM;
+				}
 				output_channel_resolution[j] = timer_banks[i].clk_rate;
 			}
 		}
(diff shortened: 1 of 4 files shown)


mlyle:
i don't understand this failure / it doens't happen on my machine.  trying again.

mlyle:
jenkins test this please

mlyle:
jenkins, test this please

mlyle:

flight/PiOS/STM32/pios_servo.c
```diff
@@ -248,7 +250,13 @@ void PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_
...
 			if (timer_banks[i].timer == chan->timer) {
 				/* save the frequency for these channels */
-				output_channel_mode[j] = (out_rate[i] == 0) ? SYNC_PWM : REGULAR_PWM;
+				if (out_rate[i] == 0) {
+					output_channel_mode[j] = SYNC_PWM;
+				} else if (channel_max[j] >= channel_min[j]) {
+					output_channel_mode[j] = REGULAR_PWM;
+				} else {
+					output_channel_mode[j] = REGULAR_INVERTED_PWM;
+				}
```
tracernz, line 259: Doesn't handle inverted SyncPWM, which is a perfectly valid configuration.

mlyle: Doesn't need to.  A brushed motor won't be behind inverted syncpwm.

mlyle:

ufoDziner:
I rebooted my dtfc 5 times while connected to usb, unplugged usb, and battery and did the same thing again, and then for a third time. No motors spun at any time.

mlyle:
@ufoDziner thanks!

mlyle:
rebase'd and force push'd

flight/PiOS/STM32/pios_servo.c
mlyle: For clarity, a real PWM channel can be dangerous to nail "active" because it might be hooked to a transistor and then a load.  For a oneshot type channel, going all-high or all-low should be safe / cause the actuator to disarm.

tracernz:
Needs test with something other than BLHeli ESCs, and ideally a quick check with scope. I intend to do this later today, with a couple of other ESCs, and servos.

flight/PiOS/STM32/pios_servo.c
tracernz: Okay.

tracernz:
Results to date:
BLHeli tested by @ufoDziner 

Dug some China ESCs out of my old parts bin, these are all tested on Sparky1 with a few cycles of soft reboots via GCS.
**YEP45 (YGE clone), 400 Hz PWM:**
next: no movement during reboot
next: no movement during reboot

**Mystery 40A, 400 Hz PWM:**
next: no movement during reboot, re-init tone every time
with this PR: no movement during reboot, sometimes double re-init tone, otherwise single

And some servos of varying quality:
**HXT900, 50 Hz PWM:**
next: no movement during reboot
with this PR: no movement during reboot

**MKS DS95i, 560 Hz/760 us PWM:**
next: no movement during reboot
with this PR: no movement during reboot

Seems like BLHeli is the only thing that has problems even without this. Is somebody able to try SimonK (@dustin maybe)? Ran out of time to scope, can do tomorrow.

ufoDziner:
To be clear, my tests were with V-Good ESCs which run a proprietary firmware.  I do have a set of blheli ESCs around I can test with if necessary.

mlyle:
@tracernz-- I think it comes down to a pretty random set of factors whether you get the problem.  Pull-up impedances and leakages all around to whether you get a couple of pulses that spin stuff.  That's why we've never had reproducibility on this issue.

The big question is--  does half a second of no-pulses disarm anything?  If so, doing this seems like a good idea.

tracernz:
@ufoDziner Thanks, we should probably have a go with BLHeli too.

@mlyle Yes, but for something so critical it's pretty important to at least smoke test a wide variety of devices IMO.

mlyle:
Sure, testing to avoid regression is good.  It making at least one setup better is enough, though, IMO.

P.S. I want to use this same kind of technique to make ESC calibration slightly safer.

tracernz:
Doesn't do anything bad on a scope.

PR #1329, opened 2016-09-09
Tools: Build breakpad on Mac and Linux, use repo copy on win

### LINKED ISSUE #1267: Need a permanent/correct location for breakpad artifacts and their source
Both for license compliance and for reproducibility.

tracernz:
I did build it with MSVC on Windows but it's not really possible in our build environment (msys doesn't handle the paths, it needs a lot more MSVC env. stuff than we have etc.). Instead the Windows binary that is already checked into upstream repo is pulled (or a newer one if we decide that is needed at some point). On Mac and Linux we need Google depot-tools to fetch all the deps and prepare the repo for configure/build.

On Windows (doesn't build, fetches pre-built binary from repo):

```
make breakpad_install
```

On Linux/macOS (installs/updates depot_tools, fetches, builds and installs breakpad):

```
make breakpad_install
```

I am thinking about adding `depot_tools_install` dep to `breakpad_install` so it only needs the one step. If depot-tools is already installed it will just fetch the rev defined in the Makefile and check it out (either doing nothing, or upgrading), so harmless or possibly even beneficial. What do you think @mlyle ?

Fixes #1267

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -14,11 +14,11 @@
 ##############################
 # Checking for $(OPENOCD_FTDI) to be sane
 ifdef OPENOCD_FTDI
- ifneq ($(OPENOCD_FTDI),yes)
-  ifneq ($(OPENOCD_FTDI),no)
-   $(error Only yes or no are allowed for OPENOCD_FTDI)
+  ifneq ($(OPENOCD_FTDI),yes)
+    ifneq ($(OPENOCD_FTDI),no)
+      $(error Only yes or no are allowed for OPENOCD_FTDI)
+    endif
   endif
- endif
 endif
 
 # Set up QT toolchain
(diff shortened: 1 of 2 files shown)


make/tools.mk
```diff
@@ -511,6 +511,101 @@ endif
...
+		cd $(BREAKPAD_BUILD_DIR) ; \
+		export PATH="$(DEPOT_TOOLS_DIR):$(PATH)" ; \
+		gclient config --name=src "$(BREAKPAD_REPO)" ; \
+		gclient sync --with_tags --no-history -r refs/tags/$(BREAKPAD_REV) ; \
+	)
+
+	$(V0) @echo " BUILD        $(BREAKPAD_DIR)"
+	$(V1) ( \
+		cd "$(BREAKPAD_BUILD_DIR)/src/src" ; \
+		$(MAKE) distclean ; \
```
tracernz: Can't do this until configure has run once, and need to do it before configure runs. Will have to check to see if Makefile exists.

mlyle: So is this TODO then?

mlyle:
This looks great-- way above and beyond what I was worried about.

This has a documentation dependency.  When you confirm this is OK for merge, I will fix up development docs and merge.

make/tools.mk
tracernz: Yah, will try to get to it today, but likely will be tomorrow.

tracernz:
I started on it and just decided to go ahead and figure out how gclient works etc. and eventually managed to make a reproducible sequence. 😉 I almost had it on Windows, but the actual build was just a bit beyond msys shell. I might revisit it sometime and see if I can start an MS shell with VC environment setup scripts and run a custom .bat file to build at least dump_syms.

PR #1336, opened 2016-09-11
transmitter_control: improvements to rcvr_activity

### LINKED ISSUE #1045: RSSI injection screws up radio setup wizard
Because we see activity as stick movement.
1.  Do we do the right thing if RSSI is already configured / is it masked from ReceiverActivity?  If so, we could document to configure RSSI first in the case of injection.
2.  The radio setup wizard could look for channels that are moving at baseline, and warn / disqualify them / do something.

mlyle:
- factor the rssi injected-receiver source stuff out
- ignore the configured RSSI channel
- pick the "most active" channel in a given group

Fixes #1045 -- if the user configures and saves the RSSI channel before running input wizard.  NEEDS TEST.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -86,6 +86,8 @@ struct rcvr_activity_fsm {
 	uint8_t sample_count;
 };
 
+extern uintptr_t pios_rcvr_group_map[];
+
 // Private variables
 static ManualControlCommandData   cmd;
 static ManualControlSettingsData  settings;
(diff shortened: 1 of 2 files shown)


mlyle:

mlyle:
Note this is implemented differently than what I had suggested in #1045-- flight side rather than ground side.  I figure it makes better sense to not send a lot of events (perhaps masking the event we actually want) than to have GCS ignore the RSSI ones.

mlyle:
rebased.

ufoDziner:
Tested today using a Taranis with RSSI on channel 8.  RSSI was setup first per the note.  Worked just fine.  Two things that I feel need attention.
1.  The font for the RSSI note should be at least 2x its current size IMO.
2. The "Back" button on the first page of the wizard should be enabled to go back to the normal input screen, or similar to set the RSSI up.  Another option would be to have the RSSI input on the first page of the wizard.  "Cancel" does not seem intuitive, as I don't want to cancel what I'm doing, only go back and enable rssi.

mlyle:
@ufoDziner Thanks for the test.  I opened #1343 for your feedback; probably won't get to it for this cycle.

@tracernz et al-- I think this is ready to merge.  I think there can be further usability work done, but at least there is a way to make it through the wizard now.

mluessi:
Code LGTM, this should make the wizard more robust :). Merging.

PR #1344, opened 2016-09-12
autotune wizard: finesse final page; warning about rates

### LINKED ISSUE #1248: New users usually don't configure rates; usability help; controls "level up"
Proposed solution:

```
20:59 < icee> A) we do expoM this cycle.
21:00 < icee> B) we have 3 parameters.. 'maximum rate', 'center stick rate', and 
              'stick break point'
21:00 < icee> and show a nice graph
21:00 < icee> C) we force people to review it as part of autotune wizard
21:00 < icee> D) we leave leveling defaults alone
21:01 < icee> E) we move this stuff ABOVE pids on teh stabilization basic pane.
21:01 < icee> F) we kill the horizon rate visualization graph
```

connects #108

mlyle:
I figure that making the rates more visible / at the front of
stabilization pane takes away 25% of the problem.. and this maybe is
visible enough to take another 25% away.

Fixes #1248, a lucky-numbered bug.

### DIFF
--- a/ground/gcs/src/plugins/config/autotunefinalpage.ui
+++ b/ground/gcs/src/plugins/config/autotunefinalpage.ui
@@ -17,7 +17,7 @@
    <item>
     <widget class="QLabel" name="lblCongrats">
      <property name="text">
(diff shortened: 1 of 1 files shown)


mlyle:
I thought about showing the rate graph in a step, or telling the user that they are expected to get 0.6 rotations/second... but I feel like less is more here.  Always can refine more.

dustin:
I like this.  It's surprising to me, but a lot of people believe tuning is
about making it spin faster.

jhitesma:
What @dustin said.  And agreed, less is more here.  Maybe in the future adding the ability to adjust rates here...but take things easy first and see how this does.

PR #1347, opened 2016-09-12
autotune: Apply gyro lowpass.

### LINKED ISSUE #1312: Autotune: consider running gyro filter on incoming points
Looking at the Librepilot import of our autotune, they run the configured gyro cutoff lowpass filter on the points as they come in.

This is marginally correct to do-- there is about 1.5ms of group delay @50Hz with a 55Hz cutoff filter.  But at the same time, we have appropriate fudge factors that a 1.5ms offset is not that big of a deal and by softening the edge we probably add dispersion to the measured tau.

BUT-- it seems from recent facebook discussion that one cause of failure of tunes to converge is vibration.  This can cause beta to run away super high and the magnitude of the wiggle to go to zero or nearly zero.  This would help address that.

glowtape:
Might result in improvements. Needs some more testing. Personally it seems a win. See issue #1312 

It is not doing anything for dealing with spilled samples, it just applies the LPF in the assumption that all data will be processed. I've checked a bunch of random CC3D results on Autotown, they aren't spilling anything (yet). Someone should test this on a CC3D, whether the additional math for the LPF is the tipping point. It sure enough caused a fuzz with the watchdog (or whatever) when I had it in at_new_gyro_data instead. On the other hand, adding more code to check and recalculate the filter coefficient for the dT between points after a spill would probably make things even worse on the F1.

--edit: This doesn't include the removal of input scaling on the other PR.

edit: fixes #1312 -mpl

### DIFF
--- a/flight/Modules/Autotune/autotune.c
+++ b/flight/Modules/Autotune/autotune.c
@@ -84,6 +84,9 @@ static void AutotuneTask(void *parameters);
 static void af_predict(float X[AF_NUMX], float P[AF_NUMP], const float u_in[3], const float gyro[3], const float dT_s, const float t_in);
 static void af_init(float X[AF_NUMX], float P[AF_NUMP]);
 
+// Get the gyro alpha from the stabilization task
+extern float gyro_alpha;
+
 #ifndef AT_QUEUE_NUMELEM
 
 #ifdef SMALLF1
@@ -265,6 +268,8 @@ static void UpdateStabilizationDesired(bool doingIdent) {
 
 #define MAX_PTS_PER_CYCLE 4
 
+float fgyro[3];
+
 /**
  * Module thread, should not return.
  */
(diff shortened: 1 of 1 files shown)


mluessi:
Interesting. I'm a bit surprised that tau decreased when the lpf is applied, shouldn't go the other way?

At first, I was thinking that we should account for this in the AT calculation. However, looking at the derivation, it seems that AT assumes there is no LPF applied (so unfiltered gyro data was used), while the actual PID controller applies the LPF. Hence, there has been a discrepancy between what AT assumes (no LPF) and what we do when running the PID controller (using LPF filtered gyro samples). So using filtered gyro samples for AT should actually be correct. I may be wrong, only had 0.5 coffee so far.

glowtape:
I don't know enough of control theory to be authoritative about it, but my intuition says that you should systemident under the conditions you're going to fly. But it's currently only a single quad this has run on.

That said, if gyro noise is the main cause of issues, it might be worthwhile to revisit the expected gyro noise parameter in the autotune routine instead? It was lowered considerably a while ago ( s_a in 00bfc502c17356ca2b080e560fb7afaebed1057e ).

glowtape:
I've done a bunch of more prop tests. The DAL Q5040 quadblade have a 2ms difference in tau between filtered and unfiltered in favor of former. The DAL 5045BNV2 twinblades have same tau. Based on that, I've rerun an unfiltered autotune on the tris from the initial test, now tau is in the same ballpark as the filtered. Seems like it was a fluke.

(Links to the tunes are in #1312 )

Regardless of that, even with same tau, the filtered tunes have higher gains. They don't introduce any ringing or oscillations on harder manouvers, like fast flips and such.

glowtape:

mlyle:
Thanks, this was a very worthwhile experiment.  It has now been superseded by #1450.

PR #1350, opened 2016-09-12
Make Vehicle Setup Wizard - Input Type selection work for AQ32

yamahabest:
Was not implemented. It is now.


yamahabest:
Already tested on my AQ32, seems to work.

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
```diff
@@ -106,11 +105,137 @@ QPixmap AQ32::getBoardPicture()
...
+		return INPUT_TYPE_SBUSNONINVERTED;	
+	case HwAQ32::UART3_IBUS: // None of the other targets have IBUS in getInputType, but seems to be no problem.
+		return INPUT_TYPE_IBUS;
+	}
+
+    switch(settings.Uart4) {
+	case HwAQ32::UART4_DSM:
+        return INPUT_TYPE_DSM;
+	case HwAQ32::UART4_HOTTSUMD:
+        return INPUT_TYPE_HOTTSUMD;
```
tracernz, line 214: I'm not sure this is a good idea since the wizard never sets SUMD etc. on this port. It will make the wizard think it doesn't need to reboot when it sets them on UART3 and they're already on UART4 or 6 (this case is broken for a different reason but we shouldn't make it worse here).

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
```diff
@@ -106,11 +105,137 @@ QPixmap AQ32::getBoardPicture()
...
+
+    HwAQ32::DataFields settings = hwAQ32->getData();
+
+    switch(settings.RcvrPort) {
+    case HwAQ32::RCVRPORT_PPM:
+		return INPUT_TYPE_PPM;
+    case HwAQ32::RCVRPORT_PWM:
+        return INPUT_TYPE_PWM;
+    default:
+        break;
```
mlyle, line 194: It looks like indentation is not correct in this file.  Coding standards for GCS code are 4 spaces per indent level, no tabs.

mlyle:
Looks reasonable-- thanks for the change!  What combinations have you tested?

mlyle:
jenkins, test this please

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
yamahabest: Ok, corrected.

yamahabest:
"combinations", I am not sure what you mean?
I have tested that each option sets the correct type on the correct port.

One problem I can see, but that might be done intentionally, for example on the AQ32, If one first runs the wizard, and select PPM. And afterwards run the wizard again, and select SBUS. Result is that RcvrPort is set to PPM and Uart3 is set to SBUS.
However, this might be hard to fix, as it is not possible to reset the ports you are not configuring on that place, as the user might have selected something specific on those ports, right?
For example: User might have configured MSP on Uart3, setting RcvrPort to PPM should then not reset Uart3 to "none".

mlyle:
So you tested a full matrix of the possible inputs, good.

>  And afterwards run the wizard again, and select SBUS. Result is that RcvrPort is set to PPM and Uart3 is set to SBUS.

That's OK, it's a valid configuration-- we support multiple configured receivers at once, whether you're using them at once or not.  Just wastes a small amount of ram.

mlyle:
Thanks much :)

tracernz:
There were remaining review comments that will need addressed in future to make this target behave properly...

mlyle:
OK, I didn't see other review comments-- I still don't see them-- am I missing something?

It looked correct to me.

yamahabest:
Review comments?
Are you talking about another PR or an issue?

tracernz:
Ah, made at the same time as github upgraded. Showing as pending, must have slipped through the cracks :disappointed:

yamahabest:
What were your comments?

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
mlyle: We should decline to set anything in the wizard when it's "already set".  That way if you have a valid sbus configuration on one port, it won't try to move it to another port, etc.)

tracernz:
That will break wiring diagrams though. Fair bit of work to make the wizard support that properly.

mlyle:
Yah, if the user already has configured it they will get a lie for a wiring diagram (as opposed to double configuring the port, etc).  But I get that's meh.

yamahabest:
I understand what you mean. But this would be a more overall issue.
Brain is doing the same stuff, I used that one as base to implement this for AQ32.

```
switch(settings.MainPort) {
        case HwBrain::MAINPORT_SBUS:
            return INPUT_TYPE_SBUS;
        case HwBrain::MAINPORT_DSM:
            return INPUT_TYPE_DSM;
        case HwBrain::MAINPORT_HOTTSUMD:
            return INPUT_TYPE_HOTTSUMD;
        case HwBrain::MAINPORT_HOTTSUMH:
            return INPUT_TYPE_HOTTSUMH;
    }

    switch(settings.FlxPort) {
        case HwBrain::FLXPORT_DSM:
            return INPUT_TYPE_DSM;
        case HwBrain::FLXPORT_HOTTSUMD:
            return INPUT_TYPE_HOTTSUMD;
        case HwBrain::FLXPORT_HOTTSUMH:
            return INPUT_TYPE_HOTTSUMH;
    }
```

tracernz:
We'll clean it all up at some point. 👍

PR #1351, opened 2016-09-12
logviewer: tolerate pyqt4 + python2.7 better

mlyle:
Reported by Troy on IRC.

Siiiiigh.

Tested with full matrix py2 vs py3, pyqt4 vs pyqt5.

### DIFF
--- a/python/dronin/logviewer/logviewer.py
+++ b/python/dronin/logviewer/logviewer.py
@@ -163,7 +163,7 @@ def handle_open(ignored=False, fname=None):
     with pg.ProgressDialog("0 objects read...", wait=500, maximum=1000, cancelText=None) as dlg:
         global t
 
-        if not isinstance(fname, six.string_types):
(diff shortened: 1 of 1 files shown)


python/dronin/logviewer/logviewer.py
```diff
@@ -163,7 +163,9 @@ def handle_open(ignored=False, fname=None):
     with pg.ProgressDialog("0 objects read...", wait=500, maximum=1000, cancelText=None) as dlg:
         global t
 
-        if not isinstance(fname, six.string_types):
+        print type(fname)
```
tracernz: Did you mean this print to be here? This will break with Python3?

mlyle: No, i didn't :P

mlyle:
Ugh needs more work, I must have failed the python3 test before.

mlyle:
OK, this is what I'm goin' with. :P

PR #1352, opened 2016-09-12
Mpl halmagandbaro

jihlein:
Here's the UI changes.  Put BMP280 1st in the list as it looks like it will be the primary device.


jihlein:
This is the wrong place to merge this.  Closing.

PR #1353, opened 2016-09-13
hw uavos: f1: disable PWM/PPM by default

mlyle:
Don't waste resources, don't get timer conflicts right off the bat.
Let the user / wizard select that if needed.


tracernz:
Don't think it's a bug per se but LGTM.

mlyle:
Kinda a bug because it results in setupwizard leaving ppm enabled and timer conflicts afterwards.

PR #1359, opened 2016-09-15
BrainRE1: Expose additional serial port, improve GCS hw config

mluessi:
Exposes an additional serial port (UART4) using PWM1 and PWM2 pins in normal mode. This is based/inspired by earlier work by @jihlein . Also changes the speed of all serial port IOs to 25MHz, as 2MHz may cause problems with OpenLager.

To keep it user friendly, the GCS plugin was improved to show the pinout and hide unused options. Also some small changes to change the output banks depending on HW config.

Status: bench tested (scoped PWM outputs and tested function of serial ports)

Normal mode:
re1_mp_normal

Mode with 2 serial ports and external mag enabled:
re1_mp_4pwm


dustin:
Ooh, that's pretty.

jihlein:
Ah, someone who knows how to fix the channel bank issue I ran into.  This is great!

mluessi:
@dustin thanks. Will it be possible to add some code to the Python cloud upgrader thing to transfer the value from `MultiPort` to `MultiPortSerial` (if the former is present)? I can look into it but don't know where the code is.

dustin:
Most of the transformer stuff is in the python directory.  The cloud bit
copies the python directory from dronin main, and adds this:
https://github.com/d-ronin/autotown/blob/master/py/convert.py

jihlein:
When using MultiPortSerial2, might want to consider labeling the TX/RX pins TX2/RX2 to help reduce the inevitable confusion between 2 sets of pins labeled TX/RX on the same header.

mlyle:
I'm inclined to avoid adding code to transform during upgrade because A) it's really, really hard to test and get right, and B) we've basically only used it for stuff that's safety/correctness critical to date.

shared/uavobjectdefinition/hwbrainre1.xml
```diff
@@ -54,6 +61,24 @@
 		</options>
 	</field>
 
+	<field name="MultiPortSerial2" units="function" type="enum" elements="1" parent="HwShared.PortTypes" defaultvalue="Disabled">
+		<options>
+			<option>Disabled</option>
```
mlyle, line 66: I assume the missing receiver telemetry options is because no inverter?

mluessi: Exactly. IMO it would be confusing to have them and lead to user error. The only scenario that I can see where this would be useful would be if the other serial port is damaged.

mluessi:
@mlyle yeah, this isn't really critical and easy for users to adjust after the update, so not adding it to the updater script is fine.

@jihlein good idea, changed the image to show TX2 / RX2

mlyle:

mlyle:
LGTM

PR #1366, opened 2016-09-18
Android GCS serial

mlyle:
- Put correct license file from upstream distro in (it is, and was, LGPL 2.1 so what we had is incorrect).
- Pick up code changes from upstream project as noted in commit comments.
- Attempt to support CP21xx correctly.
- Attempt to adjust for API changes.

Comes from: https://forum.dronin.org/forum/d/154-serial-telemetry-on-gcs/11

I noticed that CP21xx didn't seem to be hooked in properly in the old code, and in any case the usb serial code we're running is very, very old.


mlyle:
I can't find my CP2102 thing.  But I was able to confirm this works with both FT232 and PL2303.  (PL2303 was unsupported before).

Recommend for merge.

PR #1374, opened 2016-09-21
attitude: remove yaw bias rate

mlyle:
It was unused/removed  on F3/F4, and practically unused on F1 (default value
was very, very, very, very low).

### DIFF
--- a/flight/Modules/Attitude/acro/attitude.c
+++ b/flight/Modules/Attitude/acro/attitude.c
@@ -93,7 +93,6 @@ static float accelKi = 0;
 static float accelKp = 0;
 static float accel_alpha = 0;
 static bool accel_filter_enabled = false;
-static float yawBiasRate = 0;
 static const float IDG_GYRO_GAIN = 0.42;
 static float q[4] = {1,0,0,0};
 static float Rsb[3][3]; // Rotation matrix which transforms from the body frame to the sensor board frame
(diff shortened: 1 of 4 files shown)


flight/Modules/Attitude/acro/attitude.c
```diff
@@ -199,7 +198,6 @@ static void AttitudeTask(void *parameters)
 			// For first 7 seconds use accels to get gyro bias
 			accelKp = 0.1f + 0.1f * (PIOS_Thread_Systime() < 4000);
 			accelKi = 0.1;
-			yawBiasRate = 0.1;
```
mlyle, line 202: This is duplicative with accumulateGyro stuff.

tracernz: It is? I just read over it and can't see it.

mlyle: Sorry, not this one, the next one.

We set yawbiasrate to 0.1 to try and more aggressively set the yaw bias to 0 initially.

At the same time, we're accumulating gyro samples to set an offset to make them average to 0 at the arming edge.

mlyle:
Freebased

mlyle:
jenkins, test this please

flight/Modules/Attitude/acro/attitude.c
tracernz: This allows you to bias toward zero all the time, not just when arming... but the default is so weak it's not effective and I doubt anyone is using it.

PR #765, opened 2016-03-06
WIP WIP WIP : Buzzermodule

henn1001:
fixes #105 

So far this is only for digital buzzer support and is more or less functional now with these target changes [link](https://github.com/henn1001/dRonin/commit/5014ded9d1bd43465c0f62d2af1f8dc91a24c830)
Can and should probably (maybe?) adapted to pwm type output as well. I´m putting it out here so it can be discussed in which direction this pr should go before i put more work into it.

I made it customizable so people can choose which alarms are active. Options are On|Off|WasArmed.

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/765)

Issue #1370, opened 2016-09-20
RE1/DTFc: Don't brutalise UAVO settings


tracernz:
Settings should _never_ be overwritten like this (which in this case also precludes using any sensor with the default scaling factor).
https://github.com/d-ronin/dRonin/blob/1840a1461ad033fdd748f8b7c3a3bdec6dc25fb2/flight/targets/dtfc/fw/pios_board.c#L304-L308
https://github.com/d-ronin/dRonin/blob/ff3de726463edf6c5ca0d53dc5442f295229a16b/flight/targets/brainre1/fw/pios_board.c#L561-L568

Possible alternatives (in order of preference):
- Something really clever I haven't thought of
- Putting ADC settings through Hw\* UAVOs and pios_hal (this would make the GCS end a bit cleaner too)
- Only replace the defaults if the object **doesn't** exist in flash (i.e. has never been set)

mlyle:
Could declare the defaults for the uavo as a weak reference in the linker, and let a target override... but then there's the chance that something-not-intended-to-be-overriden could be overridden as defaults change.

PR #1379, opened 2016-09-22
Annunciator refactor, buzzer support

### LINKED ISSUE #105: Buzzer Support
I think we need a buzzer module with the following requirements:

* Configurable output (digital or pwm on something supported by the hardware)
* Configurable alarm conditions (beep on particular alarm conditions)
* User-initiated beeping (accessory is|isn't at a particular value)

Supporting naze's nfet thing would be nice.  [pwm buzzers](http://hobbyking.com/hobbyking/store/__14521__HobbyKing_174_Discovery_Buzzer.html) exist and should be quite easy to support.

See also:  https://github.com/TauLabs/TauLabs/issues/1349

mlyle:
### Other todo:
- [x] Fix the config to be right for when buzzer does stuff to not annoy me to death
- [x] Get the 'ManualControlOnly' class to do the right thing
## Test matrix
- [x] Arming/disarmed state is communicated on heartbeat LED
- [x] Basic alarms seem to do correct blink codes
- [x] "After arming" alarms do the right thing
### Targets:

(Don't expect to get a full matrix here)
- [x] DTFc
- [x] Lux
- [x] Naze32
- [x] Quanton
### Blink codes:

```
-... 'b' battery
-.-. 'c' system configuration
--.  'g' gps
.-.  'r' radio (the most common code)-- manualcontrol
.-   'a' any other alarm

OK alarms:
-    't' disarmed (by default only to heartbeat)
..   'i' armed
```
### Out of scope (currently)
- BrainRE1 buzzer support
- LED R/G/B alarm channels
- Actuator targets for arming/heartbeat/buzzer LEDs (especially sparky2)
- Driving the buzzer channel with accessorydesired

Fixes #105, supersedes #765


mlyle:
jenkins, test this please

mlyle:
https://www.youtube.com/watch?v=5B_7J4F0CkQ

mlyle:
This does not work on my Quanton.

mlyle:
Works on Naze32.

mlyle:

mlyle:
Everything tried passed (including now Quanton) -- and as far as i'm concerned, this one is ready to go.

tracernz:
Quanton with 1k load, manual control alarm.
newfile1

-e- zoomed a bit since it's tiny:
newfile2

mlyle:

mlyle:
ehh. maybe the windows builder will work this time.

mluessi:
Very nice. I was thinking about the best way to use this with the RE1 buzzer and other future targets that may not use GPIOs to toggle LEDs etc. One way could be to extend `struct pios_annunc` to contain two function pointers, one for querying the state and one for setting the annunciator. We could ifdef it, so it wouldn't use resources on targets that don't have it. Let me know what you think; I can give it a shot and open a PR against this one.

mluessi:
Also, using a switch to activate the buzzer is something users really want (I get asked about it a lot), so it would be awesome if we could figure out a way to support it in this release.

mlyle:
@mluessi let's get this one merged, but I'd be fine with speedily merging something afterwards like that to make RE1 work.

It shouldn't require two function pointers-- just one to set should be enough.

mlyle:
Re: switch-- I think we're just too late at this point to be introducing much more functional stuff.  We don't have everything everyone wants, but we do have a whole lot of functional improvements.

ufoDziner:
Successfully tested on DTFc.

ufoDziner:
Might want to add a note in documentation for the DTFc stating that the buzzer will be constantly sounding any time the board is powered up prior to having the buzzer capable firmware.  Could be quite the eye-opener for a person upgrading.

flight/Libraries/inc/morsel.h
```diff
@@ -0,0 +1,33 @@
+// Morse Library header
```
tracernz, line 1: Any reason for the C++ style comments around this header? Our code style guide doesn't allow them (although I know we all use them from time to time). Standardised headers make it easier to write scripts to work on them too.

flight/Libraries/morsel.c
```diff
@@ -0,0 +1,271 @@
...
+	/*   ...-	V SHORT SHORT SHORT LONG */
+	0B01001100,
+	/*   .--	W */
+	0B01110010,
+	/*   -..-	X */
+	0B01110110,
+	/*   -.--	Y */
+	0B01111000,
+	/*   --..	Z */
+};
```
tracernz, line 145: I'm gonna assume your table is correct. :stuck_out_tongue:

flight/Libraries/morsel.c
```diff
@@ -0,0 +1,271 @@
...
+	if (c <= 90) { /* 40..90 */
+		return morse_expand(morse_table[c - 40], len);
+	}
+
+	if (c < 97) { /* 91..96 */
+		return 0;
+	}
+
+	if (c < 122) { /* 97..122 */
+		return morse_expand(morse_table[c - 97 + 25], len);
```
tracernz, line 179: Maps lowercase to uppercase.

flight/Libraries/morsel.c
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	return 0;
+}
+
+enum __attribute__ ((__packed__)) character_state {
+	STATE_NEXTCHAR = 0,       // 1 time of nil, prime next character -> NEXTSYM/SPACEWAIT
```
tracernz: Small column misalignment here, up to you if you want to fix it.

flight/Modules/System/systemmod.c
```diff
@@ -310,32 +399,92 @@ static void systemPeriodicCb(UAVObjEvent *ev, void *ctx, void *obj_data, int len
...
-			DEBUG_MSG("+ 0x%08x\r\n", 0xDEADBEEF);
+	if (candidate_prio > blink_prio) {
+		// Preempt!
+		blink_state = 0;
+		blink_string = candidate_blink;
+		blink_prio = candidate_prio;
+
+		if (blink_string &&
+				!strcmp(blink_string, BLINK_STRING_RADIO)) {
+			// XXX do this in a more robust way */
```
tracernz, line 428: :wink:

flight/Libraries/morsel.c
```diff
@@ -0,0 +1,271 @@
...
+	STATE_DOT,              // Send a dot -> NEXTSYM
+	STATE_DASH,             // Send a dash -> DASH_B
+	STATE_DASH_B,           // Send a dash -> DASH_C
+	STATE_DASH_C            // Send a dash -> NEXTSYM
+};
+
+/* Returns 0 for off, 1 for on, -1 for completed */
+int morse_send(const char **c, uint32_t *state)
+{
+	uint8_t *len = ((uint8_t *) state);
```
tracernz: Not really a major fan of this vs. a struct. :/

flight/Libraries/inc/morsel.h
```diff
@@ -0,0 +1,33 @@
...
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef _MORSEL_H
+#define _MORSEL_H
+
+#include <stdint.h>
+
+int morse_send(const char **c, uint32_t *state);
+
+#endif
```
tracernz, line 33: No dox in this file.

shared/uavobjectdefinition/annunciatorsettings.xml
```diff
@@ -0,0 +1,25 @@
+<xml>
+	<object name="AnnunciatorSettings" singleinstance="true" settings="true">
+		<description>Settings for alarm annunciation (LEDs, buzzer)</description>
+		<field name="AnnunciateAfterArming" units="" type="enum" parent="SharedDefs.AlarmLevels" defaultvalue="HairOnFire,HairOnFire,ManualControlOnly">
```
tracernz: No field descriptions.

tracernz (commented):
Looks pretty good. :+1:

flight/Libraries/inc/morsel.h
mlyle: It's imported from OpenLager, which I wrote using BSD-ish conventions.   Can reformat if you want (in which case I should reformat over there, so things stay in sync).

flight/Libraries/morsel.c
mlyle: Could use a union, I guess, to preserve the bundled-in-uint32_t characteristic..

flight/Modules/System/systemmod.c
mlyle: Yah.. I think it's OK for now.  But if we're going to have fancier preemption behaviors, different alarms going different places, etc.  this whole section needs to change.

flight/Libraries/morsel.c
mlyle: Seems to work on OpenLager and have blinked lots of stuff there.  But I doubt I've fully covered it.

flight/Libraries/morsel.c
mlyle: I see why I did this.  I think I was concerned about whether enum character_state, even being packed, would pack into a 3 byte struct properly.

mlyle:
Force push'd to eliminate conflict from brain indent.

The last commit replaces the state innards with a union.  Do you like it better?  I could go either way.

mlyle:
P.S. untested in this form, so if you like it I need to do some validation.

mlyle:
jenkins, test this please

flight/Libraries/inc/morsel.h
tracernz: Okay. Up to you if you want to do it now. Probably can wait for a future occasion when it's necesssary.

flight/Libraries/morsel.c
tracernz: Fair enough. I am okay with it then. :smile:

tracernz (approved):
Looks good, thanks. :+1:

mlyle:
Exercised 'a', 'i', 't', 'c', 'r'.  All's good from what I can see.

tracernz:
I miss that the heartbeat alarm ALWAYS indicated arming state before, now it gets overridden if you have anything else, even if armed, and just duplicates the alarm LED which isn't very useful. There are some boards with only one LED, but ~~you can always just set the LED to OK to get the same info anyway~~ we should make sure they're all defined as alarm LED, and make the default "OK" in that case (another case that needs a nice solution to #1370). Otherwise, I tested a buzzer and it worked. :+1: 

-e- English better.

PR #1381, opened 2016-09-23
battery: make estimated flight time nicer

mlyle:
Use a second order filter, with longer rolloff, to come up with
remaining flight time estimates.  It will be optimistic for the first
20-30 seconds of a flight, but should be a pretty good estimate after
that.

### DIFF
--- a/flight/Modules/Battery/battery.c
+++ b/flight/Modules/Battery/battery.c
@@ -55,7 +55,7 @@ static int8_t voltageADCPin = -1; //ADC pin for voltage
 static int8_t currentADCPin = -1; //ADC pin for current
 static bool battery_settings_updated;
 
-static float time_lpf;
+static float avg_current_lpf_for_time;
 
 // ****************
 // Private functions
@@ -228,6 +228,7 @@ static void batteryTask(void * parameters)
 			float alpha = 1.0f - dT / (dT + 2.0f);
 			flightBatteryData.AvgCurrent = alpha * flightBatteryData.AvgCurrent + (1 - alpha) * flightBatteryData.Current; //in Amps
 
+			// XXX Arguably this should be to 10% capacity or 15%
 			energyRemaining = batterySettings.Capacity - flightBatteryData.ConsumedEnergy; // in mAh
 
 			// And for time estimation, smooth things much more.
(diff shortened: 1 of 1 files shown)


flight/Modules/Battery/battery.c
```diff
@@ -227,8 +229,16 @@ static void batteryTask(void * parameters)
...
 			energyRemaining = batterySettings.Capacity - flightBatteryData.ConsumedEnergy; // in mAh
-			if (flightBatteryData.AvgCurrent > 0)
-				flightBatteryData.EstimatedFlightTime = (energyRemaining / (flightBatteryData.AvgCurrent * 1000.0f)) * 3600.0f; //in Sec
+
+			// And for time estimation, smooth things much more.
+			// Second order since it incorporates the above
+			// smoothing, 12s time constant for this layer.
+			alpha = 1.0f - dT / (dT + 12.0f);
+
+			time_lpf = time_lpf * alpha + (1 - alpha) * flightBatteryData.AvgCurrent;
```
tracernz: Not sure this is a particularly good variable name. `filtered_current` or something similar might be better?
If I read `(energyRemaining / (time_lpf * 1000.0f)) * 3600.0f`, it's not clear what time_lpf is thus what the equation is actually doing.

tracernz (changes requested):

flight/Modules/Battery/battery.c
mlyle: Done :)

dustin:
This works really well in my little testing today. Punch hard and then
float around in trees and it gave me a useful indicator as to when to come
home.

tracernz (approved):

flight/Modules/Battery/battery.c
```diff
@@ -226,9 +228,18 @@ static void batteryTask(void * parameters)
 			float alpha = 1.0f - dT / (dT + 2.0f);
 			flightBatteryData.AvgCurrent = alpha * flightBatteryData.AvgCurrent + (1 - alpha) * flightBatteryData.Current; //in Amps
 
+			// XXX Arguably this should be to 10% capacity or 15%
```
mluessi, line 231: Yeah, we should make this configurable and default to 20%

tracernz: Yah, the "rule" has always been 80% discharge.

mlyle: Question is, should the time read 0 at 80% discharge?  Who does the 80%?  Probably us, but not sure.

PR #1390, opened 2016-09-29
GCS: Make updating translations work/easier

tracernz:
First commit is good, second one I'm not sure of it's wisdom. 
First one:
- Fixes translations.pro so ts target actually works
- Adds gcs_ts target to main Makefile so people can easily update translation sources

Note:
To actually perform translation, run ./tools/Qt5.6.1/gcc_64/bin/linguist and open appropriate ts file in ./ground/gcs/share/translations/dronin_*.ts


tracernz:
Dropped the second commit which updates the files, they can be updated when somebody decides to work on them.

PR #1391, opened 2016-09-29
BrainRE1: add HID/VCP telemetry warning

mluessi:
Warns user if both HID and VCP telemetry are disabled and prevents saving.


mlyle:
Thanks!

Issue #1228, opened 2016-08-24
flyingpi: no way to configure receiver type


mlyle:
Right now it's always PPM in checked in code (always sbus in my branch).

mlyle:
@mhvdrone is going to hit this soon.

PR #1392, opened 2016-09-29
flyingpio: set default to SBUS

mlyle:
This is a hack pending #1228, but at least stops my test rig from
repeatedly tripping over it.

### DIFF
--- a/flight/PiOS/posix/pios_flyingpio.c
+++ b/flight/PiOS/posix/pios_flyingpio.c
@@ -333,7 +333,7 @@ static void PIOS_FLYINGPIO_ActuatorSetMode(const uint16_t *out_rate,
 	}
 
 	/* XXX needs to be specified / configured somewhere... */
-	cmd->receiver_protocol = HWSHARED_PORTTYPES_PPM;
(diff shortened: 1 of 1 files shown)


tracernz:
PPM seemed like a better default since it has much more universal support?

mlyle:
It's just me using this, and ultimately either way it needs to be specified on the command line.

tracernz:
Alright, any reason to target Quixote if it's just for you?

mlyle:
Yes, so I can fly quixote builds on flyingpi without failing to make the receiver work each time :(

tracernz:
Haha, okay, so we are not targeting general availability this cycle then? :stuck_out_tongue:

mlyle:
Right now anyone who uses it has to compile it?  AKA cross-compiled for ARM is not in any artifacts.  You also have to build a kernel from our kernel source fork, etc, etc, etc.  There are a lot more things that have to happen to make this target turnkey-- having to edit this variable is the smallest of concerns...

tracernz:
Alright, Artifice it is.

Issue #748, opened 2016-03-05
Excessive event alarms


mlyle:
Happened on Tau, but now it happens even more I think.  Need to run down latency and bursts of excessive processing everywhere.

mlyle:
It's worth noting that it sure seems like toggling other alarms makes EVENT prone to stack up itself.

mlyle:
Been thinkin' -- is this telemetry?

tracernz:
Things like manualcontrol etc. set to onchange generate more updates than the link can handle I think? Should have a throttled onchange, but it's a little tricky because you always want the last state to go through.

mlyle:
we have a throttled onchange.

tracernz:
How does it handle the tricky bit (i.e. when there's a flurry of events then static value)? If it always sends the last state in all cases then we could replace some onchange with throttled.

mlyle:
It doesn't send the last event in al lcases, but everything high frequency has a throttle value.

We don't know that this is telemetry.  It doesn't entirely make sense that it is (e.g. we used to see EVENT on screen in the OSD in flight before we masked it)

tracernz:
So, it puts a pointer in a UAVO somewhere with who triggered it, but IIRC that pointer changes a bit.

mlyle:
It puts the uavo id.  I'll check it out ;)  Just wanted to remind myself here to look at telemetry again.

mlyle:
I am seeing this much less lately....

PR #1393, opened 2016-09-29
annunciators: ignore EVENTSYSTEM nuisance alarms

mlyle:
Else you can get .- 'a' generic alarm interspersed between your 'armed' or 'disarmed' flashes.


tracernz:
Yep, we should probably raise a bug to address the root cause of these though...

mlyle:
#748

PR #1394, opened 2016-09-30
input wizard: Allow a higher max channel value

dustin:
SUMD scaling was removed in 12075d77865c5023d7f544e349a0ded848c5df1d and
my effective SUMD range is 7872 - 16128.  In the wizard, any time you
move the stick above the middle when trying to find the range, it kind
of freaks out and considers stuff failed.


mluessi:
LGTM. This should probably make it into the release, right?

dustin:
It seems harmless enough, but I only changed the number, I didn't test that somehow things got worse.

mlyle:
Yah, I'm fine doing it on faith.  I was just waiting for CI then forgot :D

PR #1396, opened 2016-10-04
remove picoc and waypoints storage

### LINKED ISSUE #1087: PicoC abuses settings objects
Settings objects should only ever be written by GCS/the user. Changing settings in-flight opens up lots of possibilities for serious bugs. Things like setting actuators should be done through the appropriate PiOS functions rather than attempting to force values by corrupting the user's settings.

### LINKED ISSUE #1088: PicoC Writes to Disabled Actuators
This is potentially quite dangerous if you have a motor on a channel you have disabled because you don't want it to spin (e.g. when bench-testing your PicoC scripts). PicoC should have it's own actuator type and should only work on actuators which are assigned to PicoC.

mlyle:
Frees up a partition on all F3 and F4 targets for storing system ident data.

This removes marginal subsystems, frees up flash the loader knows about for MagicTune(tm) but incurs a lot of debt.  That is, things we've been planning to do-- runtime code loading to replace picoc, storing multi-instance stuff in config to replace waypoint storage, etc.

fixes #1087 
fixes #1088


lolorc:
I know that you really don't like picoC, but until a proper replacement (if ever) could it be possible to at least keep it disabled by default and enabled/compiled with a compile flag ?

mluessi:
@lolorc are you currently using PicoC? What is your application?

lolorc:
yes been using for a long time.
driving buzzer and ws2812 leds for now.
atm using data from flightbatterystate, flightstatus, gpsposition.

I was using TauLabs with a 11 flight modes patch I haven't ported  back to DR yet. so I'm wondering about using Stabilized3Set as an alternate way to support more flight modes.

more generally I'd be in favor of being able to change every single parameters from picoC :)

mlyle:
@lolorc -- the problem is that I need a big chunk of flash for something that will be on the path to significantly improve flight performance.  The only extra chunk many targets have is the waypoints partition, which PicoC also uses. :/

mlyle:
rebased

tracernz:
This seems unfortunate, but inevitable at this point.

PR #1397, opened 2016-10-04
makefile: add Release-20161004 to UAVO_GIT_VERSIONS

mlyle:


tracernz:
Something like this (gets most recent 5 releases) might be a good idea rather than adding another maintenance burden (we have done well to avoid versioning stuff in tree to date)?
`git for-each-ref --count=5 --sort='-creatordate' --format '%(refname:short)' 'refs/tags/Release-*'`

tracernz:
Also, the plumbing command (rather than porcelain command) for the one already there is:
`git rev-list --max-count=18 --merges --abbrev-commit HEAD shared/uavobjectdefinition`
😉

mlyle:
Maybe in the future?  Unfortunately, also, if the UAVOs have not changed it makes the app bigger and more bloated and extends the build time to include e.g. hotfixes

tracernz:
I'll look at it and see if I can come up with something.

PR #1402, opened 2016-10-05
gcs: gps display widget: remove NMEA support

### LINKED ISSUE #1376: RFC: Remove NMEAParser from GCS
GCS has functionality that you can hook a NMEA GPS to GCS, and display your position and satellite set in a widget.  It's not used for anything else, and only supports NMEA GPSes.  There are better programs for this-- I don't know if it even works.  I'd like to remove it, in the spirit of "remove all the code," and make the GPS gadget only work with telemetry.  Comments?

mlyle:
This isn't useful; there are better tools than us for directly talking
to a GPS.  There's a lot of configuration surface etc that is wasted
for this.

Fixes #1376


mlyle:

PR #1403, opened 2016-10-05
vbar: Move settings to separate object.

mlyle:
StabilizationSettings is getting too big, and these are decoupled
from everything else.
- StabilizationSettings is too big with the virtualbar stuff to fit acrodyne stuff too.  Makes more sense for vbar to get its own object
- Likely a subsequent PR will drop virtualbar support from F1, which will get us a fair bit of breathing room for other projects.

### DIFF
--- a/shared/uavobjectdefinition/vbarsettings.xml
+++ b/shared/uavobjectdefinition/vbarsettings.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<xml>
+	<object name="VbarSettings" singleinstance="true" settings="true">
+		<description>Settings for the virtualbar flight mode</description>
+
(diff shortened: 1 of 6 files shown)


shared/uavobjectdefinition/vbarsettings.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<xml>
+	<object name="VbarSettings" singleinstance="true" settings="true">
+		<description>Settings for the virtualbar flight mode</description>
+
+		<field name="VbarSensitivity" units="frac" type="float" elementnames="Roll,Pitch,Yaw" defaultvalue="0.5,0.5,0.5"/>
```
tracernz, line 6: You don't want to describe the fields? :stuck_out_tongue:

mlyle: I'd have to learn what they mean :D

mluessi:
LGTM, I think this can merge if it has been tested

PR #1407, opened 2016-10-07
receiver activity -- attempt to improve

mlyle:
- Fix a negative array index (not the bug we're looking for but not good)
- Don't only check "every other manual control cycle" for activity
- Check each receiver group twice in a row to catch smaller movements, be more responsive, and not be as vulnerable to beat-effects if a receiver doesn't update every time.
  cc/ @jihlein

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -82,7 +82,7 @@
 struct rcvr_activity_fsm {
 	ManualControlSettingsChannelGroupsOptions group;
 	uint16_t prev[RCVR_ACTIVITY_MONITOR_CHANNELS_PER_GROUP];
-	uint8_t sample_count;
+	uint8_t check_count;
 };
 
 extern uintptr_t pios_rcvr_group_map[];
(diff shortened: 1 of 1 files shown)


mlyle:
This next commit may be the 'receiver activity doesn't detect activity bug'.  If we have just one receiver, and it updates only every other cycle, we'll only be watching it every other time. 

Next commit will adjust to sample each receiver more than once in a row.

jihlein:
This solves the problem of no activity in the ReceiverActivity UAVO when sticks are moved on my SciSky board.  RC Input Configuration Wizard functions normally too.

mlyle:
@jihlein Thanks for the test and the confirmation.

jihlein (commented):
Code LGTM.  Tested on SciSky FC where problem originally observed, now works correctly.  If it's thought that's enough testing, ready for merge in my opinion.  Otherwise I'll work it into an F3 and F4 target this afternoon/evening.

jihlein:
Went ahead and built for DTFc and AQ32.  

Loaded code into DTFc and flew with existing configuration.  ReceiverActivity UAVO updates as expected, flew fine.  This test was with DSM.

Loaded code into AQ32, cleared settings.  Setup PPM with RC Input Configuration Wizard, all channels detected properly.  ReceiverActivity UAVO updates as expected.  Only bench test so far due to wind, but don't expect any issues.

jihlein:
Weather looks bad tomorrow, so flew the AQ32 just now.  All's good wrt this change from what I can tell.

mlyle:
@jihlein thanks much

PR #1410, opened 2016-10-07
gcs: esc calibration: send no pulses for 1.25s first

### LINKED ISSUE #1170: ESC calibration safety improvement?
Can we send no PWM for a bit to disarm the ESC, in case the user didn't listen and still has vbat connected?

mlyle:
Fixes #1170.

If the user doesn't listen and detach the battery and props, this gives
the ESC a chance to disarm by having no pulses.  No guarantees, etc...
Just slightly safer.

### DIFF
--- a/ground/gcs/src/plugins/config/configoutputwidget.cpp
+++ b/ground/gcs/src/plugins/config/configoutputwidget.cpp
@@ -366,6 +366,30 @@ void ConfigOutputWidget::startESCCalibration()
     actuatorCommand->setMetadata(mdata);
     actuatorCommand->updated();
 
+    mbox.setText(QString(tr("Please wait...")));
+
+    // Set output for motors to nil to disarm
+    // Should not be necessary; battery is supposed to be disconnected, etc.
+    // But this makes it moderately safer if the user does not obey.
+    for (unsigned int i = 0; i < ActuatorCommand::CHANNEL_NUMELEM; i++)
+    {
+        // Check if the output channel was selected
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/config/configoutputwidget.cpp
```diff
@@ -366,6 +366,29 @@ void ConfigOutputWidget::startESCCalibration()
...
+
+    QEventLoop loop;
+
+    QTimer timeout;
+    timeout.setSingleShot(true);
+    connect(&timeout, SIGNAL(timeout()), &loop, SLOT(quit()));
+    timeout.start(1250);        // 1.25 seconds
+    loop.exec();
+
+    actuatorCommand->setData(actuatorCommandFields);
```
tracernz: Don't you want to do this before loop.exec() so the output is disabled during that 1.25 s?

tracernz (changes requested):

ground/gcs/src/plugins/config/configoutputwidget.cpp
mlyle: LOL that might help huh :)

tracernz (approved):

tracernz:
Looks good now, I'm confident it will work, but have you tested at all? :stuck_out_tongue:

mlyle:
i have not yet, no :(

tracernz:
I really need to double-check this and merge. :D

tracernz:
Works great. :+1:

PR #1411, opened 2016-10-07
flight: osd: fix theoretical off by one

### LINKED ISSUE #1120: OnScreenDisplay: Boot reason can write 1 byte past end of buffer
Not a real problem as long as all the reasons are shorter than the buffer but should be fixed anyway.

mlyle:
fixes #1120

### DIFF
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -727,12 +727,16 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
 	if (PIOS_Thread_Systime() < BOOT_DISPLAY_TIME_MS) {
 		const char *boot_reason = AlarmBootReason(alarm.RebootCause);
 		strncpy((char*)buf, boot_reason, sizeof(buf));
-		buf[strlen(boot_reason)] = '\0';
+		buf[sizeof(buf) - 2] = '\0';
 		pos = strlen(boot_reason);
 		buf[pos++] = ' ';
 	}
 
 	uint8_t state;
+
+	// With the above arrangement, can pass a length of 1 to this if
+	// there's an impossibly long boot reason.
+	// which then does the right thing and just fills it with a null.
 	int32_t len = AlarmString(&alarm, buf + pos, sizeof(buf) - 1 - pos,
 			blink, &state);
 
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -728,7 +728,7 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
 		const char *boot_reason = AlarmBootReason(alarm.RebootCause);
 		strncpy((char*)buf, boot_reason, sizeof(buf));
 		buf[sizeof(buf) - 2] = '\0';
-		pos = strlen(boot_reason);
+		pos = strlen(buf);
 		buf[pos++] = ' ';
 	}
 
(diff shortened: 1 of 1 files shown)


flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -727,12 +727,16 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
 	if (PIOS_Thread_Systime() < BOOT_DISPLAY_TIME_MS) {
 		const char *boot_reason = AlarmBootReason(alarm.RebootCause);
 		strncpy((char*)buf, boot_reason, sizeof(buf));
-		buf[strlen(boot_reason)] = '\0';
+		buf[sizeof(buf) - 2] = '\0';
 		pos = strlen(boot_reason);
 		buf[pos++] = ' ';
```
tracernz, line 732: This will still write over the end if boot_reason is 100 or more chars long right?

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -727,12 +727,16 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
 	if (PIOS_Thread_Systime() < BOOT_DISPLAY_TIME_MS) {
 		const char *boot_reason = AlarmBootReason(alarm.RebootCause);
 		strncpy((char*)buf, boot_reason, sizeof(buf));
-		buf[strlen(boot_reason)] = '\0';
+		buf[sizeof(buf) - 2] = '\0';
```
tracernz, line 730: Maybe I'm missing something obvious.. but why -2 rather than -1?

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -727,12 +727,16 @@ void draw_alarms(int x, int y, int xs, int ys, int va, int ha, int flags, int fo
...
 		pos = strlen(boot_reason);
 		buf[pos++] = ' ';
 	}
 
 	uint8_t state;
+
+	// With the above arrangement, can pass a length of 1 to this if
+	// there's an impossibly long boot reason.
+	// which then does the right thing and just fills it with a null.
 	int32_t len = AlarmString(&alarm, buf + pos, sizeof(buf) - 1 - pos,
```
tracernz: This may also write past the end if boot_reason is more than 99 chars (pos = strlen(boot_reason)). Also, size_t is unsigned (~~according to Google... but I'm not sure that's correct for arm-none-eabi-gcc...~~ guaranteed by C standard) so this could go very badly if (pos + 1) is bigger than sizeof(buf)... https://github.com/d-ronin/dRonin/blob/b323ecce4bc16b80aa638fdff07793a872d2e3f7/flight/Libraries/alarms.c#L329

tracernz (changes requested):
Strings in c... :/

flight/Modules/OnScreenDisplay/onscreendisplay.c
tracernz: https://ideone.com/V61iXx

flight/Modules/OnScreenDisplay/onscreendisplay.c
mlyle: ugh, supposed to be strlen(buf).

Then buf[98] = 0
strlen(buf)==98;
buf[98] = 0;
pos=99

flight/Modules/OnScreenDisplay/onscreendisplay.c
mlyle: Room for space.

mluessi:
LGTM

tracernz (approved):

tracernz:
I still don't super like how this whole string building works but it's better than before. :+1:

Issue #70, opened 2015-11-28
Usability: Input configuration


jhitesma:
Display of current TX values is disappointing for users coming from Base/Cleanflight.  In their configuration app the values are displayed as a bargraph that updates in real time making it very easy to tell how your TX/RX are working.  

In our GCS the numeric value is shown to the right - but it updates slowly and isn't obvious so it's easy for a new user to miss.  I've always found the delay in values updating to be really annoying when trouble shooting common issues like figuring out why stick arming isn't working.  

Going into manual calibration values display in real time.  But every time you go into manual calibration the existing setup is wiped and arming is disabled so you have to reset everything.  This feels really cumbersome and annoying when you're first coming over from naze-land and I still find it frustrating when I'm having an input calibration issue.  

Also min/max values can be input directly, but neutral values can only be set with a slider - which is hard to adjust in small increments.  It would be nice to be able to input neutral values directly without having to use the UAVO browser which is intimidating to new users.

DTFUHF:
This is a pretty old issue for me that never really got addressed, partially because the Qt way of doing this easily is ugly, and the Qt way of doing this nicely is hard.

tracernz:
Would be good to get this done for Tanto. I quite like the way Libre does this.

tracernz:
From @DTFUHF in #128:

> There's an awesome virtual transmitter that aids input channel setup greatly, but it's only available in the channel input wizard. Adding this to the input channel setup screen would help immensely in understanding which controller channel does what, if it's backwards, etc.

mlyle:
@tracernz did this!

PR #1417, opened 2016-10-09
GCS:Input: Show current values on slider, add reverse button

tracernz:
As per title. The logic for neutral when reversing is to preserve it's current value except for throttle, where it is switched to the other end (so actually, it stays in the same place graphically). Feedback wanted on this behaviour.

newchans

The response is a little slow because ManualControlCommand only updates every 1s by default. Should we reduce that?

Fixes the main part of #70.

### DIFF
--- a/ground/gcs/src/plugins/config/configinputwidget.cpp
+++ b/ground/gcs/src/plugins/config/configinputwidget.cpp
@@ -122,6 +122,7 @@ ConfigInputWidget::ConfigInputWidget(QWidget *parent) : ConfigTaskWidget(parent)
         int index2 = manualCommandObj->getField("Channel")->getElementNames().indexOf(name);
         if (index2 >= 0) {
             addUAVObjectToWidgetRelation("ManualControlCommand", "Channel", inpForm->ui->channelCurrent, index2);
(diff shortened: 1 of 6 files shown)


mlyle:
I think we should do manualcontrol updates at like, 166ms or something.  Not spend tons of bandwidth on it but at least get a few frames per sec.

tracernz:
Yep, 200 ms look okay, so 166 should be better.

tracernz:

mlyle:
you want the metadata change to be a subsequent PR?

tracernz:
Probably not a bad idea since it's relatively higher risk.

mluessi:
This looks awesome :)

tracernz:
New commit that cleans up the code a little (and prepares it for re-use for something else I have in mind for a future PR), and improves appearance with anti-aliasing.

tracernz:
Could do this for output form too if desired, for only a few lines:
output

tracernz:
Haven't had any feedback, good or bad, so I'll assume good. 😀

mlyle:
I love your screenshot, i just need ought to try it to merge :/

mlyle:
'Tis wunnerful :D

PR #1422, opened 2016-10-11
Flight: BrainRE1 - Move to new buzzer infrastructure.

jihlein:
Moves BrainRE1 buzzer control to new annunciator infrastructure.

Tested failsafe after initial arm, buzzed out "R" in morse as expected.

Changed the buzzer severity level, and it buzzed out "T" when disarmed and "I" when armed, again as expected.

### DIFF
--- a/flight/Modules/System/systemmod.c
+++ b/flight/Modules/System/systemmod.c
@@ -52,10 +52,6 @@
 #include "systemstats.h"
 #include "watchdogstatus.h"
 
-#if defined(PIOS_INCLUDE_RE1_FPGA)
-#include "fpga_drv.h"
-#endif
-
 #if defined(PIOS_INCLUDE_DEBUG_CONSOLE) && defined(DEBUG_THIS_FILE)
 #define DEBUG_MSG(format, ...) PIOS_COM_SendFormattedString(PIOS_COM_DEBUG, format, ## __VA_ARGS__)
 #else
--- a/flight/Modules/System/systemmod.c
+++ b/flight/Modules/System/systemmod.c
@@ -52,6 +52,10 @@
 #include "systemstats.h"
 #include "watchdogstatus.h"
 
+#if defined(PIOS_INCLUDE_RE1_FPGA)
+#include "fpga_drv.h"
+#endif
+
 #if defined(PIOS_INCLUDE_DEBUG_CONSOLE) && defined(DEBUG_THIS_FILE)
 #define DEBUG_MSG(format, ...) PIOS_COM_SendFormattedString(PIOS_COM_DEBUG, format, ## __VA_ARGS__)
 #else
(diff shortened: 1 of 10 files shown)


flight/Modules/System/systemmod.c
```diff
@@ -363,11 +367,26 @@ static inline void consider_annunc(AnnunciatorSettingsData *annunciatorSettings,
 		}
 	}
 
+#if defined(PIOS_INCLUDE_RE1_FPGA)
```
mlyle: So hshred and I were talking about this before-- and I was wondering if you'd be willing to do this slightly differently.

We had the idea of maybe adding a callback to the pios_annunciator config that is used to change the value.  If it's not-NULL, the callback is called instead of setting a GPIO.  Then re1 could just provide a function that enables / disables the buzzer.

mlyle (changes requested):
I like the UI and hook changes.  If you don't want to do the factoring into annunciator-- no biggie, I can do it in a few days and tack it onto this PR.

jihlein:
Hate to make more work for you, but I'm not sure how to implement what you are asking........

mlyle:
I pushed a change-- does this code work and look reasonable?  thx

mlyle (approved):

flight/targets/brainre1/board-info/board_hw_defs.c
```diff
@@ -68,7 +71,12 @@ static const struct pios_annunc pios_annuncs[] = {
 		},
 		.remap = 0,
 		.active_high = true,
-	}
+	},
+#if defined(PIOS_INCLUDE_RE1_FPGA)
```
jihlein, line 75: Since this is the BrainRE1 target, and the FPGA is an integral part of it, are lines 75 and 79 necessary?

jihlein:
Looks fine, one question above, see above comment.  Will test this evening.

flight/Modules/System/systemmod.c
```diff
@@ -52,6 +52,10 @@
 #include "systemstats.h"
 #include "watchdogstatus.h"
 
+#if defined(PIOS_INCLUDE_RE1_FPGA)
```
mluessi: Shouldn't be needed now that the handler is used.

flight/targets/brainre1/board-info/board_hw_defs.c
mluessi: True, but I think it's used elsewhere in the same file, so we can also leave it there.

mluessi:
Thanks @jihlein and @mlyle. Other than the comment about the header in systemmod, it looks good to me.

flight/targets/brainre1/board-info/board_hw_defs.c
mlyle: It's necessary because the fpga support isn't linked into the bootupdater, so it gets a link error.

mluessi: It is linked in the bootloader. Without it, there wouldn't be any LEDs on RE1 when the BL is active ;)

mlyle: I said bootupdater, not bootloader.  If you remove the ifdef it blows up linking in the bootupdater.

jihlein:
Tested as above in first post.  Works correctly with the refactoring changes.

mlyle:
@tracernz / @hshred -- willing to merge?

tracernz (approved):
LGTM. @mluessi should do this one I think.

mluessi:
Thanks!

PR #1427, opened 2016-10-15
GCS:Input: Remember previous channel selection when changing rx type

tracernz:
Very frustrating (probably only) for devs regularly switching between rx protocols.


tracernz:
Hopefully won't conflict with #1417 :/

PR #1430, opened 2016-10-15
flyingpi: hw: bump to rev B.

mlyle:
Just correcting the silk.


mlyle:
Loaded gerbers, they look OK.

PR #1434, opened 2016-10-18
GCS: Increase stack size to 8mb on Windows

tracernz:
Attempt to combat stack overflow bug reports. Main thread on OSX gets 8mb, threads on typical Linux distro get 8mb.


mlyle:

PR #1436, opened 2016-10-18
CC3D: Attempt to fix PWM channel 1 input (plus some general bugfixes)

### LINKED ISSUE #1435: CC3D: PWM Rx and SyncPWM on Outputs 1-3 Incompatible
Due to timer conflict. Previous attempts to force this timer when PWM selected aren't effective.

tracernz:
Comments in commit messages. PIOS_PWM now handles timer overflow correctly which allows the force_1MHz config option to work properly. Still lose an overflow event once every few seconds, but much less frequently than before. The result is pulse measurement short by ARR (400 us on CC3D). Happens when overflow event occurs between these two lines I believe:

``` c
uint16_t flags = timer->SR;
timer->SR = ~(TIM_IT_Update | TIM_IT_CC1 | TIM_IT_CC2 | TIM_IT_CC3 | TIM_IT_CC4);
```

Fixes #1435.

### DIFF
--- a/flight/PiOS/STM32/pios_pwm.c
+++ b/flight/PiOS/STM32/pios_pwm.c
@@ -45,21 +45,25 @@ const struct pios_rcvr_driver pios_pwm_rcvr_driver = {
 /* Local Variables */
 /* 100 ms timeout without updates on channels */
 const static uint32_t PWM_SUPERVISOR_TIMEOUT = 100000;
+const static uint32_t PWM_CAPTURE_MAX = 3000;
 
 enum pios_pwm_dev_magic {
 	PIOS_PWM_DEV_MAGIC = 0xab30293c,
+};
+
+enum pios_pwm_capture_state {
+	PIOS_PWM_CAPTURE_RISING,
+	PIOS_PWM_CAPTURE_FALLING,
 };
 
 struct pios_pwm_dev {
 	enum pios_pwm_dev_magic     magic;
 	const struct pios_pwm_cfg * cfg;
 
-	uint8_t CaptureState[PIOS_PWM_NUM_INPUTS];
-	uint16_t RiseValue[PIOS_PWM_NUM_INPUTS];
-	uint16_t FallValue[PIOS_PWM_NUM_INPUTS];
-	uint32_t CaptureValue[PIOS_PWM_NUM_INPUTS];
-	uint32_t CapCounter[PIOS_PWM_NUM_INPUTS];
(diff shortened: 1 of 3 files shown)


tracernz:

mlyle:
Ah.  Well, those two lines are _wrong_ :D  Will commit a fix then ask jenkins again.

mlyle:

tracernz:
Oh yeah, that's a better idea. 😛

tracernz:
Need to check this on F4 (there are no F3 targets with PWM rx support?).

mlyle:
Yes.  I skimmed the F4 pios_tim code earlier and it looks.. more sane.

tracernz:
The F4 IRQ handler looks like a copy-paste of the previous F1 code so it will have the same issue.

mlyle:
I don't think so?

https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/STM32F4xx/pios_tim.c#L144

Or am I missing something?

mlyle:
Oh, I see, you factored these tests out of the loop?  I thought the flags thing at the top was original code and not new.

tracernz:
Yah, I was trying to minimise the time between reading the flags and resetting them, and the ST reading function has quite a bit of code. I saw a very definite improvement in the frequency of missed overflows by doing that. Looking at the code in retrospect... I'm not totally convinced. :stuck_out_tongue:

mlyle:
@tracernz -- any chance you have tried since my commit-- if so I'm glad to merge this.

If not, I can look into getting a test rig together in the next couple days.

tracernz:
No I haven't. Maybe I can do it this evening.

tracernz:
I get an occasional disarm due to a lost overflow event still. Otherwise it does actually work (arguably worse than completely not working like before :wink: ).

mlyle:
> I get an occasional disarm due to a lost overflow event still.

Yes, that sounds super bad.  I wonder why.

It's an actual -disarm-?  Not just a single-frame glitch?

mlyle:
One alternative thing we could do is keep the current PWM code (which I don't really like), and choose a oneshot output rate that guarantees we can measure up to 2.5ms pulses--e.g. maybe a fixed 400Hz.

tracernz:
That's probably the best option IMO. And yes, an actual disarm.

tracernz:

mlyle:
image

PR #1437, opened 2016-10-19
GCS: Elide serial port names

tracernz:
Can make dropdown really big on OSX with long names..

Now looks like:
untitled


mluessi:
Nice. LGTM

PR #1439, opened 2016-10-20
WS2811: bare bones DMA driver

mlyle:
Supports lots of LEDs (tested to 7, but should go well beyond).  No real software support.  Support on quanton target.

Changes to Quanton:
- Relocate DMA channel to avoid conflict
- Support new PPM-in location to eliminate dependency on TIM1

For subsequent PRs:
- prevent refreshes before the LEDs have had time to 'reset'
- hook into annunciator subsystem
- other F4 targets
- support setting baseline color
- some trivial animation

### DIFF
--- a/flight/PiOS/STM32F4xx/pios_ws2811.c
+++ b/flight/PiOS/STM32F4xx/pios_ws2811.c
@@ -177,11 +177,16 @@ int PIOS_WS2811_init(ws2811_dev_t *dev_out, const struct pios_ws2811_cfg *cfg,
 	TIM_DMACmd(cfg->timer, TIM_DMA_CC1, ENABLE);
 	TIM_DMACmd(cfg->timer, TIM_DMA_CC2, ENABLE);
 	TIM_DMACmd(cfg->timer, TIM_DMA_CC4, ENABLE);
-	//TIM_SelectCCDMA(cfg->timer, ENABLE);
 
 	TIM_SetCompare4(cfg->timer, 1);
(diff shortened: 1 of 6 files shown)


mlyle:
image
image

mlyle:
image

tracernz:
Yay. :+1: 

capture

img_0506

Shame about the PPM pin change :(.

mluessi:
Nice :); I like the double buffered DMA approach.

flight/PiOS/STM32F4xx/pios_ws2811.c
```diff
@@ -0,0 +1,386 @@
...
+
+#include "pios_ws2811.h"
+
+struct ws2811_pixel_data_s {
+	uint8_t g;
+	uint8_t r;
+	uint8_t b;
+};
+
+// Each DMA side is 6 LEDs, 24 bits per pixel, 2 bytes per bit, for 288
```
mluessi, line 55: I think multiples of four LEDs are most commonly used, so using 4 instead of 6 may save some memory in some cases. It doesn't really matter though ;)

mlyle: Yah.  I picked 6 because if people are using more, I need to manage the interrupt rate, as well as the amount of work during each interrupt.  This results in about a 3KHz interrupt rate during blitting out the pixel data, having to fill 144 bytes in an array on each one.

PR #1441, opened 2016-10-23
Refactor External Magnetometer Support, Add HMC5983

tracernz:
- Move to common shared mag type enum in HwShared
- Add HMC5983 as an option (my original goal :stuck_out_tongue:), attn @tgreer 
- Add missing HTML for Sensors-Error state in GCS SystemHealth
- Improve tooltips for ext. mag. configuration
- New configuration field for Quanton and Sparky2 to select which port ext. mag. is on.

What still annoys me with this:
- Can only override parent enum options in UAVOs, can't exclude if you just don't want a specific one (e.g. internal mag). Adding ability to exclude options to UAVOGen would reduce target maintenance e.g. when a new mag type is added, and avoid cases where it is forgotten on some targets.
- uintptr_t and friends rather than actual pointers :stuck_out_tongue: 

Tested that it boots on a few boards and detects missing sensors when I select external option. Haven't actually tested that external mags work.


tracernz:
@jihlein Would appreciate if you could have a look over this and maybe test it to make sure it hasn't broken your stuff. :smiley:

mluessi:
Nice. Code LGTM.

jihlein:

mlyle:
```
pios_board.c:643:26: error: initialization makes integer from pointer without a cast [-Werror=int-conversion]
    uint32_t i2c_mag_id = NULL; /* TODO change to a real pointer */
```

tracernz:

tracernz:
@jihlein ^Windows builder came back to life.

jihlein:
@tracernz - I'll try to get some testing done tonight.

shared/uavobjectdefinition/hwaq32.xml
```diff
@@ -146,10 +146,10 @@
 		<field name="MPU6000DLPF" units="" type="enum" elements="1" options="188,98,42,20,10,5" defaultvalue="188">
 			<description>Gyro filter cutoff frequency</description>
 		</field>
-		<field name="Magnetometer" units="function" type="enum" elements="1" options="Disabled,Internal,External" defaultvalue="Internal">
-			<description>Connected magnetometer</description>
+		<field name="Magnetometer" units="" type="enum" elements="1" parent="HwShared.Mag" defaultvalue="Internal">
+			<description>Which magnetometer to use. If external, must be connected to flexi port, and flexi port configured as I2C.</description>
```
jihlein: The concept of flexi port does not exist on the AQ32 target, it has a dedicated external I2C bus that can't be configured as a UART.  In this case, the description can simply be "Which magnetometer to use."

tracernz: Ah, copy-pasta. Thanks. :+1:

jihlein:
Tested AQ32 and Quanton with HMC5883 ext mag.

On AQ32:
All combinations of Disabled, Internal, and External Mag, with the mag both powered an unpowered worked properly.  The sensor and I2C WCA blocks displayed status correctly.  Ext mag orientation also worked.

On Quanton:
Same tests as AQ32, but on both flex port 1 and flex port 3.  Everything also worked as expected.

Can't test a revo target.  I might have an HMC5983 breakout board somewhere, will need to dig for it if testing the HMC5983 is important.

tgreer:
@tracernz can confirm RE1+HMC5983 works

PR #1452, opened 2016-11-07
stm32f0: enable the ADC clock at startup

### LINKED ISSUE #1429: FlyingPi: ADC sometimes returns all 1's for all values
I believe this is a firmware bug but could be hardware.

mlyle:
This may explain the (! intermittent !) failures of the ADC.

Fixes #1429 

Needs verification.


tracernz:
_golf clap_

😜

mlyle:
So I'm hesitant to say this fixes everything, as the underlying issue has been intermittent.  But got through 3 consecutive boots with ADC happy (have 1V unexplained of offset still, but.. meh :D

mlyle:
Also note: the old code should be OK, unless there's a lost event... this new code is more robust to lost events.

tracernz:
Looks good, thanks. 😄

PR #1456, opened 2016-11-11
GCS: Upgrade to Eigen 3.3

### LINKED ISSUE #664: GCS:Libs: Upgrade to Eigen 3.3

tracernz:
Fixes #664.

Also cleans up a bit of code in Calibration.


tracernz:

ground/gcs/src/plugins/config/calibration.cpp
```diff
@@ -387,20 +386,21 @@ void Calibration::timeout()
...
         // Do nothing
         return;
-        break;
     case YAW_ORIENTATION:
         // Disconnect appropriate sensors
         connectSensor(ACCEL, false);
         calibration_state = IDLE;
         emit showYawOrientationMessage(tr("Orientation timed out ..."));
         emit yawOrientationProgressChanged(0);
+        break;
```
tracernz, line 395: This looked like a bug before?

mlyle: Ooh.  Just a bug in reported message, but still.

PR #1457, opened 2016-11-13
GCS: Disable GNU C++ extensions

tracernz:
To ensure better portability to other compilers, e.g. MSVC. We don't usually build with anything other than GCC in our CI system so we'd miss if people introduce GNU extensions.

This causes (for GCC/clang) these flags to be inserted in Makefiles:
* Qt 5.6.1 (our current version)
    - `-std=c++0x` rather than `-std=gnu++0x`
* Qt 5.7+ (C++11 is default and enforced as a minimum)
    - `-std=c++11` rather than `-std=gnu++11`

Doesn't affect any other compilers.


tracernz:

PR #1459, opened 2016-11-16
Addition of newly added input types to radio setup wizard

yamahabest:
Open items:
- [x] It is building successfully, however on startup the GCS can't load Configd.dll: "A dynamic link library (DLL) initialization routine failed". I have looked through the changes, but can't find anything pointing to this. Help would be appreciated. This is also causing these changes to be untested yet.
- [x] According to the original code, Lux/Sparky can't do SUMH, but can do SUMD, is that correct?
- [x] On CC3D, the FlexiPort was/is always configured for telemetry when choosing SBUS, seems strange to me, but could be OK?
- [x] On Quanton, all options that have "PPM" in their naming, are now considered as being PPM input type. Would that be OK?

This still doesn't solve the comment of @tracernz on the previous PR (#1350): "I'm not sure this is a good idea since the wizard never sets SUMD etc. on this port. It will make the wizard think it doesn't need to reboot when it sets them on UART3 and they're already on UART4 or 6 (this case is broken for a different reason but we shouldn't make it worse here)."

If there is a good way to solve that, I am happy to implement that.


yamahabest:
Oh, didn't notice your comments. Will fix those!

yamahabest:
Ok, I have been able to test this when building the stuff as release.
I have tested that each selection of input type, in the Vehicle Setup wizard, results in:
- The right type set on the Hardware tab.
- The right type set on the RC Input tab.

Did this both on AQ32 and Brain.

I would appreciate an answer on the three open items in the first comment. After solving those, I can remove the WIP label.

tracernz:
I think the SUMD/SUMH thing is not a real limitation. The CC3D thing sounds strange.

yamahabest:
I could add SUMH to those two targets.
And I could remove the configuring of the flexiport on the CC3D.

What do you think of the Quanton question?

tracernz:
Quanton will be okay. Hopefully I will remove these functions in future anyway, and make the connection diagram much more general rather than tied to the setup wizard (with a view to removing the setup wizard in favour of something that guides you through the normal UI).

yamahabest:
Ok, added SUMH to Lux and Sparky, removed configuring telemetry on flexiport on CC3D.
And noticed that SUMD/SUMH were also excluded from Sparky2, added them there.

Removing the WIP text in the title, I am not able to remove the WIP label myself.

mlyle:
@tracernz -- since you looked at this before, do you have an idea as to status/enxt steps?

tracernz:
@icee, unstaged changes on my local checkout of mpl-seppuku I have changed connection diagrams a bit. It now looks for SVG elements named `hw<board>-<field>-<field.value> for field in hw<board>.fields`. This way it will handle all configurations rather than just what the setup wizard does. Of course SVGs will need the additional elements to be added for it to be useful.

ground/gcs/src/plugins/boards_taulabs/sparky.cpp
```diff
@@ -174,16 +181,46 @@ enum Core::IBoardType::InputType Sparky::getInputType()
...
+    case HwSparky::FLEXIPORT_DSM:
+        return INPUT_TYPE_DSM;
+    case HwSparky::FLEXIPORT_SBUS:
+        return INPUT_TYPE_SBUS;
+    case HwSparky::FLEXIPORT_SBUSNONINVERTED:
+        return INPUT_TYPE_SBUSNONINVERTED;
+    case HwSparky::FLEXIPORT_IBUS:
+        return INPUT_TYPE_IBUS;
+    case HwSparky::FLEXIPORT_SRXL:
+        return INPUT_TYPE_SRXL;
```
mlyle, line 215: default: break; is needed on all switch statements that are not expected to cover all possible values in an enum

tracernz:
Thanks for cleaning this up a bit. I notice on some of the switch statements you don't have/have removed the `default` case but haven't handled all the possible enumerations. That will cause compile errors (and it is a very good thing because it allows us to find out at compile time that we forgot to handle a newly added enum somewhere, e.g. when we add a new Rx type and didn't handle it in the Input configuration page). It also makes static analysis in Qt Creator noisy. ;) Otherwise, this is looking much better. :+1:

tracernz:
jenkins, test this please

ground/gcs/src/plugins/boards_dtf/dtfc.h
```diff
@@ -36,37 +36,35 @@
...
     virtual ~Dtfc();
 
-    virtual QString shortName();
-    virtual QString boardDescription();
-    virtual bool queryCapabilities(BoardCapabilities capability);
-    virtual QPixmap getBoardPicture();
-    virtual QString getHwUAVO();
-    virtual int queryMaxGyroRate();
+    QString shortName();
+    QString boardDescription();
```
mlyle: Why is the 'virtual' keyword removed in these types of places?  We don't do it everywhere, but I like it because of http://stackoverflow.com/questions/1370976/c-style-prefixing-virtual-keyword-to-overridden-methods   ... in any case this does not seem to be minimal/required

tracernz:
jenkins, ok to test

ground/gcs/src/plugins/boards_openpilot/cc3d.cpp
```diff
@@ -100,9 +99,14 @@ QString CC3D::getHwUAVO()
 
 
 //! Determine if this board supports configuring the receiver
-bool CC3D::isInputConfigurationSupported(enum InputType type = INPUT_TYPE_ANY)
+bool CC3D::isInputConfigurationSupported(Core::IBoardType::InputType type)
 {
-    Q_UNUSED(type);
+    switch (type) {
+    case INPUT_TYPE_PWM:
```
mlyle: Is this correct?  I think for CC3D, Quanton, revo PWM should be supported.

mlyle (commented):
Thanks for your work here-- and sorry for being slow on feedback with this one.

ground/gcs/src/plugins/boards_dtf/dtfc.h
yamahabest: I figured the Seppuku implementation was the way to go, as that was the newest one. Will revert!

ground/gcs/src/plugins/boards_openpilot/cc3d.cpp
yamahabest: Copy-paste error. Will fix.

ground/gcs/src/plugins/boards_dtf/dtfc.h
mlyle: Ah, if this happened on Seppuku I'm not sure it was on purpose.

ground/gcs/src/plugins/boards_taulabs/sparky.cpp
yamahabest: Done.

yamahabest:
Does this one need additional changes?

tracernz:
Looks pretty good now.

jenkins, test this please

mlyle:

mlyle:
@yamahabest -- thanks for your patience with this :D

yamahabest:
Oh no, it got merged! Just when I got used to it being open. Thanks guys!

PR #1461, opened 2016-11-19
worldmagmodel: fixes for simulation

mlyle:
So it can compile with indentation warnings on gcc 6.2.0.

### DIFF
--- a/flight/Libraries/WorldMagModel.c
+++ b/flight/Libraries/WorldMagModel.c
@@ -906,8 +906,8 @@ int WMM_SummationSpecial(WMMtype_SphericalHarmonicVariables *
 
     float PcupsS_s[NUMPCUPS];
     float       *PcupS = PcupsS_s;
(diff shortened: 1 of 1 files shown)


tracernz:
Hard to tell on phone.. But is the file a mix of tabs and spaces? Should re-indent the whole thing and fix it off so IMO. 😃

mlyle:
Problem is it's an upstream file.

tracernz:
Hmm, are we ever likely to update this?

mlyle:
Yes.  Navigation requires it to be updated every couple of years, because Earth's magnetic field drifts and this only contains a forecast for the drift vs. time.

It's enough drift that runways (which are numbered in 10s of degrees, e.g. runway 31 has a heading between 305 and 315 degrees magnetic) are having their runway number increase here when repainted.  And every degree of heading accuracy helps navigation.

PR #1468, opened 2016-11-27
USB product ID refactor

mlyle:
Main change is to move the bootloader to a common vendor/product id.

Also bump bootloader rev to 0x87.

This is beneficial so that androidgcs can refuse to try connecting
to the bootloader.  It should simplify things going forward.

It will be quite some time before most of our users pick this up, so we
can't require this ID for quite some time.


tracernz:
> This is beneficial so that androidgcs can refuse to try connecting
to the bootloader. It should simplify things going forward.

And GCS as well.

mlyle:
Yes, though GCS is smart enough to look at the version stuff.  Android is not able to see whether something is bootloader or not (serial number, product "version") with our current mechanism on the old API versions we're forced to use for backwards compat.

PR #1469, opened 2016-11-27
relocate DONT_BUILD_IF to pios.h

mlyle:
Necessary if you want to use the pios_ws2811 places where it's not safe to include openpilot.h (e.g. bootloader)


tracernz:
Nice to have available generally too. :+1: 
jenkins, test this please

tracernz (approved):

PR #1472, opened 2016-11-29
WIP: Stabilization: Add VirtualBar2 stabilization + flight modes.

ekaszubski:
The current VirtualBar mode acts like a stabilization bar; that is, it resists all rate changes whether from the pilot or the environment. Essentially, you're always fighting the flybar in this mode.

This VirtualBar2 mode acts more like a paddle flybar, specifically a bell-hiller flybar:
- A component of pilot inputs go straight to the main rotor, enabling minimal control delay (VbarGyroSupress,Rate P-term)
- A component of pilot inputs go to changing the state of the flybar (analogous to changing the pitch of the paddles)(Rate I-term)
- As the flybar accelerates due to the new pilot-commanded paddle pitch, it contributes additional control output (Rate I-component)
 - Since the flybar is rotating, it also resists rate changes relative to its current orientation
 - Over time, the flybar settles to the position of least resistance, where gyroscopic and aerodynamic forces balance out (Rate I-accumulator decay)

In terms of implementation, VirtualBar2 is just Rate mode with a constantly decaying I-accumulator, and with the same manual/stabilization mixing as AcroPlus.

Rough simulations of lift force vs angle of attack suggest that using a decaying integral should increase performance at a wide range of rotation rates.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -488,6 +488,7 @@ int32_t transmitter_control_select(bool reset_controller)
 	case FLIGHTSTATUS_FLIGHTMODE_ACROPLUS:
 	case FLIGHTSTATUS_FLIGHTMODE_LEVELING:
 	case FLIGHTSTATUS_FLIGHTMODE_VIRTUALBAR:
+	case FLIGHTSTATUS_FLIGHTMODE_VIRTUALBAR2:
 	case FLIGHTSTATUS_FLIGHTMODE_HORIZON:
 	case FLIGHTSTATUS_FLIGHTMODE_AXISLOCK:
 	case FLIGHTSTATUS_FLIGHTMODE_STABILIZED1:
@@ -979,6 +980,7 @@ static inline float scale_stabilization(StabilizationSettingsData *stabSettings,
 			return 0;
 		case SHAREDDEFS_STABILIZATIONMODE_MANUAL:
 		case SHAREDDEFS_STABILIZATIONMODE_VIRTUALBAR:
+		case SHAREDDEFS_STABILIZATIONMODE_VIRTUALBAR2:
 		case SHAREDDEFS_STABILIZATIONMODE_COORDINATEDFLIGHT:
 			return cmd;
 		case SHAREDDEFS_STABILIZATIONMODE_RATE:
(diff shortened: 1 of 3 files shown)


ekaszubski:
Tried this on the bench, swash movement looks reasonable.

ekaszubski:
An alternate model suggests that thrust vs AOA is actually much more linear than the original model predicted, which brings the performance of this flight mode into question. Need to run a data collection to know for sure.

tracernz:
I wonder if there's any reason to keep the old mode if the new one is better? Interested to see how you go with flight testing.

mlyle:
(if flight testing is wonderful and soon this could make Artifice, but unlikely)

ekaszubski:
@mlyle @tracernz I've done a bunch of analysis on cp heli flight characteristics over the last few days:
- VirtualBar:
  - I think we should ditch all forms of VirtualBar (including this PR) in favor of acro+. It appears to only be useful for very nonlinear systems (data collection verified that thrust vs AOA is mostly linear under typical flight conditions) or when people are abusing the I-term.
- CP Heli flight characteristics:
  - During a "fast" flip, as the rotor's angular velocity increases, its effective AOA decreases. As the aircraft approaches the desired rate, the P-term will decrease (with decreasing error), while the required cyclic output will actually increase (with increasing angular velocity), meaning a P-controller will necessarily undershoot the desired angular velocity. This can be partially remedied with a large I-gain, but that introduces other issues.
  - Thus, any cp heli setups out there are likely either "significantly" undershooting desired angular velocity (during acceleration maneuvers) or are abusing the I-term; a simulation of my ~3 lb heli with 380mm blades @ 2750 RPM suggests that the "real world" P-gain for roll (13.75) will undershoot the desired rate by ~15%, and the "max" P-gain for roll (26.00) will undershoot by ~7%, for any non-zero desired rate. The corresponding numbers for pitch are ~7% undershoot at a P-gain of 30 ("real-world"), and ~4% undershoot at a P-gain of 60 ("max"). The undershoot percentage decreases as a function of increasing rotor RPM and P-gain, and is roughly linear with respect to the angular velocity of the aircraft for a given rotor RPM and P-gain (increasing rotor RPM increases effective controller output scale and decreases the effects of changes in effective AOA vs angular velocity).
  - While an I-term with just the right I-gain (roughly 3x the P-gain for the conditions mentioned) _can_ approximate the undershoot, it introduces other performance issues (for example, at high positive angular momentum, a high negative desired rate causes overshoot at the new rate) and can be hard to determine experimentally.
  - A more general solution is to add a second P (or PI) controller that operates on _desired_ rate instead of _error in_ rate, since the extra control output required to eliminate the undershoot is roughly linear with respect to the desired rate, and should only change with rotor RPM (for a given aircraft). A controller using _actual_ rate instead of desired rate would likely perform similarly, but with a slightly longer settling time.
  - This applies mostly to cyclic movements, and to a lesser degree to yaw movements (tail rotor spins at higher RPM but also experiences higher relative air velocity during typical maneuvers) and collective movements (not sure if we actually want to filter these, with the exception of althold). It also has the benefit of creating symmetric _effective_ control outputs for both acceleration and deceleration by making the _actual_ control outputs asymmetric as a function of desired angular velocity.
  - Quads may see some of these effects, but only at very high angular velocity and/or very low prop RPM.
  - It should be safe to add this enhanced rate controller (can you think of a better name?) to the existing rate stabilization mode with its own separate PI gains set to 0 by default.

mlyle:
Unclear benefit, and closed for inactivity (no actionable items)

PR #1473, opened 2016-11-30
Brain/BrainRE1 OSD: set PAL/NTSC settings from OSD task not ISR

### LINKED ISSUE #1254: BrainRE1 watchdog reset
- Seen in version (release version or githash): 2016-07-20.1
- On flight controller (if applicable): BrainRE1

Issue details: Flight controller resets in-flight, message "BOOT: INDY WDOG"  shown on reboot (see video below). This has only be seen by very few users, so it could be a hardware defect (unlikely, I think), or something related to the particular setup where it occurs. Logging was enabled in the video below, but it also occurs without logging. 

https://www.youtube.com/watch?v=OBUXzPko5MA&feature=youtu.be

First thoughts on what could be causing this:
- Spurious interrupts to video driver / OSD
- BMI160 driver
- OSD code\* 
- VTX Config code*

*not sure if/how an rtos task would trigger the watchdog. Any other ideas?

mluessi:
Fixes #1254 

When the video system changed between PAL and NTSC, the ISR was setting the black and white levels. For BrainRE1, this resulted in a SPI transfer to the FPGA from the ISR, which could lock up when the SPI was in use (e.g. for logging). This PR moves the calls to the OSD task and cleans up some things. Same changes also applied for Brain for consistency.

Status: Bench tested for RE1 (could not reproduce problem anymore, even with 250Hz logging and a lot of spurious interrupts). Needs test on Brain

### DIFF
--- a/flight/PiOS/STM32F4xx/pios_video.c
+++ b/flight/PiOS/STM32F4xx/pios_video.c
@@ -95,6 +95,8 @@ volatile uint16_t active_line = 0;
 
 const struct pios_video_type_boundary *pios_video_type_boundary_act = &pios_video_type_boundary_pal;
 
+int8_t video_type_act = VIDEO_TYPE_NONE;
+
 // Private variables
 static int8_t x_offset = 0;
 static int8_t x_offset_new = 0;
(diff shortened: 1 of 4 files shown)


flight/PiOS/STM32F4xx/pios_video.c
```diff
@@ -95,21 +95,17 @@ volatile uint16_t active_line = 0;
 
 const struct pios_video_type_boundary *pios_video_type_boundary_act = &pios_video_type_boundary_pal;
 
+int8_t video_type_act = VIDEO_TYPE_NONE;
```
tracernz: Not a big fan of sharing a global between a low-level driver and high-level module.. :/

mluessi: Yeah, I agree. Any suggestions for a more elegant approach?

tracernz: Other than a getter function, none. At least that allows you to easily add locking later if needed etc. without having to search through lots of code. You could add an inline attribute if you really wanted but it's usually unnecessary.

mluessi: @tracernz code cleaned up. Oddly there was already a getter function but I somehow didn't use it ;)

mlyle (approved):
LGTM-- @tracernz ?

tracernz (approved):
Thanks @mluessi! LGTM

PR #1479, opened 2016-12-07
WIP: Stabilization: Add rate precompensation to inner loop

ekaszubski:
- During a "fast" flip, as the aircraft's angular velocity increases, its effective AOA decreases since the air entering the rotor already has some velocity. As the aircraft approaches the desired rate, the P-term will decrease (with decreasing error), while the required cyclic output will actually increase (with increasing angular velocity), meaning a P-controller will necessarily undershoot the desired angular velocity. This can be partially remedied with a large I-gain, but that introduces other issues.
- Thus, any cp heli setups out there are likely either "significantly" undershooting desired angular velocity (during acceleration maneuvers) or are abusing the I-term; a simulation of my ~3 lb heli with 380mm blades @ 2750 RPM, +/- 10 degrees cyclic pitch, suggests that the "real world" P-gain for roll (13.75) will undershoot the desired rate by ~20%, and the "max" P-gain for roll (35.00) will undershoot by ~9%, for any non-zero desired rate. The corresponding numbers for pitch are ~7% undershoot at a P-gain of 47.25 ("real-world"), and ~4% undershoot at a P-gain of 95 ("max"). The undershoot percentage decreases as a function of increasing rotor RPM and P-gain, and is roughly linear with respect to the angular velocity of the aircraft for a given rotor RPM and P-gain (increasing rotor RPM increases effective controller output scale and decreases the effects of changes in effective AOA vs angular velocity).
- While an I-term with just the right I-gain (roughly 3x the P-gain for the conditions mentioned) _can_ approximate the undershoot, it introduces other performance issues (for example, at high positive angular momentum, a high negative desired rate causes overshoot at the new rate) and can be hard to determine experimentally.
- A more general solution is to add a second P (or PI) controller that operates on _desired_ rate instead of _error in_ rate, since the extra control output required to eliminate the undershoot is roughly linear with respect to the desired rate, and should only change with rotor RPM (for a given aircraft). A controller using _actual_ rate instead of desired rate would likely perform similarly, but with a slightly longer settling time.
- This applies mostly to cyclic movements, and to a lesser degree to yaw movements (tail rotor spins at higher RPM but also experiences higher relative air velocity during typical maneuvers) and collective movements (not sure if we actually want to filter these, with the exception of althold). It also has the benefit of creating symmetric _effective_ control outputs for both acceleration and deceleration by making the _actual_ control outputs asymmetric as a function of desired angular velocity.
- Quads may see some of these effects, but only at very high angular velocity and/or very low prop RPM.

p-gain | no precompensation | precompensation
-|-|-
"real world" | sim-roll-lowp-noprecomp-cropped | sim-roll-lowp-precomp-cropped
"max" | sim-roll-highp-noprecomp-cropped | sim-roll-highp-precomp-cropped

description | image
-|-
undershoot at high p-gain | sim-roll-steep-highp-noprecomp-cropped
overshoot at high positive momentum, high negative desired rate | sim-roll-steep-highp-highi-cropped
precompensation applied | sim-roll-steep-highp-precomp-cropped

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -152,7 +152,19 @@ static float compute_inner_loop_precomp(uint32_t pid_group, uint8_t axis, float 
 
 static float compute_inner_loop_precomp(uint32_t pid_group, uint8_t axis, float desired_rate, float measured_rate, float dt)
 {
-	return pid_apply_setpoint(&pids[pid_group + axis], get_deadband(axis), desired_rate, measured_rate, dt) + settings.RatePrecompensation[axis] * desired_rate;
+	// select the rate source for precompensation
+	float precomp_rate = 0;
+	switch (settings.RatePrecompensationSrc) {
+		case STABILIZATIONSETTINGS_RATEPRECOMPENSATIONSRC_DESIRED:
+			precomp_rate = desired_rate;
+			break;
+		case STABILIZATIONSETTINGS_RATEPRECOMPENSATIONSRC_MEASURED:
+			precomp_rate = measured_rate;
+			break;
+		default:
+			break;
+	}
+	return pid_apply_setpoint(&pids[pid_group + axis], get_deadband(axis), desired_rate, measured_rate, dt) + settings.RatePrecompensation[axis] * precomp_rate;
 }
 
 static float get_throttle(StabilizationDesiredData *stabilization_desired, SystemSettingsAirframeTypeOptions *airframe_type)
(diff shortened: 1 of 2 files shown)


ekaszubski:
- Many flight modes had a copypasta for the "inner loop" calculation. This has been replaced with a `compute_inner_loop_precomp` function so that all modes dependent on rate can take advantage of the precompensation.
- RatePrecompensationSrc can be set to `Desired` or `Measured` rate to select which value will be used for precompensation.

ekaszubski:
Flew this today, using `Measured` for src. Performed some rocking motions but no fast flips; at the very least, this is not detracting from the flight performance. Will try again tomorrow with `Desired` for src.

mlyle:
Unclear benefit, and closed for inactivity (no actionable items)

PR #1481, opened 2016-12-10
Qt: Move to 5.6.2

tracernz:
Gives us some bugfixes, and works around https://forum.dronin.org/forum/d/243-building-dronin-qt-version-issue. Parameterizes the stuff in tools.mk for ease of use in future. Needs tested on Windows, and macOS (although I did verify the filenames etc. are correct, and with a suffixed release, e.g. 5.6.1-1).

My thoughts on the path forward:
 - Stick with 5.6 branch for Artifice
- Look at 5.8 branch (which should be stable right at the start of our cycle) for Post-Artifice
    - Brings us stable Qt3D, better OpenGL implementation (frequent source of crash reports on Windows), etc.

### DIFF
--- a/ground/gcs/gcs.pro
+++ b/ground/gcs/gcs.pro
@@ -14,9 +14,9 @@ unix:!macx:!isEmpty(copydata) {
 	QMAKE_EXTRA_TARGETS += bin_install
 }
 
-!equals(QT_MAJOR_VERSION, 5) | !equals(QT_MINOR_VERSION, 6) | !equals(QT_PATCH_VERSION, 1) {
+!equals(QT_MAJOR_VERSION, 5) | !equals(QT_MINOR_VERSION, 6) | !equals(QT_PATCH_VERSION, 2) {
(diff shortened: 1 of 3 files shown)


tracernz:
Another option http://blog.qt.io/blog/2016/12/14/qt-5-7-1-released/.

Note: both 5.6.2 and 5.7.1 fix the XCode 8 issue too (QTBUG-55915).

nanodude:
For windows, shouldn't we be installing the MSVC version, instead of MinGW?

tracernz:
Yes, but we need the MinGW environment for our Makefile and to build firmware so the MSVC offline installer isn't an option. Current path of least resistance for Windows is Qt online installer so you can select MSVC Qt _and_ MinGW as documented in our docs.

nanodude:
Agreed. It sucks to have to download and install 1GB worth of files if one just wants to compile firmware, though. In this case, I guess installing MinGW(-w64) separately would be sufficient, no?

PR #1449, opened 2016-11-04
Fancy LED stuff

mlyle:
* Robustness improvements to ws2811 driver
* SystemMod runs faster on RGB LED targets
* Functionality to blend throttle, accessory desired, etc. between colors
* Blending OK in HSV or RGB spaces
* Can also set other LEDs to a default color
* Functionality to direct alarms to subset of LEDs

No UI.

PR #1488, opened 2016-12-13
BrainRE1: Use Fancy LED stuff

### LINKED ISSUE #1485: RE1's WS2811 LED support very limited
Right now RE1 does its own thing for '2811 LEDs, allowing you to set a static color on startup and using a periodic timer to refresh them.

#1439 and #1449 added more general / featureful LED support for other targets, but RE1's LED driver is not hooked into these nor is the OSD configuration of the LED connected to the object (instead updating a HwBrainRe1 field).

Needs a decision from @mluessi on how to go forward-- and any implementation necessary... keep old LED support?  If not, how to migrate?

mluessi:
Makes BrainRE1 use the fancy LED stuff introduced in #1449. Removes the LED configuration from the hardware config in the GCS and the OSD. We should find a more general / better way to add this to the OSD / GCS.

Status: bench tested, works.

Fixes #1485

### DIFF
--- a/flight/targets/brainre1/re1fpga/fpga_drv.c
+++ b/flight/targets/brainre1/re1fpga/fpga_drv.c
@@ -38,6 +38,7 @@
 #include "fpga_drv.h"
 #include "stm32f4xx_rcc.h"
 #include "pios_spi_priv.h"
+#include "pios_ws2811.h"
 
 /**
 * RE1 Register Specification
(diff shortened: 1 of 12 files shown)


flight/targets/brainre1/re1fpga/fpga_drv.c
```diff
@@ -618,6 +565,90 @@ void PIOS_RE1FPGA_Set3DConfig(enum pios_video_3d_mode mode, uint8_t x_shift_righ
...
+
+void PIOS_WS2811_trigger_update(ws2811_dev_t ws2811_dev)
+{
+	uint16_t n_leds;
+	PIOS_Assert(ws2811_dev->magic == WS2811_MAGIC);
+
+	if (PIOS_RE1FPGA_ClaimBus() != 0)
+		return;
+
+	n_leds = MIN(ws2811_dev->max_idx + 1, 1024);
```
mlyle, line 644: Is this a threshold you tested this at?  (Since this is happening in System thread)

flight/targets/brainre1/re1fpga/fpga_drv.c
```diff
@@ -618,6 +565,90 @@ void PIOS_RE1FPGA_Set3DConfig(enum pios_video_3d_mode mode, uint8_t x_shift_righ
...
+
+struct ws2811_dev_s {
+#define WS2811_MAGIC 0x31313832		/* '2811' */
+	uint32_t magic;
+	uint16_t max_leds;
+	uint16_t max_idx;
+	struct ws2811_pixel_data_s pixel_data[0];
+};
+
+int PIOS_WS2811_init(ws2811_dev_t *dev_out, const struct pios_ws2811_cfg *cfg,
```
mlyle, line 585: I like this approach of replicating the signatures from the other driver to avoid conditionals in the users.

mlyle (approved):
Thanks for the quick turn on this-- getting this in early will really help the next release :D

flight/targets/brainre1/re1fpga/fpga_drv.c
mluessi: No, I tested it with 256 (the maximum the UAVO allows).

mlyle:
Thanks again!

PR #1489, opened 2016-12-14
revolution: enable internal pullups on I2C lines

mlyle:
Cheap clones do not place the resistors (and may not place the
peripherals, either).  At least allow us to verify the bus is clear in
this case so that we don't panic on boot.

Reported by IRC user tree`d


mlyle:

tracernz:
Pretty harmless, I say let's do it. 👍

tracernz (approved):
Thanks for indentation fixes :)

PR #1491, opened 2016-12-15
python: support hid on (at least) Linux

mlyle:
use like: `sudo python dronin-getconfig -u 20a0:4250`

Tested lightly with python2 and python3 on Linux.  Part of the manufacturing support for Seppuku FC.

Not perfect (especially performance) / likely to be an area of further work but is useful for basic things now.

### DIFF
--- a/python/dronin/telemetry.py
+++ b/python/dronin/telemetry.py
@@ -499,18 +499,25 @@ class HIDTelemetry(BidirTelemetry):
             try:
                 raw = str.split(vidpid, ':')
 
-                if len(raw) != 2:
+                if len(raw) > 2:
                     raise
 
                 vid = int(raw[0], base=16)
-                pid = int(raw[1], base=16)
+
+                if len(raw) == 2:
+                    if raw[1] != '':
+                        pid = int(raw[1], base=16)
+
             except Exception:
                 raise ValueError("Invalid value of vidpid")
 
         import usb.core
         import usb.util
 
-        dev = usb.core.find(idVendor = vid, idProduct = pid)
+        if pid is None:
+            dev = usb.core.find(idVendor = vid)
+        else:
+            dev = usb.core.find(idVendor = vid, idProduct = pid)
 
         if dev is None:
             raise RuntimeError("No fc found")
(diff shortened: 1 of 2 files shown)


python/dronin/telemetry.py
```diff
@@ -479,6 +479,137 @@ def _do_io(self, finish_time):
...
+
+        if vid is None:
+            try:
+                raw = str.split(vidpid, ':')
+
+                if len(raw) != 2:
+                    raise
+
+                vid = int(raw[0], base=16)
+                pid = int(raw[1], base=16)
```
tracernz: pid should be optional IMO. So you can go `20a0:` just like lsusb etc.

python/dronin/telemetry.py
```diff
@@ -479,6 +479,137 @@ def _do_io(self, finish_time):
...
+
+        # XXX do better things here
+        if dev.is_kernel_driver_active(0):
+            dev.detach_kernel_driver(0)
+        if dev.is_kernel_driver_active(1):
+            dev.detach_kernel_driver(1)
+        if dev.is_kernel_driver_active(2):
+            dev.detach_kernel_driver(2)
+
+        #dev.set_configuration()
```
tracernz, line 533: ??

python/dronin/telemetry.py
```diff
@@ -479,6 +479,137 @@ def _do_io(self, finish_time):
...
+            if f.bEndpointAddress < 0x80:
+                self.ep_out = f
+            else:
+                self.ep_in = f
+
+        if self.ep_out is None:
+            raise RuntimeError("no output endpoint found")
+
+        if self.ep_in is None:
+            raise RuntimeError("no output endpoint found")
```
tracernz: input

tracernz (commented):
Great stuff! :grin:

python/dronin/telemetry.py
mlyle: So, we're supposed to tell the device to take on a configuration.  Our stuff only supports one.

If we do tell it to change configuration, the OS often re-enumerates it, and then seizes it, which is annoying.

tracernz (approved):

PR #1494, opened 2016-12-21
[WIP] Add N-order IIR gyro filter algorithm.

ekaszubski:
- Implements an N-order (up to 3rd for now for memory constraints) IIR filter for gyro data
- Uses numerator/denominator coefficients
- Uses transposed direct form 2 based on [scipy.lfilter](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.signal.lfilter.html#scipy.signal.lfilter)
- Default coefficients are for a 3rd-order lowpass filter sampling at 1000 Hz with cutoff 35 Hz, passband ripple 1 db, and stopband attenuation 15 db
- Default a[0] coefficient of 0.0 disables the filter and uses the existing first-order IIR lowpass filter configured via GyroCutoff


ekaszubski:
Swash movement looks reasonable, will try a hover test soon.

mluessi:
One thing to be vary of is that quantization of the filter coefficients combined with a direct implementation of the IIR filter can lead to stability problems. It's usually better to implement the filter as a number of biquad sections (2nd order IIR) combined with a first order filter (if the order is odd).

ekaszubski:
@mluessi Yep, that's another reason I stopped at 3rd-order for now. I have 500Hz gyro data from the air and the bench and a 3rd-order lowpass appears to perform well on them offline.

When combining the filter sections, do you just feed the output signal of the first filter into the input of the second and so on, so each filter has its own input/output history? How are the coefficients calculated for serially combined filters? (e.g. a biquad lowpass has 3+3 b/a coefficients and a 4-th order has 5+5 so it seems that if the 4th-order is split into two biquads, 6+6 coefficients are needed when only 5+5 are available)

mluessi:
I believe you can only implement the filters as separate sections if the numerator and denominator polynomials can be factorized. This is from the top of my head, so I may be wrong ;)

mluessi:
Oh, I guess every polynomial can be factorized into quadratic factors, so you can implement any IIR filter like that.

mlyle:
Superseded #1595 --- thx for the initial work here

PR #1495, opened 2016-12-21
ws2811 LEDS: show armed status

mlyle:
Configurable, but default behavior is to fade through color range at .5Hz, and to blink super fast when arming is in progress.

Tested on seppuku.


mlyle:
Scope creep-- needed to change python API to handle parent-in-same-object.

tracernz (approved):

PR #1496, opened 2016-12-28
OSD Menu: Add RGB LED color settings

mluessi:
Add ability to change the colors using the OSD menu. More detailed settings UI should be added to the GCS.

Status: bench tested on RE1


mlyle:
Thanks much :D

PR #1497, opened 2016-12-28
bmp280: improve performance

mlyle:
* make polling intervals more conservative
* halve number of txns by putting sensor into free running mode.
* the previous temperature logic was odd ; it always retrieved it but only processed it half the time.  Now we always process it, and always lowpass filter it before using it in temperature conversion (expected to change more slowly than pressure).
* preserve more precision for temperature; we've got the bits, so use 'em.

### DIFF
--- a/flight/PiOS/Common/pios_bmp280.c
+++ b/flight/PiOS/Common/pios_bmp280.c
@@ -106,7 +106,6 @@ struct bmp280_dev {
 	int16_t  digP9;
 
 	uint8_t oversampling;
-	uint32_t temperature_interleaving;
 	enum pios_bmp280_dev_magic magic;
 
 	struct pios_semaphore *busy;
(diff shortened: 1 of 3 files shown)


flight/PiOS/Common/pios_bmp280.c
```diff
@@ -266,7 +263,7 @@ static int32_t PIOS_BMP280_StartADC(void)
 		return -1;
 
 	/* Start the conversion */
-	return(PIOS_BMP280_WriteCommand(BMP280_CTRL_MEAS, dev->oversampling));
+	return(PIOS_BMP280_WriteCommand(BMP280_CTRL_MEAS, dev->oversampling | 3));
```
tracernz: What do the two low bits do? Is it possible to use a constant/define?

tracernz (approved):
Looks good. 👍

flight/PiOS/Common/pios_bmp280.c
mlyle: Thanks-- meant to do this.

mlyle:
jenkins, test this please

PR #1500, opened 2017-01-02
python: autotune: allow specifying cutoff freq

mlyle:
Used for analysis of @ekaszubski 's helicopter as I needed to adjust for lower cutoff frequency.


tracernz (approved):
Thanks.

PR #1503, opened 2017-01-05
gcs: osd: teach "page" layout to expand horiz

mlyle:
On some machines, widgets get hidden in awkward scrolliness even when
you have the horizontal width for them.  This fixes that and gets the
osd page widget to consume whatever horiz width you have available.


tracernz (approved):
Looks great, thanks.

PR #1506, opened 2017-01-10
Simplify connection polling

mlyle:
* Simplify serial polling, which was the last model that was using a separate thread.
* While we're at it, greatly reduce overhead and eliminate synchronization problems (static variable touched by multiple threads) in the serialplugin.
* Don't defer through queue the telemetrymanager stuff, as this could be a frequent cause of Windows crashes.
* Remove a lot of code :dagger:

Requires review and test.  Tested on Linux so far-- seems very snappy :D


mlyle:

tracernz:
Works okay on Linux. Let's take the risk.

PR #1508, opened 2017-01-13
txpid: allow changing camera tilt setting

mlyle:


mlyle:

tracernz (approved):

chidibitoh:
I haven't flight tested this yet, but I did do some ground testing on a CC3D and everything worked as intended.

Video here:
https://www.youtube.com/watch?v=fDqZILFYxYY

I was kind of sad when I learned that Levelling doesn't work with the Camera Reprojection stuff.  I guess that would require changing the "virtual attitude relative to board" settings on the fly, in addition to the yaw/roll control reprojection that's already being done.  I can think of how to do it if the board's offsets are zero (just modify the "pitch" setting according to the camera angle), but when the board's mounted at anything but the perfect angle, it requires math that hurts my head...

Anyway, thanks for the awesome FC firmware!  And thanks for adding this feature!

PR #1513, opened 2017-01-13
GCS:Get some extra symbols for breakpad

tracernz:
We missed our own libraries before. This will also get Qt pdbs when available (if only they shipped them for release binaries :/).


tracernz:

tracernz:
Derp derp, operator precedence strikes back. Force-pushed with the or operator in a separate statement.. and tested on Linux this time.

tracernz:

PR #1515, opened 2017-01-14
PIOS_DAC: Implement DAC driver

mlyle:
This is WIP because it needs actual end-user useful functionality to merge.

May want to crank sample rate up too, because stair steps are fairly bad without an anti-aliasing filter.

### DIFF
--- a/flight/PiOS/STM32F4xx/inc/pios_dac.h
+++ b/flight/PiOS/STM32F4xx/inc/pios_dac.h
@@ -86,4 +86,4 @@ typedef bool (*fill_dma_cb)(void *ctx, uint16_t *buf, int len);
 bool PIOS_DAC_install_callback(dac_dev_t dev, uint8_t priority, fill_dma_cb cb,
 		void *ctx);
 
-#endif /* LIB_MAX7456_MAX7456_H_ */
+#endif /* _PIOS_DAC_H */
--- a/flight/PiOS/STM32F4xx/inc/pios_dac.h
+++ b/flight/PiOS/STM32F4xx/inc/pios_dac.h
@@ -1,0 +1,76 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS_DAC DAC Functions
+ *
+ * @file       pios_dac.h
+ * @author     dRonin, http://dronin.org Copyright (C) 2017
+ * @see        The GNU Public License (GPL) Version 3
+ *
+ *****************************************************************************/
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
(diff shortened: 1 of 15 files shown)


mlyle:
Need a decent way to turn this on and off, and I think the pios_dac priority system needs a little help.

With #1525 plus this, I get:

image

mlyle:
With #1525, this works well ---- though configuration is necessary.  An annunciator target would be really nice.

An example from seppuku is at http://jar.lyle.org/~mlyle/data.wav

A crude demodulator is at: https://gist.github.com/mlyle/4e0db1cab991cdceb364dc2859c3099e

Which yields output like:

```
595 ___m8}____=____]m_Y____c_6_'__h_)__9_G_a_~n_____o________NN
687 $TA_______
778 $TA_______
944 $TG_______________
1035 $TA____B_A
1127 $TA____B_A
1228 $TS______NN
```

(Nonprintables become underscores)

mlyle:
@peabody124 can you easily feed that data.wav to your tools to see whether we're interoperable?

mlyle:
envelope

http://jar.lyle.org/~mlyle/beepy.wav

I've tested both modes and selecting them from within the uavo browser.  works; I think this is ready for review.

peabody124:
@mlyle I will try and check Sunday. On call today. Ping me on IRC then too if you don't hear anything

mlyle:
@peabody124 Thanks, much appreciated :D

mlyle:
Good news: it works, even before the minor magnitude adjustment above:

http://streamable.com/i3rh4

Bad news: I shorted my VTX while playing with the disassembled quad and it caught fire.  Woops.  Need to rewire stuff now.

flight/PiOS/STM32F4xx/inc/pios_dac.h
```diff
@@ -0,0 +1,89 @@
...
+
+#ifndef _PIOS_DAC_H
+#define _PIOS_DAC_H
+
+#include <stdint.h>
+#include <stdio.h>
+
+#include "stm32f4xx_dac.h"
+
+typedef struct dac_dev_s *dac_dev_t;
```
tracernz, line 40: Prefix with `pios_`?

flight/PiOS/inc/pios_annuncdac.h
```diff
@@ -0,0 +1,57 @@
...
+ * must be maintained in each individual source file that is a derivative work
+ * of this source file; otherwise redistribution is prohibited.
+ */
+
+#ifndef _PIOS_ANNUNCDAC_H
+#define _PIOS_ANNUNCDAC_H
+
+#include <pios_dac.h>
+
+typedef struct annuncdac_dev_s *annuncdac_dev_t;
```
tracernz, line 39: `pios_`

flight/PiOS/inc/pios_fskdac.h
```diff
@@ -0,0 +1,51 @@
...
+ * must be maintained in each individual source file that is a derivative work
+ * of this source file; otherwise redistribution is prohibited.
+ */
+
+#ifndef _PIOS_FSKDAC_H
+#define _PIOS_FSKDAC_H
+
+#include <pios_dac.h>
+
+typedef struct fskdac_dev_s *fskdac_dev_t;
```
tracernz, line 39: `pios_`

tracernz (changes requested):
Looks pretty good. :+1:

flight/PiOS/STM32F4xx/inc/pios_dac.h
mlyle: Last several drivers we merged like ws2811, max7456, lis3mdl we had decided not to do that and to move towards changing the convention.   Want to revisit?

tracernz: Ah yes. I still think this kind of artificial namespacing is good but the horse has already bolted.

PR #1516, opened 2017-01-14
gcs: output: improve output calibration

mlyle:
Offset the minimum of the calibration range up a little bit from what we
send disarmed.

### DIFF
--- a/ground/gcs/src/plugins/config/configoutputwidget.cpp
+++ b/ground/gcs/src/plugins/config/configoutputwidget.cpp
@@ -413,17 +413,12 @@ void ConfigOutputWidget::startESCCalibration()
             if (!channelsToCalibrate[i])
                 continue;
 
-            if (actuatorSettingsData.ChannelMax[i] >
-                    (actuatorSettingsData.ChannelMin[i] + 40)) {
-                // If it is reasonable to do it, calibrate the minimum of the range
-                // to be a little off what we send at idle.  Blheli takes the value
-                // we give verbatim as the minimum, and small amounts of jitter
-                // can cause the motor to tick over.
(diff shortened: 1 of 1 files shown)


tracernz:
Won't work for Oneshot42 or that other silly short pulse one. :(

ground/gcs/src/plugins/config/configoutputwidget.cpp
```diff
@@ -413,7 +413,17 @@ void ConfigOutputWidget::startESCCalibration()
...
+                    (actuatorSettingsData.ChannelMin[i] + 40)) {
+                // If it is reasonable to do it, calibrate the minimum of the range
+                // to be a little off what we send at idle.  Blheli takes the value
+                // we give verbatim as the minimum, and small amounts of jitter
+                // can cause the motor to tick over.
+                actuatorCommandFields.Channel[i] =
+                    actuatorSettingsData.ChannelMin[i] + 3;
+            } else {
+                actuatorCommandFields.Channel[i] = actuatorSettingsData.ChannelMin[i];
+            }
```
tracernz: What about this? Handles the case where min and max are inverted, and any pulse length. I chose / 40 to get close to the 3 us you were presumably intending to add to Oneshot125. Limit to max adjustment of 5 so oldschool PWM doesn't go crazy.
https://ideone.com/2m38QK

Really it would be better to fix this in BLHeli since it's their bug...

tracernz:
@mlyle Happy with this now? I haven't actually run this code through a calibration.

mlyle:
Yes, let's do it-- full speed ahead and fuck the torpedoes.

PR #1517, opened 2017-01-14
gcs: input wizard: improve arming switch sense

### LINKED ISSUE #519: GCS input wizard should clarify arming switch sense
In the "Radio WIzard" "invert axis/switch" input screen, one has the option to invert the arm switch.  Unfortunately there is no way to see which setting results in "armed"...  Would be nice to know if you need to invert it or not by showing the status.

mlyle:
1. Don't dim the flight mode switches when we're asking user what
controls to invert.
2. Put a subtle red fringe on the outside of the arming switch in the
armed position.

Fixes #519


tracernz:
Good idea. :+1:

mlyle:
image

PR #1519, opened 2017-01-18
rawhid: handle synchronization and destruction better

mlyle:
1. Write thread is in charge of usb handle cleanup and cleaning itself
2. Main rawhid object cleans everything else before instructing write
3. Unnecessary synchronization and communications between levels
removed.

### DIFF
--- a/ground/gcs/src/plugins/rawhid/rawhid.cpp
+++ b/ground/gcs/src/plugins/rawhid/rawhid.cpp
@@ -35,7 +35,6 @@
 #include <QtGlobal>
 #include <QList>
 #include <QMutexLocker>
-#include <QWaitCondition>
 
 class IConnection;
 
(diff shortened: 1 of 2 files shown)


mlyle:
Works on Quanton and Linux.

ground/gcs/src/plugins/rawhid/rawhid.cpp
```diff
@@ -225,13 +145,10 @@ void RawHIDWriteThread::run()
         {
             QMutexLocker lock(&m_writeBufMtx);
             size = qMin(WRITE_SIZE-2, m_writeBuffer.size());
-            while(size <= 0)
+
+            while(m_running && (size <= 0))
```
tracernz, line 149: Would be nice to fix the style while you're modifying the line. :grin:  I'm ok either way though.

ground/gcs/src/plugins/rawhid/rawhid.cpp
```diff
@@ -351,28 +278,29 @@ bool RawHID::open(OpenMode mode)
...
     m_readThread = NULL;
 
-    hid_close(m_handle);
+    m_writeThread->stop();
     m_handle = NULL;
 
+    m_writeThread = NULL;       /* Write thread is in charge of tearing itself and
+                                 * m_handle down */
+
     QIODevice::close();
```
tracernz, line 304: Shouldn't this happen first before we kill the layer underneath? (I know, not your bug..)

tracernz:

(my windows build env is mingw atm :P)

tracernz (approved):
Looks good. Seems to work great on Linux and no worse than usual on mac (I saw some weird stuff when doing weird things but that's not unexpected with the current state of hidapi/rawhid). Stilling waiting on artifacts to check Windows.

ground/gcs/src/plugins/rawhid/rawhid.cpp
mlyle: You mean calling the superclass's close?  This is always tricky for me as different people follow different conventions.  Usually I prefer to "construct" or "setup" from parents up, and tear down in the opposite order.

However, when I look at it, the aboutToClose signal makes this kinda unsafe.  Meh.

tracernz: Yeah, so having `QIODevice::close` at the end can result in others trying to write after we've torn down the underlying USB stuff, especially since it emits `aboutToClose`.

tracernz:
Seems to work fine in basic Windows testing. :+1:

PR #1522, opened 2017-01-21
pios_servo: support dshot

### LINKED ISSUE #1464: Flight: DShot support
* On flight controller (if applicable): brainre1 ,BrainFPV, others(?).

Issue details:

Is there anyone here that could implement dshot 150 300 600 to dronin ?

mlyle:
fixes #1464

* add new calls to pios_tim to reinit outputs and to change timer outs
to general purpose output
* add new "inline delay" infrastructure for high resolution
timing/bitbanging requirements
* abuse the actuator rates 65532, 65533, 65534Hz for dshot 300/600/1200
* teach pios_servo how to generate a dshot signal, and factor some
things slightly more cleanly

Todo:

* [x] GCS support
* [x] Verify autotune can work on CC3D with dshot
* [x] Test on dshot ESC


flight/PiOS/inc/pios_servo.h
```diff
@@ -33,8 +33,18 @@
 
 #define PIOS_SERVO_MAX_BANKS 6
 
+/* Used in out_rate.  If 65535, we really mean 65535Hz.  Values close to that
+ * are reserved/abused for dshot support.  0 is used for sync pwm per legacy
+ */
+enum pios_servo_shot_type {
+	SHOT_ONESHOT = 0,
```
tracernz, line 40: This is called SyncPWM (from SimonK's write-up of the protocol) throughout the rest of our flight/GCS code and the UI, might be nice to follow that. :smiley:

tracernz:
I made a Saleae Logic analyzer plugin. :grin: 
image

mluessi:
Won't disabling interrupts for 26us, or even 104us (DSHOT150) cause problems with e.g. the OSD code where timing is very critical? On targets that can support it, it may be better to use DMA, especially if the DMA channels are not used otherwise.

flight/PiOS/inc/pios_servo.h
mlyle: Yes, but shouldn't we abandon it?  It's a support issue for no reason to tell everyone to use "synchronous PWM" when they want "oneshot".

mlyle:
@mluessi   Yes, OSD is the main problem.  Even 115200 UART serial is no problem, and the CPU use is modest.  Note we don't support DSHOT150 in the above code.

You're welcome to contribute a DMA implementation.  Unfortunately, a DMA approach to timers requires a lot of analysis of each target and verification, or has to be something like the '2811 code where there are bugs with DMA2.

Or you could free up TIM4 (the use of a timer channel for line counting is unnecessary) on OSD targets, and then I could use DMA1 to directly blit to BSRR's.

(TIM7 is also possible but would use up a **lot** of resources)

mlyle:
@tracernz holy shit that's awesome :D

mlyle:

mlyle:
Status: previous code tested by @glowtape .

Had to change timings to make it work with blheli_s.  Works on Sparky2 with blheli_s at 300/600, on CC3D at 300 only (hits a timeout/rearm loop on 600 from not updating frequently enough).

mlyle:

mlyle:
Also, @tracernz --- @glowtape pointed out the minimum legal neutral value is 48.

glowtape:
Yeah, a while ago they've decided that 1-47 is reserved for commands. That's what Felix lists in the original post over on RCG:

> request area is 1-47 
> the kiss24A uses 1-13 from it
> 1-5: beep (1= low freq. 5 = high freq.) 
> 6: ESC info request (FW Version and SN sent over the tlm wire)
> 7: rotate in one direction
> 8: rotate in the other direction
> 9: 3d mode off
> 10: 3d mode on (3d mode dont workes with Dshot ATM)
> 11: ESC settings request (saved settings over the TLM wire) (planed but not there yet)
> 12: save Settings

https://www.rcgroups.com/forums/showthread.php?2756129-Dshot-testing-a-new-digital-parallel-ESC-throttle-signal

mlyle:
My favorite part of this protocol is that 1111 1111 1111 1111 is a legal full throttle command :p

tracernz:
@glowtape @mlyle Would be good if they could, you know, document the protocol somewhere in a readable/concise fashion rather than having various details spread throughout different sources. :/

tracernz:
So uh... is 0 stopped, or 48? That would be dumb but wouldn't surprise me... Or is 48 the minimum pulse, and neutral is a bit higher than that? (i.e. the UI should limit min/neutral/max to [48, 2047]). I hope the latter.

mlyle:
0 is stopped.  48 is minimum active pulse but probably won't start motor.

Min should be nailed to 0, neutral should be [48, ... i dunno, 2047 i guess :P] and max should be nailed to 2047.

(Later we may want a variant for the 3D mode, which is more complicated)

tracernz:
Ah, that's dumb/requires a whole lot more special-casing. 🥇 to whoever thought it was a good idea.

tracernz:
Not really sure what to do with the min/max boxes... since they can be reversed I can't just disable one of them.

tracernz:
Just pushed what I think are minimal GCS changes. Haven't spent enough effort yet on verifying safety but quick checks seem okay.

glowtape:
I wouldn't sweat it over things like 3D mode. Given there's no return communication channel, you can't actually be sure the ESC ate the command. KISS listed 3D mode separately from motor reversal. I figure the intent is to send a motor direction command a thousand times and then assume it took and never touch it again.

mlyle:
prereq work for fancy DMA is in #1523 ...

Basic idea would be to count 84MHz, generating DMA req every 7 cycles.  Then a dshot 1200 signal would have 10 BSRR writes per bit

Cycle 0-- turn on signal
Cycle 3-- turn off signal for 0
Cycle 7-- turn off signal for 1

Update one GPIO bank at a time, using interrupt to compute next buffer and swap it over for next bank.

Results in 250ns / 583ns / 833ns timings vs  312/625/833 for KISS and 250/541/833 for betaflight.

DMA req is super tricky -- 3.5 cycles on the timer's 42MHz true clock.  Achievable but hopefully ST didn't mess it up.

mlyle:
Deferring fancy DMA version.  Have spun ESCs a bunch.  Did you fly it @glowtape?

glowtape:
Flew one battery of your second revision of non-DMA Dshot600 (the BLHeli timings), then crashed into snow. Was pilot error, tho. The quad flew fine those few minutes without any obvious twitching or anything. The OSD in the RE1 worked too.

tracernz:
Holy crap, just resolved the conflict using https://github.com/blog/2293-resolve-simple-merge-conflicts-on-github :+1:

PR #1525, opened 2017-01-24
light telemetry improvements

mlyle:
* pick up changes from TL
* slightly increase buffer size
* Simplify scheduler / make it easy to add future frame types
* Adjust scheduler to try to fill link better and be fair between
different frame types
* Control selection of G frame type better-- don't initialize objects
that don't belong to us, and use the initialization to select which of
the sources (positionactual, baro, GPS) to use.
* Remove some redundant checks (module can only run if we have a port)

### DIFF
--- a/flight/Modules/UAVOLighttelemetryBridge/UAVOLighttelemetryBridge.c
+++ b/flight/Modules/UAVOLighttelemetryBridge/UAVOLighttelemetryBridge.c
@@ -67,7 +67,7 @@
 
 #if defined(PIOS_INCLUDE_LIGHTTELEMETRY)
 // Private constants
-#define STACK_SIZE_BYTES 512
+#define STACK_SIZE_BYTES 600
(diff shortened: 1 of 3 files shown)


tracernz (approved):
Looks good. Will KipK's stuff be ok with potentially higher rate?

mlyle:
@tracernz I'm not sure.  It's potentially 20% faster.

@kipk -- do you see any problems with this?  I'm working on lighttelemetry for other reasons, but it'd be nice to keep compatibility.

tracernz:
Not a blocker for us imo, but would be nice to keep compatibility if possible.

mlyle:
This half is good to merge, I think.

KipK:
Hi.

I'm out of the community for monthes, so sorry for the lack of answers.


No problem to make evolve the protocols to your needs, LTM was only a kind of draft to my needs only :)

Best

tracernz:
Thanks Kip, appreciate your input! 👍

PR #1529, opened 2017-01-27
onscreendisplay: run video/2811 hack at startup

### LINKED ISSUE #1526: Altitude handling problems
Related: #844, #443 

In #864, @dustin reported that the baro altitude and rates do crummy things on the OSD.

It looks like cfvert does not adjust PositionActual.Down to equal 0 at arming.  This affects displayed OSD values in the OSD altitude nav mode (which IMO should be preferred).

May relate to @litterbug's comment at https://github.com/d-ronin/dRonin/pull/1480#issuecomment-275268641

mlyle:
Otherwise we can get a bad baro point, which can throw off the
baro zero point.

Also, ensure to be accumulating points during startup even if there's
no video.

Fixes #1526


tracernz (approved):
This looks sane/safe.

mlyle:
This is good-- have gotten no repro of altitude problem with this change.

PR #1534, opened 2017-01-30
gcs: autotune: used queued conn to check for flag

### LINKED ISSUE #1533: Autotune wizard auto-pop-up freezes telem session
Doesn't appear to happen with manual entry of "adjust previous autotune"

Recovers after leaving dialog box.   Wtf.

mlyle:
Otherwise, we can start doing modal stuff while processing connected
state with bad results.

Fixes #1533 

So far this is the only regression I've spotted in GCS on OSX.


mlyle:
Confirmed same problem on Linux, confirmed that this fixes it.

PR #1536, opened 2017-01-31
linux/udev: allow raw USB access in addition to HID

mlyle:
Just for 2 main ids (not existing legacy ones).  Main use right now is
Seppuku mfg, but may broaden (allows using python API without root).


tracernz:
This should make lsusb read full descriptors without root privileges again. 👍

tracernz:
jenkins, test this please

(didn't pick it up..)

PR #1538, opened 2017-01-31
python-api: handle disconnection better

mlyle:
Properly get EOF on many drivers.

Will be one more PR in this area-- need to add a "close" mechanism to clean up resources as appropriate.

The sum effect of these changes is that python stuff gets "stuck" less and it will be ultimately possible to connect to a FC several times in a session.

### DIFF
--- a/python/dronin-halt
+++ b/python/dronin-halt
@@ -1,34 +1,33 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+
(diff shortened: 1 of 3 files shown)


mlyle:
Added the close mechanism.  Turns out USB/HID doesn't need it.  We're good here.

tracernz (approved):
LGTM

PR #1541, opened 2017-02-01
Tools: Check QBS config each time it's used

tracernz:
Hopefully resolves any issues encountered. Tested with empty config, and config with desired profile removed on Linux, macOS, and Windows.


mlyle:
Ah, very good idea.

Should #1539 close, given that current things are OK?

Issue #863, opened 2016-04-01
Attitude estimates often poor in fast-forward-flight or in fixed wing


mlyle:
Shows up really blatantly on OSD, etc.

taulabs/taulabs#2021 etc

mlyle:
This is the real issue with multirotors:

![image](https://cloud.githubusercontent.com/assets/90903/22564981/6aba6cf2-e93b-11e6-9316-3385ca265277.png)

(From a simulation run, where attitude was altered to gate accelerometer data for 300ms after any fast gyro moves).  The paradoxical movement due to initial acceleration on the x/y accels biases the attitude estimate before steady state is reached.  The 300ms gate improves things a lot, but A) doesn't solve it entirely, and B) probably wouldn't play nice with autotune.

dustin:
I've been struggling with this on fixed wing lately.  Roll is OK.  I'm able to fly fine with roll in auto-leveling.  Pitch doesn't work well at all.  With default accel gains, I get something that looks like this (ST1 is leveling on roll, rate on pitch):

![image](https://user-images.githubusercontent.com/1779/27343719-9ee78c98-5598-11e7-9650-e847c115d333.png)

Which leads to a position hold attempt that looks like this:

![image](https://user-images.githubusercontent.com/1779/27343735-aefd9230-5598-11e7-9d14-b67571269689.png)

Motor's up, nose down -- which it appears to believe is the correct thing.

I reduced accelkp down to 10% its original value and it seemed mostly just kind of messier:

![image](https://user-images.githubusercontent.com/1779/27343869-2b1f3404-5599-11e7-8820-c769ef4fc1df.png)

I can't say for certain that it was net worse, but it didn't seem better.

PR #1543, opened 2017-02-01
Attitude improvements

mlyle:
Connects to #863.

* Calculate in a way that behaves the same no matter the sampling rate.
* Similarly, filter acceleration data in a sample-rate invariant way
* Adjust the default values to be a "compromise value" that behaves
something between how F1 and F3 do it currently.
* Rename accelkp/accelki to acckp/accki, and magkp/magki to mgkp/mgki
because the values have changed.

This is not sufficient to be without anomaly, but a good starting point of discussion.  Unclear if this should make it into Artifice.

### DIFF
--- a/flight/Modules/Attitude/acro/attitude.c
+++ b/flight/Modules/Attitude/acro/attitude.c
@@ -196,8 +196,8 @@ static void AttitudeTask(void *parameters)
 			(PIOS_Thread_Systime() > 1000)) {
 
 			// For first 7 seconds use accels to get gyro bias
-			accelKp = 0.1f + 0.1f * (PIOS_Thread_Systime() < 4000);
-			accelKi = 0.1;
+			accelKp = 20 + 20 * (PIOS_Thread_Systime() < 4000);
+			accelKi = 1;
 			accel_filter_enabled = false;
 
 		} else if (zero_during_arming && 
@@ -206,12 +206,12 @@ static void AttitudeTask(void *parameters)
 			// Use a rapidly decrease accelKp to force the attitude to snap back
 			// to level and then converge more smoothly
 			if (arming_count < 20)
-				accelKp = 1.0f;
-			else if (accelKp > 0.1f)
-				accelKp -= 0.01f;
+				accelKp = 50;
+			else if (accelKp > 20)
+				accelKp -= 1;
 			arming_count++;
 
-			accelKi = 0.1f;
+			accelKi = 1;
 			accel_filter_enabled = false;
 
 			// Indicate arming so that after arming it reloads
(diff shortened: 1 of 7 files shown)


mlyle:
So basically, this seems to do slightly better in tests of "hold by usb cord and flop around maniacally, then set down flat on desk and watch attitude".

ufoDziner:
Jenkins build artifacts please.

mlyle:

mlyle:

ufoDziner:
Not sure what happened, but my seppuku will not save any settings.  First artifacts appeared to work, but I didn't review them before the second set was built.  Installed second set, upgraded, and now settings won't save.  It appears that they import, but just won't save.  Tested with the first build again, and it won't save them either.  Suggestions?

mlyle:
Are you using the upgrade routine in GCS or upgrading by hand and importing?

It's possible your settings flash is full of obsolete UAVOs-- the upgrader erases the settings partition to deal with this.

Try right-clicking on settings and picking 'erase'.

mlyle:
The settings partition in the browser on the firmware pane, that is.

ufoDziner:
I ran the GCS update.  My settings were definitely old.

I entered the bootloader, right clicked on settings then "erase".  Even after that, I cannot save any changes.  Just trying to add the receiver protocol I get the red X on the save button after a save attempt, and then my setting is gone upon reboot.  I do have an unusual error for me, "There were problems with the UAVObject events or callbacks".

mlyle:

ufoDziner:
Flight test footage.  https://youtu.be/_V9RSP92Wv0

glowtape:
I flew around a bit in my backyard.

Before when doing FFF and "fast curved flight" (FCF :D ), the AH would slide off screen and then took a few seconds to get back when coasting. Going into FFF and make it go offscreen, then enter a fast curve or some other manoeuver often ended up in the AH suddenly coming in from the side at a large angle to the real horizon, just to slide offscreen again, when you change direction, in a way unrelated to the actual flying. Essentially, it was pretty useless in acro flight.

Now, the artificial horizon doesn't veer off much from the position you'd actually expect it to be. In FFF or FCF, it picks up the expected bias, but once you start coasting, the artificial horizon line snaps back to the real one pretty fast. Queueing a bunch of different manoeuvers other than FFF/FCF doesn't upset it much either (before the AH went AWOL).

I had the camera angle set up and by chance the AH movement seemed to fit the FOV of my DALRC 2.5mm well enough (I picked up in IRC yesterday that FOV could be calibrated for, but didn't really spot anything regarding it in the OSD code, tho).

This fix should help newbies learning acro, too. Back when I did it, attitude mode was a failsafe to rescue the quad. Obviously that didn't always go well, when the computed attitude was completely off.

mlyle:
@glowtape ArtificialHorizonMaxPitch should be set to half the VFOV of the camera.  VFOV is 60% of DFOV in 3:4 video, so 30% of the diagonal FOV is about correct -- so the default config of 30 is about right for a 100 degree fov camera.

The way I do it is experimentally.  I point the quad at a distant horizon-- perhaps my desk 20 feet away.  then i pitch up and down until the horizon and the desk move up and down in the video at very close to the same rate.

PR #1546, opened 2017-02-03
Flight: Work around arm errata 752770

mlyle:
chcore_v7m is broken.  uavobjectmanager's task switch isn't with current code generation, but could become so in the future without this change.

> An interrupt occurring during the data-phase of a single word load to the stack pointer
(SP/R13) can cause an erroneous behavior of the device. In addition, returning from the
interrupt results in the load instruction being executed an additional time.

> The instructions affected by this limitation are the following:

...

> • LDR SP, [Rn]
> • LDR SP, [Rn,#imm]


mlyle:
Force pushed to change commit msg.

tracernz:
Isn't that lovely.

mlyle:
just searched the emitted code for any other instances of this and there's not any.

mlyle:
Did moderately extensive test of this on seppuku and it's OK.

PR #1547, opened 2017-02-03
osd: prevent memory corruption

mlyle:
The previous fix resulted in undesired memset of low memory, because
PIOS_VIDEO was not fully initialized.

This is a partial revert and should be safe for speedy merge.  Also thoroughly tested on seppuku.


mlyle:
FFS.  Force-pushed commit message because my brain's autocorrect is lame.

PR #1548, opened 2017-02-04
GCS: Basic RGB LED UI

tracernz:
Very basic, no frills.
image

First commit (UAVO field limits derived from datatype) may want additional test as it affects other parts of GCS. It will only be harmful if limits are set for UAVO relation widgets in the widget properties rather than the UAVO relation (which we should not be doing anyway) AND the UAVO relation specifies `haslimits:yes`.


tracernz:

mlyle:
zomg i like :D

tracernz:
Picker (default QColorDialog) on Linux:
image

Custom color:
image

mlyle:
Just as wonderful as it looks here in the comments :D

PR #1549, opened 2017-02-05
Vehicle setup pane: properly keep configuration

### LINKED ISSUE #1220: Vehicle setup pane is problematic
- Seen in version (release version or githash): 20160720.1
- On flight controller (if applicable): Sparky2

Issue details:
  Vehicle type switches back to Multirotor on the Vehicle tab after saving as Fixed wing.  Further investigation shows that only Multirotor & Custom will save, Fixed wing and Helicopter default back to either Multirotor or Custom (whichever was saved last), and Ground sends the FC into a bootloop.

Edit by MPL: Connects #315

mlyle:
Fixes #1220 

This code is a huge mess and needs a rewrite.

I've banged on this quite a bit but could use additional test feedback due to criticality and how late we are-- both the vehicle setup pane and the setup wizard.


ground/gcs/src/plugins/uavobjects/uavobjectfield.cpp
```diff
@@ -616,7 +616,11 @@ QString UAVObjectField::toString()
     sout.append ( QString("%1: [ ").arg(name) );
     for (unsigned int n = 0; n < numElements; ++n)
     {
-        sout.append( QString("%1 ").arg(getDouble(n)) );
+        if (type == ENUM) {
+            sout.append( QString("%1 ").arg(getValue(n).toString()) );
+        } else {
+            sout.append( QString("%1 ").arg(getDouble(n)) );
+        }
```
tracernz, line 668: Can't you just do QVariant::toString for everything? http://doc.qt.io/qt-5/qvariant.html#toString

mlyle: I could.  What's here is still clearly broken, e.g. showing UAVO ids in scientific notation.  But I didn't want to fiddle with it too much-- this solved the outright lying that was happening before when "QuadX" became 0

tracernz: Yeah, I don't think we do the right thing with enums for QVariant to be able to convert them to int. :(

mlyle:
Rebased

mlyle:

glowtape:
Tried changing vehicle type tons of times on an OG Brain upgraded to this build. Didn't need to double save for them to stick.

tracernz (approved):

PR #1551, opened 2017-02-05
Package: Hack to work around upstream QTBUG-57265

tracernz:
Fixes bug reported by @jhitesma where QML fails to work at all because debug symbols are copied in place of the real dylibs


tracernz:

mlyle:
Beautiful :D

jhitesma:
Doh, already merged but can confirm it fixed the issue I was having :D

tracernz:
Advice is that `-always-overwrite` triggers this bug, I wonder if we actually need it?

PR #1552, opened 2017-02-05
GCS:Seppuku: Add Rx Jumper to Connection Diagram, Make Config Background Transparent

### LINKED ISSUE #1512: GCS:Seppuku: Show Rx Voltage Jumpers on Connection Diagram

tracernz:
Transparency fix makes board image on board config page render on a transparent background on macOS+Windows (already did by default on Linux).

Also added RX voltage jumper reminder to connection diagram. Fixes #1512.


mlyle:
This looks really good.

Confused me for a second because it's only when you save the hwseppuku object does the conn diagram get the correct version.

PR #1554, opened 2017-02-05
osd: adjust camera pitch tilt by roll angle

mlyle:
Thanks to @glowtape for pointing this out.


mlyle:

glowtape:
Artificial horizon stays on the real one now, even if upside down.

When the quad starts pointing vertically, around +/-5° of the vertical axis, you can see the AH swing around smoothly. Probably will go unnoticed with the default pitch steps (which is 2, -20° to 20°), I'm running 5 steps.

Works!

PR #1556, opened 2017-02-06
targets: remove hardware files

mlyle:
The history will remain in this repo, but people who are cloning
with --depth in the future can get a much smaller checkout.  Canonical
place for the hardware files and relevent history is
https://github.com/d-ronin/dronin-hw


tracernz:
This will be nice for CI. Sadly the damage is done for most other purposes. 🙃

mlyle:
Maybe in a year dev-oriented-users following our dev guide will get a clone that won't include this stuff.

PR #336, opened 2015-12-23
build: introduce new script to display credits information

mlyle:
Based on dumb metric for sorting (number of commits, exponentially
decaying with 18mo half life).

This is work in progress awaiting comment.  When complete: fixes #271

[Review on Reviewable](https://reviewable.io/reviews/d-ronin/dronin/336)

PR #1557, opened 2017-02-06
GCS: Use git log for authors dialog

### LINKED ISSUE #271: GCS: Fix Authors Credit
In the GCS "Help->About Authors.." is still referring to the TL project and the list of authors seems to be outdated. Ideally, the authors list should be created automatically from git commits.

tracernz:
Not sure if this should make current cycle or not (@mlyle to add milestone). Needs some agreement on the approach (described in https://github.com/d-ronin/dRonin/pull/336#issuecomment-231619949). `git-shortlog` is not a plumbing command so really should fix this up to use a plumbing command instead (although plenty of porcelain commands are used in version-info.py already). mailmap could also do with some help.

credits

Fixes #271


tracernz:

tracernz:
Sorted overall by commits.

PR #1560, opened 2017-02-06
F4: clear flash error flags from FLASH->SR before starting new write operation

### LINKED ISSUE #1559: Saving settings in PlayuavOSD hangs the board.
* Seen in version (release version or githash): next
* On flight controller (if applicable): PlayuavOSD

Issue details:

Saving settings ends up in PIOS_Assert() after unsuccessful FLASH_ProgramByte() in https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/STM32F4xx/pios_flash_internal.c#L175

Upon further investigation, it is not the actual flash programming that fails, but rather FLASH_WaitForLastOperation() that finds PGPERR and PGSERR set in status register https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/STM32F4xx/Libraries/STM32F4xx_StdPeriph_Driver/src/stm32f4xx_flash.c#L880

```
(gdb) print /t *(FLASH_TypeDef *)0x40023C00
$14 = {ACR = 11100000101, KEYR = 0, OPTKEYR = 0, SR = 11000000, CR = 0, OPTCR = 1111111111111010101011101101, OPTCR1 = 10001000000000000000011000000}

```

The fix appears to be simple - just clearing error flags in PIOS_Flash_Internal_StartTransaction() helps with settings save function.

```
	/* Clear error flags before starting write ops */
	FLASH_ClearFlag(FLASH_FLAG_EOP | FLASH_FLAG_OPERR | FLASH_FLAG_WRPERR | FLASH_FLAG_PGAERR | FLASH_FLAG_PGPERR | FLASH_FLAG_PGSERR);
```

mindnever:
This PR fixes https://github.com/d-ronin/dRonin/issues/1559

### DIFF
--- a/flight/PiOS/STM32F4xx/pios_flash_internal.c
+++ b/flight/PiOS/STM32F4xx/pios_flash_internal.c
@@ -113,6 +113,10 @@ static int32_t PIOS_Flash_Internal_StartTransaction(uintptr_t chip_id)
 
 	/* Unlock the internal flash so we can write to it */
 	FLASH_Unlock();
+
+	/* Clear error flags before starting write ops */
+	FLASH_ClearFlag(FLASH_FLAG_EOP | FLASH_FLAG_OPERR | FLASH_FLAG_WRPERR | FLASH_FLAG_PGAERR | FLASH_FLAG_PGPERR | FLASH_FLAG_PGSERR);
+
 	return 0;
 }
 


mlyle:
jenkins, test this please

mlyle:
@mindnever --- thanks for the PR and trying out playuav.

Unfortunately, this is bad news.  Probably the reason for this is a bug somewhere else-- an attempt to write to a NULL pointer that in turn sets the flash error flag.  We'll need to look into this further to understand the reasons why.

mindnever:
Sure, bad news because some code earlier misbehaved and tried to write into flash. However, in my opinion I find this fix actually okay to live with, since it clears the flash state before attempting completely legal operation. I see no reason to fail to save the settings, just because some other task earlier did something wrong.
Consider the same thing happens on the actual flight controller, not just this benign OSD board - the bug is there, saving the settings is disabled, yet more critical stuff is still possible, like arming, spinning the props, etc.

mlyle:
@mindnever Yes, any memory corruption is dangerous.

This is probably a null pointer dereference in playuav-- it's probably something expecting something that exists in the flight environment but not just on the OSD.

I am a bit skeptical of this fix because 1) it doesn't fix the underlying problem, 2) potentially allows the corruption of flash memory, and 3) will conceal problems like this in the future.

I will look for the bug in PlayUAVOSD.

mlyle:
(Maybe we should have an interrupt on flash errors and crash immediately, instead of during another program operation, though)

mindnever:
I have tried enabling flash interrupt on error, even though there is missing #define for ERRIE CR flag (bit 25), but nothing. However, I cannot get this board reproduce the original problem anymore. Unfortunately I have no physical access to it right now, and cannot disconnect flight controller (to check). Luckily, I have one more board that had exactly the same problem and will check what happens with interrupts enabled there.

mlyle:
@mindnever it just occurred to me-- I had a null-dereference bug in OSD startup for awhile that could be the issue--- it was corrected by #1547 / 10157d32 and introduced by #1529 / d9a3d3be   -- were you possibly between these revisions on next when experiencing the problem?

mindnever:
@mlyle I have first found about the problem in Quixote, but then used fresh next for debugging (which had same problem). The #1547 was already in.

mindnever:
@mlyle 
update - this was indeed fixed by #1547 . Sorry for false alert. I was checking next, but this was in fact just the day before #1547 and forgot that I had weird SWD stlink vs jlink issue to sort out that took some days.

To add into the mix, I did give it a go at next even before experiencing problem, as I was wondering about LED behaviour (both blinking at same time), and that is probably the time I got bitten by this bug. It wasn't Quixote that was affected. 

Thanks.

mlyle:
@mindnever -- thanks for your help and the report.   I'm going to close the PR/issue because I think we want the current behavior with flash error flags.

PR #1567, opened 2017-02-09
gcs: module: change default tab to 0

mlyle:


tracernz:
We should probably just set this in C++ code right after setupUi so it never matters which tab you leave Qt Designer on.

Issue #1570, opened 2017-02-09
No modelviewer on windows


ufoDziner:
* Seen in version (release version or githash): Preview-20170207
* On flight controller (if applicable): Sparky2

Issue details:
No model displayed in model viewer on Windows 10.

![nomodel](https://cloud.githubusercontent.com/assets/7434287/22797435/0e4dcc3e-eeb3-11e6-914e-c579bb4016ce.png)

tracernz:
![image](https://cloud.githubusercontent.com/assets/9995998/22803121/8a3ee740-ef78-11e6-8d76-af23c14fd6b9.png)

tracernz:
Background setting was empty for some reason:
![image](https://cloud.githubusercontent.com/assets/9995998/22803182/cb9633ce-ef78-11e6-9492-505f8bdbf8b7.png)

Once set all was well:
![image](https://cloud.githubusercontent.com/assets/9995998/22803212/e3ec07b4-ef78-11e6-848a-4467ee933ebc.png)

Reset my config (`drgcs -r`), and it worked fine...

ufoDziner:
Here are my modelview settings.

![modelview](https://cloud.githubusercontent.com/assets/7434287/22804750/56c48a7c-eecf-11e6-9271-a8e544e571e8.png)

ufoDziner:
I tried this on my other windows computer and the model showed up correctly.  I tried the current release on the problem computer, and it's no longer working either.  It has/was working in the past so it appears to be an issue with my video and not the GCS.

mlyle:
I just want to reopen this ot see what @tracernz says -- I'm not sure why fully-qualified paths are showing up in this dialog (instead of relative ones within the installation like works).

pug398:
Not sure it is related but I see once you extract the working directory dronin-preview-20170207 you cant move it otherwise you break the gcs. Resetting config doesn't fix.

ufoDziner:
Here are my GCS settings.  Not sure if it's actually what you're looking for.
[ufo_gcs_settings.zip](https://github.com/d-ronin/dRonin/files/767740/ufo_gcs_settings.zip)

tracernz:
Downloaded the preview build (zip), placed on desktop, ran it, closed it, and moved it to another drive, ran it. Some of the QML assets still point to the old path. Some stuff did work though, e.g. github/recent tunes/links on the welcome pane but not the buttons, and PFD is all there except the background image. Modelview was fine.
```
[...]

PR #1573, opened 2017-02-10
GCS: Disable QML disk caching

tracernz:
Stops absolute paths being saved in cached QML (breaks welcome screen, PFD...). Fixes issues discussed in #1570 (but not that issue itself). Tested on Windows and confirmed .qmlc/.jsc files are no longer generated, and that I could move the GCS install.


tracernz:

PR #1583, opened 2017-02-16
Update CMSIS includes from 3.01 to 4.5.0 and remove redundant copies

mlyle:
Gets us a few more useful headers that I plan to use and also cleans up the tree a bit.


mlyle:
There's also CMSIS5 now but it's a little too different/risky for this moment IMO.

PR #1584, opened 2017-02-18
posix/logfs: fflush after writes

mlyle:
If we fully order write operations, things are safe and we won't ever
lose data.  Otherwise, with buffers, if the simulator exits config
changes can be lost.


tracernz (approved):

mlyle:
Just verified this cures the save bug.

PR #1523, opened 2017-01-22
pios_video: eliminate line_counter

mlyle:
Test status: works on seppuku/pal

PR #1589, opened 2017-02-20
RE1: Fix bug in LED driver

### LINKED ISSUE #1587: RE1: RGB LEDs broken
* Seen in version (release version or githash): Artifice
* On flight controller (if applicable): RE1

Issue details:
From [TPhillip on the forum](https://forum.dronin.org/forum/d/295-brainfpv-re1-led-issue-seppuku):
> There seems to be an issue with the new RGB LED module on RE1. When I set any value other than zero under "Total LEDs" the LEDs don't light up. After rebooting with the changed value, the OSD no longer displays, and the status LEDs don't flash. All other flight functions seem to be un-effected. Attempting to set the "Total LEDs" back to zero afterwards fails, also erasing all the settings on the board fails as well. The only way to get the board working properly again is to enter the bootloader and clear the settings partition.

> What I've tried:
Tested on two different machines (one Linux) with two separate USB cables
Also tested with two different RE1s, both showed similar results
Regressed back to Artifice and verified that LEDs physically still worked

Need to confirm this doesn't happen on other boards to narrow scope.

mluessi:
There was a bug in the LED driver for RE1 that caused the system module to hang on `PIOS_Assert` when LEDs were enabled. Not sure why it was working earlier.

Fixes #1587

### DIFF
--- a/flight/targets/brainre1/re1fpga/fpga_drv.c
+++ b/flight/targets/brainre1/re1fpga/fpga_drv.c
@@ -651,7 +651,7 @@ void PIOS_WS2811_trigger_update(ws2811_dev_t ws2811_dev)
 
 int PIOS_WS2811_get_num_leds(ws2811_dev_t ws2811_dev)
 {
-	PIOS_Assert(dev->magic == WS2811_MAGIC);
+	PIOS_Assert(ws2811_dev->magic == WS2811_MAGIC);
 
 	uint16_t n_leds = MIN(ws2811_dev->max_idx + 1, 1024);
 


tracernz (approved):
Thanks, looks good. Agree re: hotfix tag.

mluessi:
Would it be possible to have an Artifice bugfix release with this included?

mluessi:

mlyle:
Sure, though I think I'd like to wait a little longer for other issues to show up to avoid excessive hotfixing.

mluessi:
@mlyle can we build artifacts for this PR in the meanwhile? So I can direct people here to download them.

mlyle:
I don't think there's really a way to do the release + this PR other than the hotfix process.

I appreciate you have a customer service issue, and will try to expedite a hotfix release, but... anything we ship both takes about 2-3 hours of my time/attention, creates some degree of risk of regression or bad artifacts, and presents an ongoing maintenance load.  Deferring for a time to try and amortize that effort and risk over more than one moderate severity issue makes a lot of sense.

I don't have enough hardware to test RE1 thoroughly myself (just one flying quad without '2811 / one receiver type / etc).  It appears the regression is my fault, Dec 20, and no one noticed in the 2 months from then to the release-- either in code review of changes to the RE1 target or any subsequent test at all of the functionality on RE1 prior to release.

If the target is going to be supported, it requires attention from someone with the time and resources to adequately test and maintain it.

mluessi:
@mlyle Re is suported, but it would probably make sense to have some kind of collaborative test matrix before a release is published.

Changes were made to the RE1 code and it wasn't tested before merging it, which shouldn't have happened IMO.

mlyle:
@mluessi --- when we founded the project, we discussed what a maintainer was expected to do.

The LED functionality evolved to require a new call.  I made a best effort to put it in for RE1 despite having no hardware to test.  You made no objections and apparently made no test over the month that the release was pending.  The alternative eventually becomes to merge nothing, ever.

Is your solution that I buy a bunch of RE1s to have bench models of everything and run the test matrix myself?  That doesn't sound particularly appealing to me.

mlyle:
I mean, that's the alternatives-- you aren't around and looking at PRs that affect your targets.  The choices @tracernz and I have are to merge nothing, or to be careful and merge and hope that you will actually run a decent test matrix before release.

mluessi:
That's why I suggested a collaborative test matrix, a google doc or something. So we know things have been tested before the release is published. I didn't follow what was going on very closely and it went from preview release to release fairly quickly.

mlyle:
>  I didn't follow what was going on very closely

That doesn't sound like the target is maintained, then.  The release shipped a month after its schedule, which was publicly announced (forum & milestone & IRC), so it finally making out the door shouldn't have been a surprise.

mluessi:
I'm not going to argue with you. Fact is that I fixed the bug within 2 days after it was found.

As I suggested before, it would IMO be best to have a more systematic release process where target maintainers can confirm that things have been tested on each target.

mlyle:
> Fact is that I fixed the bug within 2 days after it was found.

OK, but that doesn't mean that I need to drop everything and hotfix and take the accordant risks.

> As I suggested before, it would IMO be best to have a more systematic release process where target maintainers can confirm that things have been tested on each target.

@mluessi If you would like to organize a more structured test matrix and enroll testers, especially for your target, it would be appreciated.

I'm sorry you were surprised by release; it followed a similar time from preview to ship as all releases to date (1 week).  Development had greatly slowed / quasi-frozen for 10 days before this-- a bit shorter period than usual because of how late the Qt change we were expecting shipped.  If we ship at 90 day intervals saving ~20 days for final maturing at the end is a reasonable slice of the cycle length.  Other people involved in the project all happened to know a release was coming...   The Git milestones were set to the target dates...

To date, this is the only known regression with the release...

If you want to take a more active role in maintaining things, your assistance with #1523 would be appreciated...

mlyle:
@mluessi PS, other things suck too, like the fact that og brain/RE1 shipped without DAC support, and on the previous release shipped without buzzer support.

edit: I don't even know how to do the DAC short of reverse engineering where traces go with a multimeter or microscope.

mluessi:
I can look into adding the DAC driver to RE1 / OG brain. Sparky2 can also support it AFAIK.

I don't think it's a big deal that this wasn't included in the release, as tools for end users don't really exist at the moment.

mlyle:
It's just one thing to change in the uavo browser if they want gps / battery alarm beeps.

PR #1591, opened 2017-02-24
Milliseconds aren't microseconds.

glowtape:
Everywhere I look in the code, the output of PIOS_DELAY_GetRaw() and PIOS_DELAY_DiffuS(), especially since the "uS" in the name suggests it, gets treated as microseconds. Except in pios_rcvr.c, where it's treated as milliseconds. Which I assume is not correct.

Mostly a cosmetic fix for correctness, since there's no receivers faster than 250hz. As far as I know, anyway.


mlyle (approved):
Woops.  Thx.

mlyle:
Jenkins, test this please

PR #1592, opened 2017-02-25
TBS Crossfire driver

glowtape:
Based off the IBus driver. Could probably merge the two, might look messy. I want to eventually look into sending telemetry back, once I figure out how to best integrate the data collection, so might be better if this was its own file. Supported telemetry is battery usage, GPS coords and vehicle attitude.

The CRSF serial protocol runs at 420kbaud, non-inverted, uses CRC8 (TBS variant) and supports theoretically up to 250hz, since the minimum frame timing is 4ms per some mythical specification. Right now it appears to send at link transmission rate, which is 150hz in low-latency mode, 50hz in weak RF/close range forced telemetry mode, and potentially at 4hz in forced telemetry mode in weak RF conditions (that'd probably be 20km+ out). The 4hz thing is something I'm trying to get cleared up, whenever TBS responds to it over at FPVLab, whether the serial link drops to that rate, too, since that's a problem with the desired ~50ms receiver timeout.

Benchtested on OG Brain main and flex ports, and on the RE1 in my quad on the receiver port. Flew it a bit outside, too. Set up Seppuku already in this commit, too, since I'm going to put it in my quad shortly.

There's one issue: Whenever the port is set up with the Crossfire driver and there's something being sent on the port, the firmware freaks out and bootloops until the receiver stops sending or you set it back to CRSF. Can't figure out why. Happens on all three tested ports.


flight/PiOS/Common/pios_crossfire.c
```diff
@@ -0,0 +1,337 @@
...
+
+	if (!PIOS_Crossfire_Validate(dev))
+		goto out_fail;
+
+	for (int i = 0; i < buf_len; i++) {
+		// Ignore any incoming stuff until we're expecting new data.
+		if(!dev->bytes_expected || dev->buf_pos >= CRSF_MAX_FRAMELEN) continue;
+
+		dev->u.rx_buf[dev->buf_pos++] = buf[i];
+		dev->bytes_expected--;
```
mlyle: I'm not sure I like this structure of two parallel counters -- buf_pos increasing and bytes_expected decreasing.  Can buf_pos be compared to bytes_expected?

flight/PiOS/Common/pios_crossfire.c
```diff
@@ -0,0 +1,337 @@
...
+
+		if(dev->buf_pos == CRSF_LEN_IDX)
+		{
+			// Read length field. Denotes payload, plus type field, plus CRC field.
+			dev->bytes_expected = dev->u.frame.length;
+
+			// Sanitize because we don't want to blow up when crap gets sent on serial.
+			if(dev->bytes_expected >= (CRSF_MAX_PAYLOAD + CRSF_TYPE_LEN + CRSF_CRC_LEN))
+				dev->bytes_expected = (CRSF_MAX_PAYLOAD + CRSF_TYPE_LEN + CRSF_CRC_LEN);
```
mlyle: Resync may be more robust setting this to a low value.

flight/PiOS/Common/pios_crossfire.c
mlyle: (Might simplify the above conditional too)

flight/PiOS/Common/pios_crossfire.c
```diff
@@ -0,0 +1,337 @@
...
+		if(dev->u.frame.type == CRSF_FRAME_RCCHANNELS)
+		{
+			// From pios_sbus.c
+			uint8_t *s = dev->u.frame.payload;
+			uint16_t *d = dev->channel_data;
+
+			#define F(v,s) (((v) >> (s)) & 0x7ff)
+
+			/* unroll channels 1-8 */
+			*d++ = F(s[0] | s[1] << 8, 0);
```
mlyle: If you use indices on the right, please use indices on the left too.

flight/PiOS/Common/pios_crossfire.c
```diff
@@ -0,0 +1,337 @@
...
+}
+
+static void PIOS_Crossfire_Supervisor(uintptr_t context)
+{
+	struct pios_crossfire_dev *dev = (struct pios_crossfire_dev *)context;
+	PIOS_Assert(PIOS_Crossfire_Validate(dev));
+
+	// If more than 1.6ms passed without communication, safe to say that a new
+	// packet is inbound.
+	if (++dev->rx_timer > 1)
```
mlyle: I assume this is the best that can be gotten away with here..  It's unfortunate, as it makes resync difficult if things are a byte off, etc.

mlyle (commented):
Thanks for this!  Overall this looks good.  I don't see an obvious crash bug (it's possible that sbus at 100kbps becomes too much garbage/too many interrupts at 420kbps and saturates CPU) -- but some of the commented things improving could make the code a little easier to verify.

flight/PiOS/Common/pios_crossfire.c
glowtape: I guess easiest would be to ditch all the rest, if the length field is implausible and let the supervisor reset everything for the next packet.

flight/PiOS/Common/pios_crossfire.c
glowtape: Makes sense.

flight/PiOS/Common/pios_crossfire.c
glowtape: Practically there's only 150hz, i.e. 6.67ms, so there's some breathing room. I think the 250hz only happen between the Taranis and the transmitter. If the receiver would output at 250hz, things might get a little tight. The main draw to the protocol would mainly be shaving off 2ms control delay and having CRC.

flight/PiOS/inc/pios_crossfire.h
```diff
@@ -37,13 +37,13 @@
 extern const struct pios_rcvr_driver pios_crossfire_rcvr_driver;
 
 /**
- * @brief Initialises the IBus Rx driver with a serial port
- * @param[out] ibus_id IBus receiver device handle
+ * @brief Initialises the TBS Crossfire Rx driver with a serial port
+ * @param[out] ibus_id Crossfire receiver device handle
```
yamahabest: ibus_id should be crsf_id.

flight/PiOS/Common/pios_hal.c
```diff
@@ -884,6 +913,22 @@ void PIOS_HAL_ConfigurePort(HwSharedPortTypesOptions port_type,
...
+			usart_port_params.init.USART_BaudRate            = 420000;
+			usart_port_params.init.USART_WordLength          = USART_WordLength_8b;
+			usart_port_params.init.USART_Parity              = USART_Parity_No;
+			usart_port_params.init.USART_StopBits            = USART_StopBits_1;
+			usart_port_params.init.USART_HardwareFlowControl = USART_HardwareFlowControl_None;
+			usart_port_params.init.USART_Mode                = USART_Mode_Rx;
+
+			PIOS_HAL_ConfigureTbsCrossfire(usart_port_cfg, &usart_port_params, com_driver);
+		}
+#endif  /* PIOS_INCLUDE_IBUS */
```
yamahabest: PIOS_INCLUDE_IBUS should be PIOS_INCLUDE_CROSSFIRE,

tracernz:
Can you please make your commit messages a bit more readable/useful? See https://github.com/d-ronin/dRonin/commits/next for examples, especially tagging the component you're working on at the start helps to filter when trawling through log/history. Keep in mind that they all show up in the total repo history so "Initial commit" doesn't really make sense except for the actual initial commit back in 2010.

glowtape:
I flew eight packs (~35min) today on a build with this and #1599, no issues whatsoever. I'd wager the receive-only version is good to go.

The only gotcha is that when something else sends at a lower bitrate on the USART pin while it's set up for CRSF's 420kbit, it causes drama and makes the firmware boot loop until you disable the receiver/signal (and then disable it in the GCS) or switch it back to CRSF on the receiver. Something related also seems to happen, should you reconfigure channel maps, causing a single reboot (remapping presumably re-inits their USART, doing funny business?). That latter case actually allowed me to see that the watchdog is the reboot reason.

That may be worth a footnote, although I don't see any reason for anyone to change channel maps or wire protocol mid-flight. Switching the transmitter off and on and stuff like that causes no issues however.

I have another branch on top of this with telemetry support (attitude, plus battery and/or GPS if either are enabled). However there are some issues, which seem to happen receiver or transmitter side, where data seemingly gets corrupted while being decoded and/or repacked (I'm pretty certain it's not at my end, because I don't see how you can fuck up filling something with all zeroes). No one seemed to have noticed, because apparently plenty of Betaflight users running CRSF thought the RPY values for attitude were their PIDs (I wish that was a joke). Trying to get a hold of Perna from TBS regarding this, but that's like winning the lottery apparently.

flight/PiOS/Common/pios_crossfire.c
```diff
@@ -0,0 +1,335 @@
...
+
+		dev->u.rx_buf[dev->buf_pos++] = buf[i];
+
+		if(dev->buf_pos == CRSF_LEN_IDX) {
+			// Read length field and adjust. Denotes payload, plus type field, plus CRC field.
+			dev->bytes_expected = CRSF_ADDRESS_LEN + CRSF_LENGTH_LEN + dev->u.frame.length;
+
+			// If length field isn't plausible, ignore rest of the data.
+			if(dev->bytes_expected >= CRSF_MAX_FRAMELEN)
+				dev->bytes_expected = 0;
```
mlyle: Doesn't this result in infloop?  maybe set bytes_expected to a sane value and break?

mlyle: Sorry, not infloop, but ignoring everything forever...

glowtape: bytes_expected gets reset to CRSF_MAX_FRAMELEN via PIOS_Crossfire_ResetBuffer() in the supervisor during the inter-frame communication pause. Typical distance between frame starts is 6.67ms, sending a maximum frame takes around 0.8ms (RC channels is 0.56ms), which leaves plenty of time for the supervisor to do that up to 3 times. This should work well enough, right? --edit: Should be more efficient, too, since it doesn't even get as far as running the CRC8 routine.

mlyle: Why would you ever want to loop over bytes you're discarding?

flight/PiOS/Common/pios_crossfire.c
```diff
@@ -0,0 +1,335 @@
...
+		}
+
+		if (dev->buf_pos == dev->bytes_expected) {
+			// Frame complete, decode.
+			PIOS_Crossfire_UnpackFrame(dev);
+			PIOS_RCVR_ActiveFromISR();
+		}
+	}
+
+	dev->rx_timer = 0;
```
tracernz: ~~You should only ever do this when you reset the buffer right?~~

-e- derp, wrong timer

glowtape: I thought that's to track whether the receiver still communicating/sending? The IBus driver of yours does it the same way, which this is based on.

tracernz: Yup, I read it wrong. :wink:

tracernz:
It might be appropriate to go to a table-based CRC for speed since this is running in an interrupt.

mlyle:
Or to do other verifications first-- e.g. that message type and len are
sane for each other.

tracernz:
@mlyle, absolutely, but you already had that covered. ;)

glowtape:
Right now it looks like there's only RC channel messages, anyway, so I could simply reverse the order. Check message type first, then do CRC8 to validate correctness.

The lookup table adds another 256 bytes. That a problem? It'd also affect the VTXConfig stuff, since it's the same CRC function.

glowtape:
Folded my telemetry branch into this.

There's provisions for selecting a separate USART for telemetry, but it's not yet exposed on the F4 targets. Not sure how to handle people selecting a telemetry port but not a Crossfire receiver.

glowtape:
Flew lots of packs on this meanwhile, since I merge it into every test branch of mine. Doesn't seem to drop the quad out of the sky.

mlyle:
Updated to merge cleanly with recent changes.  Needs real review.

glowtape:
Hmmm, I've pulled these changes onto my machine. Everything compiles fine, but when I boot up the flight controller, it appears to boot and work fine (WS2812 LED does its shpiel on my Seppuku) but the GCS can't connect. The connection widget in the lower right goes full blast. --edit: Settings were erased.

mlyle:
Ugh.  suggest you rebase this :(  Next was broken, and it's not easy to merge things out, but a good rebase of just your commits should make it nice.

mlyle:
Branch `mpl-crossfirefixup` is based clean off current next.   You can pick it up (assuming i'm a remote called ml) by doing

```
git branch my-backup-name
git fetch ml
git reset --hard ml/mpl-crossfirefixup
git push -f
```

while on the branch.

flight/PiOS/Common/pios_crossfire.c
```diff
@@ -0,0 +1,388 @@
...
+	// COM device ID for sending telemetry. Might be this USART
+	// or a separate one.
+	uintptr_t telem_com_id;
+
+	// To track frame starts to track whether telemetry is OK to send.
+	uint32_t time_frame_start;
+};
+
+// COM driver ID for separate telemetry channel, via pios_hal.
+extern uintptr_t pios_com_crsf_sep_telem_id;
```
mlyle: needs to be removed, right?

flight/Modules/UAVOCrossfireTelemetry/uavocrossfiretelemetry.c
```diff
@@ -0,0 +1,235 @@
...
+MODULE_INITCALL(uavoCrossfireTelemetryInitialize, uavoCrossfireTelemetryStart)
+
+// Stupid big endian stuff.
+#define WRITE_VAL(buf,p,x)				{ typeof(x) v = x; uint8_t *q = (uint8_t*)&v, n = sizeof(v); do{ buf[p++] = q[--n]; } while(n); }
+#define WRITE_VAL16(buf,p,x)			{ typeof(x) v = x; uint8_t *q = (uint8_t*)&v; buf[p++] = q[1]; buf[p++] = q[0]; }
+#define WRITE_VAL32(buf,p,x)			{ typeof(x) v = x; uint8_t *q = (uint8_t*)&v; buf[p++] = q[3]; buf[p++] = q[2]; buf[p++] = q[1]; buf[p++] = q[0]; }
+
+static int crsftelem_create_attitude(uint8_t *buf)
+{
+	int ptr = 0;
```
mlyle: not a huge problem / doesn't need to be fixed, but I'd prefer `pos` to `ptr` since it is not, in fact, a pointer.

PR #1594, opened 2017-02-25
actuators: some small simplifications and code cleanups

mlyle:
- Existing code for LEDs and buzzer is not useful.  Redirecting output
from existing annunciator subsystem up at the mixer level could
eventually be worthwhile and cleaner.
- Reclaim horizontal space from that pulldown on the output pane / make
things less busy.
- Clarify some comments.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -457,23 +457,13 @@ static void actuator_task(void* parameters)
 #if defined(MIXERSTATUS_DIAGNOSTICS)
 		MixerStatusSet(&mixerStatus);
 #endif
-
-		// Update servo outputs
-		bool success = true;
-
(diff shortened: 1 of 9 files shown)


flight/Modules/Actuator/actuator.c
```diff
@@ -462,7 +462,7 @@ static void actuator_task(void* parameters)
 		bool success = true;
```
tracernz: Should remove this too since it no longer does anything. Can probably also remove ActuatorCommand.NumFailedUpdates?

tracernz (changes requested):

flight/Modules/Actuator/actuator.c
mlyle: Good eye.

tracernz (approved):

PR #1595, opened 2017-02-26
stabilization: Implementation of nth-order gyro lowpass filter.

glowtape:
This is a proposal for having an optional 2nd order filter available within the firmware going forward. This would be a nice option for people to mess with, if they have more troublesome frames, and I don't need to permanently maintain a separate branch. Flew it in the backyard earlier, didn't appear to blow up.

Re: PIDs and autotune... IIRC @mlyle applies the same filter used on the gyros in the flight controller to the raw data in his autotune-ng script, presumably to account for the delay caused by the filter. In that case, autotune-ng can account for the higher delay of the 2nd order filter, if necessary.

I need to figure out QTCreator eventually and expose the option as checkbox in the GCS. When autotune-ng shows up, I suppose getting a warning label to show up, indicating that you need to rerun the autotune wizard, when changing the filter, shouldn't be a problem.

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -105,6 +105,151 @@ enum {
 	PID_MAX
 };
 
+static const float butterworth_factors[16] = {
+	// 2nd order
+	1.4142f,
+
+	// 3rd order
+	1.0f,
+
+	// 4th order
+	0.7654f, 1.8478f,
+
+	// 5th order
+	0.6180f, 1.6180f,
+
+	// 6th order
+	0.5176f, 1.4142f, 1.9319f,
+
+	// 7th order
+	0.4450f, 1.2470f, 1.8019f,
+
+	// 8th order
+	0.3902f, 1.1111f, 1.6629f, 1.9616f
+};
+
+struct filter_biquad_state {
+	float x1, x2, y1, y2;
+};
+
+struct filter_biquad {
+	float b0, a1, a2;
+	struct filter_biquad_state s[MAX_AXES];
+};
+
+struct filter_first_order {
+	float alpha;
+	float prev[MAX_AXES];
+};
+
+struct filter_compound_state {
+
+	struct filter_first_order *first_order;
+	struct filter_biquad *biquad[4];
+	uint8_t order;
+
+};
+
+static void filter_construct_single_biquad(struct filter_biquad *b, float cutoff, float dT, float q)
+{
+	float f = 1.0f / tanf((float)M_PI*cutoff*dT);
+
+	// Skipping calculation of b1 and b2, since this only going to do Butterworth.
+	// These terms are optimized away in the actual filtering calculation.
+	b->b0 = 1.0f / (1.0f + q*f + f*f);
+	b->a1 = 2.0f * (f*f - 1.0f) * b->b0;
+	b->a2 = -(1.0f - q*f + f*f) * b->b0;
+
+	memset((void*)b->s, 0, sizeof(struct filter_biquad_state)*MAX_AXES);
+}
+
+static void filter_construct_biquads(struct filter_compound_state *filt, float cutoff, float dT, int o)
+{
+	// Amount of biquad filters needed.
+	int len = o >> 1;
+
+	// Calculate the address of coefficients in the look-up table.
+	// There's probably a proper mathematical way for this. Let's just count
+	// everything.
+	int addr = 0;
+	for(int i = 2; i < o; i++)
+	{
+		addr += i >> 1;
+	}
+
+	// Create all necessary biquads and allocate, too, if not yet done so.
+	for(int i = 0; i < len; i++)
+	{
+		if(filt->biquad[i] == NULL)	{
+			filt->biquad[i] = PIOS_malloc_no_dma(sizeof(struct filter_biquad));
+		}
+		filter_construct_single_biquad(filt->biquad[i], cutoff, dT, butterworth_factors[addr+i]);
+	}
+}
+
+static void filter_create(struct filter_compound_state *filter, float cutoff, float dT, uint8_t order)
+{
+	// Don't check whether filter is NULL or not. We're passing a static
+	// struct in the code. If this ends up NULL for a reason, something
+	// else is completely broken.
+
+	// Clamp order count. If zero, this bypasses the filter.
+	if(order == 0) {
+		filter->order = 0;
+		return;
+	} else if(order > 8) order = 8;
+
+	if(order & 0x1)	{
+		// Filter is odd, allocate the first order filter.
+		if(filter->first_order == NULL)	{
+			filter->first_order = PIOS_malloc_no_dma(sizeof(struct filter_first_order));
+		}
+
+		filter->first_order->alpha = expf(-2.0f * (float)(M_PI) * cutoff * dT);
+		memset((void*)filter->first_order->prev, 0, sizeof(float)*MAX_AXES);
+	}
+
+	filter->order = order;
+	filter_construct_biquads(filter, cutoff, dT, order);
+}
+
+static float filter_execute(struct filter_compound_state *filter, int axis, float sample)
+{
+	int order = filter->order;
+
+	// Order at zero means bypass.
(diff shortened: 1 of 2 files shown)


flight/Modules/Stabilization/stabilization.c
```diff
@@ -105,6 +105,15 @@ enum {
 	PID_MAX
 };
 
+typedef struct {
+	float x1, x2, y1, y2;
+} butterworth_state;
+
+typedef struct {
+	float b0, b1, b2, a1, a2;
+	butterworth_state state[MAX_AXES];
```
mlyle: I think you mean for this to be [] and do dynamic allocation.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -281,6 +322,12 @@ static void stabilizationTask(void* parameters)
 	ident_wiggle_points = (1 << (ident_shift + 3));
 	uint32_t ident_mask = ident_wiggle_points - 1;
 
+	// Setup butterworth state, if second order filter is desired.
+	if(settings.GyroButterworth)
+	{
+		butter_gyro_filter = PIOS_malloc_no_dma(sizeof(butterworth_filter)*MAX_AXES);
```
mlyle: This multiplication is incorrect.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -160,6 +172,35 @@ static float get_throttle(StabilizationDesiredData *stabilization_desired, Syste
...
+	b->a2 = -(1.0f - q*f + f*f) * b->b0;
+
+	memset((void*)b->state, 0, sizeof(butterworth_state)*MAX_AXES);
+}
+
+static float do_butterworth(butterworth_filter *filt, int axis, float sample)
+{
+	butterworth_state *b = &filt->state[axis];
+
+	float y = filt->b0 * sample + filt->b1 * b->x1 + filt->b2 * b->x2 + filt->a1 * b->y1 + filt->a2 * b->y2;
```
mlyle: How general purpose is this to be?  If things are really just to be butterworth in the future, it seems like it may be good to do filt->b0 * (sample + b->x1 * 2) -- saves some memory operations, some arithmetic, and some ram.

shared/uavobjectdefinition/stabilizationsettings.xml
```diff
@@ -27,6 +27,9 @@
 		<field name="PitchPI" units="" type="float" elementnames="Kp,Ki,ILimit" defaultvalue="2.5,0,50" limits="%BE:0:20,%BE:0:20,"/>
 		<field name="YawPI" units="" type="float" elementnames="Kp,Ki,ILimit" defaultvalue="2.5,0,50" limits="%BE:0:20,%BE:0:20,"/>
 		<field name="GyroCutoff" units="Hz" type="float" elements="1" defaultvalue="55.0"/>
+		<field name="GyroButterworth" units="" type="enum" elements="1" options="FALSE,TRUE" defaultvalue="FALSE">
```
mlyle: Any chance we could go a bit further and let the user select a filter order here, and dynamically construct the right thing?  For helis, second order will be a big improvement but I think they might really want 4th-5th order..

mlyle: (brick wall at 25Hz instead of a sloping filter at 12Hz like we have now or a 2nd order at 18Hz or whatever).

shared/uavobjectdefinition/stabilizationsettings.xml
```diff
@@ -27,6 +27,9 @@
...
+			<description>Enables the use of a second order Butterworth lowpass filter on the gyro. Can't be changed during runtime.</description>
```
mlyle: It would be really nice if it could--- isn't this just a question of moving the malloc of the butterworth filter down, like..

```
  else {
    if (!butter_gyro_filter) {
      butter_gyro_filter = PIOS_malloc_no_dma(sizeof(butterworth_filter);
    }
    if (butter_gyro_filter) {
      create_butterworth(butter_gyro_filter, settings.GyroCutoff, dT_expected);
    }
  |

File: flight/Modules/Stabilization/stabilization.c [hunk 6]
```diff
@@ -320,11 +367,18 @@ static void stabilizationTask(void* parameters)
...
+			if(!butter_gyro_filter)
+			{
+				if (settings.GyroCutoff < 1.0f) {
+					gyro_alpha = 0;
+				} else {
+					gyro_alpha = expf(-2.0f * (float)(M_PI) *
+							settings.GyroCutoff * dT_expected);
+				}
+			}
+			else
```
mlyle: Our coding convetion places brackets on the same line as if / else / everything else (other than function decls).

flight/Modules/Stabilization/stabilization.c
```diff
@@ -105,6 +105,15 @@ enum {
 	PID_MAX
 };
 
+typedef struct {
```
mlyle: In general we don't do typedefs like this, except for things we opaquify and export from modules as handles.  On this one I could go either way..

mlyle (changes requested):
Very cool!  Thanks for this change set-- I've been meaning to write something like this for awhile.

Sorry for the long list of requested changes-- but you're writing more and more stuff in more critical places so it becomes more important for convention to be followed.

mlyle:
jenkins, test this please

shared/uavobjectdefinition/stabilizationsettings.xml
glowtape: I can turn this into an uint8 to select an order. Gonna take a while to brush my math up enough to construct filters beyond the second order (I know it's stacking biquads, but the coefficients).

flight/Modules/Stabilization/stabilization.c
glowtape: I was considering trying a Chebyshev T1 for the steeper fall-off, but the use of it is questionable.

mlyle: If so the name should probably change.

mlyle:
One other thing to consider is to migrate all of this filtering infrastructure to the 'sensors' module'.  Then attitude loop automatically benefits from it.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -105,6 +105,147 @@ enum {
...
+};
+
+struct filter_first_order {
+	float alpha;
+	float prev[MAX_AXES];
+};
+
+struct filter_compound_state {
+
+	struct filter_first_order *firstOrder;
```
mlyle: convention is underscore

flight/Modules/Stabilization/stabilization.c
```diff
@@ -105,6 +105,147 @@ enum {
...
+		}
+		filter_construct_single_biquad(filt->biquad[i], cutoff, dT, butterworth_factors[addr+i]);
+	}
+}
+
+static void filter_create(struct filter_compound_state *filter, float cutoff, float dT, uint8_t order)
+{
+	if(!filter)
+	{
+		// Fail here.
```
mlyle: At least return

glowtape:
Bench verification that this code even does what it is supposed to. Ran modified logging profile at 1KHz, dumped the filtered output into the RateDesired, to fiddle with it in logview (FFT, Y-axis scaled the same every time). Quad spun idle on the desk each time. This looks good to go for a flight, whenever there's another dry period.

fft

glowtape:
Tried 3rd and 5th, as requested by @mlyle, they work as expected.

Did some hover tests in the living room, FWIW anyway, the 80hz second order filter has a similar delay (the first ~20hz) as the 55hz first order per MATLAB, buts cuts the magnitude of the harmonics at 180hz in half and gets rid of the noise at the top end. So that'd be a good new default. Tried a 3rd order at 80hz, too, for ~1ms more delay, the harmonics are cut in half again over the 2nd and everything above is practically a straight line. Hope weather will allow soon, to gather some real noisy data.

I think I'll create another branch sticking this filter into sensors.c this evening, since I can't take the quad outside the next few days, anyway.

mlyle:
Superseded #1599

PR #1581, opened 2017-02-14
Use shared USB ID definitions, move all boards to dRonin IDs

tracernz:
Move all targets to dRonin USB IDs (while GCS will still support legacy IDs for the foreseeable future). Windows installer is modified to allow installation of CDC driver on Windows <10.

P.S. target_code--

To test:
- [x] Win installer, Win 10
- [ ] Win installer, Win <10
- [x] Debian installer

PR #1596, opened 2017-02-26
Udev rules updated to support udev tags for permissions

lrvick:
Debian 6+, Ubuntu 13.04+ and Arch Linux do not rely on the 'plugdev' group and have been switching to a convention of using udev tags for device permissions instead.

This should avoid users having to create a group, add themseves to it, and logout/login. Plugdev group method still in place for backwards compat. Won't hurt anything.

I have been modifying this by hand for Arch and finally decided to PR it.

### DIFF
--- a/package/linux/deb/_package.udev
+++ b/package/linux/deb/_package.udev
@@ -3,86 +3,86 @@
 #
 
 # OpenPilot Flight Control board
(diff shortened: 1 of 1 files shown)


tracernz:
Thanks. I guess, I'll pull this into #1581.

mlyle:
That's probably the best way to sequence/handle this.  Thanks!

mlyle:
Superseded #1581

PR #1597, opened 2017-02-27
discoveryf4: remove target

mlyle:
The use cases that we envisioned for this-- leading to not removing it
when removing the other eval board targets-- never came to fruition.


mlyle:
Rebased.

tracernz (approved):
I'm find with this. Remaining use-cases are all covered by real boards or sim.

mlyle:
Rebased

PR #1598, opened 2017-02-27
sim: cross-compile improvements and sample rate hack

mlyle:
If you install gcc6, on some distributions the compiler name is
arm-linux-gnueabihf-gcc-6 ... probe for this and use gcc6 if possible,
else fall back to arm-linux-gnueabihf-gcc.

### DIFF
--- a/flight/PiOS/Common/pios_bmx055.c
+++ b/flight/PiOS/Common/pios_bmx055.c
@@ -288,8 +288,12 @@ int32_t PIOS_BMX055_SPI_Init(pios_bmx055_dev_t *dev, uint32_t spi_id, uint32_t s
 	TaskMonitorAdd(TASKINFO_RUNNING_IMU, bmx_dev->task_handle);
 
 	PIOS_SENSORS_SetMaxGyro(2000);
-	PIOS_SENSORS_SetSampleRate(PIOS_SENSOR_GYRO, 800);
-	PIOS_SENSORS_SetSampleRate(PIOS_SENSOR_ACCEL, 800);
+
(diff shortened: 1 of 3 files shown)


mlyle:
This last commit greatly reduces event system warnings on flyingpi and also seems to cure the every-other-message-bad phenomenon.

mlyle:
Successful flight test with these changes.

tracernz:
Shouldn't we prefer gcc5 since that's what we compile regular flight firmware with? :stuck_out_tongue: I'm also fine with this though, and we should probably decide if we want gcc6 for arm-none-eabi.

mlyle:
> I'm also fine with this though, and we should probably decide if we want gcc6 for arm-none-eabi.

Yes.  I've been waiting for the 2nd release on the 6 series, and it just shipped.  https://launchpad.net/gcc-arm-embedded/6.0/6-2017-q1-update

tracernz:
Great. :+1: Downloads: https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads

mlyle:
Heh all those downloads:

```
<Error>
<Code>BlobNotFound</Code>
<Message>
The specified blob does not exist. RequestId:0cda9ee9-0001-00c1-6128-920c81000000 Time:2017-03-01T01:07:48.3497287Z
</Message>
</Error>
```

tracernz:
Great, first party support not always the best huh? No way to contact them about it either AFAICT.

mlyle:
Looks like there's this at least https://bugs.launchpad.net/gcc-arm-embedded/+bug/1668379

PR #1599, opened 2017-02-27
sensors: Implement N-th order gyro and accel lowpass filter.

### LINKED ISSUE #1312: Autotune: consider running gyro filter on incoming points

glowtape:
Alternative version of #1595, that puts the filtering in sensors and applies to both gyros and accelerometers. Works on the bench, would be interesting to see how it affects current autotune, but weather and all.

Still need to fix the naming in GPL headers, need to figure out the proper group names.

Might be an idea to include the magnetometer?

fixes #1312


mlyle:
@ekaszubski This code is still relatively young, but I'm curious how this and the improvements to attitude code do for you.

flight/Libraries/math/filter.c
```diff
@@ -0,0 +1,156 @@
...
+ */
+
+#include <stdio.h>
+#include <stdint.h>
+#include <string.h>
+#include <math.h>
+#include "pios_heap.h"
+#include "filter.h"
+
+const float filter_butterworth_factors[16] = {
```
mlyle: All of the things not accessed directly by callers-- this, construct_single_biquad, construct_biquads.. should be declared 'static'

flight/Libraries/math/filter.h
```diff
@@ -0,0 +1,60 @@
...
+
+enum {ROLL,PITCH,YAW,MAX_AXES};
+
+struct filter_biquad_state {
+	float x1, x2, y1, y2;
+};
+
+struct filter_biquad {
+	float b0, a1, a2;
+	struct filter_biquad_state s[MAX_AXES];
```
mlyle: Is it possible to make this dynamic, too?  s[]; and then malloc appropriately.  Upshot there is that we can use this filter infrastructure for other things with different numbers of degree of freedom.

flight/Modules/Sensors/sensors.c
```diff
@@ -357,6 +361,10 @@ static void update_accels(struct pios_sensor_accel_data *accels)
 	    accels->z * accel_scale[2] - accel_bias[2]
 	};
 
+	accels_out[0] = filter_execute(&accel_filter, 0, accels_out[0]);
```
mlyle: Style choice-- it would be possible to just pass in accels_out and do all 3 at once.  Not sure which is better-- this is perhaps a little more versatile; the other is a bit quicker.

flight/Modules/Sensors/sensors.c
```diff
@@ -372,6 +380,8 @@ static void update_accels(struct pios_sensor_accel_data *accels)
 	accelsData.z += z_accel_offset;
 
 	accelsData.temperature = accels->temperature;
+
```
mlyle: Kill one of these lines please.

mlyle (changes requested):
Thanks for this code :D  I am super excited about merging it.  --- Though it will require some caution as this is about as critical as anything gets.

flight/Libraries/math/filter.h
```diff
@@ -0,0 +1,60 @@
...
+};
+
+struct filter_first_order {
+	float alpha;
+	float prev[MAX_AXES];
+};
+
+struct filter_compound_state {
```
mlyle: Sorry, noticed a couple more things.

This struct should be left as is because it is put on data segment inside the sensors module and that's efficient.  But the other ones should move to the filter.c.  The pointers for state then will be to opaque structures to all callers, discouraging misuse.

mlyle:
Also, please consider renaming filter.c / filter.h       Right now source names need to be unique across all directories and this just invites doubling.  ( I know we have many other problems elsewhere).

glowtape:
I've been bored as hell at work yesterday and looked into Bessel again. Apparently pulling biquad coefficients via BLT lets you use it with the caveat that its group/phase delay will only be maximally flat up to Fs/4. That'd be 250hz at 1KHz gyro rate. Is this worth investigating? Or even using it at all, since it'd require another table of coefficients to multiply Fc with.

Also, I've been thinking hiding away lpfilter_compound_state behind an uintptr_t to make it completely opaque from the outside. Can't have outside code messing with the order or width variables. Not that I expect anyone doing this in the near future, but we shuffled other internal structs already away from the header for similar reasons.

mlyle:
Please don't hide anything behind uintptr_t--- it's not a good practice and something we're trying to migrate away from.  An opaque struct is great, though -- a declaration of struct whatever;   and perhaps typedef struct whatever *whatever_t and then client code can use whatever_t but doesn't know what's inside.  Then in the actual .c file which has a definition of the structure it can all be happy.

I do not think Bessel is desirable-- and it's probably best to not spend time/code on things that it will be very difficult to establish any kind of difference between experimentally.  It's not like at second order that these filters are all that different, anyways-- or for that matter, that the actual ESCs or perhaps even the sensors maintain these phase/delay relationships.

mlyle (approved):

glowtape:
Rebased on 8dc909c335fea352f936270c2c80f431ab004f33.

mlyle:
Flew #1618, #1610, #1599... plus a change to decimate = 2 on Lux.  Set to 88Hz second order filter.  Did an autotune.  

Requires disabling the lowpass filter in the autotune script (for filtering change), but the results are good:

bokeh_plot 25
 
http://dronin.org/tunes#?key=ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA0OCU3AgM

PR #1607, opened 2017-03-02
python-api: correct a python3 regression

mlyle:


jpahka:
fixed python3 in logviewer for me

PR #1608, opened 2017-03-02
Remove support for F1 flight controllers (CC3D, Naze32)

### LINKED ISSUE #1363: Drop support for F1
For formal discussion here (and mentioning it on IRC, and it can make its way to the forum/broader community after we gather initial thoughts on it.)

We're nearing the end of Quixote, and it looks like it makes F1 a little better and fixes some regression/annoyance there.  It also shows us how limited F1 is, and it's annoying to worry quite so much about RAM, etc.

I'd like to do one more release after Quixote that supports F1, and declare the release "Long-term support"-- we'll fix critical bugs in it for some time even after subsequent releases ship, and maybe even backport trivial features.

After that, nix it.  (Except F1-based pipxtreme/oplink/taulinkmodule and F0-based flyingpio).

Thoughts?

mlyle:
Has the time come?  Fixes #1363.

There's a pleasant amount of negative diff here, though the 'acro' attitude code needs to get pulled also (with accordant changes to Lux/DTFc and eliminating the altitude-runaway-if-no-baro bug.  Since that affects more targets I think it's a good consideration for a subsequent PR.


mlyle:
I'm not dead-set on this, though some of the work I want to do in actuator and for new-autotune is noticeably harder if this doesn't happen.  Thoughts?

Tested the upgrader on Sparky2 on Linux with success.

mlyle:

jhitesma:
I've come around to accept the inevitable.  I still have one quad on Naze32 but pretty much only so I can still test on it.  

The biggest argument against is how many cc3d users we seem to have.

tracernz:
Can UAVOMSPBridge run on PipX? That would be kinda cool actually if it doesn't already. Might be okay without the SMALLF1 define anyway.

mlyle:
@tracernz yah right now pipx doesn't define SMALLF1 and there are SMALLF1 workarounds that aren't there.

What exactly is the use case you're thinking of for UAVOMSPBridge on PipX?  Ground-side OSD?

mlyle:
Currently F1 is just under half of our growth.  It's likely this will improve a little bit further in the 2.5-3 months  until we ship.

tracernz:
@mlyle Yah, OSDs, ground stations, antenna trackers. This PR doesn't stop it so non-issue here. :+1: 

Removing F1 flight boards is a tough one that we did already decide to do, but then kinda went back on after RE1 developments (which turned out to be pretty insignificant). I think on that basis I'm for this merging. Whether we should still attempt to maintain Artifice for F1 boards is the real question?

mlyle:
>  Whether we should still attempt to maintain Artifice for F1 boards is the real question?

I favor a soft approach-- that we try and keep artifice builds working for as long as we reasonably can and consider anything particularly severe for hotfix back there for like a year or so.

I don't think we have to be **overly** formal because it's not like we have paying customers with some kind of support arrangement-- if there's a severe bug 6 months from now and it's too hard to do on Artifice for whatever reason, we punt (in the completely give up US sense of the word).

glowtape:
A message in bold at the top of the release notes, over the next 3 releases, with a link to an Artifice release, should certainly do it. That's like a year or more. If someone didn't get the message by then, tough luck.

mlyle:
Was discussed on the forum-- https://forum.dronin.org/forum/d/320-dropping-f1/  -- no dissent.

This may not remove 100% of the code that it's possible to remove, but it's a good start.  Recommend for merge.

tracernz:
Any PipX test on this branch?

mlyle:
Pipxtreme confirmed to work.

PR #1609, opened 2017-03-04
attitude: cope when no baro is present (neuter cfvert)

mlyle:
Corrects the issue where if you're using a board without a baro, that altitude can run up to infinity (or down to negative infinity) just following the accelerometer.

Now if things are missing, don't even try to run cfvert.  Just peg the altitude at 0.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -728,6 +728,8 @@ static int32_t updateAttitudeComplementary(float dT, bool first_run, bool second
 	}
 
 	if (!secondary) {
+		static bool got_baro_pt = false;
+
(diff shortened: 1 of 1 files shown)


mlyle:
This isn't correct for targets that never have baro / mag.

Though if the missing sensor alarm is asserted and .. more work is done with missing sensor severity it could be OK.

Should happen after other related PRs merge

* [ ] ~~Confirm revolution~~
* [ ] ~~Confirm RE1~~
* [ ] ~~Implement DTFC sensor code~~
* [ ] ~~Implement Lux sensor code~~
* [ ] ~~Implement PikoBLX sensor code~~

mlyle:
Took a simpler approach.  Confirmed OK on Lux.  IMO OK to merge.

PR #1613, opened 2017-03-05
stabilization: Factor out failsafe provisions and run them before att…

glowtape:
…itude error calculation.

Essentially just factored the code out and run it before anything interesting gets done with control input. Overrides the values of the stabDesired variable, tho, instead of leaving them alone and relying on a separate variable ("raw_input") as before, that doesn't even get used consistently. Shouldn't make a difference practically.


glowtape:
Tested this for giggles, the quad still flies as expected.

PR #1614, opened 2017-03-06
python: include the spec file I use to bundle logview

mlyle:
To use this, I "source activate logview" to activate my Anaconda Python
3 environment, then run pyinstaller dronin-log.spec, and then I manually
zip up the dist directory.

This is not a proper build procedure, but getting something checked in is preferable to losing the knowledge about how to produce this artifact (again).

### DIFF
--- a/python/dronin-log.spec
+++ b/python/dronin-log.spec
@@ -1,0 +1,35 @@
+# -*- mode: python -*-
+
+block_cipher = None
+
+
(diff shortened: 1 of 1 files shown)


tracernz:
Good idea to put this in. I've used pyinstaller at work with a similar process (except with a .bat file to do it all for consistency).

PR #1615, opened 2017-03-07
Fix gcsauthors.html output

### LINKED ISSUE #1600: Authors list on Authors screen still empty
* Seen in version (release version or githash): Artifice
* On flight controller (if applicable): -

Issue details:
The html files are now included (#1563), however the gcsauthors.html file doesn't have any authors.

mlyle:
Fixes #1600 

It seems git shortlog -sn changes its format when run by qmake-- maybe a locale thing-- from spaces to tabs.  As a result splitting wasn't working.  Tell python string splitting to use any whitespace.


mlyle:

tracernz:
Good lesson on not using porcelain commands. Should've heeded my own advice in the original PR:
> `git-shortlog` is not a plumbing command so really should fix this up to use a plumbing command instead (although plenty of porcelain commands are used in version-info.py already).

mlyle:
Crap.  This did not actually fix things on the jenkins build.

tracernz:
Heh, I thought it seemed odd that the relative authors list worked in artifact builds even though it also uses tabs to split... which I missed, you didn't change it's split in this.

PR #1616, opened 2017-03-08
stabilization: Fixed variable masking ('float error' masking upper level 'bool error').

glowtape:


tracernz:
Why is the default case unnecessary? The compiler won't pick up a missing case because your switch is on an int, not an enum. DONT_BUILD_IF might be good though.

mlyle:
@tracernz yes, the DONT_BUILD_IF (or using an appropriate enum with a numelem value) is good.

IMO it does not make sense to have a loop that is counting 0, 1, 2, and then code that tries to detect if the loop is out of that range.

tracernz:
I would agree if `MAX_AXES` was close by but it's not even in this file. Fortunately your `DONT_BUILD_IF` is excellent for this sort of thing. :+1:

mlyle:
The thing is, if these things mismatched, we already accessed invalid memory and scribbled on all kinds of memory in this loop.... trying to do something clever and assert an error too is not likely to be helpful.

glowtape:
DONT_BUILD_IF(!defined(MAX_AXES)&&(MAX_AXES!=3)) or what?

mlyle:
Probably MAX_AXES != YAW+1 

DONT_BUILD_IF example:

```
DONT_BUILD_IF(sizeof(union morsel_packed_state) > sizeof(uint32_t),
                packedStateRep);
```

Must be outside a function, have meaningful compile error for the second term (that fits the rules for an identifier.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -105,6 +108,11 @@ enum {
 	PID_MAX
 };
 
+// Check if the axis enumeration maps properly to the things above.
+DONT_BUILD_IF((PID_RATE_ROLL+0 != ROLL)||(PID_ATT_ROLL != PID_GROUP_ATT+ROLL), stabAxisPidEnumMapping1);
```
mlyle, line 112: What is the reason for the +0's here?

glowtape: Can't compare anonymous enums without the compiler throwing a fit about it. +0 is shorter than a cast to fix this.

mlyle: I see. :P

PR #1619, opened 2017-03-09
DAC annunciation for brain og

droney:
Utilizng DAC driver, this change enables DAC annunciation on
brain og. Code compiled and verified on single quad.


flight/targets/brain/fw/pios_config.h
```diff
@@ -54,6 +54,8 @@
 #define PIOS_INCLUDE_FASTHEAP
 #define PIOS_INCLUDE_FRSKY_RSSI
 #define PIOS_INCLUDE_TBSVTXCONFIG
+#define PIOS_INCLUDE_DAC
```
mlyle, line 57: PIOS_INCLUDE_DAC_FSK would be nice here too

mlyle (commented):
Looks great-- you've tested and hear the beeps?

flight/targets/brain/fw/pios_config.h
droney: that's for telemetry... correct? I did not have a way to verify it, so I left it out. Happy to resubmit, but it would not be verified.

droney:
yep, hear the beeps, which are in sync with the on-board led blinks. Was hoping for different tones/sounds depending on severity of alarm, but it's a start. I'm happy. Thanks for the driver hooks.

PR #1620, opened 2017-03-10
version-info: make run under python3

### LINKED ISSUE #1507: Tools: Python 3ize version_info.py
* Seen in version (release version or githash): 
* On flight controller (if applicable): 

Issue details:

mlyle:
Fixes #1507

Functional change: sacrificed lowercasing in the author sort operation, as that'd be a pain to deal with in unicode across 2 & 3.

Also save a couple of passes of running the versioninfo (still is run once per flight target, though, which is a whole heck of a lot of duplicated work scanning git stuff).

### DIFF
--- a/make/scripts/version-info.py
+++ b/make/scripts/version-info.py
@@ -431,9 +431,9 @@ string given.
                         help='print repository info to stdout')
     parser.add_argument('--format',
                         help='format string to print to stdout')
-    parser.add_argument('--template',
+    parser.add_argument('--template', action='append', default=[],
                         help='name of template file')
-    parser.add_argument('--outfile',
+    parser.add_argument('--outfile', action='append', default=[],
                         help='name of output file')
     parser.add_argument('--image',
                         help='name of image file for sha1 calculation')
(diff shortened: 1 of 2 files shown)


make/scripts/version-info.py
```diff
@@ -139,7 +148,7 @@ def _get_authors_since_release(self):
 
             self._exec('shortlog -s {}..HEAD'.format(last_rel['ref']))
             if self._rc == 0:
-                self._authors_since_release = sorted([l.strip().split(None, 1)[1] for l in self._out.splitlines()], key=str.lower)
+                self._authors_since_release = sorted([l.strip().split(None, 1)[1] for l in self._out.splitlines()])
```
tracernz, line 151: You have explicitly decoded `self._out` to string in `_exec` so the string sort should should be operating on Python `str` type in both 2 & 3? Tested on 2.7.12 and 3.5.2 and both worked fine with the sort key.

tracernz:
Nice changeset. :+1:  I like the multiple input files although the help strings could do with some clarification. ;)

make/scripts/version-info.py
mlyle: Not sure as to the difference in your environment but on Python2 I get:

```
  File "/home/mlyle/dRonin/make/scripts/version-info.py", line 151, in _get_authors_since_release
    self._authors_since_release = sorted([l.strip().split(None, 1)[1] for l in self._out.splitlines()], key=str.lower)
TypeError: descriptor 'lower' requires a 'str' object but received a 'unicode'
```

tracernz: Okay, safer to leave that then. :)

tracernz (approved):

PR #1625, opened 2017-03-12
version-info: fix windows build

mlyle:
Turns out that I had somehow gotten type=int in there, when the
positional arguments are not of type int.


mlyle:

tracernz:
Thanks. :+1:

PR #1626, opened 2017-03-12
GCS:Reporter: Fix trailing ^0 (distance from ref)

tracernz:
- GIT_TAG had the distance from the ref on the end, we already have a coarser version of this ("dirty") and it makes using the tag name more difficult. This will also now result in a blank string rather than "undefined" when HEAD isn't tagged.
- Funnily enough GIT_DIRTY was broken and always returned false as the qmake function isEmpty takes a variable __name__, not a value, and of course the variable named "" is empty, as is "4 files changed, 8 insertions(+), 6 deletions(-)" if we're actually dirty.
- Additionally, use plumbing commands rather than porcelain.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@
 /tools
 /build
 
+# Ignore jenkins stuff
+/Nodes
+
 #flight/Project/Xcode.OSX
 flight/Project/Xcode.OSX/build
 flight/Project/Xcode.OSX/firmware.xcodeproj/*.mode1v3
--- a/.gitignore
+++ b/.gitignore
@@ -65,7 +65,7 @@ androidgcs/google-play-services_lib/build.xml
 androidgcs/google-play-services_lib/gen/
 androidgcs/google-play-services_lib/local.properties
 androidgcs/google-play-services_lib/proguard-project.txt
-
+androidgcs/build
 
 # Ignore build output from Eclipse android builds
 androidgcs/bin/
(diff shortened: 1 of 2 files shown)


tracernz:

tracernz:
So I found an actual bug with my fail here. :grin: Fixed, updated OP, and force-pushed.

tracernz:
Works fine (GIT_DIRTY now works correctly on my machine) and crashreporter worked. The Jenkins artifact build was detected as dirty though... I wonder what Jenkins is doing? I know android stuff dirties up the tree but that builds later right? Couldn't get GIT_DIRTY to give a false value locally (detached heads etc. were also fine) so I'm leaning towards Jenkins does actually dirty the tree.

mlyle:

Yes, jenkins dirties the tree, adding a Nodes/ subdir on master.

tracernz:
Thanks. :)

tracernz:
Still dirty :( http://dronin-autotown.appspot.com/api/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJCQ2oMKDA

mlyle:

tracernz:

mlyle:
Is this worth merging, evne with GIT DIRTY being dirty?

tracernz:
It won't really hurt anything, just the dirty status on crash reports might not be accurate (if it's a release tag which almost all reports are, we can reasonably safely assume it's not dirty anyway...).

PR #1630, opened 2017-03-13
GCS: Fix breakpad path

tracernz:
Makes me want to move everything to QBS so we don't repeat all the things 2-5 times :P


mlyle:

yamahabest:
I can confirm that this fixed the compile error.

PR #1632, opened 2017-03-15
bl: f3/f4: define BOOTLOADER in BOOTLOADER makefile

mlyle:
To get built with correct ids.


mlyle:
Comes from the PIkoBLX branch.

PR #1633, opened 2017-03-15
Flight: AQ32 - Remove PWM RX support, add WS2811 and DAC features

jihlein:
This PR adds WS2811 LED and DAC support, at the expense of removing PWM receiver support.  It also shuffled some pin definitions, but since the AQ32 silkscreen never match Dronin use anyway, this isn't a deal killer in my opinion.  I'll update the pin definition spreadsheet in the hardware repo after this PR is merged.  Pin use is:

Function         Silkscreen            Resource Use
PPM RX          Servo 3 Output    (TIM5 CH3 configured for PPM Input)
PWM 7 Out    PWM 1 Input       (TIM4 CH1 configured as PWM output)
PWM 8 Out    PWM 2 Input       (TIM4 CH2 configured as PWM output)
PWM 9 Out    PWM 3 Input       (TIM4 CH3 configured as PWM output)
PWM 10 Out  PWM 4 Input       (TIM4 CH4 configured as PWM output)
WS2811          RNG Input           (TIM1, identical to Quanton with exception of GPIO pin)
DAC Out         Modem Output   (TIM6, identical to Seppuku with exception of GPIO pin)

All outputs have been checked on a logic analyzer and work correctly.  A PPM receiver was connected and I was able to set it up properly.  I have not flown the code yet, but the CPU load looks reasonable I think it's safe to merge as is.

### DIFF
--- a/flight/targets/aq32/fw/pios_board.c
+++ b/flight/targets/aq32/fw/pios_board.c
@@ -186,7 +186,7 @@ void PIOS_Board_Init(void) {
     // Timers used for outputs (2, 3, 4, 8)
     PIOS_TIM_InitClock(&tim_2_cfg);
     PIOS_TIM_InitClock(&tim_3_cfg);
-	PIOS_TIM_InitClock(&tim_4_cfg);
+    PIOS_TIM_InitClock(&tim_4_cfg);
     PIOS_TIM_InitClock(&tim_8_cfg);
 
 #ifdef PIOS_INCLUDE_SPI
--- a/flight/targets/aq32/fw/pios_board.c
+++ b/flight/targets/aq32/fw/pios_board.c
@@ -44,6 +44,7 @@
 #include "hwaq32.h"
 #include "manualcontrolsettings.h"
 #include "modulesettings.h"
+#include "rgbledsettings.h"
 
 /**
  * Configuration for the HMC5883 chip
@@ -179,12 +180,13 @@ void PIOS_Board_Init(void) {
 
     /* Set up pulse timers */
 
-    // Timers used for inputs (4)
-    PIOS_TIM_InitClock(&tim_4_cfg);
+    // Timers used for inputs (5)
+    PIOS_TIM_InitClock(&tim_5_cfg);
 
-    // Timers used for outputs (2, 3, 8)
+    // Timers used for outputs (2, 3, 4, 8)
     PIOS_TIM_InitClock(&tim_2_cfg);
     PIOS_TIM_InitClock(&tim_3_cfg);
+	PIOS_TIM_InitClock(&tim_4_cfg);
     PIOS_TIM_InitClock(&tim_8_cfg);
 
 #ifdef PIOS_INCLUDE_SPI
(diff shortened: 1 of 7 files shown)


mlyle:
Is everyone flying AQ32 cool with the removed functionality?

If someone has a flying AQ32 now with serial RX or PPM in, will it still work the same or do they need to change the pins attached?

jihlein:
"Is everyone flying AQ32 cool with the removed functionality?"

Don't know, can't say how many users are using PWM receivers.  This is the biggest drawback to this change, but no way around it if WS2811 support is desired.

"If someone has a flying AQ32 now with serial RX or PPM in, will it still work the same or do they need to change the pins attached?"

No effect on serial RX, changes do not effect uart ports.  PPM RX input pin does change and will have to be moved.  I fall into this category and it's not a big deal IMHO.

"Seems unrelated?"

Correct, these lines need be removed, they are left over from something else I am working on and forgot to delete......

jihlein:
Here's the spreadsheet for the hardware repo that highlights (in red text) the changes in pin use from the initial AQ32 work.

[aq32Dronin.xlsx](https://github.com/d-ronin/dRonin/files/846338/aq32Dronin.xlsx)

mlyle:
Only one AQ32's most-recent shared tune is PWM:

http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAwL7ivggM

3 boards' most recent on PPM.

jihlein:
I think that's @yamahabest, and he mentioned he's moving that installation off of PWM to PPM.

mlyle:
OK.  Sounds good to me-- will leave open a few days for any comment/objection.

flight/targets/aq32/fw/pios_board.c
```diff
@@ -179,12 +180,13 @@ void PIOS_Board_Init(void) {
...
-    // Timers used for inputs (4)
-    PIOS_TIM_InitClock(&tim_4_cfg);
+    // Timers used for inputs (5)
+    PIOS_TIM_InitClock(&tim_5_cfg);
 
-    // Timers used for outputs (2, 3, 8)
+    // Timers used for outputs (2, 3, 4, 8)
     PIOS_TIM_InitClock(&tim_2_cfg);
     PIOS_TIM_InitClock(&tim_3_cfg);
+	PIOS_TIM_InitClock(&tim_4_cfg);
```
mlyle: Indentation

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
```diff
@@ -184,8 +181,6 @@ Core::IBoardType::InputType AQ32::getInputType()
     switch(settings.RcvrPort) {
     case HwAQ32::RCVRPORT_PPM:
         return INPUT_TYPE_PPM;
-    case HwAQ32::RCVRPORT_PWM:
-        return INPUT_TYPE_PWM;
     default:
         break;
     }
```
mlyle, line 187: I think in here there's a call which asks whether an input type supported-- and it should say false for PWM.

flight/targets/aq32/fw/pios_board.c
mlyle: indentation is off here.

mlyle (commented):
LGTM

flight/targets/aq32/fw/pios_board.c
jihlein: Fixed

jihlein: Fixed

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
jihlein: Fixed

mlyle (approved):

ground/gcs/src/plugins/boards_aeroquad/aq32.cpp
tracernz: Yep. I prefer to have an explicit list without a default case, so if anything new is added, you are forced to review it rather than silently breaking things or missing a board out (this has happened a lot in the past). Of course this isn't really workable for the get function when you don't use a dedicated Rx port.

yamahabest:
Yep, that is my AQ32.
Whilst I would like to keep PWM support, for broad support, there are so many other receiver types available in dRonin, that it is an easy trade-off for features that more people might be using.

And PWM to PPM/SBUS convertors are available for cheap, so it shouldn't really be a problem.

tracernz (approved):
Good work. :+1:

PR #1637, opened 2017-03-16
remove ground path follower

mlyle:
It's been unused in a very long time-- and doubtful whether it still
works.  It's also very far outside the premise of the project.

But most importantly, I don't think the concept is very useful-- blind
GPS-coordinate-based navigation is not a very good premise for doing
anything useful with a ground vehicle.


tracernz:
Sounds fun TBH, and a lot less regulatory hurdles to test nav. code with a ground vehicle. :stuck_out_tongue:

mlyle:
Yes, but if it's a completely different nav implementation how useful is that?  I don't really want to reimplement this / get it working.

tracernz:
So flying near my place is not feasible without forward plannign and prior ATC approval etc... in which case I prefer to just drive outside the city instead. When peabody124 was talking about developing the original nav. impl. with ground vehicles recently it highlighted the feasibility of easily testing builds at the park 2 houses down from me. For that reason I've been thinking about buying a ground vehicle. Merging this would save me the $. :stuck_out_tongue:

mlyle:
I'd love to come up with a more generalized nav impl, that has the vehicle dynamics / loiter behaviors etc abstracted out and riding on top.

tracernz (approved):

PR #1639, opened 2017-03-16
attitude: raise AccKI, both during startup and at baseline

mlyle:
This is based on eyeballing actual behavior in plots from gooch/irc.  He
has a quad with a severe gyro bias issue.  The initial transient at
startup isn't fixing it, and the further trimming would take
approximately forever to correct it.

The arming value is chosen to be 15; a value of 6 would deliver a time
constant of approximately the startup interval (6 seconds); 15 is about
2/5 of the arming interval or we should expect about 92% of the error to
get trimmed out.

Since that's raising by 15x, raise the "normal" value by a factor of 8x
to 0.4.  By the above rationale that's a time constant of about 90
seconds-- enough that slow, very small drift is coped with, but small
attitude errors don't do much at all

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -530,7 +530,7 @@ static int32_t updateAttitudeComplementary(float dT, bool first_run, bool second
 
 		// For first 7 seconds use accels to get gyro bias
 		accKp = MIN(accKp, 20 + 20 * (PIOS_Thread_Systime() < 4000));
-		accKi = 1;
+		accKi = 15;
 		mgKp = 1;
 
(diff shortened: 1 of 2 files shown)


tracernz:
Are you confident this won't be unstable (excepting a totally borked sensor) or would you prefer some test? Otherwise looks fine.

mlyle:
Reasonably confident and have used this.   It is indeed about a 2s time constant for reasonable sensor errors at startup.

The other one.. 90s time constant.  I don't really know at all whether it's the "right" value either way, but it seems reasonable-- a 90s time constant ought to be able to follow drift on a 5 min kinda scale.. yet be very long in duration than any individual maneuver.

PR #1642, opened 2017-03-17
stabilization: Introduce control input smoothing

glowtape:
@mlyle and I discussed this a few months ago (I did some work, however a lot of procrastrination happened). The idea is to take the stair stepped RC input signal and smooth the corners out a bit, by using some amount of prediction to prevent delay. The output signal of it would have the leading edges chamfered. This is supposed reduce the P- and D-term kicks on large control input changes, and in consequence reduce actuator saturation.

This early implementation has four modes. The chamfered steps with two different prediction duty cycles, simple linear interpolation (for whoever wants it, maybe planes for servo smoothness), and interpolation by using predicted values similar to the first two modes (same as previous, but with reduced delay).

In normal operation, it should maximally only predict as long as it detects the frame timing to be (which is per-axis, for the very rare multi-RX setup). There's a hard limit of 20ms, since that's the PPM timing, in case something goes wrong with detection of the timing (it stores the smallest number it encounters).

The final implementation would have modes selectable per axis for RPY+T at least via UAVO browser.

Right now only smoothcontrol.c/h in this PR are of interest, the rest is just barebones integration for Seppuku.

Here's what the generated signals look in the simulator code (the green stuff is FFT, forgot why I added that the last time).

Chamfered steps with 50% of the signal time predicted.
icee50

Same but with 75% duty cycle.
icee75

Linear interpolation, well, result as expected.
interpol

Interpolation using predicts. With current parameters, the overshoots are pretty tame.
predictive

In my opinion, the 75% one seems to be a nice approach, with the interpolated version for least delay maximum smoothness. The overshoots in the former are very very tame compared to plain extrapolation. And they look unspectacular in latter, too. (If you switch between 75% and interpolated predicts, the curve looks similar enough.)

### DIFF
--- a/flight/Libraries/math/smoothcontrol.c
+++ b/flight/Libraries/math/smoothcontrol.c
@@ -102,7 +102,7 @@ void smoothcontrol_run(smoothcontrol_state state, uint8_t axis_num, float *new_s
 		smoothcontrol_update(state, axis, *new_signal);
 	}
 
-	if(state->tick_counter <= axis->integrator_timeout && state->tick_counter < state->time_bomb) {
+	if(state->tick_counter < axis->integrator_timeout && state->tick_counter < state->time_bomb) {
 		// Don't fiddle with first value after update.
 		if(state->tick_counter > 0)
 			axis->current += axis->differential;
--- a/flight/Libraries/math/smoothcontrol.c
+++ b/flight/Libraries/math/smoothcontrol.c
@@ -1,0 +1,178 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <math.h>
+
+#include "openpilot.h"
+#include "pios.h"
+#include "misc_math.h"
+#include "smoothcontrol.h"
+
+struct smoothcontrol_axis_state {
+	// Last known signal value. Needed to detect change.
+	float signal;
+
+	// Current state of axis.
+	float current;
+
+	// Amount of differential per step of integration.
+	float differential;
+
+	// Stores a time-out when to stop predicting. Unit is integrator ticks.
+	uint8_t integrator_timeout;
+
+	// Which conditioning algorithm.
+	uint8_t mode;
+};
+
+struct smoothcontrol_state_internal {
+
+	// Maximum ticks to integrate. Calculate value from dT and constant. Unit is ticks.
+	uint8_t time_bomb;
+
+	// Semaphore for cheapskates.
+	bool ringer;
+
+	// Whole state tick counter
+	uint8_t tick_counter;
+
+	// Control interval determined by the ringer.
+	uint8_t control_interval;
+
+	// RPY+Thrust.
+	struct smoothcontrol_axis_state axis[4];
+};
+
+// Ran on signal change, recalculates the diffs and timeouts for the specific mode.
+static void smoothcontrol_update(smoothcontrol_state state, struct smoothcontrol_axis_state *axis, const float new_signal)
+{
+	float signal_diff = new_signal - axis->signal;
+
+	switch(axis->mode) {
+		default:
+			// Override any crap values.
+			axis->mode = SMOOTHCONTROL_NONE;
+		case SMOOTHCONTROL_NONE:
+			// Doesn't do anything in the integrator.
+			break;
+
+		// icee's proposal, creates chamfered steps with signal prediction
+		case SMOOTHCONTROL_NORMAL:
+		case SMOOTHCONTROL_EXTENDED:
+			axis->differential = signal_diff * SMOOTHCONTROL_PREDICTOR_SLOPE / (float)state->control_interval;
+			axis->current = axis->signal + signal_diff * SMOOTHCONTROL_CHAMFER_START;
+			axis->integrator_timeout = (uint8_t)MAX((float)state->control_interval*
+				(axis->mode == SMOOTHCONTROL_NORMAL ? SMOOTHCONTROL_DUTY_CYCLE : SMOOTHCONTROL_EXTENDED_DUTY_CYCLE), 255);
+			break;
+	}
+
+	axis->signal = new_signal;
+}
+
+// Resets the axis state.
+void smoothcontrol_reinit(smoothcontrol_state state, uint8_t axis_num, float new_signal)
+{
+	PIOS_Assert(state);
+
+	// Prevent integrating when stick scale changes.
+	state->axis[axis_num].differential = 0;
+	state->axis[axis_num].signal = new_signal;
+	state->axis[axis_num].current = new_signal;
+}
+
+// Sets mode for an axis.
+void smoothcontrol_set_mode(smoothcontrol_state state, uint8_t axis_num, uint8_t mode)
+{
+	PIOS_Assert(state && axis_num <= 3);
+	state->axis[axis_num].mode = mode;
+	smoothcontrol_reinit(state, axis_num, state->axis[axis_num].signal);
+}
+
+// Processes the signal, if internal state allows it.
+void smoothcontrol_run(smoothcontrol_state state, uint8_t axis_num, float *new_signal, float limit)
+{
+	PIOS_Assert(state && axis_num <= 3);
+
+	struct smoothcontrol_axis_state *axis = &state->axis[axis_num];
+
+	// No mode, bypass.
+	if(axis->mode == SMOOTHCONTROL_NONE) return;
+
+	if(!state->tick_counter) {
+		// Manual control updated, do stuff
+		smoothcontrol_update(state, axis, *new_signal);
+	}
+
+	if(state->tick_counter <= axis->integrator_timeout && state->tick_counter < state->time_bomb) {
+		// Don't fiddle with first value after update.
+		if(state->tick_counter > 0)
+			axis->current += axis->differential;
+	} else {
+		// If we passed the timebomb, force the receiver signal, too.
+		if(state->tick_counter >= state->time_bomb) axis->current = *new_signal;
+	}
+
+	*new_signal = axis->current;
+}
+
+// Separate handling of thrust.
+void smoothcontrol_run_thrust(smoothcontrol_state state, float *new_signal)
+{
+	// The -1 on zero throttle causes a huge blip when throttling up, so we need to
+	// handle it separately.
+	if(*new_signal < 0)
+	{
+		smoothcontrol_reinit(state, 3, 0);
+	}
+	else
+	{
+		smoothcontrol_run(state, 3, new_signal, 1.0f);
+
+		// If prediction undershoots while original signal is positive
+		// bound it to zero.
+		if(*new_signal < 0)
+			*new_signal = 0;
+	}
+}
+
+// Advances the tick counters.
+void smoothcontrol_next(smoothcontrol_state state)
+{
+	PIOS_Assert(state);
+
+	if(state->tick_counter < 255) state->tick_counter++;
+
+	// This function is supposed to get called at the end of the stabilization loop.
+	// If MCC rang the ringer meanwhile, ticks will be at zero next loop, as expected.
+	if(state->ringer) {
+		int x = ((int)state->control_interval + (int)state->tick_counter) >> 1;
+		state->control_interval = (uint8_t)MAX(x, 255);
+		state->tick_counter = 0;
+
+		state->ringer = false;
+	}
+}
+
+// To tell us the dT, to calculate the time bomb for the integrator.
+void smoothcontrol_update_dT(smoothcontrol_state state, float dT)
+{
+	PIOS_Assert(state);
+	state->time_bomb = (uint8_t)(SMOOTHCONTROL_TIMEBOMB / 1000.0f / dT);
+}
+
+// Duh.
+void smoothcontrol_initialize(smoothcontrol_state *state)
+{
+	PIOS_Assert(state);
+
+	if(!*state) {
+		*state = PIOS_malloc_no_dma(sizeof(struct smoothcontrol_state_internal));
+		PIOS_Assert(*state);
+		memset(*state, 0, sizeof(struct smoothcontrol_state_internal));
+	}
+}
+
+bool* smoothcontrol_get_ringer(smoothcontrol_state state)
+{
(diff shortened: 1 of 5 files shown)


flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,194 @@
...
+			axis->mode = SMOOTHCONTROL_NONE;
+		case SMOOTHCONTROL_NONE:
+			// Doesn't do anything in the integrator.
+			break;
+
+		// icee's proposal, creates chamfered steps with signal prediction
+		case SMOOTHCONTROL_NORMAL:
+		case SMOOTHCONTROL_EXTENDED:
+			axis->differential = signal_diff * SMOOTHCONTROL_PREDICTOR_SLOPE / (float)axis->control_interval;
+			axis->current = new_signal + signal_diff * SMOOTHCONTROL_CHAMFER_START;
```
glowtape: new_signal should be axis->signal. Has been fixed.

glowtape:
Fixed a bunch of stupid mistakes and did a living room test. Seems to do no harm at least.

I did have a few odd twitches during hover, but that might be clipping related (IIRC the last time that happened, I had to change a motor, need to diagnose with Openlager).

glowtape:
This OK? Code belongs to stab.c, but it makes more sense in the input configuration.

(Not sure why this empty label shows up near the buttons, I didn't touch anything outside input.ui.)

smoothingui

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,195 @@
...
+	// Which conditioning algorithm.
+	uint8_t mode;
+
+	// Track integration time.
+	uint8_t tick_counter;
+
+	// Stores the lowest amount of ticks between control input updates.
+	// uint8_t allows for 32ms at 8KHz, should be sufficient to deal with
+	// even PPM and PWM at that rates (plz no).
+	uint8_t control_interval;
```
mlyle: I'm concerned jitter has the possibility to effectively disable this.  Maybe IIR this?

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,195 @@
...
+
+	switch(axis->mode) {
+		default:
+			// Override any crap values.
+			axis->mode = SMOOTHCONTROL_NONE;
+		case SMOOTHCONTROL_NONE:
+			// Doesn't do anything in the integrator.
+			break;
+
+		// icee's proposal, creates chamfered steps with signal prediction
```
mlyle, line 58: We need a worst-case timeout that applies no matter what.  On the order of 15ms.

mlyle: (Which is somewhat difficult here because it's in terms of control Hz)

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,195 @@
...
+
+		// Linear interpolation of the signal.
+		case SMOOTHCONTROL_LINEAR_INTERPOLATOR:;
+			float interp_diff = new_signal - axis->current;
+			axis->differential = interp_diff / (float)axis->control_interval;
+			axis->integrator_timeout = axis->control_interval;
+			break;
+
+		// Linear interpolation of predicts of the type above.
+		case SMOOTHCONTROL_SUPERSMOOTH:;
```
mlyle: While it's not an issue now, some compilers don't like `label: ;`    while `(void) 0;`  is guaranteed to work, though wordier.

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Separate handling of thrust.
+void smoothcontrol_run_thrust(smoothcontrol_state state, float *new_signal)
+{
+	// The -1 on zero throttle causes a huge blip when throttling up, so we need to
+	// handle it separately.
+	if(new_signal < 0)
+	{
+		float tmp = 0;
```
mlyle: really really need to think about the safety of this, especially with fancy interpolators.  Can a negative throttle signal ever be interpolated to positive?

mlyle: (I believe it can for awhile with the chamfering stuff)

flight/Modules/Stabilization/stabilization.c
```diff
@@ -1021,6 +1043,23 @@ static void zero_pids(void)
...
 }
 
+static uint8_t remap_smoothing_mode(uint8_t m)
+{
+	// Kind of stupid to do this, but pulling in an UAVO into "library" code feels
+	// like a layer break.
+	switch(m)
+	{
+		default:
+		case STABILIZATIONSETTINGS_RCCONTROLSMOOTHING_NONE:			return SMOOTHCONTROL_NONE;
```
mlyle: newlines/indentiness..

flight/targets/pipxtreme/fw/Makefile
```diff
@@ -175,6 +175,7 @@ SRC += $(FLIGHTLIB)/circqueue.c
 SRC += $(FLIGHTLIB)/morsel.c
 SRC += $(MATHLIB)/misc_math.c
 SRC += $(MATHLIB)/lpfilter.c
+SRC += $(MATHLIB)/smoothcontrol.c
```
mlyle: Shouldn't be here, right?

mlyle (commented):
A very good start, though having lots of options makes me nervous.

flight/Libraries/math/smoothcontrol.c
glowtape: Yea. When you pass -1 as-is into the code and it stores it as last value, and then you throttle up (throttle range is 0..1), it'll predict a large value, which causes the motors to accelerate hard for a control interval, just to brake like hell when there's proper prediction. I didn't check for -1 on an earlier version, so the quad always made a *clunk* noise. It doesn't write back to new_signal on -1 and tells the throttle axis state that there's 0 throttle.

flight/Libraries/math/smoothcontrol.c
glowtape: Jitter didn't appear much of an issue. If at all, the signal is rather a tad late than on time. The only issue was when the stab loop started, it immediately converged to 1. Thus the stuff on L125.

flight/Libraries/math/smoothcontrol.c
glowtape: There's a "time bomb" being calculated down in L179. It uses a #define that's currently set to 20ms (typical PPM timing) and derives an amount of ticks based on dT, then gets applied in L139. If the time bomb is passed, the original signal is forced. dT_expected is being passed from stab.c, so it should always be some valid value, and if actual dT doesn't match, it'll create a critical alarm and doesn't let the user arm.

flight/Libraries/math/smoothcontrol.c
mlyle: If things happen at 9ms intervals, and something happens that ties up the FC for 5 ms, we could get a 14ms interval and then a 4ms one.  Then afterwards our behavior would be permanently changed.

This is probably part of how you get the 1ms at first.

flight/Libraries/math/smoothcontrol.c
mlyle: Right.  But doesn't this mean that a negative throttle command can become a positive one (briefly, at least)?  E.g. throttle was 1, stick is moved to -1, which is treated as 0 here.  Then in turn we go "most of the way there" but don't actually disable throt...

flight/Libraries/math/smoothcontrol.c
tracernz: In the rest of our code (both flight and GCS) we use scopes `{ }`. I think we should stick with that rather than introducing yet another style.

flight/Libraries/math/smoothcontrol.c
tracernz: Also nothing wrong with PPM or PWM so `(plz no)` isn't really necessary.

flight/Libraries/math/smoothcontrol.c
glowtape: new_signal might probably be an irritating variable name. A value gets fed into the function via new_signal, some math happens and it gets dumped back into new_signal (the raw_input and ActuatorDesired.Thrust fields, that are processed, get set up anew in stab.c every loop, anyway).

In this wrapper, when it crosses into negative, a zero gets forced into the predictor, while the value in new_signal will be left untouched. So when throttle changes from say 0.5 to -1 over a receiver update, it'll leave the value at -1 and predict into nirvana (the tmp float). The last known signal in the state gets set to zero by feeding tmp=0, and when throttle crosses back into positive, it'll start doing its predicting in an orderly fashion, starting from zero.

flight/Libraries/math/smoothcontrol.c
mlyle: Hm, there's a bunch of after-first-statement variable decls in cases too.  I dunno.

glowtape:
Hmmm, I'm just out on the field and it freaks out. This is on hold them I guess.

glowtape:
The code behaves properly again.

glowtape:
Flew a bunch of packs on this configured to Supersmooth mode on all axes. Flies smoothly.

The twitches mentioned a few days ago are gone, they happened due to forgetting to dereference a pointer.

Rebased to merge cleanly with the stab.c changes.

glowtape:
I think this still needs bounding the predictions to the max. rates.

And probably a limit on how much deg/s it's allowed predict ahead of the current known signal, as to prevent huge overshoot and subsequent pull back that might happen on the end of large stick throws on a slow receiver. ESCs would probably thank for that.

I'd say 90deg/s might be a nice limit?

glowtape:
Flew all packs of mine on the sloped step version with 75% prediction duty cycle this time. No weird stuff either.

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,217 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <math.h>
+
+#include "openpilot.h"
+#include "pios.h"
+#include "smoothcontrol.h"
+
+#define ASSERT(x) if(x) PIOS_Assert(0);
```
tracernz: Why not just use PIOS_Assert? This macro actually inverts the meaning of ASSERT which is a very bad idea IMO. If you really must keep it, it should be called ASSERT_FALSE.

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,217 @@
...
+#define MAX(v,x) ((v) > (x) ? (x) : (v))
```
tracernz: We should probably have a good min/max implementation in pios.h or openpilot.h rather than random ones like this. e.g. http://stackoverflow.com/a/3437484. But I thought arm-none-eabi headers already give us MIN/MAX??

tracernz: We do already have higher quality ones: https://github.com/d-ronin/dRonin/blob/ebb1954b074bc168491bf11bdb639ddced14a0f9/flight/Libraries/math/misc_math.h#L38

glowtape:
Rebased and condensed.

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,178 @@
...
+			// Doesn't do anything in the integrator.
+			break;
+
+		// icee's proposal, creates chamfered steps with signal prediction
+		case SMOOTHCONTROL_NORMAL:
+		case SMOOTHCONTROL_EXTENDED:
+			axis->differential = signal_diff * SMOOTHCONTROL_PREDICTOR_SLOPE / (float)state->control_interval;
+			axis->current = axis->signal + signal_diff * SMOOTHCONTROL_CHAMFER_START;
+			axis->integrator_timeout = (uint8_t)MAX((float)state->control_interval*
+				(axis->mode == SMOOTHCONTROL_NORMAL ? SMOOTHCONTROL_DUTY_CYCLE : SMOOTHCONTROL_EXTENDED_DUTY_CYCLE), 255);
```
mlyle: If 255 is chosen, does this ever time out?

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,178 @@
...
+		case SMOOTHCONTROL_NONE:
+			// Doesn't do anything in the integrator.
+			break;
+
+		// icee's proposal, creates chamfered steps with signal prediction
+		case SMOOTHCONTROL_NORMAL:
+		case SMOOTHCONTROL_EXTENDED:
+			axis->differential = signal_diff * SMOOTHCONTROL_PREDICTOR_SLOPE / (float)state->control_interval;
+			axis->current = axis->signal + signal_diff * SMOOTHCONTROL_CHAMFER_START;
+			axis->integrator_timeout = (uint8_t)MAX((float)state->control_interval*
```
mlyle: space before star

flight/Libraries/math/smoothcontrol.c
```diff
@@ -0,0 +1,178 @@
...
+	state->time_bomb = (uint8_t)(SMOOTHCONTROL_TIMEBOMB / 1000.0f / dT);
+}
+
+// Duh.
+void smoothcontrol_initialize(smoothcontrol_state *state)
+{
+	PIOS_Assert(state);
+
+	if(!*state) {
+		*state = PIOS_malloc_no_dma(sizeof(struct smoothcontrol_state_internal));
```
mlyle: sizeof(*state)

mlyle (commented):
Looks good.  I will make another careful pass through this after changes (since this is super-critical).

glowtape:
Will be interesting to prove an effect. On these ones attached, which is ActuatorCommand.Roll, with 75% duty cycle on the prediction, grabbed at 2ms from the GCS, you can tell, but only because I'm running an 8th order lowpass at 5hz on the gyros. At normal settings, any sloping goes under in the gyro noise that feeds into ActuatorCommand (at least visually).

(--edit: Altho these were small stick movements.)

chamfer

PR #1643, opened 2017-03-19
Scrubbing some markup and small tweaks to the UI

glowtape:
This is currently up for @tracernz 's amusement.


ground/gcs/src/inc/widgetnames.h
```diff
@@ -1,6 +1,18 @@
...
+#define CONFIGGADGETWIDGET_VEHICLETYPE					"configGadgetVehicleType"
+#define CONFIGGADGETWIDGET_INPUT						"configGadgetInput"
+#define CONFIGGADGETWIDGET_OUTPUT						"configGadgetOutput"
+#define CONFIGGADGETWIDGET_ATTITUDE						"configGadgetAttitude"
+#define CONFIGGADGETWIDGET_STABILIZATION				"configGadgetStabilization"
+#define CONFIGGADGETWIDGET_MODULES						"configGadgetModules"
+#define CONFIGGADGETWIDGET_AUTOTUNE						"configGadgetAutotune"
+#define CONFIGGADGETWIDGET_CAMERASTAB					"configGadgetCameraStab"
+#define CONFIGGADGETWIDGET_TXPID						"configGadgetTxPID"
+#define CONFIGGADGETWIDGET_OSD							"configGadgetCameraOSD"
```
tracernz: This header is a bad idea because:
a) #defines pollute namespaces, totally unnecessary in c++
b) config plugin stuff belongs in plugins/config, each plugin is basically standalone from the other parts of GCS (obv. there are dependencies between certain plugins exposed through their public APIs)
c) I don't see any actual win over just using the literals where you need them?
d) this will cause a ton of merge conflicts if you intended it to be for all plugins (which is not the way GCS is designed as per b)
e) breaks the style guide https://github.com/d-ronin/dRonin/wiki/Development-Coding-standards#gcs-code-c (see http://wiki.qt.io/Coding_Conventions#Aesthetics)

tracernz:

tracernz:
Stabilisation widget looks good on mac. 😁

ground/gcs/src/inc/widgetnames.h
mlyle: Would be better to use a constant from each of these classes instead of bare literals

tracernz: I would agree if they were used in more than 1 place..

glowtape: Right now it only contains the config widgets, because I haven't looked much into the other code-only widgets files yet.

The principal idea was to make objects addressable by name from a global stylesheet. Say if I were to want to put a background image into the config widget iconbar, without name, the selector I could use would be either MyTabbedStackWidget > QListWidget and I'd have to pray that no one else uses that same nested order, or use the the whole hierarchy and pray that no one changes it during UI work. With naming a stylesheet gets maintainable and selectors stay short. And as you suggested, it'd likely help with crashdumps.

The idea with widgetnames.h was mainly to have all names in a central place, so that you don't need to run after each code file, if you have/need to rename the object names for whatever reason.

glowtape: I can stick literals directly into the files, if ideally we don't intend to rename them in bulk anytime soon. The work is almost the same.

glowtape:
This is ongoing work. I suppose we could already merge some of the stuff when Neat is due, though.

ground/gcs/src/inc/widgetnames.h
tracernz: The problem is that you can't use the #defines in the stylesheet because css files don't go through the c pre-processor, and they certainly don't see c++ constants. 😛 I appreciate the idea but I think it's very unlikely these will ever be renamed given that they haven't even been named at any point in the past 7 years! 😀 As you say, naming QObjects is useful because [they come through in usage data](https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/usagestatsgadget/usagestatsplugin.cpp#L199).

glowtape: Right. I'll ditch the header then.

glowtape:
Ok, this is what this patch does:

- It scrubs unnecessary palette entries in the markup, which were a lot, which were also redundantly applied as a stylesheet. The net result of this is 180KB saved on libConfig.so on a Linux build. Similar applies to the other builds.

- I've redone the connection widget in the lower right, to be more in line to what I've in mind visually for some other things. It also reduces the SVG from 51KB originally to 3.8KB for the new one. Also, the tiny labels indicating data flow have been enlarged a bit.

- The round rect around the content has been removed. That looks silly from a visual standpoint.

- The bottom tab bar got a bunch of paddings to improve visuals.

- There were some tweaks to the code to make those three points actually work. Plus the GCS looks for a global QT stylesheet now, independent from the system.

vis2

vis

mlyle:
This looks good to me on OSX / Windows.

It's a massive diff.  Objections to merge, @tracernz?

ground/gcs/src/plugins/coreplugin/mainwindow.h
```diff
@@ -156,7 +157,7 @@ private slots:
     void updateFocusWidget(QWidget *old, QWidget *now);
     void destroyVersionDialog();
     void destroyAuthorsDialog();
-    void modeChanged(Core::IMode *mode);
+    void modeChanged(Core::IMode *);
```
tracernz: ??

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -375,6 +375,23 @@ void MainWindow::extensionsInitialized()
     emit m_coreImpl->coreOpened();
 }
 
+void MainWindow::readStyleSheet(QFile *file, QString name, QString *style) {
+    QString tmp;
```
tracernz, line 393: Not needed in the outer scope but meh.

ground/gcs/src/plugins/config/configgadgetwidget.cpp
```diff
@@ -58,8 +58,11 @@ ConfigGadgetWidget::ConfigGadgetWidget(QWidget *parent) : QWidget(parent)
 {
     setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::MinimumExpanding);
 
+    this->setObjectName(QString("ConfigWidget"));
```
tracernz: Would it make more sense to match the name of the class to avoid yet another set of names to remember/track?

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -181,14 +181,28 @@ MainWindow::MainWindow() :
...
     m_modeStack->setMovable(false);
     m_modeStack->setMinimumWidth(512);
+    m_modeStack->setMinimumHeight(384);
+    m_modeStack->setMaximumWidth(16777215);
+    m_modeStack->setMaximumHeight(16777215);
+    m_modeStack->setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::MinimumExpanding);
     m_modeStack->setElideMode(Qt::ElideRight);
 
+    gridLayout->addWidget(m_modeStack, 0, 0);
+    gridLayout->setContentsMargins(0,0,0,0);
```
tracernz: spaces

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -181,14 +181,28 @@ MainWindow::MainWindow() :
     registerDefaultContainers();
```
tracernz, line 181: From the commit comment it seems you are thinking the stylesheets are CSS or supposed to work the same as CSS, that's not the case. They work on the existing QWidget layout engine, so some of the CSS things would be "pants on head idiotic".

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
tracernz: spaces

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
glowtape: All elements have border, margin and padding fields. I'd expect a mark-up based layout engine to implement a proper boxing model, that's consistent. Poking the stick at it, trying to make various visuals work on an experimental branch, there's occasionally a catch that makes you scratch your head. In the case of the central widget, I had to nest it inside another widget to make it work as expected, so that I could apply paddings to the MyTabWidget.

ground/gcs/src/plugins/config/configgadgetwidget.cpp
glowtape: This can probably be dropped. I still need to verify for sure, but referring by class name seems to work, too. Names are probably not needed unless there's multiple instances, which in this case shouldn't.

tracernz: Ack. Remember also that you can quite happily have multiple Config gadgets in your layout so referring to it by class seems a much better idea.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
tracernz: Yup, so everything in the stylesheet is subject to the QWidget layout stuff, e.g. size policies etc. 😁

tracernz:
Worth checking that it works okay/doesn't throw a bunch of warnings with two of the same gadget I think. (this screenshot is NOT from this branch)

Imgur

mlyle:
updated jenkins/lyle.org, so firing a build

mlyle:
Superseded #1669 (since it includes this branch and probably makes sense to review/land them together).

PR #1646, opened 2017-03-20
GCS: Fix crash from reboot button

tracernz:
Oops, my bad. :blush:

http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJCQi4YIDA
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJCXlbMLDA

Repro'd this by clicking reboot button and them immediately disconnecting board. Confirmed it doesn't crash with this (QPointer has some special QObject sauce). Couldn't help tidying a little while I was there.

### DIFF
--- a/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
+++ b/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
@@ -949,10 +949,17 @@ void ConfigTaskWidget::defaultButtonClicked()
     }
 }
 
-// TODO: just call into Uploader instead
+/**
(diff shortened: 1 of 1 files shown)


tracernz:
Note: this only occurs on the hardware pages which are destroyed when the board disconnects, and only if you're unlucky with timing. The board shouldn't disconnect until at least 500 ms after the last IAP command... which means this bug shouldn't actually surface if everything else works properly..

PR #1652, opened 2017-03-23
pios_rcvr: handle very early updates OK

mlyle:
RX protocol inits before first PIOS_RCVR_Init, which in turn can cause
hangs/badness.  Possible this is what @ufodziner was experiencing, but
unlikely.  But in any event this is good to fix.


mlyle:
This is dangerously close to "double check idiom" but I have concluded it is safe because interrupts are masked during the allocation and assignment.

tracernz (approved):

PR #1653, opened 2017-03-23
attitude: use volatile flags for uavo changes

### LINKED ISSUE #953: Flight:Attitude: Unsafe UAVO Callback Usage
UAVO callback (https://github.com/d-ronin/dRonin/blob/next/flight/Modules/Attitude/attitude.c#L1468) touches lots of variables that aren't marked volatile. Compiler can perform unsafe optimizations. Should follow standard pattern of setting a flag and checking it each iteration of the task.

mlyle:
And process in the main loop.  Fixes #953.

Further cleanup here would be nice--- attitude is a big mess.  But this is a start, and fixes the biggest correctness problem.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -54,6 +54,7 @@
 #include "physical_constants.h"
 #include "coordinate_conversions.h"
 #include "WorldMagModel.h"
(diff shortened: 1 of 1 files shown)


tracernz (approved):
Great job. :grinning:

PR #1659, opened 2017-03-24
Moved the break within the #ifdef section 

yamahabest:
It doesn't make sense to be after the #endif, because if one would
disable the PIOS_INCLUDE_MS5611 define, there would be a single break.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -54,6 +54,7 @@
 #include "physical_constants.h"
 #include "coordinate_conversions.h"
 #include "WorldMagModel.h"
+#include "insgps.h"
 
 // UAVOs
 #include "accels.h"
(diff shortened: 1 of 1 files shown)


tracernz:
It does make sense and would stop the board from crashing...?

mlyle:
Would need to get teh case outside first.

@yamahabest  --- any chance that this can be cleaned up-- perhaps with a rebase -- to be a single commit instead of 20 merge commits plus one code change?

yamahabest:
Let me check, I am using the GitHub client. Using the sync button to get the changes from dRonin/next to my next, and then making a branch of my next, gets me all this stuff. Starting google.

tracernz:
Ah, now on desktop, I see where the ifdef was. I assumed it was inside the case, which would stop the crash and hopefully trigger a sensors alarm instead (although I don't recall the specifics of how we decide if we _should_ be getting updates from a particular sensor).

yamahabest:
Damn.. what a mess.. This is going the wrong way.. Trying more..

yamahabest:
Shhh... I don't know why it closed this one. Opened a new one: #1661. Looks better now.

PR #1660, opened 2017-03-25
Delete lots of things shared between targets

mlyle:
Minus lines are good lines.


flight/PiOS/STM32/chibi_main.c
```diff
@@ -2,12 +2,10 @@
  ******************************************************************************
  * @addtogroup Targets Target Boards
  * @{
- * @addtogroup Sparky2 Tau Labs Sparky2
- * @{
```
tracernz, line 6: Can we make this:
```
 * @addtogroup PIOS PIOS Core hardware abstraction layer
 * @{
 * @addtogroup ChibiOS ChibiOS Interface
 * @{
```
or similar for all of these files moving into PiOS?

I'd really like to keep a minimal set of main groups in the top level of the nav bar, and then subgroups so things are easier to find without scrolling through hundreds of symbols.

mlyle: OK

mlyle:
Yay. Thx.

PR #1667, opened 2017-03-26
Simulator: fixes for windows

mlyle:
* Sleep forever was wrong everywhere... which only affected windows because main thread exiting takes the whole program with it by default.
* Windows requires send/recv instead of write/read
* Windows doesn't allow IPV4 connections on IPV6 sockets by default.  There's a flag to overcome this, but it doesn't work in our build env.  FML.  Set it for now anyways.

With these changes, the simulator "works" on Windows.  Unfortunately it doesn't hit timing, and you get a Stabilization error that the update rate is too low.


mlyle:
Confirmed this is still happy on Linux.

tracernz (approved):

PR #1669, opened 2017-03-26
Some more tweaks to the UI.

glowtape:
This bases on my glowtape-ohdear branch (#1643, which seems included in this PR, too) and flattens the UI a bit and gets rid of the brownish tones.

This PR bases on the assumption that barely anyone touches the Edit Gadgets stuff and runs some sort of default configuration. If someone has analog dials or other non-default gadgets on that screen, it may look out of place.

Pudding:

New linear system health bar in Windows.
windows1

Flight data screen in Windows (developer default configuration).
windows2

Basic flight data screen configuration in Ubuntu (we ought to tweak that one a bit, doesn't look busy enough).
ubuntu1

Netbook configuration of flight data screen in Ubuntu.
ubuntu2

Linear system health bar in Ubuntu.
ubuntu3

Bunch of the linear dials in Ubuntu.
ubuntu4

The remaining buttons.
buttons


mlyle:
Tested this for about 15 minutes with sim on LInux.  It's very pretty.  Toggling to config pane is a little "blinding" in contrast.

I use a controller widget on the flight data screen which looks a little out of place.

Still a win overall I think.

glowtape:
I'm eventually going to touch more stuff, but I have to pick my battles and draw a line somewhere :P

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -375,6 +375,23 @@ void MainWindow::extensionsInitialized()
...
+    QString tmp;
+    if(file->open(QFile::ReadOnly)) {
+        /* QTextStream... */
+        QTextStream styleIn(file);
+        /* ...read file to a string. */
+        tmp = styleIn.readAll();
+        file->close();
+        style->append(tmp);
+        emit splashMessages(QString(tr("Loading stylesheet %1")).arg(name));
+        qDebug()<<"Loaded stylesheet:" << name;
```
tracernz: Missing spaces around operator.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -375,6 +375,23 @@ void MainWindow::extensionsInitialized()
...
+    }
+    else {
+        qDebug()<<"Failed to openstylesheet file" << name;
```
tracernz: Missing spaces around operator.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -375,6 +375,23 @@ void MainWindow::extensionsInitialized()
     emit m_coreImpl->coreOpened();
 }
 
+void MainWindow::readStyleSheet(QFile *file, QString name, QString *style) {
+    QString tmp;
+    if(file->open(QFile::ReadOnly)) {
```
tracernz: Missing space after if

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -386,6 +403,7 @@ void MainWindow::loadStyleSheet(QString name) {
     directory.cd("share");
 #endif
     directory.cd("stylesheets");
+    QFile global(directory.absolutePath()+QDir::separator()+"global.qss");
```
tracernz: Missing spaces around operators.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -395,20 +413,15 @@ void MainWindow::loadStyleSheet(QString name) {
...
-        /* ...read file to a string. */
-        style = styleIn.readAll();
-        data.close();
-        /* We'll use qApp macro to get the QApplication pointer
-         * and set the style sheet application wide. */
+
+    readStyleSheet(&global, "global", &style);
+    readStyleSheet(&data, name, &style);
+
+    if(style.length() > 0) {
```
tracernz: Missing space after if.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -395,20 +413,15 @@ void MainWindow::loadStyleSheet(QString name) {
...
         qApp->setStyleSheet(style);
-        emit splashMessages(QString(tr("Loading stylesheet %1")).arg(name));
-        qDebug()<<"Loaded stylesheet:" << directory.absolutePath() << name;
+    } else {
+        qDebug()<<"No stylesheets loaded.";
```
tracernz: Missing spaces around operator.

ground/gcs/src/plugins/coreplugin/alarmsmonitorwidget.cpp
```diff
@@ -41,7 +41,7 @@ void AlarmsMonitorWidget::init(QSvgRenderer *renderer,QGraphicsSvgItem *graph)
     error_sym->setParentItem(graph);
     error_txt = new QGraphicsTextItem();
     error_txt->setDefaultTextColor(Qt::white);
-    error_txt->setFont(QFont("Helvetica",22,2));
+    error_txt->setFont(QFont("Helvetica",24,2));
```
tracernz: Bunch of missing spaces between args and operators in this file (please write new code to current standards unless it's ridiculously ugly to).

ground/gcs/src/plugins/coreplugin/connectionmanager.cpp
```diff
@@ -54,7 +54,7 @@ ConnectionManager::ConnectionManager(Internal::MainWindow *mainWindow, QTabWidge
 {
     QHBoxLayout *layout = new QHBoxLayout;
     layout->setSpacing(5);
-    layout->setContentsMargins(0,0,0,0);
+    layout->setContentsMargins(0,0,9,0);
```
tracernz: Missing spaces.

ground/gcs/src/plugins/coreplugin/telemetrymonitorwidget.cpp
```diff
@@ -46,7 +46,7 @@ TelemetryMonitorWidget::TelemetryMonitorWidget(QWidget *parent) : QGraphicsView(
     setWindowFlags(Qt::FramelessWindowHint);
 
     // This comes from the size of the image
-    QGraphicsScene *scene = new QGraphicsScene(0,0,1040,110, this);
+    QGraphicsScene *scene = new QGraphicsScene(0,0,1240,144, this);
```
tracernz: Bunch of missing spaces around args and operators in this file..

ground/gcs/src/plugins/coreplugin/telemetrymonitorwidget.cpp
```diff
@@ -216,7 +217,7 @@ void TelemetryMonitorWidget::resizeEvent(QResizeEvent* event)
 
     // This offset is required because the widget is not centered while
     // it is a child of the connection selector widget
-    graph->setPos(0,-20);
+    //graph->setPos(0,-20);
```
tracernz: In future you can just delete code rather than commenting it out unless it adds some explanation. :wink:

ground/gcs/src/plugins/boards_dtf/dtfcconfiguration.ui
```diff
@@ -757,7 +757,6 @@
  <resources>
   <include location="dtfuhf.qrc"/>
   <include location="../coreplugin/core.qrc"/>
-  <include location="../coreplugin/core.qrc"/>
```
tracernz: I've stopped bothering with fixing this. It doesn't affect UIC so may as well not fight it. It's fine to leave it removed here though.

ground/gcs/src/plugins/systemhealth/systemhealthgadgetwidget.cpp
```diff
@@ -138,7 +138,6 @@ SystemHealthGadgetWidget::~SystemHealthGadgetWidget()
 
 void SystemHealthGadgetWidget::setSystemFile(QString dfn)
 {
-    setBackgroundBrush(QBrush(Utils::StyleHelper::baseColor()));
```
tracernz, line 141: You are defeating the setting on the main GCS settings page here. I think we should either remove the setting and force it everywhere, or leave it as-is and change the color in the default configs rather than end up with a hodge-podge of different things.

ground/gcs/share/stylesheets/global.qss
```diff
@@ -35,3 +35,9 @@ MyTabWidget > QTabBar::tab {
 	max-height: 36px;
 
 }
+
+SystemHealthGadgetWidget {
+	
+	background: #303030;
+
+}
```
tracernz: Missing newline.

ground/gcs/src/plugins/lineardial/lineardialgadgetwidget.cpp
```diff
@@ -153,7 +153,6 @@ void LineardialGadgetWidget::updateIndex(UAVObject *object1) {
 void LineardialGadgetWidget::setDialFile(QString dfn)
 {
     QGraphicsScene *l_scene = scene();
-    setBackgroundBrush(QBrush(Utils::StyleHelper::baseColor()));
```
tracernz, line 156: As per the other commit.

ground/gcs/share/stylesheets/global.qss
```diff
@@ -37,7 +37,7 @@ MyTabWidget > QTabBar::tab {
 }
```
tracernz: Any reason for all the extra newlines? :stuck_out_tongue:

tracernz:
Overall this is pretty good. There is an issue with overriding the user's preferences.

I would much prefer to keep all new code in current style (https://wiki.qt.io/Qt_Coding_Style) unless there's a compelling reason to do otherwise.

glowtape:
As far as the user preferences in regards to colors go, how many users do actually change these? Do we even want users to change these? That setting only affects a small part of the whole UI, the larger rest of it defers to the system colors anyway, so I guess we can just do away with the setting and enforce the dark colors in the patch?

tracernz:
So as I see it, I hate apps that try to force non-native colours. Allowing the user to change the only non-native bits is the compromise.

If you really want to remove that, please do so, don't leave it half-in, half-out IMO. We already have too many half-working things. :/

ground/gcs/src/plugins/systemhealth/systemhealthgadgetwidget.cpp
glowtape: I gave it a try and I think I'm gonna strip it. A lot of the UI in the GCS uses system colors anyway, and there are some widgets, that are probably supposed to use it, that actually don't and use their own static colors. I ran for instance into the PFD and OPMap doing this.

This is going to take some time, because I have to remove StyleHelper and see what breaks.

tracernz: Okay. Let me know if any issues and I can (hopefully) help a bit.

glowtape:
OK, I've removed the color button, StyleHelper because it was kinda useless after that, and also stripped ManhattanStyle, because it relied on StyleHelper and didn't want to fix it. Not that it affected the look noticeably to begin with. I've spun Windows and Ubuntu builds, and it still looks and works the same (seemingly), apart from the menu bar taking the system style now. Not sure about OSX, I don't have that, but I'm hopeful.

I still need to prettify the dropdown buttons in Edit Gadgets mode, but that's a small fix that can be done independently of this, so that I'm not camping the GCS source any longer.

glowtape:
AFAIK I am finished with this. I've worked around the remaining styling issues that showed up stripping ManhattanStyle. Anything else UI-wise is for some point later in time.

mlyle:

mlyle:
Looks pretty good on Linux.

mlyle:
On OSX, things are generally pretty good, but i see a couple of things

Edit gadget mode 'x' looks a little wonky.
image

Ideally the other widgets on the bottom bar would be centered vertically (with respect to the telemetry stats widget).

image

mlyle:
Looks gorgeous on Windows.  My only complaint (may not be regression) is the way flighttime is cut off (and possibly flight mode is nudged downwards in its container).

image

Thanks for putting this changeset together --- I really look forward to merging it.  I'm sorry UI stuff is so niggling/such a pain.

glowtape:
The Flight Time thing seems to be an issue with HiDPI mode, if I had to guess. How I understand that dial code, the text is supposed to take the size of the placeholder elements in the SVG, but it clearly doesn't in that screenshot. The Flight Mode being off center seems related, too. Not sure I can fix this on the quick. I'll probably have to figure out how to get the screen DPI within Qt and scale font sizes opposite to fight this.

The opposite seems to happen on OSX, the text on the stats widget is supposed to be larger than shown, it doesn't scale with retina. The letters are supposed to be roughly as tall as the bars. And for some reason the _SVG_ is _pixelated_?

As far as the widgets on the status bar go, how does it look on Windows with HiDPI? Over here on 96dpi it's centered in Windows as well as Ubuntu. If it's centered in Windows HiDPI but not in OSX, I'm at a loss as to why. (Wait, no, there's a reason why I keep complaining that Qt is inconsistent.) I can apply different offsets for OSX, question is how that affects non-Retina displaying of the app.

I'll look into overriding the colors of the X button to be the same on all platforms.

Here's Windows 10 and Ubuntu with 96dpi for comparison:

gcsui

tracernz:
http://doc.qt.io/qt-5/qfont.html#setPixelSize might be useful to fix this.

mlyle:
It could, but IMO we should just transform the text bounding box over the element we want.

In any event, I don't think it's a regression so it shouldn't block this.  

> I can apply different offsets for OSX, question is how that affects non-Retina displaying of the app.

Hopefully we shouldn't need **offsets**, can't it be pinned to the middle?  It's OK if it waits for a subsequent PR-- I want to get this landed.

tracernz:
Let's merge this because it's too big (smaller PRs are nicer PRs). Any regression/deficiencies will need to be dealt with in subsequent PRs.

tracernz:
Not sure I like the red hatching very much. It's __very__ dominant. Darker colour and/or lower opacity would make it better I think. It might be better without it though too..

image

tracernz:
Actually, I think it looks a lot better without the hatching. WDYT @glowtape?

image

image

glowtape:
Personally I've rather liked it, that's why it has that color. An argument for a gray shade instead can be made for the sake of unobtrusiveness. I wouldn't remove the crosshatch completely though. Looks nice visually, and it has that general flight deck stuff sort of look.

mlyle:
I have a weak preference for the barber-pole because it looks plane-y.

My wife says it is **obnoxious**, but she doesn't get much of a vote.

So do whatever, I guess. :D

glowtape:
Grayscale good compromise? :D

sysh2

sysh1

tracernz:
Yes, I like that, thanks. :+1: It's just the red makes it look like there's something majorly wrong IMO..

PR #1670, opened 2017-03-27
GCS:Wizard: Correct connection dropdown and button state

tracernz:
Before auto-connected devices would leave the connect button in the wrong state, wrong device would be selected etc.

Also don't show devices that can't be setup through wizard (because that would require a bunch more pointless code to decide if the connect button should be enabled depending on selected device). Before they were shown disabled in the list using __undocumented API__, but they could still be selected (probably why the API is undocumented, because it's incomplete/unintended) due to no valid devices being available.

Made some APIs a bit nicer in the process.


tracernz:
Oops, an unrelated change slipped in, force-pushed.

tracernz:
This would prevent: http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJCwpf0KDA. But we dropped Naze completely now anyway :P:

mlyle:
I wish you could re-open merged stuff :P 

Yes, this would be really nice to have.  It just broke all telemetry last time around.

tracernz:
Yup, coming soon. :grin:

PR #1676, opened 2017-03-28
linux: put simulator (both x86 and pi) in packages

mlyle:
Requires a jenkins change (already made) to run `PI_CROSS_SIM=yes make -j3 sim` on master.

### DIFF
--- a/package/Makefile.linux
+++ b/package/Makefile.linux
@@ -63,9 +63,9 @@ gcs:
 
 ground_package_os_specific: ground_package_common usb_id_udev
 	$(V1) cp -v $(BUILD_DIR)/shared/usb_ids/dronin.udev $(PACKAGE_DIR)/rules.udev
-	$(V1) mkdir -p $(PACKAGE_DIR)/sim
-	-$(V1) cp -v $(BUILD_DIR)/sim-pi/sim.elf $(PACKAGE_DIR)/sim/sim-pi.elf
-	$(V1) cp -v $(BUILD_DIR)/sim/sim.elf $(PACKAGE_DIR)/sim/sim.elf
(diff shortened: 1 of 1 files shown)


mlyle:

mlyle:

mlyle:
(Ran `sudo apt-get install gcc-arm-linux-gnueabi` to get toolchain on main build server).

mlyle:
(Ran `sudo apt-get install gcc-arm-linux-gnueabihf` to get **correct** toolchain on main build server)

mlyle:
~~jenkins, build artifacts please     this time please work~~ cancelled-- false alarm, build 1014 looks like it 's ok... looked at wrong build output.

mlyle:

xfce:
@mlyle only ef_flyingpio.hex in the firmware dir, not find flyingpio.bin and sim.elf

xfce:
@mlyle ok, sim.elf found, but howto convert ef_flyingpio.hex to flyingpio.bin , can the PI's scripts handle hex file?

mlyle:
... Yes.

tracernz:
I don't like that it goes in the ground package. Is there any reason we can't add it to https://github.com/d-ronin/dRonin/blob/next/package/Makefile#L142 ? Maybe add a flight_package_os_specific in the same way ground does and stub it out in Windows/mac Makefiles?

mlyle:
So perhaps **strictly** the pi one doesn't belong in the ground package, but I'd argue that the x86 one, that will never fly anything belongs in the ground package.  (And belongs in the Windows/Mac ones when we're a bit further along).

We'd talked about that if qt3d settles down that it's my hope to let people try out rates, etc, in GCS-- that requires getting the right platform for these.  And if we move (as bugs are open for) to building flight only once for release, that becomes problematic...

tracernz:
Makes sense.

PR #1683, opened 2017-03-31
gcs: remove HITL

mlyle:
It doesn't work on current version of the simulator protocols, and may
not work in general for other reasons.  Its also an unfortunate lump of
code to

It's also **very** limited for testing flight controllers due to low
update rates and lack of any antialiasing filtering in the network
protocols.

I think a good next step could be a simulator sensor module that speaks
flitegear or xplane protocols.  But even that will only really be
reasonable for checking leisurely plane scenarios.

The project has outgrown this-- both in maturity of development and in
our own simulator (though fairly inaccurate, it is fast and really tests
flight code).


mlyle:
Seemed like a good time to do this, with the pending reformat.

Issue #1681, opened 2017-03-30
Report from IRC that seppuku on current next gets sensor errors with NTSC


mlyle:
user Verris.  Doesn't happen with Artifice, happens with current next.

PAL is fine for him and me.

mlyle:
I've purchased the camera in question--- http://www.banggood.com/Eachine-OKKAN-EK1119-1100TVL-13-CMOS-Sensor-150-Degree-2_3mm-M8-Lens-Tiny-FPV-Cam-p-1085641.html?rmmds=myorder  so I can scope it.

mlyle:
Blocker removed because it doesn't affect other NTSC users.

mlyle:
Cannot repro with the camera in limited testing

mlyle:
Cannot repro with more extensive testing.

PR #1640, opened 2017-03-17
pios_video: change suspected bugs

mlyle:
I am unable to fully verify this-- or test this at all-- this moment, but I've noticed several suspect things in pios_video in code audit, that could do really bad things if hsync is toggling too much (for no good reason)

Also, it's possible to filter the video hsync, so let's try enabling it.

Fixes #1588

PR #1685, opened 2017-04-01
pios_video: correct PIOS_VIDEO_INPUT_FILTER

mlyle:
bad news: i'm a moron
good news: it's very likely this is a cause of #1681 

Basically, #1640 had two changes:

1. stop using a line counter with various dubious logic [this one is good]
2. attempt, on Seppuku, to use an edge filter on hsync for robustness when colorburst toggling gets through on the hsync line.   I messed this one up, and programmed the bits to the wrong register.  The net result was it had no effect.

I've tested this on PAL and it is good.  Feedback from people with NTSC problems (@litterbug ?   @pug398 ?)  would be much appreciated

### DIFF
--- a/flight/PiOS/STM32F4xx/pios_video.c
+++ b/flight/PiOS/STM32F4xx/pios_video.c
@@ -108,8 +108,6 @@ static uint16_t num_video_lines = 0;
 static enum pios_video_system video_system_act = PIOS_VIDEO_SYSTEM_NONE;
 static const struct pios_video_type_cfg *pios_video_type_cfg_act = &pios_video_type_cfg_pal;
 
-static uint16_t line_int_flag;
-
 // Private functions
 static void swap_buffers();
 static void prepare_line();
@@ -212,10 +210,10 @@ void PIOS_Line_ISR(void)
 	 *   responsible for counting lines.
 	 * - When DMA engine is done, it re-enables us and we keep counting.
 	 */
-	if(TIM_GetITStatus(dev_cfg->hsync_capture.timer, line_int_flag))
+	if(TIM_GetITStatus(dev_cfg->hsync_capture.timer, TIM_IT_Update))
 	{
 		TIM_ClearITPendingBit(dev_cfg->hsync_capture.timer,
-				line_int_flag);
+				TIM_IT_Update);
 
 		if (active_line > 10000) {
 			// Don't wrap.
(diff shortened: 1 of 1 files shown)


pug398:

mlyle:
sorry, i forgot to ask for a build :/  CI was broken until just now anyways heh.

pug398:
Works good for me on the bench with no sensor warnings/resets. The only thing I can force it to do on occasion is to lose vertical lock and that is because my camera output changes sync level with high contrast change. Don't think there is much you can do in code about that.

mlyle:
@pug398 --- thanks for the test.  I'm glad it works for you.  Just one small point of clarification-- is there any difference that you saw between this and the previous changeset/current next)? 

Verris from IRC indicates he's going to be able to test later, so it'll be interesting to see if it cures his problem.

mlyle:
Two changes, now.  1) Saw a small video quality regression, so made the filter less aggressive (but still low enough frequency to reject colorburst) 2) Verris's problem wasn't fixed, so attempt to manage rate of line counter interrupt by ensuring the timer takes *nearly* as long as a line to wrap/update.

pug398:
@mlyle actually the fix before this one took care of most of my issues and there wasn't any real difference. I Will try the latest incarnation later today.

mlyle:
This still doesn't fix Verris's problem.  It's a good change, though.

mlyle:

mlyle:
@litterbugs curious if this fixes your problem (&verris's)

LitterBugs:
@mlyle This does appear to have fixed the OSD startup issue.

LitterBugs:
One other thing I have noticed about the Seppuku OSD config is that the default NTSC Black level of 20 appears to be set too low.  The black background around the OSD elements of the screen seems to unsettle the video sync as a whole.  Depending on what display I am using (LCD Panel or Goggles) this shows up differently one Tears and ghosts, the other the Vsync jumps on the whole video image.  Bumping it up to 35 like PAL is set stabilizes the signal.

LitterBugs:
One last thought.  With all the recent changes, it is no longer necessary to use the y offset.  Image appears centered on both my ground station and goggles

mlyle:
@litterbugs -- thanks for the test and the feedback.  Unfortunately it's hard with current OSD infrastructure to set per-board levels, though I may be able to provide a small nudge that way for Seppuku.

LitterBugs:
@mlyle No worries, it is easy enough to adjust.  Couldn't remember if I'd said anything about that previously.  Remembered it when setting everything back to defaults for testing the other night.  I can't remember if it is just the Pre-release board or if the final board does it too.  Could also be a combination of the board and VTX.  I haven't hooked my released board up through a VTX yet.

pug398:
@LitterBugs I run NTSC black level from 35 to 50 to get rid of various artifacts so it is not just you. I do have to set y offset in some cases still.

mlyle:
#1692 offsets the black--- and white--- levels +12.

tracernz (approved):
:grin:

PR #1686, opened 2017-04-01
Improvements to GPS widget in GCS

### LINKED ISSUE #1698: Strange size of textbox on System tab
* Seen in version (release version or githash): a08a62cd

Issue details:
It seems to be overlapping some text?

image

mlyle:
1. Remove the message box part of the widget which is no longer used (would show NMEA sentences from a direct-connected GPS).
2. Improve the spacing and size of subwidgets-- make the graphical elements bigger.  Made a slightly bigger and better world map pixmap (which happens to be a smaller file).
3. Make the SNR bar graph widget useful.  Only show the first 18 satellites so that it is possible to read captions and see the bars-- if you have more than 18 satellites tracked you probably don't care about the last few lines :P
4. Adjust scaling of the data from flight.  The GPS widget assumes a opaque "1-99" SNR value which is not what UBX GPSes deliver; adjust our value so it's a bit more like NMEA.

Fixes #1698


mlyle:
image

mlyle:
Couldn't resist making another change--color coding constellations.

image

Yellow GPS, Red "Other" (Commie GLONASS satellites but also Beidou/Galileo in the future), Orange-y SBAS/differential

mlyle:
OK I am done tweaking this for reals :P

image

mlyle:
image

OK that's it for real.  I will stop lying.

tracernz (approved):
Very nice. :grinning:

PR #1688, opened 2017-04-02
Stronger filtering, and accounting for this in OG autotune.

glowtape:
This a proposed patch, in case autotune-ng doesn't make it to Neat. I've seen @mlyle ask on IRC whether to focus on it or actuator improvements (didn't these rely on the new data anyway?), so if former is pushed up, this might be worth a consideration.

Before, the gyro input to autotune were raw samples, and therefore the EKF filter did some modicum of filtering to account for that. Now the gyro samples are filtered globally before they go anywhere, so you can in principle turn down the filtering, this is what I did in this patch. It looks like without this reduction, OG autotune overestimates gains and underestimates tau, resulting in some weak PIDs.

The default filtering was also bumped to second order and 85hz cutoff. That should match the group delay in the important frequency range to before.

In the past, my quad in this combination tuned to roughly 16ms on the old code, with these changes it's ~18ms (versus ~22ms without the tweak to autotune.c), altho I've used 80hz cutoff not 85hz. The resulting PIDs are reasonable.

This is still out for proper flying, and I guess testing on other quads, however my backyard antics were successful. Interestingly, my ever present prop oscillations during hangtime seem to have finally completely gone away (--edit: hrm, still show up for my quad blade props, probably this problem on top of it https://www.youtube.com/watch?v=LMeJ7AUCJcY ).

Tune used outside:
http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAkLel_QsM

### DIFF
--- a/flight/Modules/Autotune/autotune.c
+++ b/flight/Modules/Autotune/autotune.c
@@ -612,7 +612,7 @@ __attribute__((always_inline)) static inline void af_predict(float X[AF_NUMX], f
 	const float q_B = 1e-6f;
 	const float q_tau = 1e-6f;
 	const float q_bias = 1e-19f;
-	const float s_a = 150.0f; // expected gyro measurement noise
+	const float s_a = 50.0f; // expected gyro measurement noise
 
 	const float Q[AF_NUMX] = {q_w, q_w, q_w, q_ud, q_ud, q_ud, q_B, q_B, q_B, q_tau, q_bias, q_bias, q_bias};
 
--- a/shared/uavobjectdefinition/sensorsettings.xml
+++ b/shared/uavobjectdefinition/sensorsettings.xml
@@ -36,10 +36,10 @@
 				<option>TRUE</option>
 			</options>
 		</field>
-		<field name="LowpassCutoff" units="Hz" type="float" elements="1" defaultvalue="55">
+		<field name="LowpassCutoff" units="Hz" type="float" elements="1" defaultvalue="85">
 			<description>Cutoff frequency for the lowpass on both the gyroscopes and accelerometers.</description>
 		</field>
-		<field name="LowpassOrder" units="" type="uint8" elements="1" defaultvalue="1">
+		<field name="LowpassOrder" units="" type="uint8" elements="1" defaultvalue="2">
 			<description>Order of the lowpass filter. Maximum 8, a value of zero bypasses the filter.</description>
 		</field>
 		<access gcs="readwrite" flight="readwrite"/>


mlyle:
> The default filtering was also bumped to second order and 85hz cutoff. That should match the group delay in the important frequency range to before.

We didn't do this (yet?)

I guess I'm a bit skeptical about this PR.  It sure takes a lot of characterization to make sure this is a good idea (I'm not sure we have even enough "study power" if everyone tests it to know if it converges the same amount of times / produces equivalent PIDs.  In any case, with the filtering the online autotune model is now wrong-- and I'm not sure fudging the gyro noise parameter is a way to fix it.

mlyle:
Oh, now i see the first commit / that you were doing the 55->85Hz in this PR.   I like this better, but I don't know that we've proven conclusively this is a good idea yet.

glowtape:
Nah, I've set the defaults in the filter patch back then to match what we had before, first order filtering at 55hz.

The issue however is that the input to the autotune EKF is different (not as noisy, plus delayed) and that the filtering on that side could be turned down a little. Back during the initial testing on the RE1 of mine, tau increased by 4-6ms and gains went up for no reason, even just with 1st order 55hz filtering, this resulted in really tame PIDs (around 2/3rd the P- and I-terms of what it tuned to before).

Lowering the s_a in autotune.c seems to bring results back down to where they are "supposed" to be. To no apparent ill effect. Obviously, it'd be nice to have more samples, but I only have one quad and two types of props here.

As far getting a conclusion, it'd be interesting if someone comes onto IRC again and complains about getting only crappy autotunes, prime target to try this on. Other than that, prod anyone on IRC to try it.

That said, if autotune-ng gets into this release, large part of this is moot, anyway.

mlyle:
> Back during the initial testing on the RE1 of mine, tau increased by 4-6ms and gains went up for no reason, even just with 1st order 55hz filtering, this resulted in really tame PIDs (around 2/3rd the P- and I-terms of what it tuned to before).

This is arguably correct, because IMO damping should be around 1.05 and noise sensitivity a little higher-- this seems to do best on the measurements from neo-autotune and indeed damping of just barely-over-unity is desirable.

So in other words, old-autotune usually measured **too aggressive** of constants (low system gain, low tau) as shown by autotune partition data and other measurements---  I believe the chosen parameters for PID calculation were fudges around this.

glowtape:

jhitesma:
Gave it a try on my piko.  Results here:
http://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAkIXQwQgM

Compare to previous results from new tune, and results from before that on current og tune.

henn1001:
Tried out autotune on current next - result was pretty under tuned on both of my quads. Tau went up by ~8ms and it had problems holding attitude in fff when surging throttle. I´ll try this patch once the weather clears up again.

glowtape:
Closed. Autotune-ng will likely make it to Neat.

PR #1690, opened 2017-04-03
flight library: build only once per arch

mlyle:
1. Saves a little bit of build time now, especially on Windows.
2. Allows saving a lot more if we add big libraries.
3. Takes out a bunch of per-target configuration of build files.

Also remove example.c from rscode source and move Simulation to INS14, like all other targets.


tracernz:
We should have basic CFLAGS (warning settings etc.) in a single .mk somewhere that all the firmware can include. :wink:

tracernz (approved):
Don't see any issues. Do some of the libraries include pios.h?

mlyle:
Yes.  For these reasons:

* PIOS_Assert (tolerable)
* PIOS mutexes, mallocs (tolerable)
* To bring in UAV objects (yuck)
* PIOS_COM in frsky packing & MSP (the worst IMO)

Nothing sensitive to PIOS_INCLUDE in there.

PR #1691, opened 2017-04-03
Initial support for SP Racing F3 Evo and FrSky XSRF3E

E1yot:
I added initial support for SP Racing F3 Evo and FrSky XSRF3E.
On board LED is heartbeat. IR is configured as alarm LED.
Baro can be switched on/off (for clone boards)
Mag can be switched between internal, none and external HMC (for clone boards)

Todo:
- connection diagrams
- maybe WS2812 support
- maybe sdcard Support

Should be reviewed and tested, also on clone boards.


tracernz:
What's the copyright/license situation for the board image?

yamahabest:
The "ground/gcs/src/plugins/boards_dronin" path seems to be incorrect, as this board is not produced by dRonin?

flight/targets/sprf3e/board-info/board-info.mk
```diff
@@ -0,0 +1,34 @@
...
+BOARD_REVISION      := 0x02
+BOOTLOADER_VERSION  := 0x87
+HW_TYPE             := 0x00		# seems to be unused
+
+CHIP                := STM32F303CCT
+BOARD               := STM32F30x_sprf3e
+MODEL               := HD
+MODEL_SUFFIX        := 
+
+USB_VEND            := "dRonin"
```
yamahabest: Shouldn't be dRonin.

mlyle:
Elyot:  Thank you for this PR.  I have not had a chance to look at this yet-- it will probably be a few days.

tracernz:
I hope to have a good look on Friday. :+1:

mlyle:
jenkins, test this please

shared/uavobjectdefinition/hwsprf3e.xml
```diff
@@ -0,0 +1,142 @@
...
+				<option>MavLinkTX</option>
+				<option>MavLinkTX_GPS_RX</option>
+				<option>MSP</option>
+				<option>OpenLog</option>
+				<option>Storm32Bgc</option>
+				<option>Telemetry</option>
+				<option>TBS Crossfire</option>
+			</options>
+		</field>
+		<field name="Uart2" units="function" type="enum" elements="1" parent="HwShared.PortTypes" defaultvalue="S.Bus">
```
yamahabest: I believe the default value is always "Disabled" in all hw definitions.

shared/uavobjectdefinition/hwsprf3e.xml
```diff
@@ -0,0 +1,142 @@
...
+				<option>MavLinkTX</option>
+				<option>MavLinkTX_GPS_RX</option>
+				<option>MSP</option>
+				<option>OpenLog</option>
+				<option>Storm32Bgc</option>
+				<option>Telemetry</option>
+				<option>TBS Crossfire</option>
+			</options>
+		</field>
+		<field name="Uart3" units="function" type="enum" elements="1" parent="HwShared.PortTypes" defaultvalue="FrSKY SPort Telemetry">
```
yamahabest: I believe the default value is always "Disabled" in all hw definitions.

shared/uavobjectdefinition/hwsprf3e.xml
```diff
@@ -0,0 +1,142 @@
+<?xml version="1.0"?>
+<xml>
+	<object name="HwSprf3e" singleinstance="true" settings="true" category="HardwareSettings">
+		<description>Selection of optional hardware configurations.</description>
+		<field name="Uart1" units="function" type="enum" elements="1" parent="HwShared.PortTypes" defaultvalue="Disabled">
+			<description/>
```
yamahabest: Isn't it possible to describe where the port is located ("Left of/Right of/Next to") ?

shared/uavobjectdefinition/hwsprf3e.xml
```diff
@@ -0,0 +1,142 @@
...
+				<option>MavLinkTX_GPS_RX</option>
+				<option>MSP</option>
+				<option>OpenLog</option>
+				<option>Storm32Bgc</option>
+				<option>Telemetry</option>
+				<option>TBS Crossfire</option>
+			</options>
+		</field>
+		<field name="Uart2" units="function" type="enum" elements="1" parent="HwShared.PortTypes" defaultvalue="S.Bus">
+			<description/>
```
yamahabest: Isn't it possible to describe where the port is located ("Left of/Right of/Next to") ?

shared/uavobjectdefinition/hwsprf3e.xml
```diff
@@ -0,0 +1,142 @@
...
+				<option>MavLinkTX_GPS_RX</option>
+				<option>MSP</option>
+				<option>OpenLog</option>
+				<option>Storm32Bgc</option>
+				<option>Telemetry</option>
+				<option>TBS Crossfire</option>
+			</options>
+		</field>
+		<field name="Uart3" units="function" type="enum" elements="1" parent="HwShared.PortTypes" defaultvalue="FrSKY SPort Telemetry">
+			<description/>
```
yamahabest: Isn't it possible to describe where the port is located ("Left of/Right of/Next to") ?

shared/uavobjectdefinition/hwsprf3e.xml
```diff
@@ -0,0 +1,142 @@
...
+				<option>IBus</option>
+				<option>TBS Crossfire</option>
+			</options>
+		</field>
+		<field name="USB_HIDPort" units="function" type="enum" elements="1" parent="HwShared.USB_HIDPort" defaultvalue="USBTelemetry">
+			<description/>
+		</field>
+		<field name="USB_VCPPort" units="function" type="enum" elements="1" parent="HwShared.USB_VCPPort" defaultvalue="Disabled">
+			<description>Virtual COM Port protocol</description>
+			<options>
```
yamahabest: I believe this whole options node can be removed.
The parent defines also MSP, which is now excluded by specifying this list.

shared/uavobjectdefinition/hwsprf3e.xml
E1yot: For FrSky XSRF3E this is hard wired, so I think to change the default is a good idea. But set it to Disabled is also an option.

shared/uavobjectdefinition/hwsprf3e.xml
E1yot: For FrSky XSRF3E this is hard wired, so I think to change the default is a good idea. But set it to Disabled is also an option.

shared/uavobjectdefinition/hwsprf3e.xml
yamahabest: Oh, that's for one of the 2 boards?

E1yot: Okay, I change it to disabled.

tracernz: We don't currently have a way to conditionally override defaults but it would sure be nice for a lot more than just this. I wonder if there's a way to tell the difference between SPRF3 and the Frsky board (any pinout variation?)?

E1yot: XSRF3E has no SDCARD and only 6 PWM-out. I2C is not available on connector. RSSI comes over S.BUS, RSSI-ADC-Pin is not available. The missing features are not implemented on board, but trying to access them will have no influence. Setting up a servo on PWM8 will work, but you could not connect it without soldering the STM-Pin. I found no safe way to detect the correct board. Testing for S.Bus is no option. We could messure VBat. On XSRF3E with USB-Power only you will read 4.xx V. SP Racing F3 Evo gives 0V. But if VBat is connected (1S-Lipo or BEC), detection fails.
I think, the user knows, which board he is setting up and how he has to config the UARTS, ...

flight/targets/sprf3e/bl/pios_board.c
```diff
@@ -0,0 +1,88 @@
+/**
+ ******************************************************************************
+ * @addtogroup Bootloader Bootloaders
+ * @{
+ * @addtogroup spf3e SP Racing F3E support files
```
tracernz: These need to be consistent through all the files. The first word is the group identifier used internally in Doxygen to link all the group members (technically you can define the group title once and just add things with the identifier but that doesn't really work for us because we don't have a central place to define all the groups), and the remainder is the title shown in the documentation, including the navigation tree. i.e.
image
"SP Racing F3E support files" is way too long for there IMO. We have used uppercase identifiers through the rest of the boards (IIRC).

Can you please make them all ` * @addtogroup SPRF3 SP Racing F3 Evo`

flight/targets/sprf3e/board-info/board_hw_defs.c
```diff
@@ -0,0 +1,1189 @@
...
+    { TIM2,  IO_TAG(PA1),  TIM_Channel_2, TIM2_IRQn,               1, IOCFG_AF_PP, GPIO_AF_1},  // PWM2
+    { TIM15, IO_TAG(PA2),  TIM_Channel_1, TIM1_BRK_TIM15_IRQn,     1, IOCFG_AF_PP, GPIO_AF_9},  // PWM3
+    { TIM15, IO_TAG(PA3),  TIM_Channel_2, TIM1_BRK_TIM15_IRQn,     1, IOCFG_AF_PP, GPIO_AF_9},  // PWM4
+    { TIM3,  IO_TAG(PA6),  TIM_Channel_1, TIM3_IRQn,               1, IOCFG_AF_PP, GPIO_AF_2},  // PWM5
+    { TIM3,  IO_TAG(PA7),  TIM_Channel_2, TIM3_IRQn,               1, IOCFG_AF_PP, GPIO_AF_2},  // PWM6
+    { TIM3,  IO_TAG(PB0),  TIM_Channel_3, TIM3_IRQn,               1, IOCFG_AF_PP, GPIO_AF_2},  // PWM7
+    { TIM3,  IO_TAG(PB1),  TIM_Channel_4, TIM3_IRQn,               1, IOCFG_AF_PP, GPIO_AF_2},  // PWM8
+*/
+
+static const struct pios_tim_channel pios_tim_servoport_v01_pins[] = {
```
tracernz: This doesn't appear to be used anywhere?

flight/targets/sprf3e/board-info/board_hw_defs.c
```diff
@@ -0,0 +1,1189 @@
...
+				.DMA_Priority           = DMA_Priority_High,
+			},
+		}
+	},
+	.half_flag = DMA2_IT_HT1,
+	.full_flag = DMA2_IT_TC1,
+	.oversampling = 32,
+	.adc_pin_count = 3,
+	.adc_pins = {
+		{GPIOA,GPIO_Pin_4,ADC_Channel_1,true}, // Voltage
```
tracernz: Missing spaces in these lines.

flight/targets/sprf3e/fw/pios_board.c
```diff
@@ -0,0 +1,479 @@
...
+
+
+#if defined(PIOS_INCLUDE_DEBUG_CONSOLE)
+#define PIOS_COM_DEBUGCONSOLE_TX_BUF_LEN 40
+uintptr_t pios_com_debug_id;
+#endif	/* PIOS_INCLUDE_DEBUG_CONSOLE */
+
+uintptr_t pios_com_aux_id;
+uintptr_t pios_com_telem_rf_id;
+uintptr_t pios_com_can_id;
```
tracernz: I don't think this board has CAN?

flight/targets/sprf3e/fw/pios_board.c
```diff
@@ -0,0 +1,479 @@
...
+uintptr_t pios_uavo_settings_fs_id;
+uintptr_t pios_internal_adc_id;
+uintptr_t pios_can_id;
```
tracernz: As above.

flight/targets/sprf3e/fw/pios_board.c
```diff
@@ -0,0 +1,479 @@
...
+#include "pios_bmp280_priv.h"
+  uint8_t hw_baro;
+  HwSprf3eBarometerGet(&hw_baro);
+    switch(hw_baro) {
+        case HWSPRF3E_BAROMETER_BMP280:
+        	if (PIOS_HAL_ConfigureExternalBaro(hw_baro,
+			       &pios_i2c_internal_id, &pios_i2c_internal_cfg))
+      			PIOS_SENSORS_SetMissing(PIOS_SENSOR_BARO);
+
+/*        	if ((PIOS_BMP280_Init(&pios_bmp280_cfg, pios_i2c_internal_id) != 0)
```
tracernz: Why is this commented out?

flight/targets/sprf3e/fw/pios_board.c
```diff
@@ -0,0 +1,479 @@
+/**
```
tracernz, line 1: Please re-indent this whole file with __tabs__, it's currently a mixture (even single lines have a mixture in places).

flight/targets/sprf3e/fw/pios_config.h
```diff
@@ -0,0 +1,96 @@
...
+#define PIOS_INCLUDE_SPI
+#define WDG_STATS_DIAGNOSTICS
+
+/* Select the sensors to include */
+#define PIOS_INCLUDE_MPU9250_SPI
+//#define PIOS_MPU6000_ACCEL
+#define PIOS_INCLUDE_BMP280
+#define PIOS_INCLUDE_HMC5883
+#define PIOS_INCLUDE_HMC5983_I2C
+#define PIOS_TOLERATE_MISSING_SENSORS
```
tracernz: Do some of the targets not have all the sensors placed?

flight/targets/sprf3e/fw/pios_config.h
```diff
@@ -0,0 +1,96 @@
...
+#define PIOS_INCLUDE_SESSION_MANAGEMENT
+#define PIOS_INCLUDE_LIGHTTELEMETRY 
+
+
+/* Supported receiver interfaces */
+#define PIOS_INCLUDE_PWM
+
+
+/* Flags that alter behaviors - mostly to lower resources for CC */
+//??#define PIOS_INCLUDE_INITCALL           /* Include init call structures */
```
tracernz: You can delete this line, it's defined in pios_flight_config.h which you have included already.

ground/gcs/src/plugins/boards_dronin/boards_dronin.pro
```diff
@@ -18,7 +18,9 @@ HEADERS += \
     pikoblx.h \
     playuavosd.h \
     seppuku.h \
-    seppukuconfiguration.h
+    seppukuconfiguration.h \
+    sprf3e.h \
```
tracernz, line 22: What do you think about sprf3evo rather than sprf3e? I find it a lot clearer what board it's referring to ([there are a __lot__ of SPR boards all of a sudden?!](http://seriouslypro.com/)).

ground/gcs/src/plugins/boards_dronin/sprf3e.cpp
```diff
@@ -0,0 +1,287 @@
...
+    default:
+        break;
+    }
+    
+    return 2000;
+}
+
+QStringList Sprf3e::getAdcNames()
+{
+    return QStringList() << "V" << "A" << "RSSI";
```
tracernz, line 259: Does this match the silkscreen markings?

ground/gcs/src/plugins/boards_dronin/sprf3econfiguration.h
```diff
@@ -0,0 +1,56 @@
+/**
```
tracernz, line 1: Surprised you elected to do a custom configuration form. Very good. :+1:

tracernz (changes requested):
Looks pretty clean and well done. :+1: Couple of things to clean up.

- Uses the Seppuku SVGs for the config page and connection diagram though?
- Again, what's the license status on the SPRF3 picture?

shared/uavobjectdefinition/hwsprf3e.xml
tracernz: Thanks, these were good changes. :+1:

flight/targets/sprf3e/bl/pios_board.c
E1yot: SVGs have to be created for the new board. It's on the todo (see initial post). Seppuku SVGs are only a placeholder.
Board picture is already exchanged by my own photo from my own board. No license problems.

flight/targets/sprf3e/fw/pios_config.h
E1yot: BMP280 is placed on original SP Racing F3 Evo and XSRF3E. Where are clone boards without baro, and most of the brushed version clones do not have it.
HMC is as option for external mag. Clone boards use sometimes a MPU without internal mag. What's why I implemented the option for external mag. Also sometimes it is better, to place a GPS/Mag combo away from the other electronics. Original SP  Racing F3 Evo (and 1:1 clones) has I2C on separate connector, so it is easy to connect external mag. I tested external mag with HMC5883 breakout board.

ground/gcs/src/plugins/boards_dronin/boards_dronin.pro
E1yot: I named it SPRF3E since the FrSky version is named XSRF3E. Since Evo are the only "E"-boards, I see no problem in the short version. FrSky has also an XMPF3E, also with onboard F3 Evo. I do not have this board because of missing telemetry. But It should work like the XSRF3E.

ground/gcs/src/plugins/boards_dronin/sprf3e.cpp
E1yot: Yes, at least on original version

flight/targets/sprf3e/fw/pios_config.h
tracernz: Thanks for clarifying. I agree this is the right way to go. 👍

flight/targets/sprf3e/bl/pios_board.c
tracernz: > SVGs have to be created for the new board. It's on the todo (see initial post). Seppuku SVGs are only a placeholder.

Okay. If you want some assistance/advice on this let me know (IRC is usually best).

> Board picture is already exchanged by my own photo from my own board. No license problems.

Great. 👍

flight/targets/sprf3e/fw/pios_board.c
tracernz: P.S. I know some of the existing stuff is a little bit of a mess so I understand that's where it came from, but would be nice to do it right for new stuff. 😉

flight/targets/sprf3e/fw/pios_board.c
E1yot: At first, I implemented it as internal option. Icee gives me on IRC the hint of the clone boards. So I configured Baro as optional external sensor (it is connected to the same I2C, which is available on external connector). I will delete the commented out lines.

tracernz: I think you still want to init this even if some boards don't have it, but just don't panic if it fails to init. There should be an example of how to do this in the Revo target.

flight/targets/sprf3e/fw/pios_board.c
E1yot: I will do some cleanup.

ground/gcs/src/plugins/boards_dronin/boards_dronin.pro
tracernz: I'm okay either way so it will be fine as-is if you have a definite preference.

flight/targets/sprf3e/fw/pios_board.c
E1yot: Right, no CAN.

flight/targets/sprf3e/fw/pios_board.c
tracernz: We do have a tool to clean up flight code, to install it:
`make uncrustify_install`
Then to clean this file:
`make uncrustify_flight FILE=flight/targets/sprf3e/fw/pios_board.c`

I think it should fix these types of things.

flight/targets/sprf3e/fw/pios_board.c
E1yot: I take a look to the revo-code. But I have no board to test this. 
But I think, it should give an Baro-failure, but will boot up. In the config it is possible to disable baro and next boot where is no failure.

ground/gcs/src/plugins/boards_dronin/sprf3econfiguration.h
E1yot: This is a own family of boards, so it should have it's own configuration form. So we are possible to place some text hints for configuring or a warning like "experimental support" in the first time.

flight/targets/sprf3e/fw/pios_board.c
mlyle: I have been really naughty and haven't read this code yet, but note that BMP280_Init allocates memory unconditionally before failing.  It would be really nice to try the onboard init iff external is not  configured.

mlyle:
It looks like the F3E that I ordered just shipped from China, so I'll probably have a specimen in about 2 weeks to play with this in detail.

flight/targets/sprf3e/fw/pios_board.c
E1yot: Should the memory not be freed, if init fails? At the moment baro can be selected between BMP280 and none. If "none" is configured, I will not init the baro. If BMP280 is selected, I try to init the baro. So there should be no waste of memory? Since where is only one I2C, it is the same if the baro is onboard or external connected.

mlyle: We-- as is pretty common for embedded realtime systems-- don't have a free() operator.

OK, so it sounds reasonable to have it configurable.  Maybe default to having the baro selected, since a failure is harmless (other than wasted memory).

tracernz (approved):

tracernz:
jenkins, test this please

tracernz:
@mlyle Jenkins doesn't seem to have right perms setup, can you make it test this please? :wink:

mlyle:
jenkins, test this please

tracernz:
```
make[1]: *** No rule to make target '/var/lib/jenkins/jobs/dron/workspace/Nodes/master/flight/Libraries/taskmonitor.c', needed by '/var/lib/jenkins/jobs/dron/workspace/Nodes/master/build/fw_sprf3e/taskmonitor.o'.  Stop.
Makefile:975: recipe for target 'fw_sprf3e_tlfw' failed
```

jhitesma:
@tracernz  - Probably unrelated...but I got that make target on taskmonitor error for the aq32 target a few times this weekend and it went away after a make all_clean.   Different target but same error so I wonder if there's something related.

E1yot:
I see this error first time. I'm on Windows (7 and 10). Since I never touched something with taskmonitor, I think, this is not a problem of PR.

tracernz:
I think it's a problem resulting from @mlyle's re-organization (https://github.com/d-ronin/dRonin/pull/1690). You might need to rebase on the current next head to pull that in (the PR builder merges your branch onto next before building IIRC). https://github.com/d-ronin/dRonin/pull/1691/files#diff-10d9f57f046d3806c9c5c6b105751e96R112 is no longer correct.

tracernz:
jenkins, test this please

mlyle:
I finally received my SPRF3E! :D

@E1yot --- I added a commit to your branch that moves to pios_mpu from pios_mpu9250.  I've confirmed with the combination of #1724 and #1728 that the magnetometer appears to do the right thing.  Once they get merged to next, this branch should get rebased on top of that version of next.

I added a comment-- I need to look at it further, but the magnetometer switch statement appears to use two different enums so that probably is not correct.  I will spend some time looking at that configuration later.

mlyle:
jenkins, test this please

LitterBugs:
Are there currently any builds that I could be testing?  I have an SPR F3 EVO in one of my owls and it would be easy for me to give it a spin.

mlyle:
@litterbugs Cool!  I'll try and get something together in the next couple of days.  I can't just ask Jenkins for artifacts because this depends on a couple other PRs and because people are busy they've been sitting and not merged.

mlyle:
Now I can :)

mlyle:
Windows builder was stuck; i nudged it.  It has a backlog but hopefully by tomorrow AM artifacts are here.

mlyle:
The rev it had expected to build had disappeared in the meantime.

LitterBugs:
mmmm....  being fickle....

mlyle:
Boy is it.  Now, since a bunch of stuff was merged, a change was needed to the target to get it to build.

mlyle:
@e1yot -- if you get a chance could you please rebase this onto current next and force-push?     e.g. 
```
git fetch upstream
git rebase upstream/next
git push -f
```

mlyle:
@E1yot  --- never mind, I was surprised to see github let me force push to your repository. :D

mlyle:
jenkins, test this please

mlyle:
Once more for the test, because I rebased to the wrong point..  This PR is cursed ;)

jenkins, test this please

(Artifacts are OK)

LitterBugs:
Cursed is OK.  I am supposed to be receiving a test owl frame made of a different material for durability testing.....

Downloading.....  Should be able to test tonight

mlyle:
jenkins, test this please

mlyle:

LitterBugs:
Downloaded newest Artifacts....  I have  a GPS on this FC too....

LitterBugs:
Initial flash and configuration going good.  This Frame has the FC rotated 45 degrees CW.  When updating YAW in the attitude tab,  Does the MAG also get adjusted since this is an internal mag, Or does that need to be done elseware?  (seems to be off 45 degrees)

mlyle:
Mags respect boardrotation, yes.  Have you done sixpoint cal?  (Uncalib'd mags aren't very accurate)

LitterBugs:
Actually, maybe it is just off by 90 degrees and this has nothing to do with board orientation.  I have done 6 point accel/mag calibration.  Model moves correctly in the GCS Flight Data screen.

Results were:
Calibration succeeded
Accelerometer bias, in [m/s^2]: x=-0.0559771, y=0.0848847, z=0.326468 
Accelerometer scale, in [-]:    x=0.983497 , y=1.01333  , z=0.993735 

Magnetometer bias, in [mG]: x=-564.005 , y=-374.973 , z=-237.607 
Magnetometer scale, in [-]: x=1.61182  , y=0.49482  , z=1.08233  

ESCs calibrated,  Motors tested, RX tested.   Looking good.  Need to button up and will be able to get an autotune soon.

dronininitialtesting

owl2droninbuildout

LitterBugs:
Is the SDcard reader supported?

tracernz:
Not at this stage. It is a future possibility but don't wait for it.

LitterBugs:
Actually, it looks like GCS does not always represent the MAG orientation.  Think I've noticed this before on my Seppuku too.  Reads right in OSD, but GCS is????    Don't have FPV/OSD hooked up to this yet to compare

mlyle:
Note the red "needle" on the PFD is the magnetometer direction.  The actual magnetometer isn't used for heading directly (by flight or PFD/GCS)-- but it is slowly (over the timescale of a minute or so) pulled towards the red line.  This is because magnetometers are very noisy and prone to interference; so in the short term we "dead reckon" heading and apply small corrections.

Bench test inside can easily result in crazy heading values, both because the indoor magnetometer environment is wacky and because it's easy to get discontinuities in gyro readings (e.g. picking up and moving on multiple axes can easily walk heading off target from clipping, integration errors with multiple axes etc).

mlyle:
If that SMA in the back is for a vtx antenna, BTW, it will interfere with the GPS enough to not get good position hold (doesn't take much power from a nearby transmitter to deafen a GPS since the satellites are already underneath the noise floor).

LitterBugs:
@mlyle Good info as always.  I was using this frame and FC for LOS with iNav firmware previously with no FPV.   Not really planning on using navigation features on this frame/FC in conjunction with FPV, but may throw the GPS fields on OSD just for testing.  I have a 360mm frame that this FC will eventually end up on.  Everything checked so far was already hooked up ready to test and figured might as well throw my $.02 of testing in.  Have a lightweight variant of this frame coming from the manufacturer for "durability testing" and will be removing the GPS and any other extra weight when it arrives.

E1yot:
I got much better compass, if I set mag-bias and mag scale manual. But the (SPRF3E) board has a bad design. The sdcard slot with its metal cover is on the same place as mag, but bottom side of the board. This may influence measurement.

mlyle:

jenkins, test this please

mlyle:

RossMay:
Got it flashed onto https://us.banggood.com/Wholesale-Warehouse-Beecore-F3_EVO_Brushed-ACRO-Flight-Control-Board-DSM2-For-Inductrix-Tiny-Whoop-Eachine-E010-wp-Usa-1089021.html?rmmds=myorder but can't get the built in dsm to work. It should be ppm, but not getting any throttle response with the radio setup wizard.

mlyle:

flight/targets/sprf3e/board-info/board_hw_defs.c
```diff
@@ -593,6 +593,14 @@ static const struct pios_tim_clock_cfg tim_8_cfg = {
 			.NVIC_IRQChannelCmd                = ENABLE,
 		},
 	},
+        .irq2 = {
```
tracernz: spaces->tabs

mlyle: Woops, teaches me to take lines from another target :P

RossMay:
After latest build, no problems with ppm dsm2 built in. Everything seems to be working fine now. I'll be auto tuning in a few minutes and will share the results. I'll play around with it tomorrow and post an update.

mlyle:
It's really weird how Github kinda acts like this is building when it's not (the "yellow dot" next to default).  I think it's the required status check that makes it do it.

Jenkins, test this please

LitterBugs:
Removed GPS since that is where I plan on mounting the battery.  Successful autotune.  
http://dronin.org/tunes#?key=ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA0NLTywsM

Getting ready for some LOS fun in the backyard after adjusting rates.

EDIT: Oh yeah, upgraded to a more appropriate battery strap:
owl2droninredux

tracernz:
What's the status of this now? Waiting for @mlyle review. Still waiting on my board to arrive (and I'm out of town all next week :() so I'm okay if we merge it before I get to test.

E1yot:
The seppuku SVG's and connection diagram needs to be replaced, but I think, this can be done after merge, since it has no influence on function. At the moment I work on sdcard-support. But this should not be included in this PR.

mlyle:
I've flown this on two rigs and I'm pleased with it.  Code looks good, too.  Only blockers to merge are A) what's the license of the board image?  and B) ... ideally we'd have a conn dia before merge (what do you think @tracernz, is this a blocker?)

E1yot:
Board picture is already exchanged by my own photo from my own board. No license problems. (first commit after pr, see comment above)

LitterBugs:
Just swapped out frames, did another autotune, and all is well on my rig as well.  I don't think this Owl has EVER flown this good!  Thanks for adding support for this FC!

mlyle:
jenkins, test this please

mlyle:

mlyle (approved):

tracernz:
Looks like this is good enough to merge. 👍  After that you can do further work (if needed) in subsequent PRs @E1yot. 😁 Thanks again for all the work on this!

mlyle:
@E1yot thanks for all your patience and hard work :D

PR #1702, opened 2017-04-07
library work in anticipation of actuator work

mlyle:
* Put cmsis base in libraries, not PiOS
* Add DSP Lib 4.5.0 to CMSIS.  There is a newer DSP lib from CMSIS 5 which in some ways is favorable, and .. strictly I am not sure it needs to match the cmsis include version.  Buuuuuut., I figured it was better to be cautious.
* Adjust DSP Lib so it can build on non-ARM
* Build DSP lib matrix functions in flight libraries and simulator targets


tracernz (approved):
Looks reasonable.

tracernz:
Gonna fire a build on all platforms just to be safe.

PR #1704, opened 2017-04-07
gcs: Grey syshealth, minor UI fixes.

glowtape:
Plus override system colors for the close button on gadgets in edit mode.


ground/gcs/src/plugins/config/stabilization.ui
```diff
@@ -498,7 +498,7 @@
       <enum>QTabWidget::Rounded</enum>
      </property>
      <property name="currentIndex">
-      <number>2</number>
+      <number>0</number>
```
tracernz, line 501: Should just do the same as https://github.com/d-ronin/dRonin/pull/1574/commits/1b762d8c40b91fd12a283279336aafcf35c23eba IMO.

PR #1706, opened 2017-04-07
gcs: Use system-style groupboxes (RFC)

glowtape:
OSX builds don't use a custom style, so I removed it for the others to see how it looks. Personally I prefer the system ones, altho they're less contrasty.

As far as trying to make them more contrasty... Same story as always, that's been my bane with anything QT. Look the wrong way at attributes like border or anything, and it suddenly falls onto an apparently entirely different control renderer.

groupbox_w_old

groupbox_w_new

groupbox_u_old

groupbox_u_new

### DIFF
--- a/ground/gcs/share/stylesheets/linux.qss
+++ b/ground/gcs/share/stylesheets/linux.qss
@@ -2,18 +2,7 @@ QListWidget {font-size: 11px;}
 MixerCurveWidget {font-size: 12px;}
 
 	QGroupBox 	{
-		border: 1px solid gray;
-		border-radius: 5px;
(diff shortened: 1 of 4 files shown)


tracernz:
> Same story as always, that's been my bane with anything QT. Look the wrong way at attributes like border or anything, and it suddenly falls onto an apparently entirely different control renderer.

You can only do so much to native widgets due to limitations of the host platform/native library. If you want to do something not possible then Qt has to fall back to rendering the whole thing itself. That is the difference between QPalette and QSS, QPalette will just silently fail to apply impossible styles whereas QSS guarantees to give you what you ask for.

PR #1707, opened 2017-04-07
Massive (mostly GCS) reformat

mlyle:
* fromdos (line endings) all of GCS plugins
* Intoduce clang-format template and apply it to all of GCS plugins
* chmod lots of things non-executable.


mlyle:

mlyle:
Woops, had stripped executable bit off python code

mlyle:
helps to push :(

tracernz:
- On quick inspection (very difficult in the web UI due to diff size), this looks to only do header files?
- Source of the style options? Qt project?
- I was hoping to do this after bringing in PMF signal/slot PR (currently many hours deep with thousands of lines of diff).

mlyle:
@tracernz -- OK I can abandon this for now--- can you please tell me specifically **what work** should sequence in before this (Glowtape's **current** PRs?  That PR?) so we can plan on when to get the "lock" on this?

Qt does not provide a set of style things for clang-format.  The values are pretty much clang-format's defaults, changed to have the same indent as most of our GCS.  It does not seem to generate dumb diff.

(Also, yes, I screwed up, and missed the cpp command.)

mlyle:
Corrected command style for next time:

```find plugins app \( -name '*.cpp' -o -name '*.h' \) -print0 | xargs -0 clang-format-3.9 -i```

tracernz:
We should also format our own libraries (e.g. ExtensionSystem).

tracernz:
Qt's clang-format style https://github.com/qt/qtrepotools/blob/master/config/_clang-format

mlyle:
> We should also format our own libraries (e.g. ExtensionSystem).

They should get re-orged somewhere other than the libraries path first, then, IMO.

I think ExtensionSystem comes from qt creator source / isn't really our own...

PR #1710, opened 2017-04-08
pios_servo,gcs: DShot900 timing added.

glowtape:
It exists apparently. Add it for something something signal integrity? Works on KISS 24A RE.

Seppuku OSD wobbles with that, too, still requires DShot1200 to be stable.

### DIFF
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -64,6 +64,7 @@ enum channel_mode {
 	SYNC_PWM,
 	SYNC_DSHOT_300,
 	SYNC_DSHOT_600,
+	SYNC_DSHOT_900,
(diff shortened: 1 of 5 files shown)


mlyle:
Needs rebase

tracernz:
Is this beneficial at all (i.e. is there anything that supports DShot900 and not DShot1200)?

glowtape:
No idea. Probably gonna defer this until I have DMA DShot working. I've only read about it over on the Betaflight wiki, so it's still obscure I guess. Since it's just a frequency setting, might be worthwhile to have, to keep people from being annoying.

tracernz:
The more options you present, the more support requests you tend to get. 🙃

mlyle:
Closing, since it's not in wide use and ... in favor of #1721 path of work.

PR #1712, opened 2017-04-08
gcs: Reformat

mlyle:
This did a few things I don't like, like:

```
 #ifdef Q_OS_LINUX
     QApplication::setAttribute(Qt::AA_X11InitThreads, true);
-    // This should have faster performance on linux
+// This should have faster performance on linux
```

But it's a vast improvement overall, I think.


mlyle:

mlyle:
Force-pushed: forgot about the templates.

mlyle:

tracernz (approved):
:grin:

tracernz:
Looks like the UAVO templates can do with some help (they have trailing whitespace etc.). Can clean them up later.

mlyle:
Yah the formatter breaks them because they're not valid C++.  But I care about them a lot less-- they're not edited much etc.

PR #1715, opened 2017-04-10
pios_mpu9250: correct bug with bottom side orientations

mlyle:
Really need to unify all this in pios_mpu so the opportunity to make these types of mistakes is limited.

(That and/or reproject in pios_sensors so things can be uniform).

### DIFF
--- a/flight/PiOS/Common/pios_mpu9250.c
+++ b/flight/PiOS/Common/pios_mpu9250.c
@@ -711,6 +711,10 @@ static void PIOS_MPU9250_Task(void *parameters)
 		struct pios_sensor_accel_data accel_data;
 		struct pios_sensor_gyro_data gyro_data;
 
+
+		gyro_data.z  = -1.0f * (int16_t)(mpu9250_rec_buf[IDX_GYRO_ZOUT_H] << 8 | mpu9250_rec_buf[IDX_GYRO_ZOUT_L]);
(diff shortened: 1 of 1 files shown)


tracernz:
> (That and/or reproject in pios_sensors so things can be uniform).

The gyro driver could supply the axes orientation to PIOS sensors at init as it does with sample-rate. Should save us doing two separate rotations.

mlyle:
Yes-- I proposed this before but @peabody124 objected, preferring to get normalized data out of the driver.

It's proven to be exceptionally errorprone.

OTOH combining the rotations is kinda scary, too, but if we think hard on that and solve the problem once...

[Note that some drivers will still need **some** degree of reprojection because their internal mag is "oriented" differently than other sensors]

In any event, I really want to kill all the other pios_mpu drivers.

tracernz:
I can understand that. It is a lot nicer to keep the interface normalised, but in reality it's not an area we will really work on much.

tracernz:
I'd be inclined to just take the [9250 SPI code](https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/Common/pios_mpu9250_spi.c#L672-L779). It's cleaner and matches PIOS_MPU.

tracernz:
Except PIOS_MPU has different orientation for bottom... https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/Common/pios_mpu.c#L976-L1102 I wonder who is correct...

tracernz:
Well they __are__ all the same...

image
image
image
image

mlyle:
Alternative plan: REMOVE THIS DRIVER.

PR #1714, opened 2017-04-10
Actuator work in preparation for smith predictor etc

mlyle:
Does a lot of things to make actuator module simpler, faster, and more flexible, in preparation for advanced controls.

PR #1721, opened 2017-04-12
pios_dmashot: Initial implementation.

glowtape:
This is the DMA implementation of DShot. It is implemented in a self-contained module that hooks into pios_servo via #ifdefs on F4 targets.

You set up a configuration struct in board_hw_defs.c that specifies which timers can use what DMA channel and stream combination (among other things), which then gets fed to PIOS_DMAShot_Init(). Timers that don't have a DMA configuration will fall back to GPIO DShot within pios_servo. Optionally you can slave DMA writes to a single CC register onto another timer, to deal with things like Seppuku's TIM14.

Each timer can run on its own DShot frequency, so you can mix DShot300 - DShot1200, if you want to, for whatever reason.

As per @mlyle's suggestion, it doesn't use the prescaler and calculates the period and duty cycles during initialization. The duty cycles chosen are the out-of-spec ones BF/BLHeli uses. They work with KISS though.

As far as memory use goes, if DMAShot is compiled in (data, not code), it uses statically:
- 4 bytes to store a pointer to the DMA configuration.
- 4 bytes to allocate an array to.

If DMAShot gets initialized with a configuration, it'll eventually allocate 32 bytes to aforementioned array, should an user configure DShot on any timer, to hold up to 8 pointers to the internal timer configuration allocations.

For each timer configured, it'll allocate 42 bytes on a struct to store a variety of data to look up.

For DMA buffers, it'll allocate at word (32-bit) or half-word (16-bit) where appropriate. 18 "bits" will be allocated, a leading and trailing one that silences the timer, and 16 actual "bits". So the buffer would be at least 36 or 72 bytes (multiplied by amount of channels per timer).

Due to the nature of DMAR writes, there might be some overallocation. If there's only two pins on a timer, but they're spaced apart as much as possible (one pin on channel 1 and another on channel 4), it has to allocate a DMA buffer for four channels (up to 288 bytes on a 32-bit timer) to fill in for in-between. If two pins are only separated by one or when pins are adjacent, memory is allocated accordingly, i.e. always the minimum.

Best situation would be four motors on a single 16-bit timer, that'd use 226 bytes of RAM (excluding code). The RE1 would use 370 bytes, because it has the first four pins on a single but 32-bit timer. The Seppuku would use 418 bytes, due four pins being spread across three but 16-bit timers. (If I got my math correctly.)

I could successfully drive DShot on all eight pins of the RE1 and Seppuku, at mixed frequencies. OG Brain was only tested on the first four, because I don't have any 8-pin JST cables anymore.

Still need to add other targets.

Old PR message:

> PR created for general appraisal and critique. Given how much I chased my tail over a bunch of stupid bugs, I'm going to leave it a day or two alone.
> 
> Everything is pretty much standalone with a bunch of hooks into pios_servo.
> 
> Tested on a BrainRE1, I can drive all eight pins with DMA DShot (specifically disabled bitbanging code to check).
> 
> Things I need to do:
> - Implement actual frequency selection. Maybe compute it per timer instead of a bunch of #ifdef'd defines (like currently for F446).
> - Test whether 16-bit timers can do away with half-word writes, to save memory in the DMA buffers. Full word writes to 16-bit timers (e.g. TIM1 and TIM8 on RE1) don't break anything, though.
> - Add code to deal with TIM14 for Seppuku (write to DMAR, if it has it, or direct write to CCR, slaved to other timer like second TIM3_UP).

### DIFF
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -43,6 +43,10 @@
 #include "pios_irq.h"
 #endif
 
+#if defined(PIOS_INCLUDE_DMASHOT)
+#include "pios_dmashot.h"
+#endif
+
 #include "misc_math.h"
 
 /* Private variables */
@@ -64,7 +68,8 @@ enum channel_mode {
 	SYNC_PWM,
 	SYNC_DSHOT_300,
 	SYNC_DSHOT_600,
-	SYNC_DSHOT_1200
+	SYNC_DSHOT_1200,
+	SYNC_DSHOT_DMA
 } __attribute__((packed));
 
 struct dshot_info {
@@ -191,11 +196,11 @@ static void ChannelSetup_DShot(int j, uint16_t rate)
 	ds_info->pin = servo_cfg->channels[j].pin.init.GPIO_Pin;
 }
 
-static int BankSetup_DShot(struct timer_bank *bank) {
+static int BankSetup_DShot(struct timer_bank *bank)
+{
 	/*
 	 * Relatively simple configuration here.  Just seize it!
 	 */
-
 	PIOS_TIM_SetBankToGPOut(servo_tim_id, bank->timer);
 
 	return 0;
@@ -275,6 +280,12 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 	if (!servo_cfg || banks > PIOS_SERVO_MAX_BANKS) {
 		return -10;
 	}
+
+#if defined(PIOS_INCLUDE_DMASHOT)
+	if (PIOS_DMAShot_IsConfigured()){
+		PIOS_DMAShot_Prepare();
+	}
+#endif
 
 	dshot_in_use = false;
 
@@ -375,9 +386,21 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 		case SHOT_DSHOT300:
 		case SHOT_DSHOT600:
 		case SHOT_DSHOT1200:
-			dshot_in_use = true;
+			ret = 1;
+#if defined(PIOS_INCLUDE_DMASHOT)
+			if (PIOS_DMAShot_IsConfigured()) {
+				uint32_t freq = rate == SHOT_DSHOT1200 ? DMASHOT_1200 : rate == SHOT_DSHOT600 ? DMASHOT_600 : DMASHOT_300;
+				ret = PIOS_DMAShot_RegisterTimer(timer_banks[i].timer, max_tim_clock, freq) ? 0 : 1;
+			}
+#endif
+			if (ret) {
+				dshot_in_use = true;
 
-			ret = BankSetup_DShot(&timer_banks[i]);
+				ret = BankSetup_DShot(&timer_banks[i]);
+			}
+			if (ret) {
+				return ret;
+			}
 			break;
 		}
 
@@ -391,7 +414,15 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 				case SHOT_DSHOT300:
 				case SHOT_DSHOT600:
 				case SHOT_DSHOT1200:
-					ChannelSetup_DShot(j, rate);
+#if defined(PIOS_INCLUDE_DMASHOT)
+					if (PIOS_DMAShot_IsConfigured() && PIOS_DMAShot_RegisterServo(chan)) {
+						output_channels[j].mode = SYNC_DSHOT_DMA;
+					} else {
+						ChannelSetup_DShot(j, rate);
+					}
+#else
+						ChannelSetup_DShot(j, rate);
+#endif
 					break;
 				case SHOT_ONESHOT:
 					output_channels[j].i.pwm_res = timer_banks[i].clk_rate;
@@ -409,6 +440,15 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 			}
 		}
 	}
+
+#if defined (PIOS_INCLUDE_DMASHOT)
+	if (PIOS_DMAShot_IsConfigured()) {
+		PIOS_DMAShot_Validate();
+		PIOS_DMAShot_InitializeGPIOs();
+		PIOS_DMAShot_InitializeTimers((TIM_OCInitTypeDef *)&servo_cfg->tim_oc_init);
+		PIOS_DMAShot_InitializeDMAs();
+	}
+#endif
 
 	return 0;
 }
@@ -459,6 +499,13 @@ void PIOS_Servo_SetFraction(uint8_t servo, uint16_t fraction,
 			 * Don't bother with min/max tomfoolery here.
 			 */
 			output_channels[servo].i.dshot.value = fraction >> 5;
+			return;
+		case SYNC_DSHOT_DMA:
+#if defined(PIOS_INCLUDE_DMASHOT)
+			if (PIOS_DMAShot_IsConfigured()) {
+				PIOS_DMAShot_WriteValue(&servo_cfg->channels[servo], fraction >> 5);
+			}
+#endif
 			return;
 		default:
 			break;
@@ -511,15 +558,19 @@ void PIOS_Servo_Set(uint8_t servo, float position)
 		case SYNC_DSHOT_300:
 		case SYNC_DSHOT_600:
 		case SYNC_DSHOT_1200:
-			/* Expect a 0 to 2047 range! */
-			if (position > 2047) {
-				output_channels[servo].i.dshot.value = 2047;
-			} else if (position < 0) {
-				output_channels[servo].i.dshot.value = 0;
-			} else {
-				output_channels[servo].i.dshot.value = position;
-			}
+			if (position > 2047)
+				position = 2047;
+			else if (position < 0)
+				position = 0;
 
+			output_channels[servo].i.dshot.value = position;
+			return;
+		case SYNC_DSHOT_DMA:
+#if defined(PIOS_INCLUDE_DMASHOT)
+			if (PIOS_DMAShot_IsConfigured()) {
+				PIOS_DMAShot_WriteValue(&servo_cfg->channels[servo], position);
+			}
+#endif
 			return;
 		default:
 			break;
@@ -724,6 +775,7 @@ void PIOS_Servo_Update(void)
 	// If some banks are oneshot and some are dshot (why would you do this?)
 	// get the oneshots firing first.
 	for (uint8_t i = 0; i < servo_cfg->num_channels; i++) {
+
 		if (output_channels[i].mode != SYNC_PWM) {
 			continue;
 		}
@@ -737,6 +789,13 @@ void PIOS_Servo_Update(void)
 			TIM_Cmd(chan->timer, ENABLE);
 		}
 	}
+
+#if defined(PIOS_INCLUDE_DMASHOT)
+	// Do it before bitbanging, so the updates line up +/-.
+	if (PIOS_DMAShot_IsReady()) {
+		PIOS_DMAShot_TriggerUpdate();
+	}
+#endif
 
 	if (dshot_in_use) {
 		if (DSHOT_Update()) {
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -389,7 +389,19 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 			ret = 1;
 #if defined(PIOS_INCLUDE_DMASHOT)
 			if (PIOS_DMAShot_IsConfigured()) {
-				uint32_t freq = rate == SHOT_DSHOT1200 ? DMASHOT_1200 : rate == SHOT_DSHOT600 ? DMASHOT_600 : DMASHOT_300;
+				uint32_t freq;
+				switch(rate) {
+					default:
+					case SHOT_DSHOT300:
+						freq = DMASHOT_300;
+						break;
+					case SHOT_DSHOT600:
+						freq = DMASHOT_600;
+						break;
+					case SHOT_DSHOT1200:
+						freq = DMASHOT_1200;
+						break;
+				}
 				ret = PIOS_DMAShot_RegisterTimer(timer_banks[i].timer, max_tim_clock, freq) ? 0 : 1;
 			}
 #endif
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -568,6 +568,11 @@ void PIOS_Servo_Set(uint8_t servo, float position)
 		case SYNC_DSHOT_DMA:
 #if defined(PIOS_INCLUDE_DMASHOT)
 			if (PIOS_DMAShot_IsConfigured()) {
+				if (position > 2047)
+					position = 2047;
+				else if (position < 0)
+					position = 0;
+
 				PIOS_DMAShot_WriteValue(&servo_cfg->channels[servo], position);
 			}
 #endif
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -444,8 +444,8 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 #if defined (PIOS_INCLUDE_DMASHOT)
 	if (PIOS_DMAShot_IsConfigured()) {
 		PIOS_DMAShot_Validate();
-		PIOS_DMAShot_InitializeGPIOs();
 		PIOS_DMAShot_InitializeTimers((TIM_OCInitTypeDef *)&servo_cfg->tim_oc_init);
+		PIOS_DMAShot_InitializeGPIOs();
 		PIOS_DMAShot_InitializeDMAs();
 	}
 #endif
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -392,6 +392,9 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 				uint32_t freq;
 				switch(rate) {
 					default:
+						// If for whatever reason new frequencies show up in the GPIO version,
+						// we oughta know about it. So fail if that happens.
+						PIOS_Assert(0);
 					case SHOT_DSHOT300:
 						freq = DMASHOT_300;
 						break;
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -418,10 +418,13 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 					if (PIOS_DMAShot_IsConfigured() && PIOS_DMAShot_RegisterServo(chan)) {
 						output_channels[j].mode = SYNC_DSHOT_DMA;
 					} else {
+						// If RegisterServo fails, the relevant timer wasn't registered, or DMAShot
+						// isn't configured, so we expect the bank to be set up already higher up in this
+						// section.
 						ChannelSetup_DShot(j, rate);
 					}
 #else
-						ChannelSetup_DShot(j, rate);
+					ChannelSetup_DShot(j, rate);
 #endif
 					break;
 				case SHOT_ONESHOT:
--- a/flight/PiOS/STM32F4xx/inc/pios_dmashot.h
+++ b/flight/PiOS/STM32F4xx/inc/pios_dmashot.h
@@ -83,7 +83,7 @@ void PIOS_DMAShot_InitializeTimers(TIM_OCInitTypeDef *ocinit);
 void PIOS_DMAShot_InitializeDMAs();
 void PIOS_DMAShot_Validate();
 
-bool PIOS_DMAShot_WriteValue(const struct pios_tim_channel *servo_channel, uint16_t throttle);
+void PIOS_DMAShot_WriteValue(const struct pios_tim_channel *servo_channel, uint16_t throttle);
 
 void PIOS_DMAShot_TriggerUpdate();
 
--- a/flight/PiOS/STM32F4xx/inc/pios_dmashot.h
+++ b/flight/PiOS/STM32F4xx/inc/pios_dmashot.h
@@ -43,20 +43,6 @@
 #define DMASHOT_600                                             600000
 #define DMASHOT_1200                                            1200000
 
-// Which register to target in master-slave mode.
-#define DMASHOT_REG_CCR1                                        0x01
-#define DMASHOT_REG_CCR2                                        0x02
-#define DMASHOT_REG_CCR3                                        0x03
-#define DMASHOT_REG_CCR4                                        0x04
-
-// Which DMA event source to use in master-slave mode. Make sure the proper
-// DMA channel and stream are selected, too.
-#define DMASHOT_EVENTSOURCE_UP                                  0x10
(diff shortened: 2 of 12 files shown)


mlyle:
> Add code to deal with TIM14 for Seppuku (write to DMAR, if it has it, or direct write to CCR, slaved to other timer like second TIM3_UP).

I think you'll need to write to the CCR directly, btw.  I don't know the exact logic for the generation of DMA requests beats and for address incrementing / DMAR magic in the hardware, but if they're not completely independent it's going to cause a problem using another timer's DMA request line.  (And even if they are independent on F405, it seems safer to be explicit).

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
```diff
@@ -0,0 +1,78 @@
...
+#define DSHOT_PERIOD_0							27
+#define DSHOT_PERIOD_1							55
+
+#else
+
+#define DSHOT300_FREQUENCY						6000000
+#define DSHOT600_FREQUENCY						(DSHOT300_FREQUENCY*2)
+#define DSHOT900_FREQUENCY						(DSHOT300_FREQUENCY*3)
+#define DSHOT1200_FREQUENCY						(DSHOT300_FREQUENCY*4)
+#define DSHOT150_FREQUENCY						(DSHOT300_FREQUENCY/2)
```
tracernz: Missing spaces around operators

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
```diff
@@ -0,0 +1,78 @@
...
+void PIOS_DMAShot_InitializeGPIOs();
+void PIOS_DMAShot_InitializeTimers(TIM_OCInitTypeDef *ocinit);
+void PIOS_DMAShot_InitializeDMAs();
+void PIOS_DMAShot_Validate();
+
+bool PIOS_DMAShot_WriteValue(const struct pios_tim_channel *servo_channel, uint16_t throttle);
+
+void PIOS_DMAShot_FireDMAs();
+
+bool PIOS_DMAShot_GreenLight();
```
tracernz: This is not a good function name IMO. We should stick to formal language.

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
```diff
@@ -0,0 +1,78 @@
...
+bool PIOS_DMAShot_RegisterServo(const struct pios_tim_channel *servo_channel);
+
+void PIOS_DMAShot_InitializeGPIOs();
+void PIOS_DMAShot_InitializeTimers(TIM_OCInitTypeDef *ocinit);
+void PIOS_DMAShot_InitializeDMAs();
+void PIOS_DMAShot_Validate();
+
+bool PIOS_DMAShot_WriteValue(const struct pios_tim_channel *servo_channel, uint16_t throttle);
+
+void PIOS_DMAShot_FireDMAs();
```
tracernz: Marginal name.

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
```diff
@@ -0,0 +1,78 @@
...
+
+bool PIOS_DMAShot_GreenLight();
+
+#endif // PIOS_DMASHOT_H
```
tracernz: Text files require newlines to terminate all character sequences.

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
```diff
@@ -0,0 +1,78 @@
...
+};
+
+struct pios_dmashot_cfg {
+
+	const struct pios_dmashot_timer_cfg *timer_cfg;
+	uint8_t num_timers;
+
+};
+
+void PIOS_DMAShot_Init(const struct pios_dmashot_cfg *config);
```
tracernz: Please add Doxygen documentation for all public interfaces.

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,423 @@
...
+
+	union dma_buffer buffer;							// DMA buffer
+
+	TIM_OCInitTypeDef *ocinit;
+};
+
+#define TIMC_TO_INDEX(c)			((c)>>2)
+#define NUM_CHANNELS(s)				(TIMC_TO_INDEX((s)->high_channel - (s)->low_channel) + 1)
+
+#define DMA_DISABLE(s)				{ DMA_Cmd(s, DISABLE); while (DMA_GetCmdStatus(s) == ENABLE); }
```
tracernz: Hiding tight loops in macros harms readability/analysis IMO.

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,423 @@
...
+	dmashot_cfg = config;
+}
+
+void PIOS_DMAShot_Prepare()
+{
+	if (dmashot_cfg && !servo_timers) {
+		// Allocate memory
+		servo_timers = PIOS_malloc_no_dma(sizeof(struct servo_timer) * MAX_TIMERS);
+		if(!servo_timers)
+			ERROR();
```
tracernz: This hurts readability IMO. Does it return, does it assert? Makes life harder when chasing bugs.

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
```diff
@@ -0,0 +1,78 @@
...
+
+#define DMASHOT_MAX_BUFFERS						8
+
+#define DMASHOT_MESSAGE_WIDTH					16
+#define DMASHOT_MESSAGE_PAUSE					2
+#define DMASHOT_STM32_BUFFER					(DMASHOT_MESSAGE_WIDTH+DMASHOT_MESSAGE_PAUSE)
+
+#ifdef STM32F446xx
+
+#define DSHOT150_FREQUENCY						11250000
```
tracernz: Most of these belong in a private header, or the implementation (.c), not a public header (public headers should be restricted to minimal public interface only).

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,423 @@
...
+		if (s_timer && s_timer->timer) return true;
+	}
+
+	return false;
+}
+
+bool PIOS_DMAShot_Enabled()
+{
+	return dmashot_cfg != NULL;
+}
```
tracernz: Missing newline.

tracernz (commented):
There a lot of code style issues in pios_dmashot.*. Please run uncrustify when you've finished writing it. :wink:

mlyle:
```
/var/lib/jenkins/jobs/dron/workspace@2/Nodes/master/flight/PiOS/STM32/pios_servo.c:46:26: fatal error: pios_dmashot.h: No such file or directory
compilation terminated.
```

glowtape:
Yeah I know, the DMA DShot stuff currently in PIOS/STM32F4xx. I have to move it outta there. It should compile with the F4 headers regardless of what the actual target is?

glowtape:
Updated original message.

glowtape:
With TBS Crossfire receiver, Autotune (w/ decimation on RE1) enabled, memory free (heap/fastheap) for SyncPWM or DShot, total memory usage when enabled:

All pins to DShot:
- Seppuku: SyncPWM (29136/51436), DShot (28740/51276) = 556 bytes.
- BrainRE1: SyncPWM (16864/0), DShot (16236/0) = 628 bytes.
- OG Brain: SyncPWM (28236/51648), DShot (27804/51568) = 512 bytes.

At least pins 1-4 to DShot:
- Seppuku: SyncPWM (29136/51436), DShot (28884/51316) = 372 bytes.
- BrainRE1: SyncPWM (16864/0), DShot (16536/0) = 328 bytes.
- OG Brain: SyncPWM (28236/51648), DShot (27948/51608) = 328 bytes.

glowtape:
Flew two packs of this at DShot600, three packs at DShot300 (OSD was not wobbly, so no accidental GPIO version). Seemed to work fine. Was too cold today, so I quit eventually. DShot1200 works on the bench using the Test Outputs tab, but I'm still going to fly it soon.

mlyle:
@Glowtape Awesome!

What FC was the flight test on?

glowtape:
A brand-new Seppuku.

mlyle:
Is this still WIP or are you ready for review?

glowtape:
Now it's ready for review.

flight/PiOS/STM32/pios_servo.c
```diff
@@ -375,9 +386,21 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 		case SHOT_DSHOT300:
 		case SHOT_DSHOT600:
 		case SHOT_DSHOT1200:
-			dshot_in_use = true;
+			ret = 1;
+#if defined(PIOS_INCLUDE_DMASHOT)
+			if (PIOS_DMAShot_IsConfigured()) {
+				uint32_t freq = rate == SHOT_DSHOT1200 ? DMASHOT_1200 : rate == SHOT_DSHOT600 ? DMASHOT_600 : DMASHOT_300;
```
mlyle: No ternary-to-death, please?  Any other structure is OK

(or if going to use ternary, please indent it like other uses of ternary-as-case)

flight/PiOS/STM32/pios_servo.c
```diff
@@ -391,7 +414,18 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 				case SHOT_DSHOT300:
 				case SHOT_DSHOT600:
 				case SHOT_DSHOT1200:
+#if defined(PIOS_INCLUDE_DMASHOT)
+					if (PIOS_DMAShot_IsConfigured() && PIOS_DMAShot_RegisterServo(chan)) {
```
mlyle, line 433: Important test case: PIOS_INCLUDE_DMASHOT on a target with no dma channels configured (To verify fallback logic)

flight/PiOS/STM32/pios_servo.c
```diff
@@ -460,6 +503,13 @@ void PIOS_Servo_SetFraction(uint8_t servo, uint16_t fraction,
 			 */
 			output_channels[servo].i.dshot.value = fraction >> 5;
 			return;
+		case SYNC_DSHOT_DMA:
+#if defined(PIOS_INCLUDE_DMASHOT)
+			if (PIOS_DMAShot_IsConfigured()) {
```
mlyle, line 523: This is for F0/flyingpio.  I wonder if it can be made to work with this. :D

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
mlyle: Ditto

mlyle: (Oh, I see they are in the C file.  Can you move them to the header, since this is the preferred way now?)

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,591 @@
...
+ */
+
+#include <pios.h>
+#include "pios_dmashot.h"
+
+#define MAX_TIMERS                              8
+
+// This is to do half-word writes where appropriate. TIM2 and TIM5 are 32-bit, the rest
+// are 16-bit. Can do full word writes to 16bit regardless, but it doubles DMA buffer size
+// unnecessarily.
```
mlyle, line 37: And to confirm-- you can't do half-word writes to the 32 bit timer?

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,591 @@
...
+// instead of the spec one. Works with KISS RE 24A though.
+#define DSHOT_DUTY_CYCLE_0                      36
+#define DSHOT_DUTY_CYCLE_1                      74
+
+#define TIMC_TO_INDEX(c)                        ((c)>>2)
+
+const struct pios_dmashot_cfg *dmashot_cfg;
+struct servo_timer **servo_timers;
+
+static inline int PIOS_DMAShot_GetFIFOCadence(int num_chan)
```
mlyle: What's the reasoning for this?

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,591 @@
...
+{
+	if (!servo_timers)
+		return NULL;
+
+	for (int i = 0; i < MAX_TIMERS; i++) {
+		struct servo_timer *s_timer = servo_timers[i];
+		if (!s_timer || !s_timer->sysclock)
+			continue;
+
+		if (s_timer->dma->timer == servo_channel->timer) return s_timer;
```
mlyle: line break before return

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,591 @@
...
+			break;
+		}
+	}
+
+	// Nothing found? Allocate a new slot for the timer.
+	if (!found) {
+		for (int i = 0; i < MAX_TIMERS; i++) {
+			if (!servo_timers[i]) {
+				s_timer = PIOS_malloc_no_dma(sizeof(struct servo_timer));
+				memset(s_timer, 0, sizeof(struct servo_timer));
```
mlyle: I'm ignoring allocation / memset for now.. in my next review pass I'll look at them closely as they require a focused review.

mlyle: (though should s_timer be checked for NULL? )

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,591 @@
...
+			// disable the timer.
+			s_timer->sysclock = 0;
+			continue;
+		}
+
+		PIOS_DMAShot_TimerSetup(s_timer, s_timer->sysclock, s_timer->dshot_freq, ocinit, false);
+
+		int f = s_timer->sysclock / s_timer->dshot_freq;
+
+		s_timer->duty_cycle_0 = f * DSHOT_DUTY_CYCLE_0 / 100;
```
mlyle: Suggest + 50 in these

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,591 @@
...
+	// If there's nothing setup, fail hard. We shouldn't be getting here.
+	PIOS_Assert(dmashot_cfg && servo_timers);
+
+	for (int i = 0; i < MAX_TIMERS; i++) {
+		struct servo_timer *s_timer = servo_timers[i];
+		if (!s_timer || !s_timer->sysclock)
+			continue;
+
+		DMA_Cmd(s_timer->dma->stream, DISABLE);
+		while (DMA_GetCmdStatus(s_timer->dma->stream) == ENABLE) ;
```
mlyle: I suggest we wait for completion --before disabling--, so that if jitter causes us to show up when another update is in progress we wait for it to be done.

mlyle: (Alternatively we could drop the update entirely and return)

mlyle:
Sorry for not bundling comments into a review--- by the time I realized I had screwed up I was "committed" heh..

This is a very good start.  I need to check the memory stuff in depth in a separate pass.

flight/PiOS/STM32F4xx/pios_dmashot.c
glowtape: From the docs: "The DMA burst size on the memory port and the FIFO threshold configuration must match."

flight/PiOS/STM32/pios_servo.c
glowtape: I already had a partial confirmation, when I was testing on Seppuku without slaving TIM14, resulting in mixed DMA and GPIO DShot. Just tested by commenting out PIOS_DMAShot_Init() in pios_board.c on Seppuku while leaving the #define intact. Motors started spinning.

Pudding for mixed DShot sources (before I changed call order in pios_servo). Channel 2 is GPIO, rest is DMA:
dualshot

flight/PiOS/STM32F4xx/pios_dmashot.c
glowtape: Just checked again, doesn't work. RE1 has pin 1-4 on TIM5, which is 32bit. When I try to do half-word DMA into the DMAR of a 32bit timer, it doesn't seem to execute the transfer.

flight/PiOS/STM32F4xx/pios_dmashot.c
glowtape: Added a check for the transfer complete flag. If it's not been set, it'll drop the update. Probably better that way, than to fear the erratic wrath of the watchdog.

flight/PiOS/STM32F4xx/pios_dmashot.c
glowtape: Not sure I understand.

flight/PiOS/STM32F4xx/pios_dmashot.c
mlyle: Message lengths are very short compared to watchdog intervals (e.g. that's why we get away with bit-banging them).

Arguments for blocking: might get the useful update out sooner. If DMA/something gets stuck, we actually **do** trigger the watchdog instead of just returning immediately.

Arguments against: Prevent a chain of 2-3 updates being slightly late; instead absorb the penalty all at once.

flight/PiOS/STM32F4xx/pios_dmashot.c
mlyle: Yes, I'm more saying.. why are you varying the DMA burst size on the memory port?  It's not related to the dma request / transfer burst length.

flight/PiOS/STM32F4xx/pios_dmashot.c
mlyle: OK, thx.

flight/PiOS/STM32F4xx/pios_dmashot.c
mlyle: (f * DSHOT_DUTY_CYCLE_0 + 50) / 100

Basically, get correct rounding when you can't hit timing exactly.

flight/PiOS/STM32/pios_servo.c
glowtape: If we wrap DMA_Cmd in a macro and place an #ifdef in _DMASetup(), it can probably work. For F0 and also F3 I guess. I don't have any of those for testing, though.

flight/PiOS/STM32F4xx/pios_dmashot.c
glowtape: Seems like I've misunderstood the docs then. Reviewing the table of burst configurations, I can rarely make it fit anyway (esp. since DMA buffer size depends on channels used per timer), which implies it shouldn't have worked, yet it didn't blow.

And I just tested it with FIFO disabled, there's still a proper DShot signal on the wire regardless (on both Seppuku and RE1, which have different F4). So this gets scrubbed I guess.

mlyle: They need to be chosen from the table to be a legal transfer for the **memory burst size** and element size.  But the memory burst size doesn't need to match the peripheral operation size, and the burst thing that the timer does is different from a DMA controller burst anyways.

Basically, every time the FIFO isn't filled, the DMA tries to get control of the memory bus, and do a sequence of N transfers in order to fill the fifo up to the desired level.  Because turning around the memory bus takes time and sequential transfers are faster, this uses the memory bandwidth more efficiently.  Some combinations are not legal (e.g. ones that risk overfilling the FIFO)

I believe at least-- this is my interpretation.

flight/PiOS/STM32/pios_servo.c
```diff
@@ -389,7 +389,19 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 			ret = 1;
 #if defined(PIOS_INCLUDE_DMASHOT)
 			if (PIOS_DMAShot_IsConfigured()) {
-				uint32_t freq = rate == SHOT_DSHOT1200 ? DMASHOT_1200 : rate == SHOT_DSHOT600 ? DMASHOT_600 : DMASHOT_300;
+				uint32_t freq;
+				switch(rate) {
+					default:
```
tracernz: default case should never happen? Should be an assert.

flight/PiOS/STM32F4xx/pios_dmashot.c
glowtape: Here's the OG Brain, which is already jittery with a lean configuration, at DShot300, with autotune and 250hz fullbore logging enabled to OpenLager at 2470000 baud (picked the worst case found):

jitter2

Looks like an argument for dropping (actually, this screenshot is already from the build with the check for transfer complete flag). Do we expect DMA to break?

mlyle: Wow.. OG brain doing so badly at 1KHz gyro is ... not great, but not unexpected given I2C i guess. 

if updates come nominally at 1ms, intervals, and you get one at 0ms, 1.99ms (should be 1ms), 2.01ms, 3.00ms, say..

If you drop, from 0-1.99ms you use the value from 0ms, from 1.99ms-3ms you use the value from 1ms,  and at 3ms you use the value from 3ms.

If you busy-wait, from 0-1.99ms you use the value from 0ms, from 1.99 to say 2.05ms you use the value from 1ms, and from 2.05ms to 3ms you use the value from 2ms.

> Do we expect DMA to break?

It shouldn't-- It's just a question of what makes any potential defect less severe (if someone stomps a timer configuration or whatever).

glowtape: OK, changed it to wait for the transfers to finish.

glowtape:
Left the OG Brain running at 90+ CPU for 45 minutes, waiting for DMA to break or delay things so much that the watchdog kicks in on the transfer complete waiting loop. Nothing happened. (Had to flash firmware to go flying, otherwise I'd have left it running.)

Looked at the output occasionally, the shortest distance I spotted was 78us between updates. Nothing that looked like an outright collision, though.

collision

glowtape:
Flew all my packs on the latest of this configured to DShot1200, on my Seppuku quad, together with #1714 and #1642 configured to extended prediction period. No issues.

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,603 @@
...
+			found = true;
+			break;
+		}
+	}
+
+	// Nothing found? Allocate a new slot for the timer.
+	if (!found) {
+		for (int i = 0; i < MAX_TIMERS; i++) {
+			if (!servo_timers[i]) {
+				s_timer = PIOS_malloc_no_dma(sizeof(struct servo_timer));
```
mlyle: sizeof(*s_timer) is preferred for these

flight/PiOS/STM32F4xx/pios_dmashot.c
```diff
@@ -0,0 +1,603 @@
...
+	PIOS_Assert(dmashot_cfg && servo_timers);
+
+	for (int i = 0; i < MAX_TIMERS; i++) {
+		struct servo_timer *s_timer = servo_timers[i];
+		if (!s_timer || !s_timer->sysclock)
+			continue;
+
+		if (!s_timer->buffer.ptr) {
+			// Allocate buffer at timer resolution.
+			s_timer->buffer.ptr = PIOS_DMAShot_AllocateBuffer(
```
mlyle, line 467: This factoring is much nicer.

flight/PiOS/STM32F4xx/inc/pios_dmashot.h
mlyle: This is still a blocker before merge.

Issue #1602, opened 2017-02-28
GCS: Rfm Bind Wizard Appears Broken


tracernz:
* Seen in version (release version or githash): Release 20170213
* On flight controller (if applicable): Revo

Issue details:
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJDA5JkJDA
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJDU59sKDA
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJCn4YoLDA
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJDt4IELDA
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJDAgMgIDA
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgJDJk54LDA
http://dronin-autotown.appspot.com/at/crash/ahFzfmRyb25pbi1hdXRvdG93bnIWCxIJQ3Jhc2hEYXRhGICAgNDAypsJDA

mlyle:
Yuck.  OK.

tracernz:
Keep in mind those are all the same guy. Needs triage. :smiley:

tracernz:
He managed to get this to work somehow. Would still be good to understand the failure but this can probably be punted from the hotfix so we can get RE1 fixed.

utoedter:
I can confirm that crash too, happens for me on Windows 7/64 Bit. If you need more information let me know.

pug398:
If you get the time could you revisit the RF Bind wizard. It works for me in 78241e0d 20170317
but appears broken in 1a2c64cc 20170330 and later. The latter is the first build I have after 20170317 but does not mean that is when problem started. Does not detect pipxtreme and uavo does not populate.

tracernz:
[...]

PR #1722, opened 2017-04-14
flight: move to newlib-nano from newlib

mlyle:
On Sparky2, this saves 2588 bytes of flash and.. 968 (!) bytes of data
segment SRAM.


tracernz (approved):
Great, didn't realise we weren't using nano. :blush:

glowtape:
Stuck this in my flying branch, seems to work. No firmware crashes or any other odd stuff. (--edit: Seppuku.)

mlyle:
@glowtape -- cool! thanks for the test

tracernz:
I guess for completeness we should test this on a Cortex-M3 before merging too? That's actually just the PipX now though I guess...

tracernz:
Test on a (manually bound re: pug's comment in #1602) PipX, all was well.

PR #1725, opened 2017-04-14
Flight: Document IMU rotation

tracernz:
This table needs careful review. According to it bottom 90 and 270 are wrong in pios_mpu.c.


tracernz:
Going to add mag diagrams...

tracernz:
TODO:
- BMX055/BMM150 (BMM150 orientation is different when embedded inside BMX055 :().
- Fix sensor drivers to common convention.

PR #1724, opened 2017-04-14
remove pios_mpu9250_spi driver (sparky2, lux)

mlyle:
Status: accel and gyro and mag work.

ID detection on MPU9250 was wrong, preventing the mag code from working.  Corrected this (need to verify good for DTFc which had a similar ID problem)

PR #1728, opened 2017-04-14
PiOS: Change IMU drivers to common rotation convention

tracernz:
Documented in #1725. This should make no functional change. The variety of conventions in these drivers highlights why standardising is a great idea!

Note: HMC5*83 rotation convention in HwShared is wonky, but I have kept it to avoid breaking existing configurations.

TODO:
- BMX055/BMM150

~~~Depends on/needs rebased after #1724 (due to PIOS_MPU changes).~~~


mlyle:
jenkins, test this please

tracernz:
Doh, inadvertent change, fixed.

mlyle:
This LGTM-- plausible at least.... and the correct changes are made for targets to behave the same.

tracernz:
Yah, I think it's quite safe. 👍

PR #1731, opened 2017-04-15
brainre1i,pios_crossfire: Add Crossfire to other serial ports.

glowtape:
Someone over on RCG noticed that it's only available on the RX port. Not sure why I omitted the others.

### DIFF
--- a/shared/uavobjectdefinition/hwbrainre1.xml
+++ b/shared/uavobjectdefinition/hwbrainre1.xml
@@ -31,6 +31,7 @@
 				<option>MavLinkTX</option>
 				<option>MavLinkTX_GPS_RX</option>
 				<option>VTX Config TBS SmartAudio</option>
+				<option>TBS Crossfire</option>
 			</options>
 		</field>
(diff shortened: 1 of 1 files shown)


tracernz (approved):
Thanks 👍

yamahabest:
I have a remark on this change.
The RE1 has a dedicated receiver port, which has only receiver protocols on it, no gps, no stormbgc, etc.
Which is probably why none of the receiver protocols have been added to those other serial ports.

What would be the reason that someone wants to connect their receiver to another serial port, and leave the dedicated receiver port unused?

If that reason is valid, all other receiver protocols (sbus, ibus, srxl, sumd, sumh) should also be added.
And the getInputType function should be extended to also check those serial ports.

glowtape:
If you want to use Crossfire telemetry, you need both RX and TX pins. The receiver port however has only an RX pin. For a short while I had some test code to run telemetry on a different UART, but we decided against it. Kind of odd to hook it up on two different UARTs.

yamahabest:
Ah ok, didn't think of the telemetry. Sounds good.
Would be odd indeed to connect to two uarts.

Leaves the question if we should be adding the other protocols also.

PR #1682, opened 2017-03-30
4wif passthrough implementation.

mlyle:
Please don't be too mad at the ugliness. :p

PR #1732, opened 2017-04-15
pios: add pios_dio digital IO system

mlyle:
For review / to show the concept.  This PR maybe shouldn't merge, until it's actually doing something useful.

Todo:

* [x] Verify config registers on F0
* [x] Verify config registers on F3
* [x] Verify config registers on F4
* [x] Verify drives on F0
* [x] Verify drives on F3
* [x] Verify drives on F4
* [x] Support for configuring alternate functions
* [x] Support for configuring analog
* [x] Support for F1

### DIFF
--- a/flight/PiOS/STM32/inc/pios_dio.h
+++ b/flight/PiOS/STM32/inc/pios_dio.h
@@ -91,7 +91,7 @@ typedef uintptr_t dio_tag_t;
  * @param[in] open_collector true for open-collector + pull-up semantics
  * @param[in] dio_drive_strength The drive strength to use for the GPIO
  */
-static inline void DIO_altfunc_output(dio_tag_t t, int alt_func,
+static inline void dio_set_altfunc_output(dio_tag_t t, int alt_func,
 		bool open_collector, enum dio_drive_strength strength);
 
 /**
@@ -101,7 +101,7 @@ static inline void DIO_altfunc_output(dio_tag_t t, int alt_func,
  * @param[in] alt_func The alternate function specifier.
  * @param[in] pull Pullup/Pulldown configuration.
  */
-static inline void DIO_altfunc_input(dio_tag_t t, int alt_func,
+static inline void dio_set_altfunc_input(dio_tag_t t, int alt_func,
 		enum dio_pull pull);
 
 /**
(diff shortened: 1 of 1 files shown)


flight/PiOS/STM32/inc/pios_dio.h
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * @brief Configures a GPIO as alternate function.
+ * Note that to configure drive strength, pullups, etc, DIO_output or
+ * DIO_input should be called first.
+ *
+ * @param[in] t The GPIO specifier tag (created with e.g. GPIOA_DIO(3) )
+ * @param[in] alt_func The alternate function specifier.
+ */
+static inline void DIO_altfunc(dio_tag_t t, int alt_func);
```
tracernz: Why uppercase? I think a verb would be a good idea for all these functions. e.g. `dio_set_altfunc`

flight/PiOS/STM32/inc/pios_dio.h
```diff
@@ -0,0 +1,313 @@
...
+ * @param[in] t The GPIO specifier tag
+ */
+static inline void DIO_low(dio_tag_t t);
+
+/**
+ * @brief Sets an output GPIO to a chosen logical level.
+ * @param[in] t The GPIO specifier tag
+ * @param[in] high Whether the GPIO should be high.
+ */
+static inline void DIO_set(dio_tag_t t, bool high);
```
tracernz: Should match the read call, e.g. `dio_write` and `dio_read`.

tracernz:
It's a shame you can't operate on multiple pins.

flight/PiOS/STM32/inc/pios_dio.h
mlyle: need to decide what we're doing with prefixen in general.  PIOS etc are uppercase prefixes still.. I don't know.

mlyle:
Might be possible to extend / wrap this for multiple in the same IO bank.   99% of what we do can work this way though.

Pushed based on your review feedback / names.

flight/PiOS/STM32/inc/pios_dio.h
glowtape: I can do without the PIOS_ prefix. Plenty of other frameworks and APIs don't qualify their functionality with vendor/framework specific prefixes either.

As far as the casing goes, "orthographically correct" mixed casing sure helps readability, but looks odd when doing shorthands. On the other hand, I prefer shorthands. _dio_altfunc_ is certainly preferable to _DIO_AlternateFunction_, at least when typing out things.

mlyle:
If this is acceptable, will close PR and then cherry-pick into 4wif branch.

mlyle:
Incorporated #1714

mlyle:
Correction;; incorporated to #1682.

PR #1733, opened 2017-04-16
Added the other radio protocols to the additional ports on RE1

yamahabest:
And added the ports to the getInputType function.

### DIFF
--- a/ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
+++ b/ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
@@ -200,6 +200,72 @@ Core::IBoardType::InputType BrainRE1::getInputType()
         break;
     }
 
+    switch (settings.SerialPort) {
+    case HwBrainRE1::SERIALPORT_SBUS:
+        return INPUT_TYPE_SBUS;
+    case HwBrainRE1::SERIALPORT_SBUSNONINVERTED:
+        return INPUT_TYPE_SBUSNONINVERTED;
+    case HwBrainRE1::SERIALPORT_DSM:
+        return INPUT_TYPE_DSM;
+    case HwBrainRE1::SERIALPORT_HOTTSUMD:
+        return INPUT_TYPE_HOTTSUMD;
+    case HwBrainRE1::SERIALPORT_HOTTSUMH:
+        return INPUT_TYPE_HOTTSUMH;
+    case HwBrainRE1::SERIALPORT_IBUS:
+        return INPUT_TYPE_IBUS;
+    case HwBrainRE1::SERIALPORT_SRXL:
+        return INPUT_TYPE_SRXL;
+    case HwBrainRE1::SERIALPORT_TBSCROSSFIRE:
+        return INPUT_TYPE_TBSCROSSFIRE;
+    default:
+        break;
+    }
+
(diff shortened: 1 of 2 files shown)


shared/uavobjectdefinition/hwbrainre1.xml
```diff
@@ -55,8 +62,16 @@
 				<option>LighttelemetryTx</option>
 				<option>MavLinkTX</option>
 				<option>MavLinkTX_GPS_RX</option>
+				<option>VTX Config TBS SmartAudio</option>
```
yamahabest, line 65: Was also on MultiPortSerial2, figured it should/could be here also.

shared/uavobjectdefinition/hwbrainre1.xml
```diff
@@ -55,8 +62,16 @@
...
 				<option>FrSKY Sensor Hub</option>
 				<option>FrSKY SPort Telemetry</option>
+				<option>S.Bus</option>
+				<option>S.Bus Non Inverted</option>
+				<option>DSM</option>
```
yamahabest, line 70: DSM should also be possible on these ports, right?

ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
```diff
@@ -200,6 +200,72 @@ Core::IBoardType::InputType BrainRE1::getInputType()
...
+        return INPUT_TYPE_IBUS;
+    case HwBrainRE1::MULTIPORTSERIAL_SRXL:
+        return INPUT_TYPE_SRXL;
+    case HwBrainRE1::MULTIPORTSERIAL_TBSCROSSFIRE:
+        return INPUT_TYPE_TBSCROSSFIRE;
+    default:
+        break;
+    }
+    
+    if (settings.MultiPortMode == HwBrainRE1::MULTIPORTMODE_DUALSERIAL4PWM)
```
yamahabest, line 245: Not sure if this check is needed?
Where can I see what's the default value of settings.MultiPortSerial2, if the MultiPortMode is set to "Normal"?

mluessi:
What is the use case for this? The only reason I can see why someone would want to run the receiver on a different port is that the RxPort is damaged or for a Rx protocol that includes telemetry and needs the Tx pin (e.g. TBS Crossfire). Other than that, it will just confuse users and lead to problems, e.g., the GCS gives them the option to connect S.Bus to the serial port but it doesn't have an inverter..

yamahabest:
Ok, ok, figured that it would also be strange to only have Crossfire on there.
No need to merge.

Would propose to keep the additional switch-cases with the Crossfire option from this change set, though.

yamahabest:
If purging the addition of the protocols, would the "VTX Config TBS SmartAudio" addition on MultiPortSerial still be ok?

mluessi:
It should work on MultiPortSerial2; it doesn't on MultiPortSerial due to the non-standard signal levels it uses which are not compatible with the bi-directional inverter on that port.

yamahabest:
Ok, created a new PR ( https://github.com/d-ronin/dRonin/pull/1734 ) for just the addition of crossfire to the getInputType function.

yamahabest:
https://github.com/d-ronin/dRonin/pull/1734

PR #1734, opened 2017-04-16
Extended getInputType checks with Crossfire

yamahabest:
Was added to the additional ports, but was not added here yet.

### DIFF
--- a/ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
+++ b/ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
@@ -200,6 +200,30 @@ Core::IBoardType::InputType BrainRE1::getInputType()
         break;
     }
 
+    switch (settings.SerialPort) {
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
```diff
@@ -200,6 +200,30 @@ Core::IBoardType::InputType BrainRE1::getInputType()
...
+    }
+
+    switch (settings.MultiPortSerial) {
+    case HwBrainRE1::MULTIPORTSERIAL_TBSCROSSFIRE:
+        return INPUT_TYPE_TBSCROSSFIRE;
+    default:
+        break;
+    }
+    
+    if (settings.MultiPortMode == HwBrainRE1::MULTIPORTMODE_DUALSERIAL4PWM)
```
yamahabest, line 217: Still not sure if this check is really needed.
Would be safe though.

tracernz (approved):
Matches recent firmware changes. :+1:

PR #1735, opened 2017-04-17
exti: factor common between all STM32 targets

mlyle:
Saves code, and makes it easier for exti to start moving towards the dio infrastructure.


tracernz (approved):

tracernz:
...

mlyle:
rebased, because the pios_exti lines were next to the pios_mpu lines in Makefile

PR #1736, opened 2017-04-18
f3: don't double-depend upon pios_dma.c

mlyle:
F3 PiOS stuff is brought in by wildcard, and pios_dma is newly factored
there but wasn't removed from their makefiles.  Eliminates an annoying
message about duplicate deps during build.


tracernz (approved):
Yes, also noticed this. Thanks. :+1:

PR #1738, opened 2017-04-18
f3 targets: enable autotune decimation

mlyle:
There's just too many configs that can hit low memory with autotune now
witout it.


mlyle:
Flew this on Lux, btw.

Issue #1697, opened 2017-04-06
Handling of upgrade of unitialized FC could be improved


yamahabest:
Issue details:
When saving the "upgraded" settings, the GCS complains that it has no settings.
Only way to get out of the wizard is to abort.

If there are no settings, the last step of saving the "upgraded" settings can be omitted probably.

tracernz:
I thought there was already a bug for this but I can't find one. Have encountered it many times but it's a little messy to fix.

PR #1739, opened 2017-04-18
GCS: Launch setup wizard when unconfigured board is connected

### LINKED ISSUE #1737: GCS needs to funnel unconfigured vehicles through setup wizard
Like the upgrade check- spot no mixer / output configuration and prompt to run the vehicle setup wizard, which in turn will take them to input wizard and then arming settings.

tracernz:
Got bored, wrote a little bit of code. This __always__ pops the setup wizard when a board without actuators configured is connected. Thoughts?

"unconfigured" means no non-zero channels in ActuatorSettings.

Fixes #1737.

### DIFF
--- a/ground/gcs/src/plugins/config/configplugin.cpp
+++ b/ground/gcs/src/plugins/config/configplugin.cpp
@@ -40,7 +40,7 @@ ConfigPlugin::ConfigPlugin()
 
 ConfigPlugin::~ConfigPlugin()
 {
-    // Do nothing
+    removeObject(this);
 }
 
 bool ConfigPlugin::initialize(const QStringList &args, QString *errMsg)
@@ -75,9 +75,9 @@ bool ConfigPlugin::initialize(const QStringList &args, QString *errMsg)
     connect(telMngr, &TelemetryManager::connected, this, &ConfigPlugin::onAutopilotConnect);
     connect(telMngr, &TelemetryManager::disconnected, this, &ConfigPlugin::onAutopilotDisconnect);
 
-    // And check whether by any chance we are not already connected
-    if (telMngr->isConnected())
-        onAutopilotConnect();
+    cmd->action()->setEnabled(false);
+
+    addObject(this);
 
     return true;
 }
@@ -95,7 +95,6 @@ UAVObjectManager *ConfigPlugin::getObjectManager()
 
 void ConfigPlugin::extensionsInitialized()
 {
-    cmd->action()->setEnabled(false);
 }
 
 void ConfigPlugin::shutdown()
@@ -108,7 +107,23 @@ void ConfigPlugin::shutdown()
   */
 void ConfigPlugin::onAutopilotConnect()
 {
+    // erase action
     cmd->action()->setEnabled(true);
+
+    auto pm = ExtensionSystem::PluginManager::instance();
+    auto uavoUtilManager = pm->getObject<UAVObjectUtilManager>();
+
+    if (uavoUtilManager->firmwareHashMatchesGcs()) {
+        // check if the board has been configured for flight
+        // if not, let's pop setup wizard automatically
+        if (!uavoUtilManager->boardConfigured())
+            Q_EMIT launchSetupWizard();
+    } else {
+        /** @todo upgrade firmware
+         * Currently the uploader plugin does this indepenently
+         * but it needs a big refactor because everything is done
+         * inside a UAVGadget */
+    }
 }
 
 /**
--- a/ground/gcs/src/plugins/setupwizard/setupwizard.cpp
+++ b/ground/gcs/src/plugins/setupwizard/setupwizard.cpp
@@ -51,7 +51,7 @@
 #include "vehicleconfigurationhelper.h"
 #include "actuatorsettings.h"
 
-SetupWizard::SetupWizard(QWidget *parent)
+SetupWizard::SetupWizard(bool autoLaunched, QWidget *parent)
     : QWizard(parent)
     , VehicleConfigurationSource()
     , m_controllerType(NULL)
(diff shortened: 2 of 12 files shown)


mlyle:
Holy crap that was fast.

My one thought is--- how does this play with the auto-upgrade wizard?  It would be nice if A) the auto-upgrade wizard's dialog always "won" and was presented first, so if someone plugs in an old seppuku and then starts GCS they don't get two weird things triggered (I understand the right thing probably happens if we catch the booatloader.  and B) if we presented a similar dialog for this and allowed a similar "ignore" to take place.

tracernz:
The auto-upgrade dialog wins if you plug the board in when GCS is running because it happens at the bootloader. This fires once the telemetry connection is actually established and the settings (and meta, and update-on-change) objects are all fetched. I will take a look at how it will play out if GCS is started when the board is already running+plugged-in. 

The reason I didn't want to add another dialog is because I didn't want to add yet another thing to click through. I agree it would be good so I propose to add an ignore button on the first screen of the setup wizard (but only if it was launched automatically). Do you think that's a good solution?

-e- Also need to verify how it plays with the upgrader when it boots firmware to import settings!

mlyle:
> The reason I didn't want to add another dialog is because I didn't want to add yet another thing to click through. I agree it would be good so I propose to add an ignore button on the first screen of the setup wizard (but only if it was launched automatically). Do you think that's a good solution?

That sounds great.

This is a bit of a tangent, but I'm thinking about replacing the 'Rescue' button / timer with a pulldown that is "On board connection:" 'Stop in bootloader' / 'Prompt for upgrades' / 'Immediately boot'.  Immediately boot would probably also bypass the post-boot firmware prompting.  Thoughts?

tracernz:
Just to keep this up-to-date, I have some WIP on my laptop to play nice with upgrader etc. Hoping to make upgrader offer setup wizard if no settings are imported (would fix #1697 too). Am out of town next week so likely to spend a bit of time on my laptop. ;)

tracernz:
This now only pops setup wizard if the firmware version matches GCS (in which case upgrader doesn't pop). Monkey-patched it so that it won't pop after the upgrader boots the new firmware too (very icky with current state of uploader). Doesn't go to setup wizard after upgrade if the board is unconfigured.

mlyle:
Sorry, has conflicts, needs rebase :/

mlyle:
Hey look I'm allowed to rebase

ground/gcs/src/plugins/config/configplugin.cpp
```diff
@@ -75,9 +75,9 @@ bool ConfigPlugin::initialize(const QStringList &args, QString *errMsg)
     connect(telMngr, &TelemetryManager::connected, this, &ConfigPlugin::onAutopilotConnect);
     connect(telMngr, &TelemetryManager::disconnected, this, &ConfigPlugin::onAutopilotDisconnect);
 
-    // And check whether by any chance we are not already connected
```
mlyle, line 78: So we're positive that initialization of all plugins (and all connections made) before autopilot connection?  I know it's generally slow/takes a few seconds.. I'm just curious if this is guaranteed by design.

ground/gcs/src/plugins/setupwizard/setupwizard.cpp
```diff
@@ -366,3 +368,25 @@ bool SetupWizard::saveHardwareSettings() const
...
     return helper.setupHardwareSettings();
 }
+
+void SetupWizard::boardIgnorePrompt()
+{
+    if (!m_autoLaunched)
+        return;
+
+    QMessageBox prompt(QMessageBox::Question, tr("Ignore Board?"),
+                       tr("Would you like to ignore the setup wizard for this board?"),
```
mlyle: This should be something like.. "This board doesn't appear to be configured.  Would you like to automatically configure it now with the dRonin Setup Wizard?"

ground/gcs/src/plugins/setupwizard/setupwizard.cpp
```diff
@@ -366,3 +368,25 @@ bool SetupWizard::saveHardwareSettings() const
...
+                       QMessageBox::Yes | QMessageBox::No);
+    if (prompt.exec() == QMessageBox::No)
```
mlyle: (and of course invert the sense of this if that wording is used)

mlyle (changes requested):
Looks good overall.  I think the dialog wording is the only thing that needs to change.

mlyle:
OK, I tested this and hadn't quite understood the dialog-- it comes up on cancel.

Maybe something like "dRonin Setup Wizard Cancelled!"   "Would you like to prevent the Setup Wizard from automatically opening for this board? (Only applies for this session)."

Otherwise, tested a few cases, seems good :)

ground/gcs/src/plugins/config/configplugin.cpp
tracernz: We don't have a device list until after all the plugins are loaded: https://github.com/d-ronin/dRonin/blob/next/ground/gcs/src/plugins/coreplugin/connectionmanager.cpp#L466

ground/gcs/src/plugins/setupwizard/setupwizard.cpp
tracernz: I'll take your wording in your later comment and apply it later today. :+1:

tracernz:

mlyle:

mlyle (approved):

tracernz:

and bring me a shrubbery

tracernz:
Tested on Win 10 with Seppuku (incl. upgrade). No problems.

PR #1741, opened 2017-04-19
uavtalk: tidiness

mlyle:
Had hoped to get further down the path of this one, but got waylaid prettying up uavtalk a bit first.


flight/UAVTalk/uavtalk.c
```diff
@@ -391,7 +377,7 @@ UAVTalkRxState UAVTalkProcessInputStreamQuiet(UAVTalkConnection connectionHandle
 		}
 
 		// Check the lengths match
-		if ((iproc->rxPacketLength + iproc->instanceLength + iproc->timestampLength + iproc->length) != iproc->packet_size) { // packet error - mismatched packet size
+		if ((iproc->rxPacketLength + iproc->instanceLength + iproc->length) != iproc->packet_size) { // packet error - mismatched packet size
```
tracernz, line 343: Need to keep this so we can still receive timestamped packets okay (even if we just discard the bytes further down). Either that or we need to bump protocol version because removing this would break BC (and our current public protocol definition, would need to update that too) (I really don't think we should do that).

tracernz (commented):
Ah, had thought this was just a reformat from the title but it's actually a functional change that breaks the protocol spec. Everything else is fine except that we need to still receive packets with timestamps correctly, and just discard the timestamp bytes. 

Could bring #1758 into this (I wrote that before reading this, and had thought this was just a reformat).

flight/UAVTalk/uavtalk.c
mlyle: I don't think flight code ever receives timestamped information-- the only place timestamped information is used is in the interpretation of logs.  Removing the ability for flight code to receive timestamped information seems like a good idea.

That is, there's messages flight will no longer accept with this change, but they were messages that were never sent [except to logs] and didn't make sense.  Bumping the version would actually break BC.

Updating the specification to clarify that timestamping is only for logs makes a lot of sense, though.

tracernz: > I don't think flight code ever receives timestamped information-- the only place timestamped information is used is in the interpretation of logs. Removing the ability for flight code to receive timestamped information seems like a good idea.

Not from us but it's a publicly documented protocol that's been around for some time. There exist a number of 3rd party implementations in the wild and they are all free to send timestamped packets of any type (perfectly valid under the protocol spec). For the sake of a few lines it makes no sense to change the protocol. There is a special place in hell for people who break their own public protocol specs IMO. Wastes lots of time trying to dig through other peoples code trying to figure out why their stuff doesn't do what it said it would do (and used to actually do) on the box.

> Bumping the version would actually break BC.

That's the point, if you make a BC change you should also change the version. How to reconcile that with the multiple other forks of this codebase is another issue that comes up when you do that.

mlyle: > There exist a number of 3rd party implementations in the wild and they are all free to send timestamped packets of any type (perfectly valid under the protocol spec)

If there's any non-logging implementation that sends these timestamps I would be more interested.

I don't want to be caught with design mistakes made 5 years ago forever.  We can change uavtalk.  There's a big advantage from simplifying this down, both in message bits we're reclaiming for possible future use (though this has potential caveats and has to be handled carefully), simplifying state machines, and enabling further simplification.

"Just for a few lines of code" also happened to fix a severe bug that you've fixed in another PR, for instance.

> That's the point, if you make a BC change you should also change the version.

You should change the version if you do something that will cause information from past revisions to be misinterpreted.

With the protocol version change: all messages are rejected from old implementations.

Without the protocol version change: messages we do not believe sent anywhere are rejected.

There are far bigger changes we've made with interoperability considerations.  E.g. the way some things are quasi-pipelined now that were serialized before.  The new implementation (as of a few months ago) is still compliant with the specification, but broke interop with most old implementations.

tracernz:
We debated this on IRC a bit and pretty much came to the conclusion that it should go ahead. Will need follow-up with protocol doc changes.

mlyle:
Thank you for this merge-- really helps / unblocks me / takes away my excuses for slacking.

PR #1744, opened 2017-04-22
Add DMAShot to Revo.

glowtape:
Needs testing I guess. I don't have those boards.

### DIFF
--- a/flight/targets/revolution/board-info/board_hw_defs.c
+++ b/flight/targets/revolution/board-info/board_hw_defs.c
@@ -1432,12 +1432,12 @@ static const struct pios_dmashot_timer_cfg dmashot_tim_cfg[] = {
 	},
 	{
 		.timer = TIM9,
-		.master_timer = TIM4,
+		.master_timer = TIM1,
 		.master_config = TIM_DMA_Update | TIM_DMABase_CCR2,
 
-		.stream = DMA1_Stream6,
-		.channel = DMA_Channel_2,
-		.tcif = DMA_FLAG_TCIF6,
+		.stream = DMA2_Stream5,
+		.channel = DMA_Channel_6,
+		.tcif = DMA_FLAG_TCIF5,
 	},
 	{
 		.timer = TIM2,
--- a/flight/targets/revolution/board-info/board_hw_defs.c
+++ b/flight/targets/revolution/board-info/board_hw_defs.c
@@ -1407,6 +1407,60 @@ static const struct pios_tim_channel pios_tim_servoport_rcvrport_pins[] = {
 };
 
 #if defined(PIOS_INCLUDE_SERVO) && defined(PIOS_INCLUDE_TIM)
+
+#if defined(PIOS_INCLUDE_DMASHOT)
+
+#include <pios_dmashot.h>
+
+/*
+TIM3 C3,C4
+TIM9 C2
+TIM2 C3
+TIM5 C2,C1
+
+DShot is mostly just popular with miniquads, those being actual quads,
+at most hex. So six is enough. Rest is gonna use bitbanging, if user
+wants to use these pins.
+*/
+
+static const struct pios_dmashot_timer_cfg dmashot_tim_cfg[] = {
+	{
+		.timer = TIM3,
+		.stream = DMA1_Stream2,
+		.channel = DMA_Channel_5,
+		.tcif = DMA_FLAG_TCIF2,
+	},
+	{
+		.timer = TIM9,
+		.master_timer = TIM4,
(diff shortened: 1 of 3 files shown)


yamahabest:
Could Sparky2 and AQ32 also do DMAShot?

mlyle:
Maybe.  Would take analysis to figure out DMA maps and then test, which seems like a bit of a waste for targets that are no longer commercially available and that people aren't using for new builds.

yamahabest:
True, didn't know the effort required. Thanks.

glowtape:
Sparky2 has a really wild DMA map. Ideally, you'd at least want the first four pins all to be doing DMA DShot, but Sparky2's mapping throws a huge wrench into it already.

AQ32 looks tame, but I don't have one to stick my logic analyzer on it to verify everything.

jihlein:
AQ32 is straightforward, I can implement and test via logic analyzer.

jihlein:
Opened #1755 for DMAShot on AQ32

mlyle:
I'm not inclined to spend time personally testing this until after neat (have a lot of stuff to get through).  If it gets tested (making sure all outputs are happy), I'm glad to merge.

That is, this a very-nice-to-have, but reducing CPU use is not mandatory for rel like some other things in progress.

glowtape:
I've ordered a Revo clone a while back on Banggood (together with a SPRF3E to implement DMA on F3), so it's just a matter of time until I can get to it.

glowtape:
Is the Quanton target just for the board itself, or does it act as stand-in for other ones?

mlyle:
Colibri (as is on Gemini)  is a pin-compat to quanton

tracernz (approved):
Didn't check the datasheet because I'm lazy/assume you got the mapping right. 😁

glowtape:
Revo has been checked, since I've received a clone today. The Quanton, well, I don't have one :V But I've just noticed I need to fix the comments, it's a copypasta from Seppuku.

tracernz:
Okay. I can check Quanton for you if you remind me in a few hours, or later in the week.

tracernz:
Sorry for slowness. Quanton doesn't look right:
image

image

mlyle:
I hope what that shows is -- A) channels 5-8 being implemented by the software path (boo).. and B) marginal grounding make it look like those other channels are toggling when other edges happen.  -- they'd better not be togglng.

tracernz:
I thought that but the logic is solidly grounded to the Quanton. Might have to check it out with an analog scope.

glowtape:
Weird that the parser says the first channel is zero, despite having bits set. Probably the zero character instead of value, since for some reason Logic actually shows the ASCII character instead of the value for 0-255 and '0' happens to be 48. Channel 2 is borked due to those stray pulses.

And yeah, TIM10-12 aren't configured for DMA, because they all require a master timer.

flight/targets/quanton/board-info/board_hw_defs.c
```diff
@@ -1816,6 +1816,60 @@ static const struct pios_tim_channel pios_tim_servoport_rcvrport_pins[] = {
...
+		.channel = DMA_Channel_5,
+		.tcif = DMA_FLAG_TCIF2,
+	},
+	{
+		.timer = TIM1,
+		.stream = DMA2_Stream5,
+		.channel = DMA_Channel_6,
+		.tcif = DMA_FLAG_TCIF5,
+		// Shares DMA channel with WS2811!
+		// DMAShot requests very high priority, so should get preference.
```
mlyle: Channels are per-stream.. I don't see any use of DMA2_Stream5 elsewhere.

Priority is just used to select which of multiple streams that want to DMA at the same time... who wins.

mlyle:
It may not be worth doing dma dshot for quanton-- if channels 5-8 can't do it and requires the software impl anyways.  I assume most flying Quantons are hexes and octos.

glowtape:
I think I might be able to make something work, probably just need to add some changes to force the lowest DShot rate, if there's a shared master timer.

mlyle:
But what's the point?  Bit-banging is only a problem with OSD targets.  DMA on F3 might be beneficial-- save some scarce cycles..  but otherwise?

glowtape:
Yea you're right. Figured it'd be nice to have, in case someone wants to support a future third party with another fun configuration, but they're all probably going to be designed with DShot in mind. The two channels on one timer was an extra special case that got too annoying.

tracernz:
~~~So what's the consensus on this? Dropping Quanton?~~~
-e- Already done. :+1:

PR #1749, opened 2017-04-27
Neo-Autotune

mlyle:
This works.  However, the code is very ugly still, and ~~there is no mechanism to retrieve autotune data from the flight controller.~~


mlyle:

mlyle:
Doesn't build on Windows (because of C++ abuse-- VLAs, 0 len struct element)

mlyle:

mlyle:

Hoped that this A) calculates averages properly on Windows [need to see warnings output], B) is a little less noise sensitive.

ufoDziner:
Tested this on windows with 2 different quads, both with Seppuku FCs.  Saw a weird issue on both where it showed the FC 5.86deg off in pitch, and the other 7.36 degrees off in pitch.  For the first tunes, I allowed these amounts to be set by the GCS so that the board showed level with the artificial horizon.  The results are the attached "1" autotunes.  
purple1 - https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA0MH7yQkM

white1 - https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA0KbyxQsM

I then set roll and pitch back to 0 (where they should be for these installations, and ran autotune again on both.  The results are the "2" autotunes.

purple2 - https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA0KG0xQoM

white2 - https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA0IjQ3AoM


[autotunes.zip](https://github.com/d-ronin/dRonin/files/965851/autotunes.zip)

glowtape:
I think the suggested values to use here are dampening 1.05 and noise sensitivity 1.3?

With these values, I get a pretty snappy tune, however it seems like the D-term is too high. In hangtime and at the very low end of the throttle range, my motors start to rattle. Which goes away when I disable the D-term (altho the quad becomes "springy" in faster manoeuvers) or becomes relatively negligible at sensitivity 1.0.

As far as motor temps go, can't tell much, because black anodized motors in bright sunshine don't help.

I need to make another pass of tests when it's overcast and with OpenLager. If per latter the actuators don't do much drama at normal throttle range with the 1.05/1.3 settings, there'd likely need something to be done about the D-term at the bottom of throttle.

glowtape:
Addenum, with the 1.0 sensitivity setting, it seems like the quad also has a harder time to get up to the desired rate on the heavier axis, in my case pitch. Bumping the sensitivity raises P-terms, might be related.

glowtape:
So, where do the 1.05 dampening and 1.3 sensitivity come from?

I went out flying today and set it to 1.0/1.0 on a whim, half expecting drama. Surprisingly, the response of the quad is very sharp and no obvious oscillations (apart from the shit my prerelease BLHeli_32 cause*). Tried quick medium blips on the pitch and roll sticks, it follows it nicely and doesn't oscillate. Dropping from up high and throttling up hard doesn't cause none of this propwash crap everyone is yapping about. Practically no rattling and the motors barely reach body temperature.

Quad specs: Tweaker 5", Emax RS2205 2300kv, Seppuku, Airbot Wraith32+, HQProp 5x4.3x3, ~620gr, LPF: 2nd 80hz, RC smoothing to Extended on Axes and Throttle.

(*: I still need to solder in the release BLHeli_32. The prerelease firmware on mine doesn't recognize old school PWM signals, and by what sskaug told me, the DMA for DAC and CC capture are interfering with each other, so the thing twitches oddly once a while. There's certainly something up. Gonna have to cross-check with my OpenLager eventually, so ensure the gyro isn't feeding crap, though. However it started as soon as I switched the KISS with the BLHeli ones, on the same firmware build and config, so I'm fairly certain it's the ESCs.)

We should get the other guys to try 1.0/1.0 with neotune, just to get some more opinions.

glowtape:

mlyle:
This is now code complete.  deps #1810 #1813 

Out of scope-- I think there needs to be a little more UI and some factoring cleanup.  Hope to be subsequent PR?

Also need code for AQ32 to erase autotune parts.

I have only tested this with sim-- need to test on real hardware, etc-- will over the next couple days.

mlyle:

mlyle:
rebased

ground/gcs/src/plugins/scope/scope.pro
```diff
@@ -6,6 +6,8 @@ DEFINES += QWT_DLL
 
 include(../../gcsplugin.pri)
 include (scope_dependencies.pri)
+
+INCLUDEPATH += ../../libs
```
tracernz: gcs.pri (which you're getting through gcsplugin.pri) already gets you this https://github.com/d-ronin/dRonin/blob/next/ground/gcs/gcs.pri#L99

flight/Modules/Autotune/autotune.c
```diff
@@ -240,55 +194,23 @@ static void at_new_actuators(UAVObjEvent * ev, void *ctx, void *obj, int len) {
...
 
-	q_item->throttle = actuators.Thrust;
-
-	if (circ_queue_advance_write(at_queue) != 0) {
-		last_sample_unpushed = true;
-	} else {
-		last_sample_unpushed = false;
-	}
+	update_counter++;
+	throttle_accumulator += 10000 * actuators.Thrust;
```
tracernz, line 199: Not even close to saturating. :D

flight/Modules/Autotune/autotune.c
```diff
@@ -240,55 +194,23 @@ static void at_new_actuators(UAVObjEvent * ev, void *ctx, void *obj, int len) {
...
-	system_ident.Bias[SYSTEMIDENT_BIAS_ROLL]    = X[10];
-	system_ident.Bias[SYSTEMIDENT_BIAS_PITCH]   = X[11];
-	system_ident.Bias[SYSTEMIDENT_BIAS_YAW]     = X[12];
-	system_ident.Tau                            = X[9];
-
-	system_ident.Noise[SYSTEMIDENT_NOISE_ROLL]  = noise[0];
-	system_ident.Noise[SYSTEMIDENT_NOISE_PITCH] = noise[1];
-	system_ident.Noise[SYSTEMIDENT_NOISE_YAW]   = noise[2];
-
+	system_ident.NewTune  = new_tune;
```
tracernz: Is this supposed to be aligned (has more than 1 space)?

flight/Modules/Autotune/autotune.c
```diff
@@ -369,37 +281,31 @@ static void AutotuneTask(void *parameters)
...
-#ifdef AUTOTUNE_AVERAGING_MODE
 				if (autotune_save_averaging()) {
 					// Try again in a second, I guess.
 					PIOS_Thread_Sleep(1000);
 					continue;
 				}
-#endif /* AUTOTUNE_AVERAGING_MODE */
 
-				// Save the settings locally.
+				float hover_throttle = ((float)(throttle_accumulator / update_counter))/10000.0f;
```
tracernz: needs spaces

flight/Modules/Autotune/autotune.c
```diff
@@ -408,426 +314,43 @@ static void AutotuneTask(void *parameters)
...
-				}
-
 			case AT_RUN:
-				while (true) {
-					struct at_queued_data *pt;
+				(void) 0;
 
-					/* Grab an autotune point */
-					pt = circ_queue_read_pos(at_queue, NULL, NULL);
+				float hover_throttle = ((float)(throttle_accumulator / update_counter))/10000.0f;
```
tracernz: Needs spaces. Do we even need to update the UAVO while the tune is in progress?

flight/Modules/Stabilization/stabilization.c
```diff
@@ -782,7 +782,10 @@ static void stabilizationTask(void* parameters)
 							// number of ident cycles.
 							measure_remaining &= ~ident_mask;
 						}
+					}
 
+					if(flightStatus.Armed != FLIGHTSTATUS_ARMED_ARMED) {
```
tracernz: Space

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -407,23 +395,309 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
     }
 }
 
+// XXX appropriately encapsulate all this crap
+#define ATFLASH_MAGIC 0x656e755480008041 /* A...Tune */
```
tracernz, line 399: It's really unfortunate for GCS to know about stuff like this. The FC firmware should be responsible for low-level filesystem stuff like checking that a partition is valid/coherent.

I must say I'm having second thoughts about adding the file extension to UAVTalk, because it significantly weakens the strength of the design (which was one of the things that brought me to the project). Using it like this loses the well-defined protocol between GCS and the flight board. :/

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -407,23 +395,309 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
...
+    struct at_flash_header hdr;
+
+    struct at_measurement data[];
+};
+
+/* Run a Butterworth biquad filter on a circular buffer.  First go around once
+ * to "prime" the filter, then actually filter in place.  This is derived from
+ * @glowtape's excellent flight implementation
+ */
+void biquad_filter_in_place(float cutoff, int pts, float *data)
```
tracernz, line 428: Plans to refactor all this C into C++? Algorithms like this probably don't really belong within an MVC class in any case.

flight/Modules/Autotune/autotune.c
```diff
@@ -50,28 +50,18 @@
 #include "pios_thread.h"
 #include "systemsettings.h"
 
-#include "circqueue.h"
 #include "misc_math.h"
 
 // Private constants
-#define STACK_SIZE_BYTES 1340
+#define STACK_SIZE_BYTES 640
```
tracernz, line 56: :+1:

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -140,58 +153,32 @@ QString ConfigAutotuneWidget::systemIdentValid(SystemIdent::DataFields &data, bo
...
-    if (data.Tau < -5.3) {
+    if (data.tau[0] < 0.005) {
         // Too low of tau to be plausible. (5ms)
         retVal.append(tr("Error: Autotune did not measure valid values for this craft (low tau).  "
                          "Consider slightly lowering the starting roll/pitch rate P values or "
                          "slightly decreasing Motor Input/Output Curve Fit on the output pane."));
         retVal.append("<br/>");
         *okToContinue = false;
-    } else if (data.Tau < -4.9) {
+    } else if (data.tau[0] < 0.0074) {
```
tracernz, line 163: Nice to deal with this stuff in eng. units. :+1:

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -407,23 +395,309 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
...
+}
+
+/* Returns number of samples of delay between series */
+float get_sample_delay(int pts, const QVector<float> &delayed, const QVector<float> &orig)
+{
+    ffft::FFTReal<float> fft(pts);
+
+    /* XXX avoid all these VLAs on stack */
+    /* Convert to frequency domain */
+    QVector<float> delayed_fft(pts);
```
tracernz: Keeping an eye on the Qt dev mailing list... I'm weary of using Qt containers that are more or less identical to STL containers unless they do something the STL one cannot.

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -407,23 +395,309 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
...
+        float sample_tau = get_sample_delay(pts, gyro_deriv, actu_desired);
+
+        float tau = sample_tau / flash_data->hdr.sample_rate;
+
+        biquad_filter_in_place(1 / (sample_tau * M_PI * 1.414), pts, actu_desired.data());
+
+        QVector<float> gyro_sorted = gyro_deriv;
+        QVector<float> actu_sorted = actu_desired;
+
+        qSort(gyro_sorted.begin(), gyro_sorted.end());
```
tracernz: Use std::sort. qSort has been deprecated for some time http://doc.qt.io/qt-5/qtalgorithms.html#porting-guidelines.

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -407,23 +395,309 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
...
+    if (precalc_vals) {
+        av = *precalc_vals;
+    } else {
+        av.valid = false;
+    }
+
+    if (!av.valid) {
+        // XXX needs factoring to a step
+        ExtensionSystem::PluginManager *pm =
+            ExtensionSystem::PluginManager::instance();
```
tracernz: This can easily fit on one line (100 cols)?

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -407,23 +395,309 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
...
+
+        QByteArray *data = telMngr->downloadFile(4, 32768);
+
+        if (data) {
+            av = processAutotuneData(data);
+        }
+    }
+
+    if (!av.valid) {
+        // XXX do something smart
```
tracernz, line 691: We're doomed then! :stuck_out_tongue:

ground/gcs/src/plugins/config/configautotunewidget.h
```diff
@@ -122,17 +129,20 @@ class ConfigAutotuneWidget : public ConfigTaskWidget
     ConfigGadgetWidget *parentConfigWidget;
     static const QString databaseUrl;
 
-    QString systemIdentValid(SystemIdent::DataFields &data, bool *okToContinue);
+    QString tuneValid(AutotunedValues &data, bool *okToContinue);
```
tracernz: Method can be const? It's a bit of a weird method tbh. It would be much better to return a bool and pass a ref to an error string;

ground/gcs/src/plugins/config/configautotunewidget.h
```diff
@@ -122,17 +129,20 @@ class ConfigAutotuneWidget : public ConfigTaskWidget
...
+    QJsonDocument getResultsJson(AutotuneFinalPage *autotuneShareForm, AutotunedValues *av);
 
     void stuffShareForm(AutotuneFinalPage *autotuneShareForm);
     void persistShareForm(AutotuneFinalPage *autotuneShareForm);
     void checkNewAutotune();
 
 private slots:
     void openAutotuneDialog();
-    void openAutotuneDialog(bool autoOpened);
+    void openAutotuneDialog(bool autoOpened, AutotunedValues *precalc_vals = 0);
```
tracernz: nullptr or Q_NULLPTR

tracernz (commented):
First pass, looks like a good start. :+1: 

No longer blocked-on-deps?

flight/Modules/Autotune/autotune.c
mlyle: The plan is for the task to go away, and it's not much more work to do so.  I just wimped out and decided to go this way for now.

flight/Modules/Autotune/autotune.c
mlyle: Strictly no, but it's a nice to have if you have a telemetry link or logger.

ground/gcs/src/plugins/config/configautotunewidget.cpp
mlyle: Files is really only for two things, IMO:

1. This--- a giant lump of samples in a special-purpose log file-- it's not really a filesystem, just a header + set of points.  It isn't really a clean design to do it over uavobjects (requires something like asking for a range of samples by uavo and them showing up in some other object, plus receiving the control information).
2. Shuttling logs back (so uh, uavtalk stream in uavtalk heh).

Maybe also for downloading partitions without being in the bootloader, and for tunneling stream-oriented stuff like RTK.

ground/gcs/src/plugins/config/configautotunewidget.cpp
mlyle: Plans to encapsulate it in classes.. I'm not sure about how much further to go.

Kinda **regret** doing it for the other stuff, rather than just doing it c style ,given that this code may need to get factored back to flight in the future.

ground/gcs/src/plugins/config/configautotunewidget.cpp
mlyle: OK, will in future :(

tracernz:

pug398:
I tried the latest NeoTune build on both a ZMR250 and a 280 Robocat (Sparky) and tune worked great first time. Didn't have to change anything.

tracernz:
Okay, this is ready to go in, and further work to clean-up improve will come in another PR.

raemin:
Same story for me tried "dronin-win_20170619-559de04f.zip", made a small flight with previous settings, then changed the props launched the autotune and quad is now perfectly smooth with stock noise and damping settings. Flew it 7minutes with these new settings.

Aircraft described in the autotune log (BrainRE1)
https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAkP7t7gkM. 

The autotune process took longer than usual, but there was quite some wind.

The graphics of the GCS aren't as polished as usual (no graphics in the health gadget for instance), but I like it better actually.

Thanks guys!

PR #1753, opened 2017-04-29
GCS:CrashReporter: Include Qt version info

tracernz:
I could infer this from dRonin rev but this way is simpler.


mlyle:
I would tend to want to keep version the same, and just make the field "optional"-- its own presence self-describes and there's no backwards or forwards compat issues.

If the meaning of a field changes, or things that have been mandatory for the whole life of a version disappear, or there's a gross reformatting of things-- that's when I'd bump versioning.

What do you think?

tracernz:
I'm okay with that, changed back to 1.

PR #1754, opened 2017-04-29
AQ32: Add ComBridge to UART 1 and UART 2

yamahabest:
Don't know why it was omitted, but my MinimOSD is connected to UART1...

Tested and works @ UART1.
No issues expected at UART2.


tracernz:
Thanks. :+1:

PR #1755, opened 2017-04-29
Flight: AQ32 - Implement DMAShot

jihlein:
Adds DMAShot to AQ32.  All 10 channels supported.  I have no DShot compatible ESCs to test with, but logic analyzer traces appear correct for all channels.

Channels 1 - 6:
dmashot1-6

Channels 7 - 10:
dmashot7-10


glowtape:
LGTM.

Given that the pulsing between different timers doesn't line up completely, you know it's the DMA stuff running and not GPIO.

PR #1757, opened 2017-04-30
Uncrustify: Explictly specify C, Fix pointer spacing

tracernz:
Uncrustify defaults to C++ for headers which doesn't make sense for `uncrustify_flight`. Default pointer/star config doesn't match our coding standard, add it to our config.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1099,7 +1099,7 @@ ifneq ($(strip $(filter uncrustify_flight,$(MAKECMDGOALS))),)
 endif
 
 .PHONY: uncrustify_flight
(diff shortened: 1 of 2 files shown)


tracernz:
Actually the uncrustify config doesn't match the documented coding style on several places just from a quick check. First two things I looked at: cmt_cpp_to_c should be true, code_width should be 80, not 144...

mlyle:
I made those last couple things a bit more permissive than the actual standard to control the amount of diff to be somewhat sane... Blindly changing some of that stuff automatically doesn't necessarily have a pretty result (can result in interleaved short-long lines, excessive diff, etc).

PR #1759, opened 2017-04-30
Packaging: deb package only works on Ubuntu, tag it as such

tracernz:
Due to pulseaudio package being named differently between Debian and Ubuntu repos or some such (reported by IRC user OneSplot).
Also update the mac package name while there.

@mlyle Might want to run this on the current release because it's already caused people problems.


mlyle:
I'm a bit confused by the report-- isn't the package in question in Jessie?

https://packages.debian.org/jessie/libpulse-mainloop-glib0

mlyle:

PR #1760, opened 2017-05-02
Preliminary support for OmniPreSense mm-wave radars on pi/sim

mlyle:
New driver, etc.  Some enhancements to hardware initialization on pi to allow it to work via callbacks instead of dedicated tasks (e.g. like receiver protocols work on other targets).

https://omnipresense.com/index.php/product/ops241-a/


mlyle:
Status so far: tested in sim mode on Linux, with me typing in ranges over telnet manually.  About to build and get on Pi... hopefully integrated with and flight tested tomorrow.

The current modules provide just doppler (velocity) information, so utility is limited.  But it will be interesting to compare the data against our current vel estimates.  There is no integration with actual state estimation or control path yet.

No effort has been made to correct velocities or ranges for actual attitude of craft.  (It's not clear to me we should do this in the driver, either, instead of up in the attitude estimation loop).   TBD.

mlyle:
image

Ranges make it end to end (with manual entry) on Pi.

mlyle:
@yamahabest fixed thx :P

mlyle:
This doesn't work in practice, for poorly understood reasons having to do with USB-CDC serial and tcgetattr/tcsetattr hanging on the module.

tracernz:
> This doesn't work in practice, for poorly understood reasons having to do with USB-CDC serial and tcgetattr/tcsetattr hanging on the module.

😢

tracernz (approved):
Didn't spot anything. 👍

mlyle:
Yah, the device in question does odd things when you stty it.. though screen works.  It's similar to our CDC bugs, really.  Will be looking to work-around it.

mlyle:
Rebased.  Still needs verification.

mlyle:
image

Successfully flight tested.

tracernz:
Got an unrelated commit in here somehow (88ac460).

mlyle:
Good eye.

I'm not sure we want to merge this anyways-- right now the sensors are not plugged into the flight loop, and only provide doppler (vel) information.

tracernz:
> I'm not sure we want to merge this anyways-- right now the sensors are not plugged into the flight loop, and only provide doppler (vel) information.

Still neat though. Unmerged things tend to rot.

mlyle:
image

tracernz:
Thoughts on merging? I'm happy to press the button if you think you can use this in the future.

mlyle:
Sure, might as well do ti, I guess :D

PR #1764, opened 2017-05-07
sensors, pios_sensors: Ditch sensor queues and read them out in Sensors module.

glowtape:
Putting this up for critique. Probably quite a bunch of issues I'm not seeing (also, kind of terri-bad).

The code co-habitates the pios_sensor.c|h files, because it uses a similar name and is supposed to replace it eventually. I've loosely followed the DIO naming scheme, all lower case, no PIOS_ prefix.

This current patch only processes the old queued sensors, but an earlier test version ran fine with converted sensor drivers (those that run on Seppuku), but I started over to get a clean base and left them out (also I needed to test whether it processes old queued stuff properly).

Here's roughly how things are supposed to work:
- Sensors module sits on the gyro semaphore and waits (via sensors_wait_on()).
- Gyro driver triggers semaphore and causes Sensors to run.
- Sensors are being read out by running a callback. First gyro, then accel.
- Then it reads out only _one_ more of the auxilliary sensors, based on LRU. And whether it is allowed to run anyway, some drivers do sleep in their sensor task, there's sort of scheduling in the new code.

There's provisions to tell a driver to do something and then wait for the result ("cueing"), since that option was desired for future stuff. But I'm not yet sure whether that works properly yet (need to set up a testing case). Certainly doesn't break the normal stuff. It currently uses hints whether the sensor messes with I2C or SPI (maybe more hints if needed), potentially to run two cues in parallel, if they don't interfere.

Eventually the Attitude module is going to get decimated, IIRC, so it might be an idea to stick the accel into the LRU switch, too. Then again, most IMUs usually pair the gyro and accel readouts.

I'm not entirely sure I have this missing sensors toleration stuff right in sensors.c, I had to scratch my head a few times.

### DIFF
--- a/flight/Libraries/sanitycheck.c
+++ b/flight/Libraries/sanitycheck.c
@@ -319,7 +319,7 @@ static int32_t check_safe_autonomous()
 static int32_t check_stabilization_rates()
 {
 	const float MAXIMUM_SAFE_FRACTIONAL_RATE = 0.85;
-	int32_t max_safe_rate = PIOS_SENSORS_GetMaxGyro() * MAXIMUM_SAFE_FRACTIONAL_RATE;
+	int32_t max_safe_rate = PIOS_Sensors_GetRange(PIOS_Sensors_GetSensor(PIOS_SENSOR_GYRO)) * MAXIMUM_SAFE_FRACTIONAL_RATE;
 	float rates[3];
 
 	StabilizationSettingsManualRateGet(rates);
--- a/flight/Modules/AltitudeHold/altitudehold.c
+++ b/flight/Modules/AltitudeHold/altitudehold.c
@@ -108,7 +108,7 @@ int32_t AltitudeHoldInitialize()
 	}
 #endif
 
-	if (!PIOS_SENSORS_GetQueue(PIOS_SENSOR_BARO)) {
+	if (!PIOS_Sensors_Available(PIOS_SENSOR_BARO)) {
 		module_enabled = false;
 		return -1;
 	}
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -324,7 +324,7 @@ static void AttitudeTask(void *parameters)
 	last_algorithm = 0xfffffff;
 	last_complementary = false;
 
-	uint16_t samp_rate = PIOS_SENSORS_GetSampleRate(PIOS_SENSOR_GYRO);
+	uint16_t samp_rate = PIOS_Sensors_GetUpdateRate(PIOS_Sensors_GetSensor(PIOS_SENSOR_GYRO));
 
 	if (samp_rate) {
 		dT_expected = 1.0f / samp_rate;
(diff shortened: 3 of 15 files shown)


glowtape:
pios_mpu is tentative. For some reason it blows up everything on SPRF3e, while it works fine on OG Brain with the I2C MPU9250, as well as on that Revo clone of mine with the whatever it is (SPI MPU6xxx). Tried excluding mag to have it act like a MPU6xxx sensor, still blows up.

glowtape:
Heh, fixed. Some memalloc bug that I should be ashamed of.

glowtape:
I think I'm done with this for now. I need someone to point out any terrible flaws I might be missing at this point.

CPU savings are modest at best. It's around 1% on both F4 and F3, how much exactly I'm not sure, with this rounding thing. Given that the per-sensor tasks are gone, there should be less scheduling drama going on and be less jittery.

The BMI160, LIS3, BMP280 and MPU drivers are converted, the rest are still using the queues, that'll get wrapped. I don't have hardware for the others to test, but there shouldn't be any noticeable overhead.

If Jenkins finishes compiling this successfully, I'll yank the WIP tag.

--edit:
And oh, there's a bunch of provisions to deal with multiple gyros at some point. Should F7 support make it in eventually, I'll get Airbots dual gyro board and finalize the stuff, or if there's gonna be an F4 board with two (or more) gyros, whatever comes first. Some of the drivers need a little untangling to run multiple instances, anyway.

glowtape:
I need to redo pios_mpu after #1783.

mlyle:
I need to look at this in depth.  It's a bit terrifying in scope :D  

Any useful memory savings?

glowtape:
Looks way worse than it actually is in regard to code complexity.

As far as memory savings go, on my Seppuku with its flying config loaded, it saves 1632 bytes of heap and 380 bytes of fast heap. 36 bytes short of 2KB.

glowtape:
Rebased, optimized some duplicate code stuff in the BMI160 and MPU callbacks, other tiny fixes.

mlyle:
Rebased

glowtape:
Pushed a copy to have a backup. Made some changes in regards to backwards compatibility for old drivers, need to ditch driver conversion commits to test.

glowtape:
Surprisingly these changes work fine.

OK, renamed the namespace and functions as requested, dropped the multisensor prep work, ditched the semaphore reset, turned the old code needed to run unconverted drivers some more into an abstraction.

glowtape:
Flew this in the current state earlier. Still works with no apparent issues after the namespace change.

glowtape:
Rebased.

Probably gonna cut the cueing stuff out. Doesn't seem to be a need for it, and it'd make things a little bit easier to read. If there's gonna be a need eventually, I'll cross the bridge when it comes to that.

glowtape:
Cueing is out. Wasn't really tested, anyway.

mlyle:
Superseded #1987

PR #1762, opened 2017-05-04
sim/pi: support I2C

mlyle:
status: hmc5883 tested, px4flow not (have a cbaling issue)

PR #1765, opened 2017-05-09
Synchronous gyro interrupts on flyingpi

mlyle:
depends on ~~#1714~~ , ~~#1762~~, and ~~#1779~~ 

Still a little hackish (as this call is not universal between stm32 and posix yet, so bmx055 can't work on STM32).  But this seems to work, and is most of the way tehre.


tracernz:
Can you rebase please? Should make it easier to see what actually belongs to this branch too. :grin:

mlyle:
Rebased :) thanks for merges, etc.

I am not sure why it needed rebase as the actual rebase was conflict free.

mlyle:
Note that... this should sit at least a short time while we get a better handle on what @glowtape's doing... it's quite possible we need something more "interrupt-like" for sim and shouldn't be moving to this "wait on pinchange" impl.

tracernz:
> Rebased :) thanks for merges, etc.

No prob. Sorry I've been unavailable for a few days. I'll catch up with you on IRC when I'm at home (run out of mobile data :'().

> I am not sure why it needed rebase as the actual rebase was conflict free.

Github seems quite random with when it sees a conflict.

mlyle:
I am not sure I like this and want to defer.

mlyle:

mlyle:

mlyle:
Putting this forward for review.  It makes flyingpi better, even though it breaks abstraction a little currently.

mlyle:
Sorry.  I changed my mind again.  defer for now.

PR #1769, opened 2017-05-10
simulation/posix: try to report serial number to GCS

mlyle:
Checks for serial number in /proc/cpuinfo, and if that's no good checks the DMI info.  This should cover most linux systems (though the DMI info requires root).

Makes it so usage data, autotune, etc, is useful.


mlyle:

mlyle:

tracernz:
Did you consider linking OpenSSL rather than importing code (should be available on *nix and we install it in our tools dir on Windows)?

mlyle:
I did think about it, yes.. libcrypto is kind of a pain in various ways and.. it'd require 32 bit openssl which we don't currently have and would create issues for cross-compilation, etc.

Hashing has really reached the point the c library should just provide it, but sadly it hasn't happened yet.

tracernz:
Good points, I agree with your conclusion. :+1:

mlyle:
(I also don't love the wordy parsing code in here, but don't really see a better way :/

tracernz (approved):

PR #1774, opened 2017-05-16
treeitem: fix a dumb bug from the changes

mlyle:
regression/blocker.  meh.

My fault.  Confirmed GCS happy with this change.

### DIFF
--- a/ground/gcs/src/plugins/uavobjectbrowser/treeitem.cpp
+++ b/ground/gcs/src/plugins/uavobjectbrowser/treeitem.cpp
@@ -45,7 +45,7 @@ HighLightManager::HighLightManager(long checkingInterval, QTime *currentTime)
     m_expirationTimer.start(checkingInterval);
     connect(&m_expirationTimer, &QTimer::timeout, this, &HighLightManager::checkItemsExpired);
 
-    if (currentTime == NULL)
+    if (m_currentTime == NULL)
         m_currentTime = new QTime;
 
     *m_currentTime = *currentTime;


jihlein:
Don't know if it's related, but after rebasing to next, GCS is hanging at the "Initializing Welcome plugin" step on Windows 7.

dustin:
This seems to have solved the issue I had.

jihlein:
Deleted my original post.  For some reason my 1st attempt did not pick up the necessary change.  Issue is solved for me too.

mlyle:
Thx all.  Sorry for the dumb couple of regressions.

ground/gcs/src/plugins/uavobjectbrowser/treeitem.cpp
```diff
@@ -45,7 +45,7 @@ HighLightManager::HighLightManager(long checkingInterval, QTime *currentTime)
     m_expirationTimer.start(checkingInterval);
     connect(&m_expirationTimer, &QTimer::timeout, this, &HighLightManager::checkItemsExpired);
 
-    if (currentTime == NULL)
+    if (m_currentTime == NULL)
         m_currentTime = new QTime;
```
tracernz, line 49: Should probably be QTime::currentTime()? But that is problematic because we stupidly try to use pointers to QTimes. QTime default ctor returns a null time. But uhhh.... the check looked correct before? It just needed `*m_currentTime = *currentTime;` to be an else statement.

mlyle: m_currentTime could be null before, so assigning to it in that case would be bad.  need to read this for reals.

tracernz: Ahhh... lol. We should really just make it not a pointer. currentTime could also be nullptr (and should not be a pointer to begin with).

mlyle: So I've read the surrounding code.  The idea is, one class has an idea of the current time.. that it updates with a timer.. that everyone else has pointers to cuz LOL.

tracernz: Ahhhh... haha.

PR #1777, opened 2017-05-19
brain: Switch to generic MPU driver.

glowtape:
Did that to test some pios_mpu changes. Works on the bench.


glowtape:
Seems like the OG Brain and Sparky are the last hold-outs not using pios_mpu.c. If latter also gets converted to the new driver, we could essentially ditch the model-specific drivers, if we want to.

mlyle:
We want to ;) of course this is the type of thing we need to be super careful with.

@mluessi ?

tracernz:
> If latter also gets converted to the new driver, we could essentially ditch the model-specific drivers, if we want to.

That's exactly why I wrote PIOS_MPU.

mlyle (approved):
Looks good to me-- only other thing I see worth considering is deleting the pios_mpu9250_brain brain-specific header.

mluessi:
LGTM too. Thanks @glowtape !

flight/targets/brain/fw/pios_board.c
```diff
@@ -531,79 +530,73 @@ void PIOS_Board_Init(void) {
...
+		(hw_mpu9250_dlpf == HWBRAIN_MPU9250ACCELLPF_41) ? 41 : \
+		(hw_mpu9250_dlpf == HWBRAIN_MPU9250ACCELLPF_20) ? 20 : \
+		(hw_mpu9250_dlpf == HWBRAIN_MPU9250ACCELLPF_10) ? 10 : \
+		(hw_mpu9250_dlpf == HWBRAIN_MPU9250ACCELLPF_5) ? 5 : \
+		184;
+
+	pios_mpu_dev_t mpu_dev = NULL;
+	int retval;
+	retval = PIOS_MPU_I2C_Init(&mpu_dev, pios_i2c_internal_id, &pios_mpu_cfg);
+	if (retval == -10)
```
tracernz: Redundant.

flight/targets/brain/fw/pios_board.c
```diff
@@ -531,79 +530,73 @@ void PIOS_Board_Init(void) {
...
+		PIOS_HAL_CriticalError(PIOS_LED_ALARM, PIOS_HAL_PANIC_IMU);
+
+	PIOS_MPU_SetGyroBandwidth(gyro_lpf);
+	PIOS_MPU_SetAccelBandwidth(accel_lpf);
+
+	// To be safe map from UAVO enum to driver enum
+	uint8_t hw_gyro_range;
+	HwBrainGyroFullScaleGet(&hw_gyro_range);
+	switch(hw_gyro_range) {
+		case HWBRAIN_GYROFULLSCALE_250:
```
tracernz: Indentation not right... but it matches the rest of the file so meh?

flight/targets/brain/fw/pios_board.c
```diff
@@ -531,79 +530,73 @@ void PIOS_Board_Init(void) {
...
+	if (retval != 0)
+		PIOS_HAL_CriticalError(PIOS_LED_ALARM, PIOS_HAL_PANIC_IMU);
+
+	PIOS_MPU_SetGyroBandwidth(gyro_lpf);
+	PIOS_MPU_SetAccelBandwidth(accel_lpf);
+
+	// To be safe map from UAVO enum to driver enum
+	uint8_t hw_gyro_range;
+	HwBrainGyroFullScaleGet(&hw_gyro_range);
+	switch(hw_gyro_range) {
```
tracernz: Missing space.

flight/targets/brain/fw/pios_board.c
```diff
@@ -531,79 +530,73 @@ void PIOS_Board_Init(void) {
...
-			case HWBRAIN_ACCELFULLSCALE_8G:
-				PIOS_MPU9250_SetAccelRange(PIOS_MPU60X0_ACCEL_8G);
-				break;
-			case HWBRAIN_ACCELFULLSCALE_16G:
-				PIOS_MPU9250_SetAccelRange(PIOS_MPU60X0_ACCEL_16G);
-				break;
-		}
+	uint8_t hw_accel_range;
+	HwBrainAccelFullScaleGet(&hw_accel_range);
+	switch(hw_accel_range) {
```
tracernz: Missing space.

flight/targets/brain/fw/pios_board.c
glowtape: For future reference, are case statements supposed to be indented or on the same column as the switch one?

(If it's about how deep it's indented, it's a single tab.)

tracernz: Yeah, same level: https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst#1-indentation

tracernz (approved):

tracernz:
Thanks.

PR #1779, opened 2017-05-21
4WIF: Change 1: Add pios_dio subsystem

mlyle:
Also add a call to pios_servo to get the pins used.


tracernz (approved):
Already reviewed this one and just gave it another quick read. Great. 👍

PR #1780, opened 2017-05-21
4WIF: Change 2: Fixes to USB CDC implementation

### LINKED ISSUE #1679: VCP Oddities
* [ ] If you read with a long timeout (e.g. PIOS_QUEUE_TIMEOUT_MAX), the port locks up for at least Linux users.
* [ ] The port is unusable for several seconds after connection when using Liunx
* [ ] VCP to BLHeliSuite doesn't seem to work (cabled serial does).

All experienced on F4 on Seppuku

mlyle:
1. Just send a sane line state report always; don't try and omit it and
confuse the other side (wtf!)
2. Don't inhibit TX based on "serial lines" like DTE.
3. Simplify and make tx_active tracking logic in CDC layer correct.

edit: fixes #1679

### DIFF
--- a/flight/PiOS/STM32F30x/pios_usb_cdc.c
+++ b/flight/PiOS/STM32F30x/pios_usb_cdc.c
@@ -375,7 +375,7 @@ const struct usb_cdc_serial_state_report uart_state = {
 	.wValue        = 0,
 	.wIndex        = htousbs(1),
 	.wLength       = htousbs(2),
-	.bmUartState   = htousbs(3),
+	.bmUartState   = htousbs(2 /* DSR */ | 1 /* DCD */),
 };
 	
 static void PIOS_USB_CDC_CTRL_EP_IN_Callback(void)
--- a/flight/PiOS/STM32F30x/pios_usb_cdc.c
+++ b/flight/PiOS/STM32F30x/pios_usb_cdc.c
@@ -320,8 +320,7 @@ static bool PIOS_USB_CDC_Available (uintptr_t usbcdc_id)
 	bool valid = PIOS_USB_CDC_validate(usb_cdc_dev);
 	PIOS_Assert(valid);
 
-	return (PIOS_USB_CheckAvailable(usb_cdc_dev->lower_id) &&
-		(control_line_state & USB_CDC_CONTROL_LINE_STATE_DTE_PRESENT));
+	return PIOS_USB_CheckAvailable(usb_cdc_dev->lower_id);
 }
 
 static struct usb_cdc_line_coding line_coding = {
@@ -370,13 +369,13 @@ const uint8_t *PIOS_USB_CDC_GetLineCoding(uint16_t Length)
 	}
 }
 
-struct usb_cdc_serial_state_report uart_state = {
+const struct usb_cdc_serial_state_report uart_state = {
 	.bmRequestType = 0xA1,
 	.bNotification = USB_CDC_NOTIFICATION_SERIAL_STATE,
 	.wValue        = 0,
 	.wIndex        = htousbs(1),
 	.wLength       = htousbs(2),
-	.bmUartState   = htousbs(0),
+	.bmUartState   = htousbs(3),
 };
 	
 static void PIOS_USB_CDC_CTRL_EP_IN_Callback(void)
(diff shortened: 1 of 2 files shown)


flight/PiOS/STM32F30x/pios_usb_cdc.c
```diff
@@ -386,22 +385,10 @@ static void PIOS_USB_CDC_CTRL_EP_IN_Callback(void)
...
-	 *      4:  bFraming    framing error
-	 *      3:  bRingSignal RI
-	 *      2:  bBreak      break reception
-	 *      1:  bTxCarrier  DSR
-	 *      0:  bRxCarrier  DCD
-	 */
-	uart_state.bmUartState = htousbs(0x0003);
-
-	UserToPMABufferCopy((uint8_t *) &uart_state,
+	UserToPMABufferCopy((const uint8_t *) &uart_state,
```
tracernz, line 388: This changed nothing except removing documentation (not good) and putting the struct in flash/not writing it each time (good)? :/

flight/PiOS/STM32F4xx/pios_usb_cdc.c
```diff
@@ -463,39 +464,9 @@ static bool PIOS_USB_CDC_CTRL_EP_IN_Callback(uintptr_t usb_cdc_id, uint8_t epnum
...
-	/* Currently, we only handle TxCarrier and RxCarrier reporting */
-	uint16_t new_uart_state = 0;
-	if (usb_cdc_dev->tx_out_cb) {
-		/* Someone is going to providing FC->PC data, advertise an RxCarrier to the host */
-		new_uart_state |= 0x1;
-	}
-	if (usb_cdc_dev->rx_in_cb) {
-		/* Someone is consuming PC->FC data, advertise a TxCarrier to the host */
-		new_uart_state |= 0x2;
-	}
```
tracernz, line 487: This was actually correct before from my reading of the spec?

flight/PiOS/STM32F4xx/pios_usb_cdc.c
```diff
@@ -463,39 +464,9 @@ static bool PIOS_USB_CDC_CTRL_EP_IN_Callback(uintptr_t usb_cdc_id, uint8_t epnum
...
-
-		uart_state.bmUartState = htousbs(new_uart_state);
-
-		PIOS_USBHOOK_EndpointTx(usb_cdc_dev->cfg->ctrl_tx_ep,
-					(uint8_t *)&uart_state,
-					sizeof(uart_state));
-	}
+	PIOS_USBHOOK_EndpointTx(usb_cdc_dev->cfg->ctrl_tx_ep,
+			(uint8_t *)&uart_state,
+			sizeof(uart_state));
```
tracernz, line 469: From the spec:
> SerialState is used like a real interrupt status register. Once a notification has been sent, the device will reset and reevaluate the different signals. __For the consistent signals like carrier detect or transmission carrier, this will mean another notification will not be generated until there is a state change__. For the irregular signals like break, the incoming ring signal, or the overrun error state, this will reset their values to zero and again __will not send another notification until their state changes__.

flight/PiOS/STM32F4xx/pios_usb_cdc.c
mlyle: The way this is presented to the application on all OSes as a DCD level.  There's no reason if, say, logging data is coming out here that someone shouldn't be able to connect with kermit etc, which this would prevent.

flight/PiOS/STM32F30x/pios_usb_cdc.c
mlyle: Yes, it's just changed to be consistent with F4.

flight/PiOS/STM32F4xx/pios_usb_cdc.c
mlyle: This was necessary to get rid of hanging behavior.   F1, F3, and ST ref code notified every time, F4 didn't.  It's unclear how you'd ever get the state of the line if you connected and it never changed.  In a protocol analyzer without this change I saw repeated setcontrollinestate's with no meaningful response-- I guess we could do it change-triggered (never) plus after this, but...  seems safer to send it every time

mlyle:
Added some minimal control line docs.

flight/PiOS/STM32F4xx/pios_usb_cdc.c
mlyle: Just to be clear-- repeated control messages with then blhelisuite bailing out and saying it couldn't connect.

I referred to the spec and other ref code (linux kernel).  I believe, but can't be certain, that this is the minimal changeset required for things to consistently work.  Ideally we'd better notice connect/disconnect and ensure to notify at least once in the beginning-- but I'm not even positive this is fully safe.

In any case, -- it doesn't increase the number of transfers (notifications are polled, and declining to notify is a transfer in itself), and client code properly looks for edges (it has to-- if there's another interrupt it'll still receive a duplicate on the other bits).  If we want to lower bus overhead, we could lower bInterval on this mgmt endpoint, which is currently 4ms.

tracernz: Okay. I guess this part is more an issue for Windows? I've used CDC successfully on Linux many times without any issues.

mlyle: Yes, 100% windows-- some of the traces I have are in the bug.

I do think it would cause control lines to get "stuck" on linux in default state e.g. on re-plug--- because we would never send the status back as it'd never change.

mlyle:
Thx for merge

flight/PiOS/STM32F4xx/pios_usb_cdc.c
tracernz: Yes, that part was a bit wonky in that it wouldn't send a notification upon every connection (the spec doesn't really say that explicitly but it seems to be implied, and is the only way it could make sense).

PR #1782, opened 2017-05-21
pios_spi: Optional DMA transfers on F4 (WIP).

glowtape:
I was suggested on IRC this might be useful, to maximize SPI bandwidth. Apparently it does something. One does DMA, one doesn't:

timing

Gonna look into F3 eventually, since apparently it does the same thing.

### DIFF
--- a/flight/PiOS/STM32/pios_spi.c
+++ b/flight/PiOS/STM32/pios_spi.c
@@ -55,6 +55,74 @@ typedef enum {
 	PIOS_SPI_PRESCALER_256 = 7
 } SPIPrescalerTypeDef;
 
+#if defined(STM32F4XX)
+
+#include "pios_quickdma.h"
+
+void PIOS_SPI_InitializeDMA(struct pios_spi_dev *dev)
+{
+	const struct pios_spi_dma *dma = &dev->cfg->spi_dma;
+	PIOS_Assert(dma);
+
+	if(!dev->dma_send) {
+		dev->dma_send = quickdma_initialize(&dma->send);
+		dev->dma_recv = quickdma_initialize(&dma->recv);
+
+		/* If an IRQ channel was defined on the receive channel, it'll go yield.
+		   Remember to implement an IRQ handler to run quickdma_irq_trigger. */
+		quickdma_enable_yielding(dev->dma_recv);
+	}
+}
+
+bool PIOS_SPI_StartDMA(struct pios_spi_dev *dev, const uint8_t *send, uint8_t *recv, uint16_t len)
+{
+	const struct pios_spi_dma *dma = &dev->cfg->spi_dma;
+	PIOS_Assert(dma);
+
+	quickdma_stop_transfer(dev->dma_send);
+	quickdma_stop_transfer(dev->dma_recv);
+
+	SPI_I2S_DMACmd(dev->cfg->regs, SPI_I2S_DMAReq_Rx, DISABLE);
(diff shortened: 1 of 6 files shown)


glowtape:
F3 has no gaps on the scope, DMA or not. I suppose this was a typo, @mlyle?

`<icee> because killing the gaps on SPI on F3 would be interesting IMO`

flight/PiOS/STM32/pios_spi.c
```diff
@@ -89,8 +89,8 @@ void dma_init(DMA_Stream_TypeDef *stream, SPI_TypeDef *spi, uint32_t channel, bo
 
 void PIOS_SPI_InitializeDMA(struct pios_spi_dev *dev)
 {
-	dma_init(dev->cfg->dma_send_stream, dev->cfg->regs, dev->cfg->dma_send_channel, TRUE);
-	dma_init(dev->cfg->dma_receive_stream, dev->cfg->regs, dev->cfg->dma_receive_channel, FALSE);
+	dma_init(dev->cfg->dma_send_stream, dev->cfg->regs, dev->cfg->dma_send_channel, true);
+	dma_init(dev->cfg->dma_receive_stream, dev->cfg->regs, dev->cfg->dma_receive_channel, false);
```
tracernz: Should really fix the original commit so the change is atomic, otherwise bisection is sad. :/

mlyle:
It surprises me F3 SPI is gapless since it is doing the same thing as F4, and, if anything, should be slower.

(F3 doesn't need DMA to be gapless, just to effectively use the FIFO.. which maybe we already are, need to look at what the default is...)

mlyle:
And it's not, for sure.  how weird.  Are you sure it's gapless?  Is the SPI on that target just very slow?

glowtape:
The interesting thing is, doing DMA, the waveform got all weird on F3, with slow rises and ringing. If I'm going to add DMA to F3 regardless, I have to write an abstraction of the whole DMA stuff, because the earlier attempt ended up in an #ifdef horror show.

glowtape:
As far as slowness goes, this is very possible. I just checked the scope capture from yesterday, a gyro data transfer takes like ~300us there, vs 12us on the Revo.

mlyle:
I'm not sure F3 should get DMA, but it should get pipelining.  I can implement that.

What target is this on, with the slow SPI transfers on F3?

glowtape:
It was on SPRF3e actually. Thinking about it, it may just be a configuration SNAFU that it's that slow.

glowtape:
Also, as far as DMA goes, yea, it's either-or for SPI DMA or DShot, given the channel map on F3.

mlyle:
There's only power of 2 divisors for SPI speed, but with F3 @ 72MHz and a desire for a 20MHz clock, pios_mpu should be choosing 72 / 4 = 18MHz...

glowtape:
OK, this sits on top of that quickdma stuff.

Currently only Seppuku and Revo are enabled, in DMA fire and wait mode. SPI pauses are gone, the transfer times for the Revo go from ~20us to 12us, for the Seppuku it's similar, from ~18us to 12us.

--edit: For review only. Need to fix all makefiles. :|

glowtape:
This one is getting held up a bit, because I can't figure out why the gyro rates are just a fraction when this is used on the F3 (saves 3% CPU, though, without yielding).

glowtape:
Scrapping this for the time being. Gets ugly as hell quickly trying to setup DMA without a helper class, which in turn seems to blow the flash data budget on F3.

PR #1783, opened 2017-05-22
Sparky: use pios_mpu for 9150

mlyle:
1. Move the target to pios_mpu
2. Adjust pios_mpu to be smarter about reading out mag--- needs validation on '9250 etc.
3. Teach pios_mpu to use SLV1 to trigger magnetometer
4. Sparky1 shouldn't be decimating autotune data because it's only at 500Hz.

THOUGH-- I do believe sparky1 can move to 1KHz soon.  It's really close with this changeset-- in autotune CPU use goes just occasionally into the yellow.  I think upcoming attitude decimation code + new sensors code will take it up to 1KHz no problem.


mlyle:
Rebased.

Tested Brain (MPU9250 on I2C).  CPU load 31% with no video etc.  vs. 33% on bare next.  Mag appears functional/good.

mlyle:
Tested sparky2 (MPU9250 on SPI).  CPU load toggles between 29% and 30%, vs toggling between 30% and 31% on bare next.  A minimal savings but still surprisingly noticeable.  Mag OK.

mlyle:
Retested sparky.  CPU load 44%. vs 48% on bare next (hooray).  Mag OK.

tracernz (approved):
Looks good. :+1:

PR #1784, opened 2017-05-22
pios_quickdma: DMA helper code to deal with SPI DMA and DShot across both F3/F4 (and future?) (WIP!)

glowtape:
This commit is currently F4 only. Works for SPI DMA on my test Revo. Next step is to move DShot onto it, to verify it keeps working as expected. Then I'll implement the F3 side of the code.

Also, this doesn't use StdPeriph.

This PR also wraps ChibiOS' task yielding function in a PIOS_Thread_Yield function, to call from within pios_quickdma, if desired. Say if you have long and slow SPI transfer, like on SPRF3E, you can let the system do other stuff instead of sitting the wait function.

(Currently up just for rough appraisal.)

### DIFF
--- a/flight/PiOS/STM32/inc/pios_quickdma.h
+++ b/flight/PiOS/STM32/inc/pios_quickdma.h
@@ -1,0 +1,46 @@
+#ifndef PIOS_QUICKDMA_H
+#define PIOS_QUICKDMA_H
+
+#include <pios.h>
+
+typedef struct quickdma_transfer* quickdma_transfer_t;
+
+#define QUICKDMA_SIZE_BYTE				1
+#define QUICKDMA_SIZE_HALFWORD			2
+#define QUICKDMA_SIZE_WORD				4
+
+#define QUICKDMA_PRIORITY_LOW			0
+#define QUICKDMA_PRIORITY_MEDIUM		1
+#define QUICKDMA_PRIORITY_HIGH			2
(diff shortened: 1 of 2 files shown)


glowtape:
OK, per IRC, the yielding stuff as-is doesn't exactly appear to be fine operating system design, so I'm going to drop it for the time being, until I can do it with semaphores.

Turning this into plain helper code to setup and run DMA transfers, to avoid doing the unwieldy DMA_Init_TypeDef stuff over and over.

Right now I have patches of both DShot and SPI DMA on F4 sitting on this, both working. Pretty much ready to go for testing.

Next plan is to move WS2811 onto this and try to do it DMAShot style. If that works, I'll write the F3 counterpart of quickdma to get WS281x on F3.

On the very very longterm, having it manage all DMA stuff including IRQs (and subsequently allow to yield properly with semaphores) would be interesting.

glowtape:
I think I have a solution for proper yielding without much fuzz. The recent change to a configuration struct (which would usually be defined in board_hw_defs) would allow (a pointer of) it to be used as reference in a traditional IRQ handler to pass to a quickdma function for releasing a semaphore. Either by storing a pointer back to the quickdma state struct in it, or by looking up a linked list of those. Not really elegant, but it'd work.

glowtape:
Scrapping this for the time-being. Need to rethink it all to make F3 fit into this.

PR #1785, opened 2017-05-24
actuator: fix helicopter throttle

mlyle:


mlyle:
Spotted during reading of fixed wing path follower for @dustin

tracernz:
Only affects failsafe (only, hah!) on quick inspection. Nice spotting and fix looks good. :+1:

mlyle:
It was actually worse than that-- the throttle_val is not static, so, between manual control updates the throttle val was lost.  yuck :(

PR #1787, opened 2017-05-26
pios_ws2811: Make sure ChibiOS IRQ epilogue runs.

glowtape:
For correctness?


tracernz:
Thanks, we've been meaning to look at that.

Side note: can you please make sure your commits all build, and if not fix them up? The broken ones will make bisection a pain later. For this one we can happily squash it when merging.

PR #1788, opened 2017-05-26
Flight: Add IRQ Pro/Epilogue to ADC handler

glowtape:
Essentially this was an issue I, and probably everyone else running at least the Seppuku, had likely since the beginning:

sds1202x-e13

Every 200-300ms the servo output signal jitters like this. Turns out it was the ADC IRQ handler doing this and making the system have a ball. Sticking the (missing?) ChibiOS IRQ pro- and epilogue stuff (which also does some thread rescheduling from what I've seen) into the IRQ handler of the ADC mitigates the issue.

So as a precaution, I've also stuck it in the PIOS DMA IRQ handlers on F3.

### DIFF
--- a/flight/PiOS/STM32F30x/pios_dma_hooks.c
+++ b/flight/PiOS/STM32F30x/pios_dma_hooks.c
@@ -72,39 +72,122 @@ static inline void PIOS_DMA_Mapper(uint32_t index)
 }
 
 static void PIOS_DMA_11_irq_handler(void){
+#if defined(PIOS_INCLUDE_CHIBIOS)
+	CH_IRQ_PROLOGUE();
+#endif /* PIOS_INCLUDE_CHIBIOS */
 	PIOS_DMA_Mapper(0);
+#if defined(PIOS_INCLUDE_CHIBIOS)
+	CH_IRQ_EPILOGUE();
+#endif /* PIOS_INCLUDE_CHIBIOS */
 }
+
 static void PIOS_DMA_12_irq_handler(void){
+#if defined(PIOS_INCLUDE_CHIBIOS)
+	CH_IRQ_PROLOGUE();
+#endif /* PIOS_INCLUDE_CHIBIOS */
 	PIOS_DMA_Mapper(1);
+#if defined(PIOS_INCLUDE_CHIBIOS)
+	CH_IRQ_EPILOGUE();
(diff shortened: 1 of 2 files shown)


tracernz:
The ADC ISR should __never__ wake anything. This is masking something else if it makes any difference. We should abandon this and find the real cause.

glowtape:
The real cause might be tricky. I've hacked together some code that hijacks the ChibiOS IRQ pro- and epilogues to track IRQ calls/sec and runtime of PiOS devices (except pios_video), and there's nothing out of the ordinary from that side. Actually, in terms of PiOS devices, there's surprisingly way less activity than I've anticipated. I'd figure that threads won't be able to disturb the system to such a degree, and there isn't much that disables IRQs at all, or even extended periods. A Seppuku with no configuration and just four SyncPWM channels already showed this problem. 

The results are these for my regular setup/config, FWIW:

1) Crossfire, doing ~3961 calls/sec, ~3.57us per call
2) ADC*, doing ~2413 calls/sec, ~13.80us per call
3) BMI160, doing ~1610 calls/sec, ~1.97us per call
4) USB-OTG, doing ~1080 calls/sec, ~2.33us per call
5) RTC, doing ~620 calls/sec, ~2.95us per call
6) WS2811, doing ~69 calls/sec, ~1.13us per call

(* with IRQ epilogue, unpatched.)

There might be things not tracked, due to not using the ChibiOS stuff, but from what I can tell, most of the PiOS stuff does. There isn't much left to assign blame IMO.

As far as the IRQ epilogue goes, I'd have figured every ISR, except the really performance sensitive ones (like pios_video) should use them? From what I can tell from ChibiOS, if I understood it correctly, it just makes sure the scheduler runs the next thread, instead of returning to the current one, should the ISR runtime have blown past the allotted timeslot of the running thread.

glowtape:
Oh, by the way, #1789 made it practically go away. Everything is even with this, only if you look hard and long enough, you can see a single servo signal here and there be delayed.

tracernz:
While we're looking at prologue/epilogue, rather than ifdeffery throughout the code, we should introduce PIOS_IRQ_Pro/Epilogue macros (epilogue might want to take a woken arg which might be useful for other RTOS e.g. FreeRTOS) and concentrate the ugliness to one place (and also lessen churn later). We currently have 2 OS (ChibiOS and linux-rt) and who knows what we will look like in a few years.

PR #1789, opened 2017-05-27
adc: make interrupt rates sane, clean up

mlyle:
Needs test and review.

Came across from #1788-- It's good to have the epilogue on the IRQ, but since the IRQ doesn't wake anyone it can't be the real root cause for jitter / things not waking.  OTOH, caused us to observe that the ADC interrupt is called **fast** on F4 because of suck, wasting a whole ton of CPU.  (And running reschedules frequently undoubtedly papers over whatever other problem there is).

### DIFF
--- a/flight/PiOS/STM32F4xx/pios_internal_adc.c
+++ b/flight/PiOS/STM32F4xx/pios_internal_adc.c
@@ -55,13 +55,8 @@ enum pios_adc_dev_magic {
 };
 
 struct pios_internal_adc_dev {
-	const struct pios_internal_adc_cfg * cfg;
-	volatile int16_t *valid_data_buffer;
-	volatile uint8_t adc_oversample;
-	uint8_t dma_block_size;
-	uint16_t dma_half_buffer_size;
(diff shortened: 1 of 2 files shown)


glowtape:
With this change and _without_ the epilogue, the servo output signal is still evenly spaced. Only occasionally a signal is late enough to be noticeable on the scope, but it's a lone one, not a whole group of them being jumbled up like in #1788. The occasional late straddlers in this patch go away with the epilogue is back in, too.

Also, the CPU usage for my configuration on the quad dropped from 50-51% to 47% with this.

tracernz:
Thanks, this is good. :+1: I vaguely remember working on the F1 driver once upon a time and doing something with the sample rate or buffer length too but never looked much at F4... https://github.com/TauLabs/TauLabs/pull/2035

> Also, the CPU usage for my configuration on the quad dropped from 50-51% to 47% with this.

Expected since we don't run the ISR anywhere near as often (about 1/11th as often roughly), saving us quite a few context switches too.

tracernz (approved):

PR #1790, opened 2017-05-28
pios,chibios: No flash instruction/data cache for buggy F4 revisions.

glowtape:
Apparently there's older revisions of the F4 MCU out there, that don't get flaky if the caches are enabled. ChibiOS can deal with it, if you specify a certain #define.

https://github.com/d-ronin/dRonin/blob/b323ecce4bc16b80aa638fdff07793a872d2e3f7/flight/PiOS/Common/Libraries/ChibiOS/os/hal/platforms/STM32F4xx/hal_lld.c#L258

### DIFF
--- a/flight/PiOS/STM32/inc/board.h
+++ b/flight/PiOS/STM32/inc/board.h
@@ -29,6 +29,10 @@
 
 #include "pios_chibios_transition_priv.h"
 
+/* Some older STM32F4 revisions might get antsy when flash prefetching is enabled.
+   This makes sure the affected revs don't get enabled by ChibiOS. */
+#define STM32_USE_REVISION_A_FIX
+
 /*
  * Board oscillators-related settings.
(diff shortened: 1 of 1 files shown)


mlyle:
Does pios.h get included somehow by hal_lld.c where this is used?   I didn't see a path looking quickly.  If not, board.h might be a better place.

glowtape:
Not sure what I was thinking.

flight/PiOS/STM32/inc/board.h
```diff
@@ -29,6 +29,13 @@
 
 #include "pios_chibios_transition_priv.h"
 
+#if defined(PIOS_INCLUDE_CHIBIOS)
```
tracernz: Don't need this, this _is_ a ChibiOS header.

mlyle:
Thanks!

PR #1791, opened 2017-06-02
telemetry: significant cleanup and performance improvement

mlyle:
1. Remove infrastructure for acks and related synchronization from
uavtalk; let the telemetry module sequence this.
2. Add context pointer infrastructure.
3. Bundle up most of telemetry's state, so that it is much closer to
be able to run multiple instances.
4. Periodic event handlers were never called in Telemetry-- this code
got lost at some point and was reinstated.
5. No longer require an expensive ack for the HwSimulation object, so
that we don't incur a lot of extra traffic for blinkenlights.
6. Allow multiple acked-things to be in flight at once-- don't bind up the
entire telemetry channel when one thing requiring an ack is in flight.
7. Move radiocombridge's PIOS_COM calls to blocking variants, to
avoid a bug where we just spin on a full buffer for a short time then
give up.

Has passed cursory test; requires further test and review.


tracernz (approved):
This looks good. Should make things a lot nicer. Have you noticed better performance?

mlyle:
I have not measured it at all yet.  I need to do a bunch of test on this-- including radio cases to see if that stuff is any better.

mlyle:
With this, telemetry sessions come up at about flightTime = 7; overall CPU load with telem session up over USB on revolution is 25-26%.  Telemetry thread CPU usages come in at 0's.  A couple hundred bytes of RAM are saved.

Without this, telemetry sessions come up at about flightTime = 8-9; overall CPU load is 26-27%.  

With this, radio sessions come up at about flightTime = 11.

Without this, radio telemetry sessions come up at flightTime = 18.

n=4 for each of these tests.  Results probably just barely statistically significant for USB; -definitely- real and significant and a few sigma out for radio.

mlyle:
So, all in all, this is much better.... but oplink still has use cases where configuration doesn't come down right.  Seems like less of the time, but...

flight/Libraries/inc/uavtalk_priv.h
```diff
@@ -103,6 +103,8 @@ typedef struct {
 #define UAVTALK_TYPE_OBJ_ACK   (UAVTALK_TYPE_VER | 0x02)
 #define UAVTALK_TYPE_ACK       (UAVTALK_TYPE_VER | 0x03)
 #define UAVTALK_TYPE_NACK      (UAVTALK_TYPE_VER | 0x04)
+#define UAVTALK_TYPE_FILEREQ   (UAVTALK_TYPE_VER | 0x10)
```
tracernz, line 106: I also had something like this in mind. Probably doesn't really fit the commit description though. 🙃

mlyle: Assign new message types for file data is in the commit message? :p 

The big thing (not documented yet) is atomically shoving ~3 responses in every time there's a req to keep overhead down, and to be able to detect loss.

tracernz (approved):
Looks good, thanks! :grin: I didn't go too deeply into verifying locking.

tracernz:
Tested with HID, CDC, and with a PipX talking to a Revo. All seemed well. The session negotiation is noticeably faster on CDC in particular! :+1:

mlyle:
:) thx

PR #1794, opened 2017-06-03
PipX: Don't pass FirmwareIAP commands through

tracernz:
The flight board cannot do anything useful with them through a radio link, and they are actually quite dangerous if we're in-flight etc. Fixes (for us) https://github.com/TauLabs/TauLabs/issues/1860


mlyle:
Hmmm-- I use this (to reboot after changing settings on battery pwoer).  An alternative would be for firmwareIAP to not listen when armed.

tracernz:
Ah, I guess that's fair. Ignoring IAP when armed seems better.

tracernz:
Well, we already have that. ;) https://github.com/d-ronin/dRonin/blob/next/flight/Modules/FirmwareIAP/firmwareiap.c#L172

PR #1795, opened 2017-06-03
GCS:Uploader: Fix up layout and resizing behaviour, artwork for TauLink

tracernz:
Now behaves a lot better. Also added new artwork for TauLink (didn't notice until after creating new artwork that the path was incorrect and it actually already had some.. sigh..).

Near minimum usable size:
screenshot_2017-06-04_03-09-01

Stupidly large size:
screenshot_2017-06-04_03-09-30
screenshot_2017-06-04_03-10-16


ground/gcs/src/libs/utils/scaledpixmaplabel.cpp
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, see <http://www.gnu.org/licenses/>
+ *
+ * Additional note on redistribution: The copyright and license notices above
+ * must be maintained in each individual source file that is a derivative work
+ * of this source file; otherwise redistribution is prohibited.
+ */
+
+#include <QScreen>
```
tracernz: Oops, don't need this.

mlyle:
Glad to merge once qscreen is gone.

PR #1805, opened 2017-06-06
Support for SITL for flightgear sim.

mlyle:
Seems to work.  Of course, there's other objects that can be sent, etc-- this is just the general framework.


tracernz (approved):
Wonderful. :+1:

PR #1806, opened 2017-06-06
flight & python: file transfer functionality

mlyle:
And assorted small cleanups.

If/when architecture for this is approved, will set about A) GCS implementation, and B) updating uavtalk protocol docs.


mlyle:
Tested on Sparky1 & Sparky2.  Gets about 10k/sec in transfer over USB HID, which is not blazingly fast.

OTOH we put 4 in bInterval, which means 250 messages/sec.  At 64 bytes payload each, maximum theoretical transfer rate is 63 (len prefixed) * 250 = 15750 bytes/sec, before any of the uavtalk overhead, channel turnaround time, etc.

python/dronin/uavtalk.py
```diff
@@ -171,7 +171,9 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=None,
         # Search for object.
         uavo_key = '{0:08x}'.format(objId)
         if not uavo_key in uavo_defs:
-            #print "Unknown object 0x" + uavo_key + "type = ", pack_type
+            #print "Unknown object 0x" + uavo_key + " type = ", pack_type
+            obj_len = pack_len - MIN_HEADER_LENGTH
```
tracernz: This is defeated at line 192 (and it looks more correct)?

flight/Libraries/inc/uavtalk.h
```diff
@@ -31,8 +31,10 @@
 #define UAVTALK_H
 
 // Public types
-typedef int32_t (*UAVTalkOutputCb)(void *ctx, uint8_t* data, int32_t length);
+typedef int32_t (*UAVTalkOutputCb)(void *ctx, uint8_t *data, int32_t length);
```
tracernz, line 34: :+1:

flight/Libraries/inc/uavtalk_priv.h
```diff
@@ -88,24 +98,28 @@ typedef struct {
...
+	UAVTalkFileCb fileCb;
 	void *cbCtx;
 } UAVTalkConnectionData;
 
 #define UAVTALK_CANARI         0xCA
 #define UAVTALK_WAITFOREVER     -1
 #define UAVTALK_NOWAIT          0
 #define UAVTALK_SYNC_VAL       0x3C
-#define UAVTALK_TYPE_MASK      0x78
+#define UAVTALK_TYPE_MASK      0x70
```
tracernz, line 109: Seems this is really UAVTALK_VER_MASK. :/

flight/Libraries/uavtalk.c
```diff
@@ -254,6 +261,26 @@ UAVTalkRxState UAVTalkProcessInputStreamQuiet(UAVTalkConnection connectionHandle
 		if (iproc->rxCount < 4)
 			break;
 
+		if (iproc->type == UAVTALK_TYPE_FILEREQ) {
```
tracernz, line 264: Do we want support for receiving files on the flight side? Could be useful for PicoC replacement if it ever eventuates. Might be safer to handle it here and error out in any case.

flight/Libraries/uavtalk.c
```diff
@@ -617,6 +640,95 @@ static int32_t sendBuf(UAVTalkConnection connectionHandle, uint8_t *buf, uint16_
...
+			total_len += cb_numbytes;
+
+			file_offset += cb_numbytes;
+
+			if (i == 5) {	/* 6 messages per chunk */
+				resp->flags = UAVTALK_FILEDATA_LAST;
+			} else {
+				resp->flags = 0;
+			}
+		} else {
```
tracernz, line 701: So if no callback registered or an error getting the file, 0 bytes response with EOF.. seems reasonable. ENOENT could also be nice to have?

flight/Modules/Telemetry/telemetry.c
```diff
@@ -393,6 +416,67 @@ static void addAckPending(telem_t telem, UAVObjHandle obj, uint16_t inst_id)
...
+ * id to the actual unit of information, and returns/copies it.
+ * Currently only operates on partitions.
+ *
+ * \param[in] ctx Callback context (telemetry subsystem handle)
+ * \param[in] file_id The requested file_id
+ * \param[in] offset The offset of information requested
+ * \param[in] len The maximum amount of information to return.  A short read
+ * is permissable, as long as some bytes are returned.
+ * \returns The number of bytes filled, 0 on EOF, negative on error.
+ */
```
tracernz, line 430: Nice work on all the Doxygen. :grin:

python/dronin/telemetry.py
```diff
@@ -193,23 +197,70 @@ def __handle_handshake(self, obj):
 
             if obj.Status == obj.ENUM_Status['Disconnected']:
                 # Request handshake
-                print("Disconnected")
+                #print("Disconnected")
```
tracernz, line 200: These should probably all use python logging so they can be enabled at runtime... For another PR. :)

python/dronin/uavtalk.py
```diff
@@ -170,9 +173,12 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=None,
         # Search for object.
         uavo_key = '{0:08x}'.format(objId)
         if not uavo_key in uavo_defs:
-            #print "Unknown object 0x" + uavo_key
+            #print "Unknown object 0x" + uavo_key + " type = ", pack_type
+            obj_len = pack_len - MIN_HEADER_LENGTH
```
tracernz: Isn't this overritten at line 195? header_fmt.size seems better too.

flight/Libraries/inc/uavtalk_priv.h
mlyle: Heh yup.  And no idea why it was malformed this way.

flight/Libraries/uavtalk.c
mlyle: Yes, the plan is to allow pushing files the other way, though the state machine of the initiator is slightly trickier.

flight/Libraries/uavtalk.c
mlyle: Yes flags exist to extend this.

flight/Modules/Telemetry/telemetry.c
mlyle: Thx :)

python/dronin/telemetry.py
mlyle: Yah, these were problematic-- corrupted my received files.  Ditto for the bad CRC messages.

PR #1808, opened 2017-06-09
Make: Add rule to check GCS code with clazy

tracernz:
Allows the compiler to understand Qt semantics and check code for compliance + recommended practice. https://github.com/KDE/clazy


tracernz:
It gets really angry even at level0. :cry: 

^ changed Makefile so want to make sure

tracernz:
Needs #1809 for build to succeed, but since the Makefile parsed fine I'm calling this one okay.

mlyle:

PR #1810, opened 2017-06-10
gcs: uavtalk: SIGNIFICANT performance improvement

mlyle:
Rewrote uavtalk to remove state machine, not read just a byte at a
time, and be cleaner.

I am not really clear why this is such a large change-- the old code
was inefficient but didn't look **that** bad.

Without: GCS over USB comes up at about flighttime=7.  With this,
GCS over USB comes up at about flighttime=3, and is obviously
significantly more responsive.

Over TCP, the differences are even more extreme.

### DIFF
--- a/ground/gcs/src/plugins/uavtalk/telemetry.cpp
+++ b/ground/gcs/src/plugins/uavtalk/telemetry.cpp
@@ -377,7 +377,7 @@ void Telemetry::transactionTimeout(ObjectTransactionInfo *transInfo)
     transInfo->timer->stop();
     // Check if more retries are pending
     if (transInfo->retriesRemaining > 0) {
-        qDebug() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 Retrying")
+        qInfo() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 Retrying")
                 .arg(transInfo->obj->getName()
                      + QString(QString(" 0x")
                                + QString::number(transInfo->obj->getObjID(), 16).toUpper()))
@@ -386,7 +386,7 @@ void Telemetry::transactionTimeout(ObjectTransactionInfo *transInfo)
         processObjectTransaction(transInfo);
         ++txRetries;
     } else {
-        qDebug() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 no more retries. FAILED "
+        qInfo() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 no more retries. FAILED "
                     "TRANSACT")
                 .arg(transInfo->obj->getName()
                      + QString(QString(" 0x")
@@ -439,7 +439,7 @@ void Telemetry::processObjectUpdates(UAVObject *obj, EventMask event, bool allIn
             ++txErrors;
             obj->emitTransactionCompleted(false);
             obj->emitTransactionCompleted(false, false);
-            qDebug() << QString(tr("Telemetry: priority event queue is full, event lost (%1)")
+            qWarning() << QString(tr("Telemetry: priority event queue is full, event lost (%1)")
                             .arg(obj->getName()));
         }
     } else {
@@ -449,7 +449,7 @@ void Telemetry::processObjectUpdates(UAVObject *obj, EventMask event, bool allIn
             ++txErrors;
             obj->emitTransactionCompleted(false, false);
             obj->emitTransactionCompleted(false);
-            qDebug() << QString(tr("Telemetry: event queue is full, event lost (%1)")
+            qWarning() << QString(tr("Telemetry: event queue is full, event lost (%1)")
                             .arg(obj->getName()));
         }
     }
--- a/ground/gcs/src/plugins/uavtalk/telemetry.cpp
+++ b/ground/gcs/src/plugins/uavtalk/telemetry.cpp
@@ -377,23 +377,21 @@ void Telemetry::transactionTimeout(ObjectTransactionInfo *transInfo)
     transInfo->timer->stop();
     // Check if more retries are pending
     if (transInfo->retriesRemaining > 0) {
-        TELEMETRY_QXTLOG_DEBUG(
-            QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 Retrying")
+        qDebug() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 Retrying")
                 .arg(transInfo->obj->getName()
                      + QString(QString(" 0x")
                                + QString::number(transInfo->obj->getObjID(), 16).toUpper()))
-                .arg(transInfo->obj->getInstID()));
+                .arg(transInfo->obj->getInstID());
         --transInfo->retriesRemaining;
         processObjectTransaction(transInfo);
         ++txRetries;
     } else {
-        TELEMETRY_QXTLOG_DEBUG(
-            QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 no more retries. FAILED "
+        qDebug() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 no more retries. FAILED "
                     "TRANSACT")
                 .arg(transInfo->obj->getName()
                      + QString(QString(" 0x")
                                + QString::number(transInfo->obj->getObjID(), 16).toUpper()))
-                .arg(transInfo->obj->getInstID()));
+                .arg(transInfo->obj->getInstID());
         transInfo->timer->stop();
         transactionFailure(transInfo->obj);
         ++txErrors;
@@ -441,9 +439,8 @@ void Telemetry::processObjectUpdates(UAVObject *obj, EventMask event, bool allIn
             ++txErrors;
             obj->emitTransactionCompleted(false);
             obj->emitTransactionCompleted(false, false);
-            TELEMETRY_QXTLOG_DEBUG(
-                QString(tr("Telemetry: priority event queue is full, event lost (%1)")
-                            .arg(obj->getName())));
+            qDebug() << QString(tr("Telemetry: priority event queue is full, event lost (%1)")
+                            .arg(obj->getName()));
         }
     } else {
         if (objQueue.length() < MAX_QUEUE_SIZE) {
@@ -452,6 +449,8 @@ void Telemetry::processObjectUpdates(UAVObject *obj, EventMask event, bool allIn
             ++txErrors;
             obj->emitTransactionCompleted(false, false);
             obj->emitTransactionCompleted(false);
+            qDebug() << QString(tr("Telemetry: event queue is full, event lost (%1)")
+                            .arg(obj->getName()));
         }
     }
     // Process the transaction queue
--- a/ground/gcs/src/plugins/uavtalk/telemetry.h
+++ b/ground/gcs/src/plugins/uavtalk/telemetry.h
@@ -91,8 +91,10 @@ signals:
 
 private:
     // Constants
-    static const int REQ_TIMEOUT_MS = 250;
-    static const int MAX_RETRIES = 2;
+    static const int REQ_TIMEOUT_MS = 1500;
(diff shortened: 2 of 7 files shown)


ground/gcs/src/plugins/uavtalk/telemetrymonitor.cpp
```diff
@@ -297,14 +308,20 @@ void TelemetryMonitor::transactionCompleted(UAVObject *obj, bool success)
         if (!success && (retries < IAP_OBJECT_RETRIES)) {
             ++retries;
             obj->requestUpdate();
+
+            return;
         }
     }
+
     // Disconnect from sending object
+    requestsInFlight--;
```
PTDreamer: are you sure this is thread safe?

PTDreamer:
:+1:

ground/gcs/src/plugins/uavtalk/telemetry.cpp
```diff
@@ -377,23 +377,21 @@ void Telemetry::transactionTimeout(ObjectTransactionInfo *transInfo)
     transInfo->timer->stop();
     // Check if more retries are pending
     if (transInfo->retriesRemaining > 0) {
-        TELEMETRY_QXTLOG_DEBUG(
-            QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 Retrying")
+        qDebug() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 Retrying")
```
tracernz: I'll clean this up with categories that can be en/disabled at runtime in a future PR. :) FWIW I think a timeout is worthy of a higher level, probably qInfo.

ground/gcs/src/plugins/uavtalk/telemetry.cpp
```diff
@@ -377,23 +377,21 @@ void Telemetry::transactionTimeout(ObjectTransactionInfo *transInfo)
...
                                + QString::number(transInfo->obj->getObjID(), 16).toUpper()))
-                .arg(transInfo->obj->getInstID()));
+                .arg(transInfo->obj->getInstID());
         --transInfo->retriesRemaining;
         processObjectTransaction(transInfo);
         ++txRetries;
     } else {
-        TELEMETRY_QXTLOG_DEBUG(
-            QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 no more retries. FAILED "
+        qDebug() << QString("[telemetry.cpp] Transaction timeout:%0 Instance:%1 no more retries. FAILED "
```
tracernz: This should be a warning IMO.

ground/gcs/src/plugins/uavtalk/telemetry.cpp
```diff
@@ -441,9 +439,8 @@ void Telemetry::processObjectUpdates(UAVObject *obj, EventMask event, bool allIn
             ++txErrors;
             obj->emitTransactionCompleted(false);
             obj->emitTransactionCompleted(false, false);
-            TELEMETRY_QXTLOG_DEBUG(
-                QString(tr("Telemetry: priority event queue is full, event lost (%1)")
-                            .arg(obj->getName())));
+            qDebug() << QString(tr("Telemetry: priority event queue is full, event lost (%1)")
```
tracernz: Also a warning.

ground/gcs/src/plugins/uavtalk/telemetry.cpp
```diff
@@ -452,6 +449,8 @@ void Telemetry::processObjectUpdates(UAVObject *obj, EventMask event, bool allIn
             ++txErrors;
             obj->emitTransactionCompleted(false, false);
             obj->emitTransactionCompleted(false);
+            qDebug() << QString(tr("Telemetry: event queue is full, event lost (%1)")
```
tracernz: qWarning

ground/gcs/src/plugins/uavtalk/telemetry.h
```diff
@@ -91,8 +91,10 @@ class Telemetry : public QObject
 
 private:
     // Constants
-    static const int REQ_TIMEOUT_MS = 250;
-    static const int MAX_RETRIES = 2;
+    static const int REQ_TIMEOUT_MS = 1500;
+    // Would be nice to dynamically scale the request timeout... this 1500 value
+    // is about what's necessary for 9600bps uavtalk links.
```
tracernz: Indeed, maybe mark TODO so it gets collected in the list.

ground/gcs/src/plugins/uavtalk/telemetrymonitor.cpp
```diff
@@ -269,17 +270,27 @@ void TelemetryMonitor::retrieveNextObject()
...
+        UAVObject *obj = queue.dequeue();
+        // Connect to object
+        TELEMETRYMONITOR_QXTLOG_DEBUG(QString("%0 requestiong %1 from board INSTID:%2")
+                                          .arg(Q_FUNC_INFO)
+                                          .arg(obj->getName())
+                                          .arg(obj->getInstID()));
+
+        requestsInFlight++;
+
+        connect(obj, QOverload<UAVObject *, bool>::of(&UAVObject::transactionCompleted), this,
```
tracernz, line 289: Would be nice to go to C++14 so we can make these a bit shorter with qOverload :wink: .

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
```diff
@@ -123,23 +111,26 @@ UAVTalk::ComStats UAVTalk::getStats()
...
 {
-    quint8 tmp;
-
-    if (io && io->isReadable()) {
-        while (io->bytesAvailable() > 0) {
-            io->read((char *)&tmp, 1);
-            processInputByte(tmp);
+    while (io && io->isReadable()) {
+        if (startOffset > (sizeof(rxBuffer) - MAX_PACKET_LENGTH)) {
+            /* If we're not sure there's room for a frame, shift things left in
```
tracernz: Would be better off to use a std::vector IMO and let it grow if need be. Then you can also do things like std::find to get the first sync byte in the buffer, and no more pointer arithmetic. That could be a refactor for a future PR.

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
```diff
@@ -123,23 +111,26 @@ UAVTalk::ComStats UAVTalk::getStats()
...
-{
-    QUdpSocket *socket = qobject_cast<QUdpSocket *>(sender());
-    QByteArray junk;
-    while (socket->hasPendingDatagrams()) {
-        junk.resize(socket->pendingDatagramSize());
-        socket->readDatagram(junk.data(), junk.size());
+        int bytes = io->read((char *) (rxBuffer + filledBytes),
+                sizeof(rxBuffer) - filledBytes);
+
+        if (bytes <= 0) return;
```
tracernz: newline

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
```diff
@@ -191,244 +182,128 @@ bool UAVTalk::objectTransaction(UAVObject *obj, quint8 type, bool allInstances)
...
-            } else if (rxObj == NULL) {
-                // This is a non-existing object, just skip to checksum
-                // and we'll send a NACK next.
-                rxState = STATE_CS;
-                UAVTALK_QXTLOG_DEBUG("UAVTalk: ObjID->CSum (no obj)");
-                rxInstId = 0;
-                rxCount = 0;
-                break;
-            }
+    quint8 ourCrc = updateCRC(0, rxBuffer + startOffset, hdr->size);
```
tracernz: Just do an extra byte and check the result is zero?

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
```diff
@@ -191,244 +182,128 @@ bool UAVTalk::objectTransaction(UAVObject *obj, quint8 type, bool allInstances)
...
-            // Determine data length
-            if (rxType == TYPE_OBJ_REQ || rxType == TYPE_ACK || rxType == TYPE_NACK) {
-                rxLength = 0;
-            } else {
-                rxLength = rxObj->getNumBytes();
-            }
+    if (ourCrc != *theirCrc) {
+        /* Since we can't trust hdr->size for sure, we should just skip
+         * forward one byte.
+         */
```
tracernz, line 239: :+1:

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
```diff
@@ -191,244 +182,128 @@ bool UAVTalk::objectTransaction(UAVObject *obj, quint8 type, bool allInstances)
...
-            UAVTALK_QXTLOG_DEBUG("UAVTalk: InstID->InstID");
-            break;
-        }
+    quint32 rxObjId = qFromLittleEndian(hdr->objId);
 
-        rxInstId = (qint16)qFromLittleEndian<quint16>(rxTmpBuffer);
+    UAVObject *rxObj = objMngr->getObject(rxObjId);
 
-        rxCount = 0;
+    if (rxObj == NULL) {
```
tracernz: Q_NULLPTR (which is nullptr, NULL or 0 depending on compiler).

ground/uavobjgenerator/uavobjectparser.h
```diff
@@ -41,7 +41,7 @@
  * The flash slot size is 256 bytes which is comprised of the FlashFS header (12 bytes)
  * and the UAVO. This leaves a maximum of 244 bytes for the UAVO.
  */
-#define UAVO_MAX_SIZE 244
+#define UAVO_MAX_SIZE 243
```
tracernz: Need to update comment too.

ground/gcs/src/plugins/uavtalk/telemetry.cpp
mlyle: Will promote all of these.  We don't **really** use levels now but it would be nice to in the future.

ground/gcs/src/plugins/uavtalk/telemetry.h
mlyle: Just TODO: -- does doxygen scrape that up somewhere tehn?

ground/gcs/src/plugins/uavtalk/telemetrymonitor.cpp
mlyle: Yah.  Feels kinda early to go C++14 still from a tools, etc, standpoint though.

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
mlyle: I disagree here-- I like vectors but the objective here is three things: A) copy almost all bytes once (which vector may or may not do for you depending on impl and semantics), B) don't call into the io device too much, and C) use a fixed, relatively small amount of buffer memory [but enough that copying can be limited for A and number of calls can be limited for B],

The cost here in skipping to a sync byte is very high, though-- I pretty much ignored it because I figured it should be rare.

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
mlyle: ergh I hate Q_\*.  ;)

Will change-- in the future we may want to consider nullptr since we're C++11 anyways.

ground/uavobjgenerator/uavobjectparser.h
mlyle: OK

ground/gcs/src/plugins/uavtalk/uavtalk.cpp
mlyle: Could-- but that's both slightly move expensive and relies on a property of CRCs that not all may know.

mlyle:
Push'd a new commit to clean these things up.

tracernz (approved):
Thanks. :+1:

Issue #1671, opened 2017-03-27
GCS: UAVBrowser Spinbox Zeros Values > 2^31-1


tracernz:
Ideally need this fixed to make a reasonable SpinBox64 impl. :( https://bugreports.qt.io/browse/QTBUG-45627

PR #1811, opened 2017-06-11
gcs: hwsettings: handle uint32 data

mlyle:
Namely, radio IDs.  If the value exceeded 2^31, it was clipped to 0.
This meant if you saved the hardware page it'd unbind your OpLink, etc.

Now, use a double spinbox to represent these values.  Though there is
still an unsightly thing of the units "hex" being displayed when the
value is decidedly not hexy.


mlyle:
This is something users have reported before and that I had blamed on telemetry-suck.  Turns out it's not!

tracernz:
This is another case of https://github.com/d-ronin/dRonin/issues/1671. I'd like to fix this with a 64-bit spinbox but unfortunately the implementation of QSpinbox doesn't trivially allow this. Perhaps I should just go ahead and do it to solve this everywhere.

mlyle:
Would be nice to have a real hex variant, too.

tracernz:
Yeah, I have some WIP for that, and the ability to specify the default display radix in UAVO field attributes (so it would work for default HW form etc.).

PR #1813, opened 2017-06-15
gcs: file transfer functionality

mlyle:
builds upon #1810

### DIFF
--- a/ground/gcs/src/plugins/uavtalk/uavtalk.h
+++ b/ground/gcs/src/plugins/uavtalk/uavtalk.h
@@ -48,7 +48,7 @@ class UAVTALK_EXPORT UAVTalk : public QObject
     Q_OBJECT
 
 public:
-    typedef struct
+    struct ComStats
     {
         quint32 txBytes;
         quint32 rxBytes;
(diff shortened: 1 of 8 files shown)


mlyle:
With most recent commit, 32768 byte autotune part (and we're capped to not transfer more than this) file transfer is about 2,5 seconds over USB from Revolution, and about 40 seconds over a 64kbps oplink connection (about 13% of theoretical channel bandwidth).

Most of the problem with oplink is that other stuff takes over the channel between "chunks" and fills up rather large buffers that we have.  I've explored making changes to curtail some of telemetry when GCS is actively requesting stuff, but this is problematic for a number of reasons.  I think the real solution is lower-latency buffer management and link prioritization.

mlyle:
Rebased

ground/gcs/src/plugins/uavtalk/uavtalk.h
```diff
@@ -114,6 +121,14 @@ private slots:
         quint8 resv;
         quint32 objId;
     } UAVTalkHeader;
+
+    typedef struct {
```
tracernz: Why typedef? No need in C++.

mlyle: Yah good point-- right next to other code that did it :P

tracernz (approved):

PR #1814, opened 2017-06-15
PiOS: Introduce generic IRQ prologue/epilogue macros

tracernz:
Clean up a lot of ugly ifdef

Follows up #1787


mlyle:
Dullest diff to review ever :D

PR #1816, opened 2017-06-15
UAVO: Remove incorrect "hex" unit from Rfm CoordID fields

tracernz:
Clearly bogus. 

TODO: Future PR should make it possible to set default display base that is used in UAVO Browser, default Hardware form, and ConfigTaskWidget relations.


mlyle:
This doesn't change uavo IDs right?  (Units aren't part of the hash?)

mlyle:
Seems not.

PR #1817, opened 2017-06-15
pios: f4: flash_internal: erase flash from RAM

### LINKED ISSUE #1482: Erasing internal flash at runtime on STM32F4 targets problematic
tErase128KB is 1.3s typical, 2.6s max.  Erasing smaller pages takes a long, long time too.  During this time flash reads stall (e.g. we can't load new code to run).

This is problematic for AQ32 and Seppuku, which don't have outboard flash.

Proposed solution:

* [ ] Move interrupt handlers to SRAM on F4
* [ ] Move RTC interrupt handler to SRAM on F4---- requires infrastructure for code-in-ram
* [ ] Teach RTC interrupt to reset IWDG when A) flash operation in progress and B) disarmed

mlyle:
Need to have this one little routine in RAM, so that we can do
lengthy flash operations without hanging long enough for watchdog
to expire.

Still needs to have a way to test arming status at the outset. I'm
thinking probably a weak symbol or something like that (as this code can
be used in the loader, too).

fixes #1482

### DIFF
--- a/flight/PiOS/STM32F4xx/pios_flash_internal.c
+++ b/flight/PiOS/STM32F4xx/pios_flash_internal.c
@@ -7,6 +7,8 @@
  *
  * @file       pios_flash_internal.c  
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2013
+ * @author     dRonin, http://dRonin.org/, Copyright (C) 2017
+ *
  * @brief Provides a flash driver for the STM32 internal flash sectors
  *****************************************************************************/
 /* 
@@ -33,6 +35,9 @@
 #include "pios_wdg.h"
 #include "pios_semaphore.h"
 #include <stdbool.h>
+
+/* Todo: consider relocating to a good header */
+bool __attribute__((weak)) vehicle_is_armed = false;
 
 static const uint16_t sector_to_st_sector_map[] = {
 	[0] = FLASH_Sector_0,
@@ -171,8 +176,7 @@ static int32_t PIOS_Flash_Internal_EraseSector(uintptr_t chip_id, uint32_t chip_
 
 int32_t PIOS_Flash_Internal_EraseSector_FromRam(uint16_t st_sector)
 {
-	/* XXX armed-check infrastructure */
-	if (0) {
+	if (vehicle_is_armed) {
 		return -1;
 	}
 
--- a/flight/PiOS/STM32F4xx/pios_flash_internal.c
+++ b/flight/PiOS/STM32F4xx/pios_flash_internal.c
@@ -164,7 +164,11 @@ static int32_t PIOS_Flash_Internal_EraseSector(uintptr_t chip_id, uint32_t chip_
 		return ret;
 	}
 
-	return FLASH_GetStatus();
+	if (FLASH_GetStatus() != FLASH_COMPLETE) {
+		return -1;
+	}
+
+	return 0;
 }
 
 int32_t PIOS_Flash_Internal_EraseSector_FromRam(uint16_t st_sector)
--- a/flight/PiOS/STM32F4xx/pios_flash_internal.c
+++ b/flight/PiOS/STM32F4xx/pios_flash_internal.c
@@ -132,6 +132,11 @@ static int32_t PIOS_Flash_Internal_EndTransaction(uintptr_t chip_id)
 	return 0;
 }
 
+// XXX eliminate section attribute warning by adding new section to
+// linker scripts
+int32_t PIOS_Flash_Internal_EraseSector_FromRam(uint16_t st_sector)
+	__attribute__ ((section (".data"),long_call));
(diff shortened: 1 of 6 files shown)


tracernz (approved):

mlyle:
(doesn't work according to @yamahabest )

yamahabest:
Well, there was the issue that the board was not recognized by Windows after auto-upgrading and rebooting. Even after flashing manually it didn't work.
When going back to the previous working firmware, and skipping the auto-upgrade, and flashing manually (through Rescue) the board was recognized by Windows.

The other issue is that after autotune, and disarming, all ESCs start beeping. I believe the FC on disarm reboots/hangs during reboot?
If I leave the GCS connected during the autotune, the representation of the aircraft gets updated irregularly after disarming, just if the telemetry feed is also affected. After disarming, that is. So that would say that the FC doesn't reboot?

mlyle:
The code obviously doesn't work :P

yamahabest:
I'm just writing down what I noticed ;-)

mlyle:
@yamahabest any chance you can try again?  Seems to work on seppuku, at least...

yamahabest:
Sure, can you make Jenkins do a build? Just to rule out any local stuff on my side...
Will take a few days though. Can probably test on Wednesday.

tracernz:

mlyle:

mlyle:
Rebased, new commits.  jenkins, build artifacts please

flight/PiOS/STM32F4xx/pios_flash_internal.c
```diff
@@ -34,6 +36,9 @@
 #include "pios_semaphore.h"
 #include <stdbool.h>
 
+/* Todo: consider relocating to a good header */
+bool __attribute__((weak)) vehicle_is_armed = false;
```
tracernz: This is kinda ugly IMO. Would be much better to keep the flash API clean and have the module call into it.

mlyle: To be safe, we have to check after disabling interrupts.

mlyle: Oh, you mean a direct cross-module call from manualcontrol -> flash? But f4 flash internal isn't present on all targets, so we could have a weak call symbol or..

tracernz: Yes? What I'm proposing is making this a static local var, and adding a public method to set it which the module can call.

mlyle: OK, but this file isn't in many targets-- do we stub it out?  what if someone doesn't PIOS_INCLUDE_FLASH_INTERNAL etc

tracernz: ifdef like we do everywhere else?

mlyle: I looked at the pios_sys idea and it's not good because it's also not shared everywhere (would need to duplicate code between arches).  Putting it in pios_flash_internal isn't good because the code in manualcontrol would need two ifdefs-- F4 and PIOS_INCLUDE_FLASH_INTERNAL which IMO is destroying a lot of abstraction.

Can we defer?  I think we need a new module to keep track of state like this, and I feel like we need to get into shipping cadence soon.

tracernz: Yep. Not sure the doxygen thing will pick up Todo (vs. TODO) though. :stuck_out_tongue:

mlyle: Improved the TODO comment :P

tracernz:
Just needs test at this point.

mlyle:
Yup--- I've mocked up the AQ32 conditions as best as I can on Revolution, but @yamahabest --- testing on actual AQ32 would really be appreciated.

yamahabest:
Yep, will do, this evening.

yamahabest:
I have tested this.
The first time during upgrading, the board didn't reboot completely, but after unplugging and replugging, the wizard continued. Tried again, and it worked without problems. First time could have been an anomily on my system though.

ESCs still beep, only once, directly after disarming, FC still rebooting?
During my previous test, the ESCs kept on beeping.

The FC does seem to store the Autotune measurements, as the Autotune Wizard comes up when connecting the FC?

mlyle:
I assume the beep is because output PWM stops for the couple seconds while flash is resetting.

yamahabest:
Ok, is there a way I could actually check that?

mlyle:
Sure, you could connect USB and look at uptime after it saves the tune data.

yamahabest:
Ah, right. Will test that once I get to the FC again, will take a couple of days again...

PR #1818, opened 2017-06-16
GCS:LongLongSpinBox: Fix use of C limits

tracernz:
C standard doesn't include nearly as much, which means it doesn't require LONG_LONG_MIN (it is a GNU and MSVC extension apparently). Broke build on macOS.

`qint64` is a typedef for `long long` except with MSVC where it is `__int64`. MSVC defines `std:numeric_limts<__int64>` and linux + mac have `std::numeric_limits<long long>`.


tracernz:

PR #1819, opened 2017-06-17
GCS: Support preferred integer display base other than 10 for UAVO fields

tracernz:
Supports dec (0), hex (0x0), bin (0b0), oct (0o0).


tracernz:

PR #1820, opened 2017-06-17
GCS:Browser: Propagate default value state up to objects

tracernz:
Show objects in bold when any field is non-default.

### DIFF
--- a/ground/gcs/src/plugins/uavobjectbrowser/treeitem.h
+++ b/ground/gcs/src/plugins/uavobjectbrowser/treeitem.h
@@ -174,8 +174,15 @@ public:
         return 0;
     }
 
-    bool isDefaultValue();
-    void setIsDefaultValue(bool isDefault);
+    virtual bool isDefaultValue() const { return true; }
+
+protected:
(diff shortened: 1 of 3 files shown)


tracernz:
Ahhh, this doesn't work for multi-element fields...

tracernz:
And fixed by overriding isDefaultValue in ArrayFieldTreeItem too.

ground/gcs/src/plugins/uavobjectbrowser/treeitem.h
```diff
@@ -174,8 +174,15 @@ class TreeItem : public QObject
...
-    bool isDefaultValue();
-    void setIsDefaultValue(bool isDefault);
+    virtual bool isDefaultValue() const { return true; }
+
+protected:
+    bool childrenAreDefaultValue() const {
+        bool ret = true;
+        for (const auto item : treeChildren())
+            ret &= item->isDefaultValue();
+        return ret;
```
tracernz, line 184: Should have used std:all_of :P

PR #1824, opened 2017-06-17
GCS: Fix name of .clang_format symlink

tracernz:
Needs to be .clang-format for clang-format to find it.


tracernz:
IMO it should move up to ./ground because the same code style applies to all ground code. Thoughts?

mlyle:
Ah, sorry I messed this up.  I"m fine with either place.

tracernz:
No problem! :smiley: I tried moving it up and for some inexplicable reason it didn't work... The relevant piece of clang hasn't changed for 4 years so I don't think it's related to that: https://github.com/llvm-mirror/clang/blob/master/lib/Format/Format.cpp#L2028. Let's leave it where it is for now.

PR #1834, opened 2017-06-21
pios_flash_posix: Don't keep a transaction flag

mlyle:
Multiple things want to do stuff with flash at once.  On other targets,
a strong synchronization mechanism is used (semaphore), which ensures
waiting.  This was just a flag and would blow up if you did tightly
timed operations.

pios_flash_posix should get moved out of tests sometime soon.


tracernz:
Add TODO to each bit so they get added to the list IMO.

mlyle:
Doc fixup made.

PR #1835, opened 2017-06-22
neotune: factoring improvements, progress notification

mlyle:
Pretty close to complete.

### DIFF
--- a/ground/gcs/src/plugins/config/config.pro
+++ b/ground/gcs/src/plugins/config/config.pro
@@ -105,6 +105,7 @@ FORMS += airframe.ui \
     modules.ui \
     osd.ui \
     osdpage.ui \
+    autotunebeginning.ui \
     autotuneproperties.ui \
     autotunesliders.ui \
     autotunefinalpage.ui
--- a/ground/gcs/src/plugins/config/config.pro
+++ b/ground/gcs/src/plugins/config/config.pro
@@ -5,6 +5,7 @@ DEFINES += QWT_DLL
 
 QT += svg
 QT += network
+QT += charts
 
 include(../../gcsplugin.pri)
 
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -706,6 +706,7 @@ AutotuneBeginningPage::AutotuneBeginningPage(QWidget *parent,
     av = autoValues;
     this->autoOpened = autoOpened;
     dataValid = false;
+    setupUi(this);
 }
 
 QString AutotuneBeginningPage::tuneValid(bool *okToContinue) const
@@ -776,12 +777,11 @@ void AutotuneBeginningPage::initializePage()
     setTitle(tr("Preparing autotune..."));
 
     if (autoOpened) {
-        setSubTitle(
-            tr("It looks like you have run a new autotune since you last connected to the flight "
+        subtitle->setText(tr("It looks like you have run a new autotune since you last connected to the flight "
                "controller.  This wizard will assist you in applying a set of autotune "
                "measurements to your aircraft."));
     } else {
-        setSubTitle(tr("This wizard will assist you in applying a set of autotune "
+        subtitle->setText(tr("This wizard will assist you in applying a set of autotune "
                                   "measurements to your aircraft."));
     }
 
@@ -791,7 +791,14 @@ void AutotuneBeginningPage::initializePage()
         TelemetryManager *telMngr = pm->getObject<TelemetryManager>();
 
         if (!av->data) {
-            av->data = telMngr->downloadFile(4, 32768);
+            av->data = telMngr->downloadFile(4, 32768, [&](quint32 progress) {
+                        float fraction = progress / 32768.0;
+
+                        if (fraction > 0.8f) {
+                            fraction = 0.8f;
+                        }
+                        qDebug() << "Progress " << fraction;
+                    });
         }
 
         if (av->data) {
@@ -801,12 +808,7 @@ void AutotuneBeginningPage::initializePage()
 
     QString initialWarnings = tuneValid(&dataValid);
 
-    QLabel *status = new QLabel(initialWarnings);
-    status->setWordWrap(true);
-
-    QVBoxLayout *layout = new QVBoxLayout;
-    layout->addWidget(status);
-    setLayout(layout);
+    status->setText(initialWarnings);
 }
 
 bool AutotuneBeginningPage::isComplete() const
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -36,21 +36,22 @@
 #include "configautotunewidget.h"
 
 #include <QtAlgorithms>
+#include <QChartView>
+#include <QClipboard>
 #include <QCryptographicHash>
 #include <QDebug>
-#include <QFileDialog>
-#include <QStringList>
-#include <QWidget>
-#include <QTextEdit>
-#include <QVBoxLayout>
-#include <QMessageBox>
-#include <QPushButton>
 #include <QDesktopServices>
-#include <QUrl>
+#include <QFileDialog>
 #include <QList>
 #include <QMessageBox>
-#include <QClipboard>
+#include <QPushButton>
+#include <QStringList>
+#include <QTextEdit>
+#include <QUrl>
+#include <QValueAxis>
+#include <QVBoxLayout>
 #include <QVector>
+#include <QWidget>
 #include <QWizard>
 #include <QtNetwork/QNetworkAccessManager>
 #include <QtNetwork/QNetworkRequest>
@@ -463,6 +464,26 @@ AutotuneMeasuredPropertiesPage::AutotuneMeasuredPropertiesPage(QWidget *parent,
     this->av = autoValues;
 }
 
+QChart *AutotuneMeasuredPropertiesPage::makeChart(int axis) {
+    QChart *chart;
+
+    chart = new QChart();
+
+    av->actual[axis]->setName(tr("Actual"));
+    chart->addSeries(av->actual[axis]);
+    av->model[axis]->setName(tr("Predicted"));
+    chart->addSeries(av->model[axis]);
+
+    chart->createDefaultAxes();
+
+    chart->axisX()->setTitleText(tr("Time"));
+    dynamic_cast<QValueAxis *>(chart->axisX())->setLabelFormat("%.0f ms");
+    chart->axisY()->setTitleText(tr("Angular acceleration"));
+    chart->axisY()->setLabelsVisible(false);
+
+    return chart;
+}
+
 void AutotuneMeasuredPropertiesPage::initializePage()
 {
     measuredRollGain->setText(QString::number(av->beta[0], 'f', 2));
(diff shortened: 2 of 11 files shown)


mlyle:
Not quite ready, need to send a couple signals.  Woops.

mlyle:
OK, fixed up

mlyle:
image

mlyle:
rebased

mlyle:
jenkins, test this please

tracernz:
Plot looks good. :+1:

tracernz:
Plot looks good. :+1:

jenkins, test this please

tracernz:
No ox again.

tracernz:

tracernz:
@mlyle there you go. 🙃

pug398:
@mlyle I got this...
neotune_warn

mlyle:
#1840 made WIndows GCS not work-- reverted.

mlyle:
oh crap, back to having conflicts, so i don't even want those set of artifacts.. re-rebasing

mlyle:

mlyle:

pug398:
@mlyle Getting there. It only wipes out the config gadget now :)
neo_warn

amstan:
I ran an autotune on my quadcopter inside. Here's the autotune partition: [jun24.autotune.bin.zip](https://github.com/d-ronin/dRonin/files/1100040/jun24.autotune.bin.zip)

pug398:
@amstan Are you on windows and get above warning message? The path looks good so I am unsure if it might be my machine.

mlyle:
@pug398 I think it's the effects of auto-updates on the windows builder.  Actual code is fine, it seems.

mlyle:

mlyle:

mlyle:
Still doesn't work.  I'm fairly stumped.  Here's what we know:

1. Windows ran updates on its own on the builder a couple days ago.  Yuck.
2. Before the revert earlier today, a huge set of modules didn't build.  With the revert, it is only this PR that doesn't build/only config plugin.
3. I tried rm -rf of the .ccache directory to see if it helped.  It doesn't.
4. I tried including charts at toplevel because it looks like the windeployqt stuff looks for it.  No dice.

ground/gcs/src/app/app.pro
```diff
@@ -6,6 +6,8 @@ TARGET = $$GCS_APP_TARGET
 DESTDIR = $$GCS_APP_PATH
 QT += xml
 QT += widgets
+QT += charts
```
tracernz: ~~What's this for?~~ :confused:
Oh, I see your comment above. This is definitely not needed. These are the modules drgcs.exe uses, not what the plugins use (they are completely independent shared libraries).

tracernz:
Need to put charts in here I think https://github.com/d-ronin/dRonin/blob/next/package/Makefile.win#L14

tracernz:
So, running in build environment works, but packages don't, or? And next packages are ok but not this branch?

ground/gcs/src/plugins/config/config.pro
```diff
@@ -5,6 +5,7 @@ DEFINES += QWT_DLL
 
 QT += svg
 QT += network
+QT += charts
```
tracernz, line 8: This is correct.

tracernz:

tracernz:

tracernz:
@mlyle Those artifacts work fine. :wink: Just need to remove `Qt += charts` from app.pro.

tracernz:
Patch required to fix the issue: https://github.com/tracernz/qttools/commit/da9cb922532723891416214501d99ffeb21ff37a
I should probably attempt to push it upstream for 5.9.[2?].

mlyle:
Rebased out the QT += charts

mlyle:
No, I dind't, I screwed up the force push and didn't notice.  Now I did.  So much fail :(

pug398:
@mlyle no fails it's all good :)

mlyle:
finally.

@pug398 -- thanks for your patience.  :)

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -395,417 +333,162 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
...
+                       | Qt::WindowCloseButtonHint);
 
-    for (int i = 0; i < pts; i++) {
-        float y = b0 * (data[i] + 2.0f * x1 + x2) + a1 * y1 + a2 * y2;
+    wizard.setPixmap(QWizard::BackgroundPixmap, QPixmap(":/configgadget/images/autotunebg.png"));
 
-        y2 = y1;
-        y1 = y;
+    // Used to track state across the entire wizard path.
+    AutotunedValues av;
```
tracernz: Shadows the member variable? I would strongly prefer the member variable to have a better name since it has a _long_ lifetime and large scope, so probably best to change it to `tuneValues` or something.

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -395,417 +333,162 @@ void ConfigAutotuneWidget::stuffShareForm(AutotuneFinalPage *autotuneShareForm)
...
-        product[i] = -(u * x) - (v * y);
-        product[i + fpts] = (v * x) - (u * y);
-    }
+    if (av.valid && (wizard.result() == QDialog::Accepted) && av.converged) {
+        // Apply and save data to board.
+        StabilizationSettings *stabilizationSettings =
+            StabilizationSettings::GetInstance(getObjectManager());
+        Q_ASSERT(stabilizationSettings);
+        if (!stabilizationSettings)
+            return;
```
tracernz: Can you add a qWarning here? That way we will know if there's a problem rather than it happening silently.

ground/gcs/src/plugins/config/configautotunewidget.cpp
```diff
@@ -1036,3 +732,357 @@ AutotuneFinalPage::AutotuneFinalPage(QWidget *parent)
...
+    } else {
+        subtitle->setText(tr("This wizard will assist you in applying a set of autotune "
+                                  "measurements to your aircraft."));
+    }
+
+    progressBar->setRange(0, 100);
+    progressBar->setValue(0);
+
+    QMetaObject::invokeMethod(this, "doDownloadAndProcess", Qt::QueuedConnection);
+
```
tracernz: Superfluous empty line.

ground/gcs/src/plugins/config/configautotunewidget.h
```diff
@@ -27,28 +27,36 @@
...
+#include "ui_autotuneproperties.h"
+#include "ui_autotunesliders.h"
+#include "ui_autotunefinalpage.h"
 #include "configgadgetwidget.h"
 
 struct AutotunedValues
 {
     bool valid;
 
+    QByteArray *data;
```
tracernz: Reference preferred over pointer IMO.

ground/gcs/src/plugins/uavtalk/telemetry.h
```diff
@@ -84,7 +84,9 @@ class Telemetry : public QObject
     Telemetry(UAVTalk *utalk, UAVObjectManager *objMngr);
     ~Telemetry();
     TelemetryStats getStats();
-    QByteArray *downloadFile(quint32 fileId, quint32 maxSize);
+    QByteArray *downloadFile(quint32 fileId, quint32 maxSize,
+            std::function<void(quint32)>progressCb = nullptr);
```
tracernz, line 88: This is a bit ugly. Telemetry really needs to be in a different thread. :/

ground/gcs/src/plugins/uavtalk/telemetry.cpp
```diff
@@ -398,7 +398,12 @@ void Telemetry::transactionTimeout(ObjectTransactionInfo *transInfo)
     }
 }
 
-QByteArray *Telemetry::downloadFile(quint32 fileId, quint32 maxSize)
+/* This is synchronous, so we use a primitive callback mechanism
+ * instead of signal/slot.  Can have a future async variant if
+ * necessary
```
tracernz, line 403: Necessary to shift to async in future IMO.

ground/gcs/src/plugins/uavtalk/telemetry.h
mlyle: I don't know if I agree, but if we had a fully-async version it can all be signals (whether same thread or different).

it's just when it's synchronous, it seems dumb to connect/disconnect.

ground/gcs/src/plugins/uavtalk/telemetry.cpp
mlyle: Probably for some things.

I really like intuitive code describing the sequence things will happen in instead of ... signals cascading to each other willy nilly and efforts to jam a crude state machine onto them.  So even if this becomes async, I think we'd want one level up to be synchronous/fairly imperative.

But conversely, we can use a lot of stack if there is any real concurrency between things using this model.

ground/gcs/src/plugins/config/configautotunewidget.cpp
mlyle: I don't think it has a member variable of type AutotunedValues?

ground/gcs/src/plugins/config/configautotunewidget.cpp
mlyle: OK

ground/gcs/src/plugins/config/configautotunewidget.cpp
mlyle: OK

ground/gcs/src/plugins/config/configautotunewidget.h
mlyle: If it's a reference, it needs to be known at the moment of struct construction, doesn't it?  I don't think it's easy to structure the code that way.

mlyle: (I go back and forth on whether downloadFile should pass back a pointer to a constructed byte array or fill in something the user provides.  There's yuckiness each way)

tracernz: No, the default initialiser will be used, which is an empty container (sizeof = sizeof(void *)). Assignment operator https://github.com/qt/qtbase/blob/d1ff9b419da5e3ace68479b02582222486ec9703/src/corelib/tools/qbytearray.h#L177 which simply assigns the d pointer from the other container (i.e. it costs the same as assigning a pointer). This is the way Qt containers are designed to be used.

tracernz: Pardon me, should just be a QByteArray, not a pointer or ref.

ground/gcs/src/plugins/uavtalk/telemetry.h
tracernz: Not really, it would be consistent with the rest of the codebase whereas this is not.

ground/gcs/src/plugins/config/configautotunewidget.cpp
tracernz: https://github.com/d-ronin/dRonin/pull/1835/files#diff-803cd9c265c5c58af660006e345bf45eR105

tracernz: Ah, that's the page. Anyway, av is a bad name for that member variable IMO. Short meaningless names like that should be reserved for very tight scope like a loop var.

ground/gcs/src/plugins/uavtalk/telemetry.h
mlyle: I think QEventLoop is newer than a lot of the codebase, as are lambdas.  We also do relatively few things in the main UI  that are multistep-oriented... uploader, calibration, log download... and the wizards.  I think these would all be better written as eventloops (though cal should really be modal)

ground/gcs/src/plugins/config/configautotunewidget.h
mlyle: OK, that makes more sense-- will do

mlyle:
OK, this was a fair bit of work, but fixed up per comments.

Also made two minor functional changes:

1. For autotune, use the average of roll and pitch tau for now, instead of just roll tau.
2. Turn the long-standing experimental parameters that we tell everyone to turn on-- Yaw & Outer Ki-- on by default

jhitesma:
> Turn the long-standing experimental parameters that we tell everyone to turn on-- Yaw & Outer Ki-- on by default

👍   should help cut down on a good bit of confusion/concern from end users

tracernz (approved):

mlyle:
Rebased to fix conflict.

Issue #1689, opened 2017-04-02
bootloader-updater not working for revolution target


henn1001:
* Seen in version (Noticed on Artifice - quixote and onwards?): 
* On flight controller (revolution): 

Issue details:
I was trying to update the bootloader on a airbot f4 because i reconfigured the alarm led to buzzer (to make  it shut up while booting or flashing :p). I noticed that the bu-updater didn't work. After flashing & pressing boot, it just boots into apparently nothing - no heartbeat, nothing.  

I retested it on another board cloner board ( "cc3d revolution"). Same outcome. 
So i tried a different board and for dtfc and sparky2 it´s updating the bootloader as expected.

While testing older releases i found that the last one where it´s working, is the Samsara release.

Maybe someone can cross check it :)

pug398:
@henn1001 I have a revolution I can check it on but was wondering if you have flashed ef_revolution.hex to the board (working?) and then flashed bu_revolution.tlfw to the board (still working?) both from the current build? (I take it you mean the bootloader is corrupt and you cannot rescue)

henn1001:
I flashed the ef_revolution.hex, yes. The bootloader on that combination is running fine. I can rescue and update firmware just fine and everything is working. It´s just the bu_revolution that is not working.

The bu_ does nothing. After i flash the bu_ and reboot it just boots into an empty loop or something alike.
I can still rescue though and flash a normal firmware - the bootloader is not getting corrupt.

pug398:
Ok so yes I have had to flash the main code back after flashing bu updater on revolution. My understanding was bu updater loads into main memory and extracts and writes the bootloader from there. I assume it is overwriting key part of main code. Whether that is a bug or just the process I am not sure.

henn1001:
Yes that is expected. That bu_ code basically replaces the "main code" i.e. flight firmware. So after you flash the bu_, you reboot the controller. From there the old bootloader jumps into the new application, in this case the bootloaderUpdater. The updater overwrites the old bootloader and replaces it with the new one. After this you need to reflash a normal flight fw_

So did it work for you? You need to check if the bootloader version changes in the recovery screen.

pug398:
[...]

PR #1837, opened 2017-06-22
bu: move to -lc instead of -lc_nano

mlyle:
Relates to #1689 , though I don't think it can be cause because the commit in question was after the reports started coming in.

With this change, bu on sparky2 works.  Without it, it doesn't.  I don't understand why-- perhaps some kind of weird alignment thing or minimum image size thing that we're triggering.


mlyle:
This is not enough to make revolution work.

mlyle:
Moving things in memory breaks BU.  need to understand why.

mlyle:
I don't believe this is root cause, so closing.

PR #1839, opened 2017-06-23
flight: uavtalk: handle reqs for unknown multiinst objs

mlyle:
This may be a factor with the pipx bugs-- since it doesn't know
about a lot of kinds of objects.  And it causes connections to
hang initially when GCS requests multi-inst stuff that the board
doesn't know about.

### DIFF
--- a/flight/Libraries/uavtalk.c
+++ b/flight/Libraries/uavtalk.c
@@ -286,8 +286,19 @@ UAVTalkRxState UAVTalkProcessInputStreamQuiet(UAVTalkConnection connectionHandle
 
 		// Determine data length
 		if (iproc->type == UAVTALK_TYPE_OBJ_REQ || iproc->type == UAVTALK_TYPE_ACK || iproc->type == UAVTALK_TYPE_NACK) {
-			iproc->length = 0;
+			iproc->length = iproc->packet_size - iproc->rxPacketLength;
 			iproc->instanceLength = 0;
+
+			/* Length is always pretty much expected to be 0
+			 * here, but it can be 2 if it's a multiple inst
+			 * obj requested.  Don't peer into metadata to
+			 * figure this out-- use the packet length
+			 * [so we can properly NAK objects we don't know]
+			 */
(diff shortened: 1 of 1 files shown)


flight/Libraries/uavtalk.c
```diff
@@ -286,8 +286,19 @@ UAVTalkRxState UAVTalkProcessInputStreamQuiet(UAVTalkConnection connectionHandle
...
+			iproc->length = iproc->packet_size - iproc->rxPacketLength;
 			iproc->instanceLength = 0;
+
+			/* Length is always pretty much expected to be 0
+			 * here, but it can be 2 if it's a multiple inst
+			 * obj requested.  Don't peer into metadata to
+			 * figure this out-- use the packet length
+			 * [so we can properly NAK objects we don't know]
+			 */
+			if (iproc->length == 2) {
```
tracernz: Could do a stronger check if we know about the object and it's always an error if the length isn't 2 according to our protocol spec (given that flight doesn't accept timestamps):
```c
if (iproc->length == 2 && (!iproc->obj || !UAVObjIsSingleInstance(iproc->obj))) {
    iproc->length = 0;
    iproc->instanceLength = 2;
} else if (iproc->length > 0) {
    iproc->state = UAVTALK_STATE_ERROR;
    break;
}
```

mlyle: Hm.  That's an interesting tradeoff.  We have to go to ERROR on the other side (data path) because there's nothing sane to do with a wrong number of bytes, and we do better going to ERROR if it's a protocol desync issue (vs.line noise on objid or a collision with a different objid in the past).

On this side, I think we want to be eager to accept the frame so we can NACK it.  I think my plan is to take the second half of this conditional but to always allow length 2.

Note one thing we don't handle presently-- no NAK is sent when the instance id is unknown, nor is there a field in the NAK for the instance id.

mlyle:
Fixed up to pick something halfway between feedback and what I had before.

PR #1840, opened 2017-06-23
GCS:UAVSettingsImportExport: Disable commands when not connected

### LINKED ISSUE #1499: Exporting UAVOs with no board attached potentially confusing
* Seen in version (release version or githash): d0ab1073

Issue details:
Autoupdate failed leaving me in bootloader.  Didn't notice I was in bootloader and tried to export my UAVOs - it "worked" but I got an essentially empty UAVO dump:

<uavobjects>
    <version>
        <hardware revision="9f" serial="000000000000000000000000" type="f0"/>
        <firmware date="" hash="" tag=""/>
        <gcs date="detached at 19" hash="19bd0c53" tag="(HEAD detached at 19bd0c5)"/>
    </version>
    <settings/>
</uavobjects>

Which while not wrong isn't exactly very helpful to the user.  I think this may have happened to me once before and resulted in me trying to restore from an empty dump, and I think it may have happened to another user I spoke with.

Perhaps it would be better if the GCS were to give an error and not allow exporting when there's no actual firmware to dump from?  That way no one can accidentally export an empty dump and then wind up having to start from scratch if they try to restore from it.

tracernz:
Fixes #1499

### DIFF
--- a/ground/gcs/src/plugins/plugins.pro
+++ b/ground/gcs/src/plugins/plugins.pro
@@ -227,6 +227,7 @@ plugin_uavsettingsimportexport.subdir = uavsettingsimportexport
 plugin_uavsettingsimportexport.depends = plugin_coreplugin
 plugin_uavsettingsimportexport.depends += plugin_uavobjects
 plugin_uavsettingsimportexport.depends += plugin_uavobjectutil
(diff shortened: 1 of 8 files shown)


mlyle:
Nice.  I like hte cleanups too

PR #1841, opened 2017-06-23
F3: Support WS2811 LEDs

mlyle:
Initially just F3EVO.  Need to look at other targets and determine if it can reasonably work.  Also need to test-- the waveforms look good on a scope but have not been used to drive LEDs.


flight/PiOS/STM32F30x/inc/pios_ws2811_target.h
```diff
@@ -0,0 +1,53 @@
+/**
+ ******************************************************************************
+ *
+ * @file       pios_ws2811_cfg.h
```
tracernz: This needs to be `STM32F30x/inc/pios_ws2811_target.h` so it doesn't conflict with the F4 one. And _cfg.h matches our conventions (and the file contents) better?

flight/PiOS/STM32F4xx/inc/pios_ws2811_target.h
```diff
@@ -0,0 +1,63 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS_WS2811 WS2811 Functions
+ * @{
+ *
+ * @file       pios_ws2811_target.h
```
tracernz: This needs to be `STM32F4xx/inc/pios_ws2811_target.h` so it doesn't conflict with the F3 one. And _cfg.h matches our conventions (and the file contents) better?

flight/PiOS/STM32F30x/inc/pios_ws2811_target.h
```diff
@@ -0,0 +1,53 @@
+/**
```
tracernz: On all files, please add:
```
 * @addtogroup PIOS PIOS Core hardware abstraction layer
 * @{
 * @addtogroup PIOS_WS2811 WS2811 RGB LED Driver
 * @{
```

and the corresponding
```
 * @}
 * @}
```

flight/PiOS/STM32F30x/pios_ws2811.c
```diff
@@ -0,0 +1,226 @@
...
+	ws2811_dev->max_leds = max_leds;
+
+	TIM_TimeBaseInitTypeDef TIM_TimeBaseStructure;
+	TIM_OCInitTypeDef TIM_OCInitStructure;
+	DMA_InitTypeDef DMA_InitStructure;
+
+	GPIO_InitTypeDef gpio_cfg = {
+		.GPIO_Pin = ws2811_cfg->gpio_pin,
+		.GPIO_Mode = GPIO_Mode_AF,
+		/* Drive hard-- the pin is loaded on F3E, and there could be
```
tracernz: F3E shouldn't really matter in a general PiOS driver, the latter part of the comment is appropriately general :P

flight/PiOS/STM32F30x/pios_ws2811.c
```diff
@@ -0,0 +1,226 @@
...
+
+	GPIO_PinAFConfig(ws2811_cfg->led_gpio,
+			__builtin_ctz(ws2811_cfg->gpio_pin), ws2811_cfg->remap);
+
+	/* Compute the prescaler value */
+	uint16_t prescalerValue = (PIOS_SYSCLK / WS2811_TIMER_HZ) - 1;
+
+	/* Time base configuration */
+	TIM_TimeBaseStructInit(&TIM_TimeBaseStructure);
+	TIM_TimeBaseStructure.TIM_Period        = WS2811_TIMER_PERIOD;// 800kHz
```
tracernz: That comment belongs with the def (assuming it's accurate).

flight/PiOS/STM32F30x/pios_ws2811.c
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	TIM_CtrlPWMOutputs(ws2811_cfg->timer, ENABLE);
+
+	DMA_DeInit(ws2811_cfg->dma_chan);
+
+	DMA_StructInit(&DMA_InitStructure);
+	DMA_InitStructure.DMA_PeripheralBaseAddr = (uint32_t)&ws2811_cfg->timer->CCR1 + ws2811_cfg->timer_chan;
+	DMA_InitStructure.DMA_MemoryBaseAddr     = (uint32_t)ws2811_dev->dma_buffer;
+	DMA_InitStructure.DMA_DIR = DMA_DIR_PeripheralDST;
```
tracernz: These should either all be aligned, or none.

flight/PiOS/STM32F30x/pios_ws2811.c
```diff
@@ -0,0 +1,226 @@
...
+		return;
+	}
+
+	if (led >= dev->max_leds) {
+		return;
+	}
+
+	int offset = led * WS2811_BITS_PER_LED;
+
+	uint32_t grb = (g << 16) | (r << 8) | (b);
```
tracernz: Redundant parentheses.

flight/PiOS/STM32F30x/pios_ws2811.c
```diff
@@ -0,0 +1,226 @@
...
+		return;
+	}
+
+	if (dev->dma_active) {
+		if (!DMA_GetFlagStatus(dev->config->dma_tcif)) {
+			return;
+		}
+
+		dev->dma_active = false;	/* LOL */
+		DMA_ClearFlag(dev->config->dma_tcif);
```
tracernz, line 209: Doesn't the if above stop you getting here if that flag is set? This whole thing seems wrong. :/

flight/PiOS/STM32F30x/pios_ws2811.c
mlyle: Yah, but the reason I sped it up was the loading on F3E, figuring it was harmless for the rest of targets.  Seems OK to have the comment here (else someone might test a lower speed on e.g. sparky1 and like it more, and think this could change).

flight/PiOS/STM32F30x/pios_ws2811.c
mlyle: No.  We return before if the flag is clear.

Basically, there used to be an interrupt handler in this code-- that would, when transmit complete was set, clear the TC flag and set dma_active to false.

Now instead we lazily do that work when someone wants to update the LED-- if we think we're in a transaction, we check the hardware-- if it says it's done, we clear that flag and mark ourselves in our internal state as not being so.  (And then immediately set it again heh).

flight/PiOS/STM32F30x/pios_ws2811.c
mlyle: I like paranoid extra parens, else every time I go through this code when it's not working I'm going to squint at this line and then double check my order of operations chart.

mlyle:
Will tidy the rest of these later

mlyle:
Fixed up per feedback

mlyle:
IMO, this needs, before merge:

* At least one target tested.
* Colibri Race / Lux support (I have a rig I can easily test for this!)

flight/PiOS/STM32F4xx/inc/pios_ws2811_cfg.h
```diff
@@ -5,7 +5,7 @@
  * @addtogroup PIOS_WS2811 WS2811 Functions
  * @{
  *
- * @file       pios_ws2811_target.h
+ * @file       flight/PiOS/STM32F4xx/pios_ws2811_target.h
```
tracernz: inc/ :wink:

flight/PiOS/STM32F30x/pios_ws2811.c
tracernz: Heh, I was meaning the variable by itself. :wink:

flight/PiOS/STM32F30x/pios_ws2811.c
tracernz: > (And then immediately set it again heh)

That's the main reason it doesn't really seem right. :stuck_out_tongue: It seems like the extra layer of state tracking is redundant and you could just:
```c
void PIOS_WS2811_trigger_update(ws2811_dev_t dev)
{
	if (!dev) {
		return;
	}

	if (DMA_GetCmdStatus(dev->config->dma_chan) == ENABLE
			&& !DMA_GetFlagStatus(dev->config->dma_tcif)) {
		return;
	}

	DMA_ClearFlag(dev->config->dma_tcif);
	DMA_Cmd(dev->config->dma_chan, DISABLE);

	DMA_SetCurrDataCounter(dev->config->dma_chan, dev->buffer_size);
	DMA_Cmd(dev->config->dma_chan, ENABLE);
}
```
or am I missing something (didn't read the datasheet :stuck_out_tongue: )?

-e- Ah, F3 stdperiph doesn't have a function to check the cmd status. :/

mlyle: Yah, basically that's it-- I need to know whether it's cued or not.  And it's this way in case it gets factored out and used at another time.

tracernz (approved):

mlyle:
LEDs blink properly on TBS Vendetta tail light (colibri race/Lux).

tracernz:
Tested on LUX, and tested the PiOS driver on Sparky1 (although that target is not supported by this PR, it needs more work because it would require a change to the output channel configuration/stealing output 8).

PR #1847, opened 2017-06-24
Revert "GCS:UAVSettingsImportExport: Disable commands when not connected"

### LINKED ISSUE #1846: GCS: Doesn't load properly on Windows
Something must have happened recently, that affects the Windows builds. Older builds of mine start up fine, but I compiled a next build from scratch and it complained about missing dependencies all over the place. Downloaded a build from Jenkins, same story.

drama

mlyle:
Reverts d-ronin/dRonin#1840

Fixes #1846 

This had broken Windows-- circular dep between modules?  DLLs don't load


tracernz:
lame -_-

I wonder if there's a dep that isn't in plugins.pro because I checked that to see what depends on what, and thus chose the approach I used here to avoid circular deps (and added a new dep to plugins.pro).

PR #1849, opened 2017-06-24
Change Revo's DShot back to GPIO

glowtape:
On Revo, you can configure WS2811 to output to pins PA0, PB12 and PD2.

When WS2811 is enabled and set to pin PD2, DMAShot works just fine on the Revo. When however pins PA0 or PB12 are configured, the DMA stream for servo pin 3, which shares a DMA channel with WS2811, appears to get interrupted if their DMA requests collide. The DMA update function waits for all previous DMAShot updates to finish by checking the transfer complete flag (IMO unnecessary), and when the DMA request gets stopped, it'll never trigger, thus the board crashes in the watchdog because it's stuck in that wait. See this:

drama2

The initial naive fix was to ditch the TC flag wait, which kept the board working. Except above still happens, at least until it gets reset by the next update from DMAShot. If you could rely on ESCs noticing that there's too much data incoming, this would be a kludgy fix. Practically, this will likely not be the case, and thanks to XOR checksums, the ESC will likely parse it as disarm when the DMA got interrupted sending an 0-bit, or go full throttle if it happened on an 1-bit.

I'm at a loss as to why this happens. The documentation says not to run two DMA streams on the same DMA channel, but why it only causes drama for two third of the WS2811 pins, I'm not sure.

There aren't any relevant CPU savings between DMAShot and GPIO shot, anyway. Former just allows for IRQs to run, e.g. to prevent Seppuku video from wobbling, otherwise there's no real advantage (other than mixed DShot rates).

So disabling DMAShot is the best option. GPIO works just fine on the Revo.

### DIFF
--- a/flight/targets/revolution/board-info/board_hw_defs.c
+++ b/flight/targets/revolution/board-info/board_hw_defs.c
@@ -1432,12 +1432,12 @@ static const struct pios_dmashot_timer_cfg dmashot_tim_cfg[] = {
 	},
 	{
 		.timer = TIM9,
-		.master_timer = TIM1,
+		.master_timer = TIM4,
 		.master_config = TIM_DMA_Update | TIM_DMABase_CCR2,
 
-		.stream = DMA2_Stream5,
-		.channel = DMA_Channel_6,
-		.tcif = DMA_FLAG_TCIF5,
+		.stream = DMA1_Stream6,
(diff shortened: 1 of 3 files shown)


mlyle:
Probable underlying cause-- conflict for TIM1.

PR #1856, opened 2017-06-28
GCS:Attitude: Disable temperature calibration accept until done

### LINKED ISSUE #1854: Temperature compensation without a sufficient temperature swing can make NaNs
* Seen in version (release version or githash): Release-20170213.1
* On flight controller (if applicable): BrainFPV RE1

Issue details:

After performing Temperature Compensation, it is no longer possible to perform an Automatic Level Detection. Looking at the Sensor Settings on the browser on the RHS, a few of the values are NANs. After these NANs are changed to zero, Automatic Level Detection works again.

tracernz:
Must now wait for the minimum temperature change to occur before calibration can be accepted.

Half of #1854.


mlyle:
Nice, fast :D  Really would rather not close 1854 though, because of @nanodude 's comment and the need to look at tempcal in general.

(I have made the GCS code work but I also have experienced oddness)

tracernz:
Yeah, I've seen the firmware crash several times testing this code too.

mlyle:
Thx :D

PR #1861, opened 2017-06-30
GCS: Move plugin errors into detailed text

tracernz:
Adding them into the message box main text creates a message box much taller than my display (at least on Windows), so I can't see all of it. The detailed text gets a scroll view so it should always be okay.


tracernz:

Need this to see what's up with next.

tracernz:
What it looks like on Linux:
image

And before:
image

PR #1863, opened 2017-06-30
pios: move from uint32 handles

mlyle:
1. Ensure compile-time type checking in the future so fewer bugs sneak
up on us.
2. The handle was uint32_t, so make at least this one uint64_t safe
(there's quite a few other that are not uintptr so this is just the
beginning, unfortunately...)
3. Remove BMP085, which was unused and not 64 bit safe.
4. Remove F1 I2C drivers, which are not safe to new handle and are unused.
5. For now, disable ETASV3 sensor.  It was only used in a few targets and we don't have a safe mechanism to provide its handle to the airspeed subsystem.

Moves I2C, SPI, ADC, RFM22, COM_MSG


mlyle:
Rebased

flight/PiOS/Common/pios_mpu.c
```diff
@@ -98,7 +98,8 @@ struct pios_mpu_dev {
 	const struct pios_mpu_cfg *cfg;             /**< Device configuration structure */
 	enum pios_mpu_type mpu_type;                /**< The IMU device type */
 	enum pios_mpu_com_driver com_driver_type;   /**< Communication driver type */
-	uint32_t com_driver_id;                     /**< Handle to the communication driver */
+	pios_spi_t spi_driver_id;
+	uint32_t i2c_driver_id;                     /**< Handle to the communication driver */
 	uint32_t com_slave_addr;                    /**< The slave address (I2C) or number (SPI) */
```
tracernz: Hmm, this one is a bit uglier now. :(

flight/PiOS/Common/pios_mpu.c
```diff
@@ -216,8 +217,6 @@ static int32_t PIOS_MPU_Validate(struct pios_mpu_dev *dev)
 		return -1;
 	if (dev->magic != PIOS_MPU_DEV_MAGIC)
 		return -2;
-	if (dev->com_driver_id == 0)
-		return -3;
 	return 0;
```
tracernz, line 220: ?

tracernz (approved):

flight/PiOS/Common/pios_mpu.c
mlyle: I dislike code that checks for things that should never happen (something being null that is not allowed to be null / would have accessed invalid memory during initialization if it was null, etc) and tries to salvage the situation when we're already deeply into undefined behavior.

tracernz: Fair enough. The only thing it can really protect is if someone steps over the dev struct with 0s, we'd not try to access it, but that's very much an edge-case.

PR #1865, opened 2017-06-30
posix: remove pios_udp

mlyle:
1. It's unused
2. It has 32 bit IDs and... the way it frames things if of limited
utility.


tracernz:
Need to remove it from GCS including default config then too IMO.

mlyle:
Let me look at the GCS half to see if it's as dumb as this was.

PR #1867, opened 2017-07-01
GCS:UAVSettingsImportExport: Hax to avoid circular dll dep on Windows

tracernz:
Just disable functionality for Windows users, while everyone else can still enjoy it.


tracernz:

PR #1873, opened 2017-07-02
GCS:Utils: Implement minimumSizeHint for LongLongSpinBox

tracernz:
Otherwise widgets with fixed/minimum size policies end up too short. If only Qt made QAbstractSpinBox able to take care of things like this without private headers.. Also made a small performance improvement by caching the text interpretation result (as the upstream QSpinBox impl. does).

This is useful for default hw widget on macOS.


mlyle:
Code looks good, as does linux run of this.

PR #1877, opened 2017-07-05
Autotune: Send compressed raw tune data to cloud

### LINKED ISSUE #1876: GCS/Autotown: Record tune sample data in Autotown
Would be nice to grab at very least the raw tune data (a few kb binary blob) and stick it in autotown as a blob (cloud storage, linked to the datastore entries we already do?). @dustin

mlyle:
Also display per-axis taus properly.

Fixes #1876


mlyle:
Untested.

dustin (approved):
Seems fine.

Can we do an isolated test of what this might look in a bad case just to make sure there's not a column limit hidden somewhere?

I assume you're just using the normal API to extract it and don't need any other special support.

tracernz (approved):
Looks ok.

mlyle:
http://dronin-autotown.appspot.com/api/tune?tune=ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAsPibygkM is a tune shared with this

mlyle:
Added a couple of python commits --- including the jupyter notebook we've been passing around -- successfully pulls/processes the autotune directly from the tune API.

https://nbviewer.jupyter.org/github/mlyle/dronin/blob/mpl-tunedatasend/python/nb_autotune.ipynb

mlyle:
This is the last remaining thing to merge before cutting a preview build.

PR #1878, opened 2017-07-09
brainre1: add MSP bridge

mlyle:


mlyle:
@mluessi --- it would be really nice if you could come support your target.  Passthrough's been working on everything else for a couple of months, but RE1 fell through the cracks because A) you'd omitted the MSP bridge originally as having not made sense, and B) no one's verifying that stuff is OK etc.

If this kind of thing continues to happen with new features on RE1, it raises the question as to whether the target should get dropped.

mlyle:
Accessed the ESCs on my RE1 rig, found they had mismatched versions, flashed them.  Wonder if it will fly better :p

PR #1884, opened 2017-07-12
actuator: clamp throttle when unarmed

mlyle:
Makes sure that you can't throttle up on a "servo" output when disarmed.

**longstanding** bug but it is much better if fixed.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -489,6 +489,9 @@ static void normalize_input_data(uint32_t this_systime,
 			&(*desired_vect)[MIXERSETTINGS_MIXER1VECTOR_ACCESSORY0]);
 	}
 
+	*armed = flightStatus.Armed == FLIGHTSTATUS_ARMED_ARMED;
+	*spin_while_armed = actuatorSettings.MotorsSpinWhileArmed == ACTUATORSETTINGS_MOTORSSPINWHILEARMED_TRUE;
+
 	if (airframe_type == SYSTEMSETTINGS_AIRFRAMETYPE_HELICP) {
 		// Helis set throttle from manual control's throttle value,
 		// unless in failsafe.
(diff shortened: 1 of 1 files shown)


flight/Modules/Actuator/actuator.c
```diff
@@ -497,13 +500,14 @@ static void normalize_input_data(uint32_t this_systime,
 		}
 	} else {
 		throttle_val = desired.Thrust;
+
+		if (!*armed) {
+			throttle_val = -1;
+		}
```
tracernz: You know, I actually prefer a ternary here/think it is clearer, i.e.:
```c
throttle_val = *armed ? desired.Thrust : -1;
```

flight/Modules/Actuator/actuator.c
```diff
@@ -497,13 +500,14 @@ static void normalize_input_data(uint32_t this_systime,
...
 	}
 
 	static uint32_t last_pos_throttle_time = 0;
 
-	*armed = flightStatus.Armed == FLIGHTSTATUS_ARMED_ARMED;
-	*spin_while_armed = actuatorSettings.MotorsSpinWhileArmed == ACTUATORSETTINGS_MOTORSSPINWHILEARMED_TRUE;
```
tracernz: But actually, it could go here and cover helis too?

tracernz (commented):
P.S. It's also very weird/ugly API the way armed and spin_while_armed work. I guess it's worth it to save passing 1 extra arg (flight mode) or fetching a UAVO twice?

flight/Modules/Actuator/actuator.c
mlyle: Moved it down so it would hit helis.

mlyle:
Yes, the API is strange-- i don't mind the armed bit, but the spin_while_armed is wonky.  A lot of that was existing structure to help control risk a bit during the refactor-- spin while armed is really a configuration thing that belongs in a global and belongs getting checked on conf updates.

tracernz (approved):

mlyle:
jenkins, test this please

Issue #1875, opened 2017-07-04
ChibiOS: Invalid MSP from bootloader causes main stack to overlap process stack


tracernz:
Follow-up from https://github.com/d-ronin/dRonin/pull/1874

Some bootloaders, due to unintended code generation, end up incrementing (positive) the main stack pointer by a number of bytes (8 for LibrePilot v6, 16 for dRonin recent releases). This causes the first few entries in the main stack to step on whatever happened to follow it in memory. In the bootloader updater this happened to be the flash partition table which caused the updater to fail. In firmware, this is the process stack (thread stacks): https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/STM32F4xx/sections_chibios.ld#L127. Fortunately this is at the end of the stack of whatever thread happens to be first, so unless it completely fills it's stack it's probably only a theoretical issue (and filling stacks with the pattern used to detect stack usage happens in ChibiOS's `ResetHandler`, before any stack is ever used).

Suggested fix is either:
- Set MSP to correct location in `__early_init` hook.. this is a bit fragile if upstream ChibiOS `ResetHandler` changes, and happens to use the main stack before calling `__early_init`, so not the preferred option.
- Change `ENTRY` function in linker script to our own function which just sets MSP, then jumps to ChibiOS `ResetHandler`. This is the safest and most preferred option I think. -e- Duh, do still need to edit ChibiOS sources to place our reset handler in the vector table. -_-

@mlyle, thoughts?

mlyle:
Does that initial stack survive to later run, or is it just used to kickstart chibios and get the init task running?

(Does chibios just blindly reuse the initial stack as an irq stack?)

tracernz:
That is the IRQ stack:
https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/Common/Libraries/ChibiOS/os/ports/GCC/ARMCMx/crt0.c#L280-L359

tracernz:
Did a little digging. The main thread (i.e. main), uses the "process stack" (switches to it at https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/Common/Libraries/ChibiOS/os/ports/GCC/ARMCMx/crt0.c#L308), while interrupts use the "main stack". So the OS is running in the process stack... I guess it never gets too close to full.

[...]

PR #1886, opened 2017-07-13
Flight: Implement Stack Unused Checks

tracernz:
Chibi uses the following types of stacks:
- main stack, used by interrupts
- process stack, used by main thread from very early init
- per-thread stacks, used by normal threads

This adds missing code to measure the first, and a new measurement for the second. Shows up #1875 when using an old bootloader.

### DIFF
--- a/flight/PiOS/STM32F0xx/pios_sys.c
+++ b/flight/PiOS/STM32F0xx/pios_sys.c
@@ -191,6 +191,11 @@ int32_t PIOS_SYS_SerialNumberGet(char *str)
 	return 0;
 }
 
+size_t PIOS_SYS_IrqStackUnused(void)
+{
+	return 0;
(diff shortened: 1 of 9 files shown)


flight/PiOS/STM32F0xx/pios_sys.c
```diff
@@ -191,6 +191,16 @@ int32_t PIOS_SYS_SerialNumberGet(char *str)
 	return 0;
 }
 
+size_t PIOS_SYS_IrqStackUnused(void)
```
tracernz, line 194: Doesn't have any code to fill the stack with a pattern... and looking at startup.c it uses different symbols for irq/msp stack for some reason.

mlyle:
jenkins, test this please

PR #1888, opened 2017-07-14
GCS/Flight: Fix build failures on recent compilers

tracernz:
This branch builds with GCC 7.1.

Note: some of these were genuine bugs while a few were merely theoretical.


tracernz:
jenkins, test this please

PR #1889, opened 2017-07-14
GCS:Libs: Roll Eigen to 3.3.4

tracernz:
Numerous fixes since we last rolled. Didn't go into detail to see how much they affect us. Unfortunately still didn't fix the issues mitigated in #1888.


tracernz:
jenkins, test this please

PR #1891, opened 2017-07-14
OSD: Use GPS ground speed scalar rather than calculating from vectors

tracernz:
UBLOX provides ground speed in both polar (2D) and cartesian (3D) co-ordinates. OSD was calculating the magnitude from the cartesian co-ordinates rather than using the scalar magnitude from the polar co-ordinates. NMEA only provides velocity in polar form and our impl. doesn't do the conversion to cartesian for GPSVelocity UAVO.

Summary: saved some math and made it work for both UBLOX and NMEA.

### DIFF
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -1380,11 +1380,11 @@ void render_user_page(OnScreenDisplayPageSettingsData * page)
 				sprintf(tmp_str, "%s", "GND");
 				break;
 			case ONSCREENDISPLAYPAGESETTINGS_SPEEDSCALESOURCE_GPS:
-				if (GPSVelocityHandle()) {
-					GPSVelocityNorthGet(&tmp);
-					GPSVelocityEastGet(&tmp1);
-					tmp = sqrt(tmp * tmp + tmp1 * tmp1);
-					speed_valid = has_gps;
+				if (has_gps && GPSPositionHandle()) {
+					GPSPositionGroundspeedGet(&tmp);
+					uint8_t fix;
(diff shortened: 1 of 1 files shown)


tracernz:

tracernz:
jenkins, test this please

tracernz:
0xnone but nothing merged recently...

jenkins, test this please

mlyle:
Yah I wish I better understood the 0xnone thing.

There's the initial (before build) failed to get a reference thing.  That's pretty clearly tree skew.  And then 0xnone, which **seems** to be somewhat tied to force pushes and merges.

tracernz:
Yeah, except this branch didn't even exist when the last merge occured prior to the failure, and has never been force-pushed... Do we even care about ancestor hash on CI builds? Should only matter for artifact builds.

mlyle:
>  Do we even care about ancestor hash on CI builds? 

No, but we care about ancestor hash on builds that are built like CI builds.  Primary use case is me doing a quick parallel build, without packaging, and installing on a FC... and then later upgrading from that point. :/

mlyle:
Have you confirmed this actually works?

tracernz:
Nope, just checked the UBLOX and NMEA protocol specs. On the other hand it currently __doesn't work__ for the majority of GPSs (UBLOX is pretty niche, NMEA very universal).

mlyle:
;) Except we basically require ublox in practice

tracernz:
I do have a branch improving NMEA support (baud-rate auto-config etc.). Needs more work on other areas before I open a PR though.

mlyle (approved):
LGTM, but ideally I'd like someone to see it work before merge.

mlyle:
Tested OK ublox

PR #1893, opened 2017-07-14
Build system improvements (ccache, qbs, clazy)

### LINKED ISSUE #1880: Tools: Setup our own qbs config/profiles
Rather than running `qbs setup-toolchains --detect` and messing with the global system config.

tracernz:
# ccache
Makes ccache a CONFIG option. This means it can be used for release builds as well, which is very handy when working on packaging etc., and it can be disabled for debug builds, even more handy to avoid polluting your cache when testing alternate compilers etc. The default is still debug with ccache. The cases are:
- make gcs
- GCS_BUILD_CONF="debug ccache" make gcs 
    * debug with ccache
- GCS_BUILD_CONF=debug make gcs
    * debug without ccache
- GCS_BUILD_CONF=release make gcs
    * release without ccache
- GCS_BUILD_CONF="release ccache" make gcs
    * release with ccache

i.e. behaves almost exactly the same as before, and with some new options you can build release with ccache and debug without ccache.

# nits
- `which clazy` was emitting warnings on stderr when not found on some platforms (Fedora and Windows at least)
- Use our own qbs settings dir rather than altering the global one

Fixes #1880

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -319,7 +319,7 @@ gcs_ts: tools_required_qt
 gcs_clazy: CLAZY_CHECKS ?= level0
 gcs_clazy: $(UAVOBJECT_MARKER) | tools_required_qt
 	echo $(CLAZY)
-ifeq ($(shell which clazy),)
+ifeq ($(shell which clazy 2>/dev/null),)
 	$(error Please install clazy and ensure it is on PATH first. https://github.com/KDE/clazy#build-instructions)
 endif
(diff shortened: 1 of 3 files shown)


tracernz:
jenkins, test this please

mlyle:
jenkins, test this please

mlyle:

mlyle:

PR #1895, opened 2017-07-17
python-api: support sending multi-instance objects

mlyle:
Seems to work.

### DIFF
--- a/python/dronin/uavtalk.py
+++ b/python/dronin/uavtalk.py
@@ -179,10 +179,16 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=None,
             # XXX check length vs pack_len
             obj = uavo_defs[uavo_key]
 
+        instance_len = 0
+
         # Determine data length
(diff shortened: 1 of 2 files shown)


python/dronin/uavtalk.py
```diff
@@ -179,10 +179,16 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=None,
...
             obj = uavo_defs[uavo_key]
 
+        instance_len = 0
+
         # Determine data length
         if (pack_type == TYPE_OBJ_REQ) or (pack_type == TYPE_ACK) or (pack_type == TYPE_NACK):
             obj_len = 0
             timestamp_len = 0
+
+            if obj is not None and not obj._single:
```
tracernz, line 189: So we can't actually send a multi-instance data object, only req/ack/nak?

mlyle: No, sending works fine.  It just moves the instance field into the object itself for normal send/receive.  In the case that it's a request we still need to consume a separate instance id.

mlyle: (tested/confirmed to send/receive waypoint objects)

PR #1899, opened 2017-07-19
Package:Linux: Fix udev rules not included in tar

tracernz:
*sigh*

(untested)

### DIFF
--- a/package/Makefile.linux
+++ b/package/Makefile.linux
@@ -62,7 +62,7 @@ gcs:
 	$(V1) $(MAKE) -C $(ROOT_DIR) GCS_BUILD_CONF=release $@
 
 ground_package_os_specific: ground_package_common usb_id_udev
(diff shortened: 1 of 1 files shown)


tracernz:

mlyle:
On road.  Yuck.  For now can we put it somewhere and kludge relnotes?  Seems better than hotfixing to include one file.

mlyle (approved):
Untested, looks sane

tracernz:
Tested this and it works correctly.

> On road. Yuck. For now can we put it somewhere and kludge relnotes? Seems better than hotfixing to include one file.

Added as a release asset, and amended relnotes as appropriate.

mlyle:
Thx, I think that's the best way (avoid the risk of cutting new artifacts).

tracernz:
Tagged this as hotfix so it will show up for inclusion if we end up doing one later.

PR #1900, opened 2017-07-21
Annunciator: Add manual buzzer option, activated by accessory chan

tracernz:
Have seen a lot of requests for this.

### DIFF
--- a/flight/Modules/System/systemmod.c
+++ b/flight/Modules/System/systemmod.c
@@ -45,6 +45,7 @@
 
 #include "annunciatorsettings.h"
 #include "flightstatus.h"
+#include "manualcontrolcommand.h"
 #include "manualcontrolsettings.h"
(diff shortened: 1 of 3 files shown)


tracernz:
Tested on Seppuku. :+1:

flight/Modules/System/systemmod.c
```diff
@@ -1095,6 +1107,9 @@ static uint32_t processPeriodicUpdates()
 	return timeToNextUpdate - now;
 }
 
+DONT_BUILD_IF(ANNUNCIATORSETTINGS_MANUALBUZZER_MAXOPTVAL >
+	MANUALCONTROLCOMMAND_ACCESSORY_NUMELEM, TooManyManualBuzzers);
```
tracernz, line 1111: Arguably want to check not equal, to make sure we support all the acc chans in future?

PR #1894, opened 2017-07-17
manualcontrol: improvements

mlyle:
* Significantly improve safety and simplicity of arming mechanism (fixes #1236  fixes #943 )
* Eliminates low-throttle disarms during autonomy entirely.
* Eliminates non-delay switch arming modes.
* Eliminates posix-specific GCS receiver.
* Creates a posix "RTC" implementation so that the common impl of receiver protocols can run on posix.
* Fixes a bug so that the main GCS receiver has failsafe work

PR #1901, opened 2017-07-22
ManualControl: Prevent arming, if TX is sending arming signal during craft power-up.

glowtape:
When the transmitter is sending an arming signal, e.g. if you have switch arming enabled and the switch is flipped on the TX, while you're powering up your quad, it immediately goes to arm. Not so much a problem, if +Throttle is also enabled, because idle motor spin won't cut your fingers, but without +Throttle, there's a possibility for drama.

This patch makes sure that the craft has been in a disarmed state, before allowing arming.

Tested on my Seppuku quad with a Crossfire system. If I power the quad up, while the arming switch is flipped on the transmitter, I need to disarm and rearm for the motors to spin. Otherwise, if the transmitter was disarmed (powering up the transmitter before or after quad power-up), everything works as before, quad arms fine.

Still gonna have to dig out one of my old FrSky receivers to test, whether it immediately sends transmitter commands, or does other silly stuff during boot.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -106,6 +106,7 @@ enum arm_state {
 	ARM_STATE_DISARMED_STILL_HOLDING
 };
 static enum arm_state arm_state = ARM_STATE_DISARMED;
+static bool disarmed_on_powerup = false;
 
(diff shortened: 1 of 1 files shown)


glowtape:
Nevermind, collides with #1894.

PR #1902, opened 2017-07-23
smoothcontrol: Fix bad bounds checking on integrator timeout.

glowtape:
I have no excuse. :(

--edit: That said, there's no UI for it and it defaults to off, so it shouldn't affect existing crafts.

### DIFF
--- a/flight/Libraries/math/smoothcontrol.c
+++ b/flight/Libraries/math/smoothcontrol.c
@@ -60,7 +60,7 @@ static void smoothcontrol_update(smoothcontrol_state state, struct smoothcontrol
 		case SMOOTHCONTROL_EXTENDED:
 			axis->differential = signal_diff * SMOOTHCONTROL_PREDICTOR_SLOPE / (float)state->control_interval;
 			axis->current = axis->signal + signal_diff * SMOOTHCONTROL_CHAMFER_START;
-			axis->integrator_timeout = (uint8_t)MAX((float)state->control_interval *
+			axis->integrator_timeout = (uint8_t)MIN((float)state->control_interval *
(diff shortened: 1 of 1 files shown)


tracernz (approved):
Heh, so these were pegged at 255 unless you configured them bigger, then they overflowed. :D

glowtape:
Yeah, that caused funny behavior, because the calculated differential value depends on the amount of cycles between ManualControl updates. On Crossfire with a 1600hz loop, that's usually 10 cycles. But when the differential is calculated on 255, the prediction is practically disabled since the differential is just a 25th of what it should be.

In terms of the chamfered steps prediction, it was then essentially just trailing the actual control signal by a ManualControl update cycle. Quasi the same thing happened to the linear interpolation (--edit: actually, I'm not so sure it worked right either, was months ago). So I didn't really notice it, because the quad handled mostly normally. The only odd thing was interpolating between predicts, which went haywire and kind of acted like a huge lowpass on control, the quad flew like it was slow and drunk. Couldn't figure it out last time, so I dropped the whole interpolation stuff. I went about reimplementing interpolation today, when the same thing happened, when I finally noticed it trying to resolve that issue.

glowtape:
So things are working properly now, I guess. I forced the receiver into 50hz mode and sampled in the GCS at 500hz/2ms, that's what RateDesired looks like:

Smoothing to none/default:
rcs_none

Smoothing to normal:
rcs_normal

Smoothing to extended:
rcs_extended

I've also reimplemented both the linear interpolation of regular control values and predicted control values, on another branch. Still need to do some UI for it. Normal mode looks like a safe bet for making it default eventually (no overshooty bits that jump out).

Still need to fly it, waiting for a dry spell.

PR #1904, opened 2017-07-26
flight,gcs: New warning for working RX but botched channel configuration.

glowtape:
Spurred by OhmEyes configuration issue on IRC.

If none of the relevant channels are in error (RPYT, Flight mode and arming), it'll show a different error message than No RX.

gcserror


tracernz (approved):
Looks good. :+1:

PR #1905, opened 2017-07-26
win32: don't use invalid compiler flag

mlyle:


mlyle:

tracernz:
:blush:

ground/gcs/src/libs/eigen/eigen.pri
```diff
@@ -1,4 +1,4 @@
 INCLUDEPATH *= $$PWD
 
 # work around Eigen's abuse of enums to allow compiling with recent compilers
-QMAKE_CXXFLAGS += -Wno-int-in-bool-context
+!win32-msvc*:QMAKE_CXXFLAGS += -Wno-int-in-bool-context
```
tracernz, line 4: The correct check is probably actually gcc or clang but this will do. :wink:

PR #1906, opened 2017-07-26
flight: pid refactor: move dt to configuration time

mlyle:
No longer allow time to vary from timestep to timestep.  This doesn't do
much right now, but it's for two upcoming reasons-- to allow
precomputing the derivative filter, and to allow precomputing a leaky
integral.


mlyle:
In support of #1747

mlyle:
(also one other note-- it seems some of the stabilization rate loops were using dT instead of dT_expected, so it's a minor bug fix for those cses).

PR #1908, opened 2017-08-02
brainre1: Enable mag calibration on RE1, when one's configured.

### LINKED ISSUE #1896: BrainRE1 calibrate mag doesn't work
This is the only nav-capable target that doesn't have an onboard mag, so BOARD_HAS_MAGNETOMETER is false.  Multiple users are disappointed that they can't calibrate the mag.

Any solution needs to properly deal with existing non-nav capable boards (e.g. Lux), and have a sane default for RE1 (e.g. don't just turn on calibrate magnetometer by default).

A forum thread: https://forum.dronin.org/forum/d/364-magnetometer-cal-box-grey-d-out/

@mluessi is this something you'd be willing to handle, since it's unique to your target?

glowtape:
Supposed to fix #1896. Low-hanging fruit. Should work, someone with a RE1 test this, mine is kaputt.

### DIFF
--- a/ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
+++ b/ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
@@ -77,6 +77,22 @@ bool BrainRE1::queryCapabilities(BoardCapabilities capability)
     case BOARD_CAPABILITIES_OSD:
     case BOARD_CAPABILITIES_UPGRADEABLE:
         return true;
+    case BOARD_CAPABILITIES_MAGS:
+        {
+            ExtensionSystem::PluginManager *pm = ExtensionSystem::PluginManager::instance();
+            UAVObjectManager *uavoManager = pm->getObject<UAVObjectManager>();
+            HwBrainRE1 *hwBrainRE1 = HwBrainRE1::GetInstance(uavoManager);
+
+            if (!hwBrainRE1) {
+                qWarning() << "Failed to get HwBrainRE1";
+                Q_ASSERT(false);
+                return false;
+            }
+
+            if (hwBrainRE1->getI2CExtMag() != HwBrainRE1::I2CEXTMAG_NONE)
+                return true;
+        }
+        Q_FALLTHROUGH();
     default:
         return false;
     }


ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
```diff
@@ -77,6 +77,22 @@ bool BrainRE1::queryCapabilities(BoardCapabilities capability)
...
+
+            if (!hwBrainRE1)
+                return false;
+
+            HwBrainRE1::DataFields settings = hwBrainRE1->getData();
+
+            if (settings.I2CExtMag != HwBrainRE1::I2CEXTMAG_NONE)
+                return true;
+        }
+        // Fall through.
```
tracernz: Please use Q_FALLTHROUGH or you _will_ break compilation on recent compilers (see #1888).

ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
```diff
@@ -77,6 +77,22 @@ bool BrainRE1::queryCapabilities(BoardCapabilities capability)
     case BOARD_CAPABILITIES_OSD:
     case BOARD_CAPABILITIES_UPGRADEABLE:
         return true;
+    case BOARD_CAPABILITIES_MAGS:
+        {
+            ExtensionSystem::PluginManager *pm = ExtensionSystem::PluginManager::instance();
+            UAVObjectManager *uavoManager = pm->getObject<UAVObjectManager>();
+            HwBrainRE1 *hwBrainRE1 = HwBrainRE1::GetInstance(uavoManager);
+            Q_ASSERT(hwBrainRE1);
```
tracernz: Would prefer:
```c
if (!hwBrainRE1) {
    qWarning() << "Failed to get HwBrainRE1";
    Q_ASSERT(false);
    return false;
}
```
The qWarning notifies the cloud service (via UsageStats via DebugGadget) so we can track problems users are encountering without requiring them to actively feedback. It also prints to the console; handy if you don't have a debugger connected when you hit the Q_ASSERT (in a debug build). Finally, the return will be hit in release builds (Q_ASSERT is empty statement), which you did already handle on the next line.

ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
```diff
@@ -77,6 +77,22 @@ bool BrainRE1::queryCapabilities(BoardCapabilities capability)
...
+
+            if (!hwBrainRE1)
+                return false;
+
+            HwBrainRE1::DataFields settings = hwBrainRE1->getData();
```
tracernz: You can just getField the one you want rather than everything. The upside of that method is less work later to move away from hard-coded/compiled UAVOs (assuming you also use the getObject method to get the object on the line above). Will accept either way so up to you.

ground/gcs/src/plugins/boards_brainfpv/brainre1.cpp
```diff
@@ -77,6 +77,22 @@ bool BrainRE1::queryCapabilities(BoardCapabilities capability)
...
+
+            if (settings.I2CExtMag != HwBrainRE1::I2CEXTMAG_NONE)
+                return true;
+        }
+        // Fall through.
     default:
```
tracernz: Existing code, but not really a fan of default case here, better to list them all explicitly so compilation breaks when a new capability is added, thus nothing gets missed (a very common problem!). Feel free to ignore this. :wink:

tracernz:
Not really sure why @mlyle mentions Lux etc. in the issue?

mlyle:
Other boards without onbuilt mag can have ext mag, no?

tracernz:
> Other boards without onbuilt mag can have ext mag, no?

Right, yes. I was thinking we were addressing an issue specific to RE1 sorry, but we should fix the general case indeed.

glowtape:
Would probably best to ungrey the checkbox, if at least one of of the mag axes in the UAVO isn't default zero then?

tracernz:
Hopefully all the boards have the same field name for external mag selection, and shared values from HwShared (I think this one is at least true)... Then we could add a function to IBoardType called externalMagEnabled or something that looks for ExtMag field in the HwUAVO and checks if it's enabled. Saves having to add code to every board.

mlyle:
Maybe this shouldn't be a board capability at all-- if mag data is present, enable it.

jimpriest:
I'd be willing to test this is someone can point me to some instruction on how to compile things?  Just got a Brain RE and ran into this issue myself.

tracernz:
build artifacts please

@jimpriest In about 45 mins-1 hour links to built packages will show up here.

mlyle:
I'm not sure testing this is useful because it's not the correct fix, right?

tracernz:
Will test mag calibration on RE1, regardless of how it's enabled?

mlyle:
Since it's an external mag I don't really expect it to be different from
any other ext mag targets.

Important thing is getting in a proper fix to enable this when applicable.

tracernz:
I'm OK with this fix in the absence of any better fix. Perfect is the enemy of good.

mlyle:
I'm pretty strongly opposed.  I hate target specific hacks, especially for
a problem which many targets have.

It is not much code to check for presence of the mag object and that it is
nonzero.

tracernz:
Tbh I'd prefer something something more positive than a heuristic like that, E.g. a new field MagPresent. We may as well close this though if it has no chance of being merged?

mlyle:
Yah I don't think this is the way.  E.g. uploader reads capabilities but
uavos aren't around yet.

Right now I think most of flight is conditional on object init.

mlyle:
Doesn't build currently either :/

glowtape:
Superseded by #2034.

PR #1915, opened 2017-08-12
fixedwingpathfollower: add position feedforward

mlyle:
When evaluating path segments and path error, feed forward the current
velocity 0.5 seconds forward into position.  This way, we begin turning
early, since a plane cannot instantly translate in the desired
direction.


mlyle:
Before:

before

After:

after

Sorry about different zoom levels :/

mlyle:
(We have similar code in vtolpathfollower where there's less of a turning delay and it's a win)

tracernz:
Yippee, another knob that needs wound. :stuck_out_tongue: Looks like it makes a nice improvement, and seems sane. :+1:

PR #1912, opened 2017-08-10
GCS: Cleanup UAVObjectWidgetUtils

tracernz:
I've had this sitting around for weeks so time to get it upstream before it rots.

- Use .ui files for UAVO relations wherever appropriate.
  - Cleanup a few things relating to these.
- Remove most of the addUAVObjectToWidgetRelation overloads
- Move some of the public API to private to prevent it making it's way back in
- Cleanup includes a little

PR #1916, opened 2017-08-13
GCS:Enable apply/save buttons on vehicle+output pages

tracernz:
Oops. :/ My fail in #1912.

### DIFF
--- a/ground/gcs/src/plugins/config/configoutputwidget.cpp
+++ b/ground/gcs/src/plugins/config/configoutputwidget.cpp
@@ -102,18 +102,6 @@ ConfigOutputWidget::ConfigOutputWidget(QWidget *parent)
     connect(m_config->calibrateESC, &QAbstractButton::clicked, this,
             &ConfigOutputWidget::startESCCalibration);
 
-    // Track the ActuatorSettings object
-    addUAVObject("ActuatorSettings");
-
-    // Associate the buttons with their UAVO fields
-    addWidget(m_config->cb_outputRate6);
-    addWidget(m_config->cb_outputRate5);
-    addWidget(m_config->cb_outputRate4);
-    addWidget(m_config->cb_outputRate3);
-    addWidget(m_config->cb_outputRate2);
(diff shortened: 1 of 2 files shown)


ground/gcs/src/plugins/config/configoutputwidget.cpp
```diff
@@ -102,18 +102,6 @@ ConfigOutputWidget::ConfigOutputWidget(QWidget *parent)
     connect(m_config->calibrateESC, &QAbstractButton::clicked, this,
             &ConfigOutputWidget::startESCCalibration);
 
-    // Track the ActuatorSettings object
-    addUAVObject("ActuatorSettings");
-
-    // Associate the buttons with their UAVO fields
-    addWidget(m_config->cb_outputRate6);
```
tracernz, line 109: These cause a segfault when ConfigTaskWidget adds the relations itself... ~~and don't actually do anything AFAICT~~ it does make us track their "dirty" state...

ground/gcs/src/plugins/config/configoutputwidget.cpp
```diff
@@ -102,18 +102,6 @@ ConfigOutputWidget::ConfigOutputWidget(QWidget *parent)
     connect(m_config->calibrateESC, &QAbstractButton::clicked, this,
             &ConfigOutputWidget::startESCCalibration);
 
-    // Track the ActuatorSettings object
-    addUAVObject("ActuatorSettings");
```
tracernz, line 106: Well, no point tracking an object unless you actually have a relation to it... which we do anyway.

tracernz:
Checked that the PWM rates and arming checkbox still work as expected.

tracernz:
Keeping some debt around in the interests of getting this blocker fixed. I missed that I still had some work to do in at least output page to really cleanup UAVO relations. :(

mlyle (approved):

tracernz:
jenkins, test this please

mlyle:
jenkins, test this please

PR #1920, opened 2017-08-18
GCS:UAVObjectField: Return QChar for bitfields

### LINKED ISSUE #1919: Metadata "modes" in uavo browser invalid
Something has regressed here.

tracernz:
quint8 == unsigned char, so QVariant::toString (used in UAVObjectBrowser TreeModel code) treats it as a char rather than an int. Give it an actual char so it can continue to use the same code for enums and bitfields.

Fixes #1919


mlyle:
Tests good, code looks good.

PR #1921, opened 2017-08-19
(presumably) uncontroversial changes from 1764.

mlyle:


flight/PiOS/posix/pios_delay.c
```diff
@@ -166,3 +166,21 @@ uint32_t PIOS_DELAY_DiffuS2(uint32_t raw, uint32_t later) {
...
+}
+
+/**
+ * @brief Calculate time in microseconds since a previous time
+ * @param[in] t previous time
+ * @return time in us since previous time t.
+ */
+uint32_t PIOS_DELAY_GetuSSince(uint32_t t)
+{
+	return PIOS_DELAY_GetuS() - t;
```
tracernz, line 185: Doesn't handle overflow.

mlyle: Isn't this the best you can do and still conform to the existing API?

tracernz: Ah, unsigned arithmetic does handle it correctly already.. carry on. 👍

tracernz (approved):

PR #1922, opened 2017-08-20
flight: use gcc6.3.1 tools

mlyle:
From gcc 5.2 before.

Needs analysis to look for any benefit, etc.  So far, firmware sizes seem similar (a few bytes saved on F3, a few bytes extra cost on F4).  Stack sizes seem similar-- a few tasks using 8 more bytes, a few tasks using 8 fewer bytes.

Most likely disappointing overall, but barring any sudden revelations picking up new tools every 18mos seems to make sense.


tracernz (approved):
Finds unused globals? And empty blocks?

flight/targets/seppuku/board-info/cmsis_system.c
```diff
@@ -404,9 +402,7 @@ static void SetSysClock(void)
     RCC->CFGR |= RCC_CFGR_SW_PLL;
 
     /* Wait till the main PLL is used as system clock source */
-    while ((RCC->CFGR & (uint32_t)RCC_CFGR_SWS ) != RCC_CFGR_SWS_PLL);
-    {
-    }
+    while ((RCC->CFGR & (uint32_t)RCC_CFGR_SWS ) != RCC_CFGR_SWS_PLL);;
```
tracernz: Empty statement at the end there. :wink:

mlyle:
> and empty blocks

Not quite.

It really doesn't like

```c
while (foo);
{
   [something]
}

or even

while (foo);
    something;
```

because the while is not guarding the block.

flight/targets/seppuku/board-info/cmsis_system.c
mlyle: Woops, rebased and forcepushed.

tracernz:
> Not quite.
> 
> It really doesn't like

Ah yes, that's quite a good catch actually.

tracernz:
Are you going to install it on Jenkins? I expect this should be a fairly safe change (after all, GCC 6 is kinda like GCC 5.1 under their old numbering scheme right?) but would be nice to get some CI builds to test.

mlyle:
Yes, slowly but surely-- going to need to smoke test it on some hardware first.

I always get suckered thinking that there's going to be a significant benefit, and there never is... empirically, compiler performance seems to be exponentially doubling -- on like a 15 year time constant.

mlyle:
Step 1-- smoke test / a lot of development myself with this stuff.  Seems OK.
Step 2-- Linux builder has tools installed, so testing base CI build.

jenkins, test this please.

mlyle:

mlyle:
Now across all 3 builders, and seems to work.  I'd consider this early/risky, but OK for merge (we're early in wired)

PR #1923, opened 2017-08-22
flight: doxygen: debrand subsystem titles

mlyle:
Basically, I feel awkward creating new files with old strings in them...


tracernz:
So we already have `Libraries` and `Modules` for this (which we have used for new modules/libraries for a while):
image

`@addtogroup Libraries Libraries`
rather than the long string because the long one is the one used in the menu tree and it's not nice for that.

mlyle:
Amended, force pushed.

tracernz:
Danke

PR #1925, opened 2017-08-26
Loadable extensions

mlyle:
Ready for preliminary review


mlyle:
@ernieift  --- this might interest you-- it supersedes PicoC in dRonin.

flight/module/Makefile
```diff
@@ -0,0 +1,29 @@
...
+ARM_SDK_PREFIX = $(ARM_SDK_DIR)/bin/arm-none-eabi-
+TCHAIN_PREFIX  = $(ARM_SDK_PREFIX)
+CC             = $(TCHAIN_PREFIX)gcc
+OBJCOPY        = $(TCHAIN_PREFIX)objcopy
+
+SRC=modtest.c
+
+MCU = cortex-m4
+
+FLIGHTPATH = /home/mlyle/dRonin/flight
```
peabody124: shouldn't be hardcoded

flight/module/Makefile
```diff
@@ -0,0 +1,29 @@
+TOOLS_DIR      = /home/mlyle/dRonin/tools
```
peabody124: shouldn't be hardcoded

mlyle: Thx for review  :)

Yah-- The build system is completely wrong (in description as WIP and the individual commit comments as "DONTCOMMIT")  There's a lot of little things to figure out about how to find includes.

The modules stuff has to reside out of tree, and ... shouldn't depend on so many include dirs, etc.  It should also be considerably less complicated than our overall build system.  So there's work to be done on how the build system works for them.

mlyle:
Overall I think the plan here is to let this stall awhile, and get a few commits like #1927 through that clean up the build system so that the module builds can ultimately be cleaner / nicer.

mlyle:
jenkins, test this please

mlyle:
Rebased.  Need to work on build system, and export a few more calls so it's minimally useful.

mlyle:
@tracernz -- can you please review -- this is an acceptable / useful state.

tracernz (approved):
I have read the early commits several times, and the most recent (fairly trivial) commits once. Looks fine to me.

PR #1926, opened 2017-08-26
androidgcs: support ft231xx uarts in device filter

mlyle:
Per forum discussions.  Needs test / validation.

### DIFF
--- a/androidgcs/AndroidManifest.xml
+++ b/androidgcs/AndroidManifest.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-	package="org.dronin.androidgcs" android:versionCode="20170717"
-	android:versionName="Neat">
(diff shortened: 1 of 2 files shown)


androidgcs/AndroidManifest.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-	package="org.dronin.androidgcs" android:versionCode="20170717"
-	android:versionName="Neat">
+	package="org.dronin.androidgcs" android:versionCode="20170826"
+	android:versionName="Neat.1">
```
tracernz: Should decouple this from the main versions because a real Neat.1 might exist later, perhaps -1 suffix instead (which is commonly used for small repackaging type things).

mlyle: good point-- done.

mlyle:
(rebase and pushed)

mlyle:
Might as well merge IMO

PR #1927, opened 2017-08-27
flight: big code refactor

mlyle:
1. slim down openpilot.h -- getting closer to eliminate this.
2. factor some startup code to chibi_main.c from individual targets,
as part of ongoing work to simplify the target files.
3. include taskmonitor.h from pios_thread, since they're pretty much
always used together.
4. adjust orderings in taskmonitor.h/pios.thread so that idempotency
works as intended here.
5. include pios_thread.h several additional places.


mlyle:
(requires some test too)

flight/PiOS/STM32/chibi_main.c
```diff
@@ -90,13 +92,62 @@ int main()
...
 }
+
+#if defined(STM32F4XX)
+
+#include <stm32f4xx_flash.h>
+/**
+ * Check the brown out reset threshold is 2.7 volts and if not
+ * resets it.
+ */
+void check_bor()
```
tracernz, line 103: This really has nothing to do with ChibiOS. :/

tracernz (commented):
Good cleanup. 👍

flight/PiOS/STM32/chibi_main.c
mlyle: Neither does anything here, really-- it's the main() function and init task for stuff running under chibios.

tracernz: 3 lines in main, and __early_init are the only chibi specific bits. P.S. https://github.com/mlyle/dronin/blob/9740da6184dde413a7e72c8dcedbe4fd63180edd/flight/PiOS/STM32/chibi_main.c#L6 is incorrect. 😉

PR #1928, opened 2017-08-28
pipxtreme: cope with slow clients

mlyle:
radiocombridge: use time-blocking-variants of calls

Technically this isn't sufficient, because we could be unlucky and have very slow throughput that is sufficient to prevent this from tripping but still hits the watchdog.  In practice it's fine, as the principal issue (USB never draining at all while it is present) is fixed.

### DIFF
--- a/flight/PiOS/inc/pios_com.h
+++ b/flight/PiOS/inc/pios_com.h
@@ -55,6 +55,7 @@ extern int32_t PIOS_COM_ChangeBaud(uintptr_t com_id, uint32_t baud);
 extern int32_t PIOS_COM_SendCharNonBlocking(uintptr_t com_id, char c);
 extern int32_t PIOS_COM_SendChar(uintptr_t com_id, char c);
 extern int32_t PIOS_COM_SendBufferNonBlocking(uintptr_t com_id, const uint8_t *buffer, uint16_t len);
+extern int32_t PIOS_COM_SendBufferStallTimeout(uintptr_t com_id, const uint8_t *buffer, uint16_t len, uint32_t max_ms);
 extern int32_t PIOS_COM_SendBuffer(uintptr_t com_id, const uint8_t *buffer, uint16_t len);
 extern int32_t PIOS_COM_SendStringNonBlocking(uintptr_t com_id, const char *str);
 extern int32_t PIOS_COM_SendString(uintptr_t com_id, const char *str);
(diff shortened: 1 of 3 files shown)


flight/PiOS/inc/pios_com.h
```diff
@@ -55,6 +55,7 @@ extern int32_t PIOS_COM_ChangeBaud(uintptr_t com_id, uint32_t baud);
 extern int32_t PIOS_COM_SendCharNonBlocking(uintptr_t com_id, char c);
 extern int32_t PIOS_COM_SendChar(uintptr_t com_id, char c);
 extern int32_t PIOS_COM_SendBufferNonBlocking(uintptr_t com_id, const uint8_t *buffer, uint16_t len);
+extern int32_t PIOS_COM_SendBufferProgressWatch(uintptr_t com_id, const uint8_t *buffer, uint16_t len, uint32_t max_ms);
```
tracernz: With this name I'd expect a progress callback, and it tells me nothing about what actually differentiates it from PIOS_COM_SendBufferNonBlocking (or even that they're similar either). PIOS_COM_SendBufferAsyncTimeout would make a lot more sense to me (but only if /s/PIOS_COM_SendBufferNonBlocking/PIOS_COM_SendBufferAsync/, so alternately the seriously long PIOS_COM_SendBufferNonBlockingTimeout to minimise diff :/).

mlyle: So, the problem is-- it's blocking, and there's not a strict timeout.  Rather, there's a timeout when there's no progress made for a defined interval.  Indeed, it's the exact same thing as SendBuffer (the blocking call) was when 5000 is passed for the last arg.  Better name very welcome :/

tracernz: Ah, I was looking at the wrong one :P. Sometimes I can see the wins of C++ and overloading would be one here (-e- or better the timeout as a default param). PIOS_COM_SendBufferTimeout says more about it IMO but... :/ I definitely think ProgressWatch isn't right though, because it provides no way to watch progress.

mlyle: It's not too bad, because I took the existing SendBuffer code and just have it call this with one more arg-- so it's pretty much a default value.

I have StallTimeout now...

tracernz: That's a good name. :+1:

PR #1929, opened 2017-08-29
hangtime: implement a leaky bucket algorithm

mlyle:
This allows power added to exceed the configured value for a short
time-- to prevent tumbling when chopping throttle from
fast-forward-flight (where aerodynamic loads require high stabilization
power) and to allow entry to aggressive maneuvers while in hangtime
regimes.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -560,26 +560,7 @@ static void normalize_input_data(uint32_t this_systime,
 		throttle_val = -1;
 	}
 
-	static uint32_t last_pos_throttle_time = 0;
-
-	*stabilize_now = *armed && (throttle_val > 0.0f);
-
-	if (*stabilize_now) {
-		if (actuatorSettings.LowPowerStabilizationMaxTime) {
-			last_pos_throttle_time = this_systime;
-		}
-
-		// Could consider stabilizing on a positive arming edge,
-		// but this seems problematic.
-	} else if (last_pos_throttle_time) {
-		if ((this_systime - last_pos_throttle_time) <
-				1000.0f * actuatorSettings.LowPowerStabilizationMaxTime) {
-			*stabilize_now = true;
-			throttle_val = 0.0f;
-		} else {
-			last_pos_throttle_time = 0;
-		}
-	}
+	*stabilize_now = throttle_val > 0.0f;
 
 	float val1 = throt_curve(throttle_val, curve1,
 			MIXERSETTINGS_THROTTLECURVE1_NUMELEM);
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -311,14 +311,61 @@ static void fill_desired_vector(
 	 */
 }
 
-static void post_process_scale_and_commit(float *motor_vect, float dT,
-		bool armed, bool spin_while_armed, bool stabilize_now)
+static void post_process_scale_and_commit(float *motor_vect,
+		float *desired_vect, float dT,
+		bool armed, bool spin_while_armed, bool stabilize_now,
+		float *maxpoweradd_bucket)
 {
 	float min_chan = INFINITY;
 	float max_chan = -INFINITY;
 	float neg_clip = 0;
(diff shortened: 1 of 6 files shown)


mlyle:
This is -critical- code, so I need to review this myself first before going any further.

mlyle:
Here's how it looks if you're at 90% throttle, chop, and control loops want to use 100% authority:

image

mlyle:
Here's how it looks if you are at 90% throttle, chop, and the stabilization loop wants 30% power, steady state.

image

mlyle:
Here's what it looks like if you chop from 90% power, and the control loop wants to use 60% power, 50ms on, 50ms off.

image

mlyle:
X axis, all graphs: time.  y axis, all graphs: power added.  Power is chopped / demands start being made at 2s.  Maxpoweradd configured as 15%.

mlyle:
One last plot, probably the most realistic:

image

90% power chopped;control loop wants to use 45% authority (2 motors at 90%) 450ms on, 150ms off.  Seems to have a pretty sane result for what the control loop is actually granted.

glowtape:
I'm curious as to how the chopped power method flies and the effect it has on the motors.

mlyle:
They're not different methods.. The 'chopped' are showing what happens to delivered power when there are intermittent stabilization power demands (control deflections, etc)

mlyle:
Just for clarity--- old way: hangtime allows up to 15% power for stabilization, which means it can e.g. if maxpower add is set to 15%, spin 2 motors up to 30%.  But the stabilization power demands are intermittent, so the average is guaranteed to be less than 15% and you can't do any maneuver that takes more power than this.

The new way: the long term **average** is guaranteed to be 15% or less, and the maximum instantaneous power used after some time in hangtime is 30% (double 15%).  But for a short time the limits are much looser, depending on how much power you were using before.

If you're in fast-forward-flight and the rear motors are at 45% more power than the fronts to stop from tipping over, the old way would reduce in the thing aerodynamically tipping over if you went to hangtime and not that much power was available.  Now there's a little bit of time when the limits are much more lax, so hopefully the vehicle can slow down and aerodynamically unload in that time.

mlyle:
Further extended to do the hangtime decision making up in stabilization so that the integral term can still do integrally stuff.

This requires careful verification.  I have already read diff and exercised some in simulator.

flight/Modules/Actuator/actuator.c
```diff
@@ -311,15 +311,62 @@ static void fill_desired_vector(
...
+	 * previously did, while still providing similar safety properties.
+	 * A secondary motivation is to prevent tumbling when throttle is
+	 * chopped during fast-forward-flight and more than hangtime power
+	 * levels are needed for stabilization (because of aerodynamic forces).
+	 *
+	 * The "maximum" stored is based on recent throttle history-- it decays
+	 * with time; at high throttle it corresponds to 300ms of the current
+	 * power; at lower throttle it corresponds to 300ms of double the
+	 * configured value.
+	 */
```
tracernz, line 341: 👍

tracernz (approved):
Looks good. 👍

glowtape:
Just went for a test flight. During hangtime, my quad maintains attitude nicely, now that the I-term works. When cutting quickly, you see a tiny jolt in the goggles, but the I-term seems to catch up in a fraction of a second and things seem stable.

However, the time limit doesn't kick in. I can put my quad on the ground, power it up, and when I arm, the motors spin idly. A quick throttle jolt and back to zero throttle (so it sits on the ground), hangtime kicks in, as expected, but doesn't time out on the 4.5s I've configured. It keeps slowly spooling up on the gyro noise, until it eventually wants to flip. If you disarming and rearming, it just continues. Until I power cycle.

glowtape:
On the bench, after hangtime, throttle only goes to 1,00001e-37 according to GCS, not true zero. Might be that.

glowtape:
Nevermind, it's my ManualControl interpolator mode. I have to fix that.

mlyle:
Rebased to be on top of #1952 for safety reasons.

mlyle:
Ready for merge, I think.

PR #1930, opened 2017-08-29
Remove incorrect (and unused) unit conversions.

mlyle:


mlyle:
Reported by @auturgy at TauLabs/TauLabs#2265.  Thanks for the report!  Would be unfortunate if someone decided to use this and got lies back as a result.

PR #1933, opened 2017-09-02
Windows build env: download latest Win32OpenSSL

gregorydulin:
Added code to search Win32OpenSSL download page for the latest
Win32OpenSSL version, vs. hard-coding the version.

Depends on curl, grep, and sed (I *think* these are included with
Windows git bash)

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -697,7 +697,11 @@ endif
 
 # OPENSSL download URL
 ifdef WINDOWS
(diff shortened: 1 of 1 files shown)


mlyle:
this doesn't test it, but it's a required checklist item too

jenkins, test this please

gregorydulin:
I did some research on the wayback machine, and this will fail if there are two versions available with a lowercase letter in the version number (which was the case in May, at least).  I'll fix it.

PR #1935, opened 2017-09-02
Make: only check for clazy when we need it

tracernz:
And suppress which error output if it's not found (only happens on some dists).

@gregorydulin


gregorydulin:
This worked for me.  I no longer get scary clazy messages when I make this or that.  Thanks!

@tracernz

mlyle:
lgtm, want to run it through a build

PR #1936, opened 2017-09-02
make: tools: find latest Win32OpenSSL_1_0_*

gregorydulin:
Eliminates the need to modify make/tools.mk every time slproweb.com
updates their Win32OpenSSL packages

use curl to get https://slproweb.com/prodcuts/Win32OpenSSL.html
use grep to find lines with 'Win32OpenSSL-1_0_*'
use head to grab the first matching line (in case they decide to make
two versions available)
use sed to get the .exe's URL

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -697,7 +697,12 @@ endif
 
 # OPENSSL download URL
 ifdef WINDOWS
-  openssl_install: OPENSSL_URL  := https://slproweb.com/download/Win32OpenSSL-1_0_2k.exe
+  openssl_install: OPENSSL_URL  := https://slproweb.com$(shell \
+          curl -s https://slproweb.com/products/Win32OpenSSL.html \
+          | grep -P 'Win32OpenSSL-1_0_[0-9]*[A-Za-z].exe' \
+          | head -1 \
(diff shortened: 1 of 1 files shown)


tracernz:
jenkins, test this please

@gregorydulin note that you can force-push over the previous branch and continue with the old PR if you like next time. 😉

mlyle:
Works like magic.  @gregorydulin thanks for this, will fix one bit of annoying maintenance. :fireworks:

mlyle:
One minor problem-- this is invoked on every use of make on windows--- would be better to use an = instead of a := assignment

mlyle:
jenkins, test this please

mlyle:
I'm lazy-- as this stands it's not perfect-- it evaluates the remote web stuff twice per openssl install-- but at least it doesn't evaluate it every make invocation.

tracernz:
Silly US imposed laws are the reason we even need this. :( http://doc.qt.io/qt-5/ssl.html#import-and-export-restrictions

mlyle:
We couldn't tamper with and make everyone's crypto weaker without ITAR, so give it credit.

The real big thing is how much openssl is broken, so there can't be a stable version/location.

Of course, we barely use SSL at all, anyways..

Does the new version look OK to you, @tracernz?

tracernz:
Yep. 👍

PR #1937, opened 2017-09-05
Serial receivers: Bumped failsafe timer to prevent intermittent radio issues from entering failsafe for just fractions of a second.

glowtape:
Whenever Crossfire switches from 50hz link rate to 150hz, say because it's switching between the two due to difficult radio conditions, there appears to be a gap of ~68ms in serial communication, before it actually bumps back to 150hz. The previous failsafe timeout was 50ms in the Crossfire driver, and might force the quad into very brief amounts failsafe.

This patch raises it to 80ms. That should still be acceptable I think.

I've been tracking down a source of these odd unpredictable twitches. Hoping this might be it.

sds00001

### DIFF
--- a/flight/PiOS/Common/pios_crossfire.c
+++ b/flight/PiOS/Common/pios_crossfire.c
@@ -374,7 +374,8 @@ bool PIOS_Crossfire_IsFailsafed(uintptr_t crsf_id)
 	if(!PIOS_Crossfire_Validate(dev))
 		PIOS_Assert(0);
 
-	return dev->failsafe_timer > 32;
+	/* ~250ms to failsafe. */
+	return dev->failsafe_timer > 156;
 }
 
 #endif // PIOS_INCLUDE_CROSSFIRE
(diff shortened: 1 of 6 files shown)


mlyle:
Can we maybe make this even bigger?  IMO there is no harm in holding a previous command on a radio failure for 250ms or even a little more-- I'd rather have my last command held for 250ms than have 170ms of failsafe in there if there's a little quarter second blip of connectivity while behind a tree.

glowtape:
Certainly have no issue with that. Stuck to ~50ms before, because the IBus driver this was derived of, had it too. So does S.Bus, and similar in one or two of the others. Figured there was a reason.

(Might get a little awkward to keep the last command that long in a 900°/s flip, though. Then again, if you're doing that sort of stuff, attitude has probably gotten biased quite a bit, anyway, and trying to level out will lead to hilarious results.)

Changing the other serial drivers, too. I presume that's OK. RFM22 and OpenLRS seem to have configurable timeouts.

glowtape:
Definitely does something positive. I was out earlier today, and there was exactly zero twitching and the control value hold during these micro-failsafes wasn't noticeable except for one occasion where I did notice a small delay (in a situation I wouldn't have wanted it to enter level mode even for just a bit).

PR #1938, opened 2017-09-06
simulation: fix regression from pr #1927

mlyle:
This was not supposed to touch the simulation target--- but somehow I
managed to anyways.  Whoops.


tracernz (approved):
Not convinced that sim is a blocker 😛 . Looks good. 👍

jihlein:
Looks good, now executes cleanly.

PR #1941, opened 2017-09-09
PiOS:MPU: Handle disabled MPU9250 magnetometer correctly

### LINKED ISSUE #1940: Boot loop on SPR F3 EVO if mag is set to disabled
* Seen in version (Neat Release-20170717): 
* On flight controller (Sprf3e): 

Issue details:
I have verified this on two different FCs.  Originally found it with an EMAX Femto SPRF3E clone that does not have an internal mag.  Was trying to get rid of the sensor error that was being thrown for the missing mag.  As soon as I disabled the mag and rebooted, the board would go into a boot loop.  Had to clear the settings partition to get the boot loops to stop.  Tried several times, and every time the mag is disabled, the board will get stuck in a boot loop.
Also verified it with a genuine SPR F3 Evo board.  Board was working fine with mag enabled.  As soon as mag is disabled, it gets into the same boot loop and can only be recovered by clearing the settings partition.

tracernz:
If the the mag was intentionally disabled (e.g. by the user configuring mag to none in HwUAVO), we would fail out erroneously if the sensor was known to have an internal magnetometer (MPU9250 only sadly, as MPU9150 doesn't have it's own WHOAMI). Still doesn't seem to be enough for #1940 😢 .

### DIFF
--- a/flight/PiOS/Common/pios_mpu.c
+++ b/flight/PiOS/Common/pios_mpu.c
@@ -429,16 +429,25 @@ static int32_t PIOS_MPU_Common_Init(void)
 
 #ifdef PIOS_INCLUDE_MPU_MAG
 	/* Probe for mag */
-	if (mpu_dev->cfg->use_internal_mag && PIOS_MPU_Mag_Probe() == 0) {
-		PIOS_MPU_Mag_Alloc(mpu_dev);
-		if (mpu_dev->mpu_type == PIOS_MPU60X0)
-			mpu_dev->mpu_type = PIOS_MPU9150;
-		if (PIOS_MPU_Mag_Config() != 0)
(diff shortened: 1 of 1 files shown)


tracernz:
Something fishy happening with SPF3E... I put `PIOS_HAL_CriticalError(PIOS_LED_ALARM, 4);` at the end of PIOS_Board_Init and it manages to boot without hitting that... Whereas before I put that line it would fail with two flashes somewhere else..

tracernz:
This patch at least makes the MPU driver behave correctly though so it's good regardless of what the SPF3 is doing.

tracernz:
build artifacts please

mlyle:
Yes, this looks good on its own-- thanks!

PR #1944, opened 2017-09-17
DShot telemetry support (WIIIIIP)

glowtape:
Early version, posted for appraisal while I'm trying to figure out something unrelated, and to see whether we want or need this (UAVO is 150 bytes). Seems to work with a single ESC. Need to do some more work on handling timeouts.

Surprisingly there's at least one thing not up to spec with BLHeli32 (current isn't output in the units the KISS specs wants).


glowtape:
Hooked up telemetry wires to the other ESCs, since I had to switch out motors anyway. Seems to capture telemetry for the appropriate actuators properly.

tracernz:
Neat. I'd been wondering about doing this. 👍

glowtape:
Voltage and mAh Consumed could probably turned into a single field. Since all ESCs usually draw from the same power bus, it doesn't make sense to keep separate voltages. Similar for consumed energy. Would also shrink the UAVO size.

While trying to make a screenshot for pudding, I noticed that it keeps timing out (currently 7.5ms in my local branch) when the motors are spinning, but doesn't appear to when they're disarmed. Not sure where to move the blame, need to put the scope on it.

Also, once the timeout business is solved, and I got some clarification from the BLHeli guys regarding the current values, might be an idea to let the Battery Monitoring module tap alternatively into the telemetry UAVO, if there aren't any sensors.

Either way, here's the aforementioned pudding:

esctelem

glowtape:
Collapsed voltage and mAh consumed, UAVO dropped from 150 to 96 bytes. Polls every 200ms now. Some history rewriting.

flight/PiOS/STM32/pios_servo.c
```diff
@@ -99,6 +99,8 @@ static bool resetting;
 
 static bool dshot_in_use;
 
+static int telemreq_servo;
```
mlyle: At the least this needs to be volatile, but it probably needs to be atomic in actuality.  This also deepens the debt for flyingpi.

glowtape:

glowtape:
@ufoDziner Seems to work on the desk. Since this doesn't contain the LQG stuff, it'll trash some of the configuration. Best to save the settings partition of the LQG configured quads before switching to this.

ufoDziner:
Appears to work with X-Racer Quadrant 25A escs.  Will continue to test.

tracernz:
KISS is the reference implementation right? Should be able to rig that up later in the week.

glowtape:
Yeah, KISS. BLHeli implements it the same (for now) from what I can tell.

ufoDziner:
@glowtape Does this require any further testing for blheli_32?

glowtape:
Not that I know of. I had it running on _32, you on _s, so it works. I need to eventually prettify it a bit.

flight/Modules/Actuator/actuator.c
```diff
@@ -251,6 +282,12 @@ int32_t ActuatorInitialize()
 	}
 #endif
 
+#if defined(PIOS_INCLUDE_ESCTELEMETRY)
+	if (ActuatorTelemetryInitialize() == -1) {
```
tracernz: Safe to go `< 0` or `!= 0` IMO, *Initialize might return multiple error codes at some point. :wink: Actually, [it already can return `-2`](https://github.com/d-ronin/dRonin/blob/next/flight/UAVObjects/uavobjecttemplate.c#L57), so this code is wrong.

flight/Modules/Actuator/actuator.c
```diff
@@ -611,6 +648,80 @@ static void post_process_scale_and_commit(float *motor_vect,
...
 
+#if defined(PIOS_INCLUDE_ESCTELEMETRY)
+	if (PIOS_ESCTelemetry_IsAvailable()) {
+
+		/* Process telemetry request stuff before servo settings. */
+
+		if (!tlm) {
+			tlm = PIOS_malloc_no_dma(sizeof(*tlm));
+			if (!tlm)
+				PIOS_Assert(0);
```
tracernz: PIOS_Assert(tlm);?

flight/Modules/Actuator/actuator.c
```diff
@@ -611,6 +648,80 @@ static void post_process_scale_and_commit(float *motor_vect,
...
+			/* If active servo is 0xFF, no request is active, either because we're starting over,
+			   the servo doesn't exists, or isn't configured to DShot. */
+
+			tlm->active = tlm->servo;
+			tlm->servo++;
+
+			tlm->req_start = PIOS_DELAY_GetuS();
+
+			/* We don't care about ongoing requests, because we shouldn't even be here, if one is. */
+			if (PIOS_Servo_RequestTelemetry(tlm->active) < -1)
```
tracernz: What does -1 mean? I always prefer to use enums when actually differentiating errors. e.g.
`if (PIOS_*() == -EBUSY)` or similar is a lot more readable than `if (PIOS_*() == -5)`.

flight/PiOS/Common/pios_esctelemetry.c
```diff
@@ -0,0 +1,244 @@
...
+				// Unfreeze telemetry driver.
+				break;
+			}
+		}
+	}
+
+	dev->rx_timer = 0;
+
+	// Keep pumping data.
+	if(headroom) *headroom = ESC_TELEMETRY_LENGTH;
```
tracernz: New lines, and space after if (applies to next few lines too).

flight/PiOS/Common/pios_esctelemetry.c
```diff
@@ -0,0 +1,244 @@
+/**
```
tracernz: This design will not work for multiple different ESC protocols? This should be `PIOS_DShot_Telemetry` or something? And other ESC telemetry drivers need a way to hook in somehow. Other protocols existed before DShot, and we have already discussed supporting them, so a generic `PIOS_ESCTelemetry` implementation that can hook into multiple different backend drivers is essential IMO.

flight/PiOS/Common/pios_esctelemetry.c
```diff
@@ -0,0 +1,244 @@
...
+	PIOS_ESCTelemetry_ResetBuffer(dev);
+
+	uint8_t crc = PIOS_CRC_updateCRC(0, dev->rx_buf, ESC_TELEMETRY_LENGTH-1);
+
+	if (crc == dev->rx_buf[ESC_TELEMETRY_LENGTH-1]) {
+
+		dev->telem.temperature = dev->rx_buf[0];
+		dev->telem.voltage = (float)(dev->rx_buf[1]<<8 | dev->rx_buf[2]) * 0.01f;
+		dev->telem.current = (float)(dev->rx_buf[3]<<8 | dev->rx_buf[4]) * 0.01f;
+		dev->telem.mAh = (uint16_t)(dev->rx_buf[5]<<8 | dev->rx_buf[6]);
```
tracernz: Missing some spaces around operators on these lines.

tracernz (changes requested):
Nice work overall. :+1: Apologies for taking so long to review.

ufoDziner:
@glowtape Actually mine was _32 as well. I don't think _s supports it.

flight/Modules/Actuator/actuator.c
glowtape: Value -2 means already initialized, which isn't a reason we'd like to fail for. The all other UAVO inits do the == -1 thing, too.

flight/PiOS/Common/pios_esctelemetry.c
glowtape: This round gonna do it just DShot specific and rename the current driver accordingly. Later on we could revisit this and split things into generic parts and actuator specific telemetry. Haven't thought it through where to draw the line yet.

glowtape:
Rebased, applied review stuff. Added actuator "pins" to the Battery module to get data from the ActuatorTelemetry UAVO instead.

ufoDziner:

jihlein:
Were does this effort stand, does it need more testing?

glowtape:
Everything's moving to Bi-DShot, or whatever the hell it is called, which sends telemetry back over the PWM wire instead of a separate serial wire. Currently slowly dabbling with making it work, so this is obsolete.

PR #1950, opened 2017-09-24
smoothcontrol: Add linear interpolation, user-specified duty cycle. Rename UAVO to ManualControlInterpolation.

glowtape:
This patch enables the Normal mode by default, which works by capping the differential to 80% and then starts predicting a 80% slope of the differential at 50% of the manual control update cycle (say with S.Bus updating at 9ms, it'll predict for 4.5ms). There's two limits in play, for one, the 50% of the update cycle, and if for whatever reason that value is completely wrong (say the receiver troddling along drawing out the measured cycle time), there's a 20ms hard-limit (incidentally like PPM).

The linear interpolation modes, which seemed to be mysteriously broken a long while back, until I've found the issue and fixed with it #1902, are back. What regular linear interpolation does is obvious (if anyone wants BFs auto RC smoothing, it's basically that), the psychic mode does linear interpolation between the end values predicted by the Extended mode (does the same things as Normal, except it does 75% duty cycle on predicting).

Normal and Extended modes seems to work fine, without any apparent ill-effects. I'm however flying Crossfire and it's fast update rate. Feedback from someone used to flying slower protocols would be appreciated.

Been flying RPY on Psychic mode and throttle on linear interpolation for a while, though, results in smooth flying.

Gonna add UI in another patch.

### DIFF
--- a/flight/Libraries/math/smoothcontrol.c
+++ b/flight/Libraries/math/smoothcontrol.c
@@ -86,6 +86,12 @@ void smoothcontrol_reinit(smoothcontrol_state state, uint8_t axis_num, float new
 	state->axis[axis_num].current = new_signal;
 }
 
+// Resets the thrust state.
+void smoothcontrol_reinit_thrust(smoothcontrol_state state, float new_signal)
+{
+	smoothcontrol_reinit(state, 3, new_signal);
+}
+
 // Sets mode for an axis.
 void smoothcontrol_set_mode(smoothcontrol_state state, uint8_t axis_num, uint8_t mode, uint8_t duty_cycle)
 {
--- a/flight/Libraries/math/smoothcontrol.c
+++ b/flight/Libraries/math/smoothcontrol.c
@@ -38,6 +38,9 @@ struct smoothcontrol_state_internal {
 	// Control interval determined by the ringer.
 	uint8_t control_interval;
 
+	// Duty cycle to use.
+	float duty_cycle;
+
 	// RPY+Thrust.
 	struct smoothcontrol_axis_state axis[4];
 };
@@ -46,6 +49,7 @@ struct smoothcontrol_state_internal {
 static void smoothcontrol_update(smoothcontrol_state state, struct smoothcontrol_axis_state *axis, const float new_signal)
 {
 	float signal_diff = new_signal - axis->signal;
+	float cycles = MIN((float)state->control_interval * state->duty_cycle, state->control_interval);
 
 	switch(axis->mode) {
 		default:
@@ -57,11 +61,14 @@ static void smoothcontrol_update(smoothcontrol_state state, struct smoothcontrol
 
 		// icee's proposal, creates chamfered steps with signal prediction
 		case SMOOTHCONTROL_NORMAL:
-		case SMOOTHCONTROL_EXTENDED:
 			axis->differential = signal_diff * SMOOTHCONTROL_PREDICTOR_SLOPE / (float)state->control_interval;
 			axis->current = axis->signal + signal_diff * SMOOTHCONTROL_CHAMFER_START;
-			axis->integrator_timeout = (uint8_t)MIN((float)state->control_interval *
-				(axis->mode == SMOOTHCONTROL_NORMAL ? SMOOTHCONTROL_DUTY_CYCLE : SMOOTHCONTROL_EXTENDED_DUTY_CYCLE), 255);
+			axis->integrator_timeout = (uint8_t)cycles;
+			break;
+
+		case SMOOTHCONTROL_LINEAR:
+			axis->differential = (new_signal - axis->current) / cycles;
+			axis->integrator_timeout = (uint8_t)cycles;
 			break;
 	}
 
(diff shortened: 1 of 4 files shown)


shared/uavobjectdefinition/stabilizationsettings.xml
```diff
@@ -52,8 +52,8 @@
 		<field name="DeadbandSlope" units="%" type="uint8" elementnames="Roll,Pitch,Yaw" defaultvalue="60,60,50" limits="%BE:0:100,%BE:0:100,%BE:0:100">
 			<description>Sets the slope of the deadband area in the PID controller.</description>
 		</field>
-		<field name="RCControlSmoothing" units="" type="enum" elementnames="Axes,Thrust" options="None,Normal,Extended" defaultvalue="None">
-			<description>Enables different ways of input signal smoothing, to reduce excessive P- and D-term excitation in the PID controller. Normal and Extended modes chamfer the leading edges in combination with some amount of prediction, to avoid delay.</description>
+		<field name="RCControlSmoothing" units="" type="enum" elementnames="Axes,Thrust" options="None,Normal,Extended,Linear,Psychic" defaultvalue="Normal">
```
tracernz: Shouldn't this be ManualControlSmoothing?

glowtape: There wasn't any opposition to the current name when it first was committed. But changing it should be OK, if ManualControlSmoothing sounds better and is more in line with naming. The setting is basically unused currently, apart from me, and renaming it shouldn't break things on upgrades I'd figure.

glowtape:
The interpolators never reach true zero, instead just an infinitesimally small number. New hangtime doesn't seem to agree with it. The latest commit in this patch fixes it.

mlyle:
I'm not sure this latest change is enough...

I think if the user provides a 0 throttle, the 0 throttle should get passed through directly and shouldn't wait for an interpolator to get there.

glowtape:
Hmmm. It did that before, because no thrust means -1 in StabilizationDesired.Thrust and that value got passed through to ActuatorDesired.Thrust, which smoothcontrol acts on. Going negative reset the state instantly. ActuatorDesired doesn't go negative anymore. Gonna fix that.

mlyle:
Before, I think it would be something between -1..0 -- depending on how far below the "Neutral" value you were on the throtttle channel.

glowtape:
Dropped the now conflicting commit that moved to #1952.

shared/uavobjectdefinition/stabilizationsettings.xml
tracernz: It's a usability issue to use different terminology for the same thing in user-facing stuff like UAVOs.

glowtape:
Now there's a duty cycle variable, that applies to normal mode that chamfers the outer edges of the stair-stepped signal. Dropped extended mode because the duty cycle is configurable. Duty cycle also applies to the interpolation modes.

Tried linear interpolation at 50% duty cycle. Chucking around the quad in attitude mode sounds definitely a little smoother, when motor throttle swings back and forth. Still need to go outside and create logs from this.

(The other camp flies with full interpolation enabled by default and they all seem happy about it. So linear interpolation with 50% duty cycle sounds like a nice compromise between delay and smoothness.)

glowtape:
Removed psychic mode. Considering most people use serial receivers now, they're relatively low latency, tau is still above 10ms, there's probably nothing to be had here. Things like FrSky half-channel flipflopping is kind of counterproductive here, when someone has it configured, the waveform will end up triangular, because of overshooting every other time when there's stick movement. Crossfire has some minor aliasing because the transmitter gets updated at 240hz, but only 150hz go OTA, so there's also intermittent overshoots.

glowtape:
Rebased and some history rewriting.

flight/Libraries/math/smoothcontrol.c
```diff
@@ -80,15 +87,17 @@ void smoothcontrol_reinit(smoothcontrol_state state, uint8_t axis_num, float new
...
 
 // Sets mode for an axis.
-void smoothcontrol_set_mode(smoothcontrol_state state, uint8_t axis_num, uint8_t mode)
+void smoothcontrol_set_mode(smoothcontrol_state state, uint8_t axis_num, uint8_t mode, uint8_t duty_cycle)
 {
 	PIOS_Assert(state && axis_num <= 3);
-	state->axis[axis_num].mode = mode;
+	state->duty_cycle = bound_min_max((float)duty_cycle, 0, 100) / 100.0f;
+	/* Below 10% duty cycle, all this is likely useless, so disable. */
+	state->axis[axis_num].mode = duty_cycle >= 10 ? mode : SMOOTHCONTROL_NONE;
```
glowtape: Need to give this yet another test. I flew it already without this line/override above present, but I added that check later on and figured it'll work without testing, except I picked the duty_cycle from within the state struct instead of the argument. Should work now, but need to be sure.

mlyle:
How do you feel about merging this, without the enabling-normal-by-default thing... and then with more audit enabling it.

I'm most concerned about SC's interaction with other control modes, with throttle, etc.

mlyle:
Also-- does this entirely fix #1974?

glowtape:
I've been flying extended mode (normal mode with longer prediction cycle) and that now removed psychic mode before without apparent issues in the past. In the kitchen sink build of mine for the last week, I've been flying linear interpolation at 50% just fine (with nice results on the ActuatorDesired envelopes), on rate and attitude each with both controllers.

When an axis is set to none, smoothcontrol_run exits without doing anything. So it's like it wasn't there.

I removed the bounding parameter, since it wasn't used, and I seem to remember that we concluded the checking should happen over in stab.c, anyway. Since I'm currently hanging over it, some modes don't actually do that, and I might just add it. It'd be temporary until the refactor.

Regarding failsafe, it's not aware about it. I need to fix this.

glowtape:
Rebased to fix latest conflicts, tested on two quads.

flight/Libraries/math/smoothcontrol.c
mlyle: Is there any reason to do this?  (Allow 10% duty cycle but prohibit 9%?)  It's counterintuitive for it to just magically do nothing.  (And for a radio that updates at 50Hz, 2ms chamfers might still do something).

glowtape: Yea makes sense. This happened more under the assumption that people fly fast protocols these days and it'd save a few cycles exiting early. Gonna drop it.

glowtape:
Stopped force-disabling smoothcontrol when duty cycle is below 10%.

mlyle (approved):
Thanks much!

PR #1953, opened 2017-10-01
Support for 3D flight

mlyle:
**~~No one should try this!  I have not even read through the diff myself yet!~~**

This is based off #1929 because it's somewhat related.

## Known todo:

* [x] ~~Need to reimplement throttle curve somewhere appropriate and fixup UI and remove from current UAVO.~~ For now rip out thrust curve support-- it belongs on "input" in the future, and we may need to add a 'curve 3' for vcurve stuff.  Curve1 will be saved for legacy reasons (so that upgrades go smoothly)
* [x] No DShot implementation of 3D stuff yet.
* [x] Figure out ESC calibration for 3D stuff?

## What this does:

* Remove mavlinkbridge to make some space on sprf3e.
* Remove -fast-math from all targets, because I wanted to use NANs for stuff and convinced myself it was dangerous.
* Changes thrust behavior in manualcontrol
  * Auto-determine whether thrust channel is bidirectional.
  * Don't use vehicle type to determine thrust channel
  * Always use ranges -1 .. 1 for autonomy (althold, vtolpathfollower) inputs
* Stabilization hangtime -- keep the last sign for throttle when doing hangtime.
* Significant actuator changes
  * Use NAN to communicate "this motor shouldn't spin"
  * Add negative thrust saturation logic-- allow only negative values for motors when throttle is negative, allow only positive values for motors when throttle is positive.
  * Add a deadband value.  If configured for a channel, the channel is a 3D motor.  Neutral value is sent to not spin, and the motor ranges are neutral+deadband .. max for positive thrust values, min..neutral-deadband for negative thrust values.  Otherwise our existing behavior is maintained.
  * Remove throttle curve from actuator, because it conflicts with this and the known problems with it (#671) 

## Configuration:

* Configure ESCs to be bidirectional.
* Input: set throttle's neutral value to the middle.
* Output: set neutral values to middle value.
  * 3D mode is enabled for a motor when ChannelDeadband >= 1.  If the neutral value is in the middle this is enough to get the ESCs to arm beep. 
  * You can experimentally find values with "test outputs" for when motors start to spin... and then set the neutral value to the midpoint of reverse and forward spinning direction.  ChannelDeadband should be set to 15% more or so than the span between the reverse and forward spinning stuff.
  * Alternatively one can use the values from BLHeli configuration-- e.g. 144/186/229
  * Calibration may take care of this ultimately.  (Uh, blheli source code is hard to read but it looks like calibration is the same)
* If motors spin the wrong way, reverse min and max value.


tracernz:
Removing throttle curve makes us much worse on non-multirotors.

mlyle:
Where throttle curve is implemented is wrong (#671) and has really bad effects.  (The path followers, althold don't expect it).

The todo is to put it back up in manualcontrol.

Curious what the big negative issue is though with other aircraft-- planes?  Presumably CCPM you'll be using the collective curve which is still there.  Multirotors have it really bad because often hover is 12% or whatever which people want to move closer to the middle.

tracernz:
CCPM needs V-curve unless you're running a governor.

mlyle:
I don't think we would have supported doing that ourselves before-- because you'd need a linear-ish collective curve and then another curve that uses collective as input.  While before you could have a throttle curve defined for input 0..1 that looks at the throttle source channel and another curve defined over range -1..1.

tracernz:
?
screenshot from 2017-10-02 17-33-35

mlyle:
The left one would be full throttle from middle stick down, I think.  (it's domain 0..1)

tracernz:
You're right. Ick.

mlyle:
First bit of progress / sanity check... In the simulator,

image

Which isn't exercising actuator yet (though I'm gonna start eyeballing actuator values to sanity check the next phaase.

mlyle:
Setting rollmax to 180 works with this hack patch.

image

mlyle:
Inspection of actuator values inside simulator looks plausible now...

mlyle:
Played with hangtime some, too.

Time for a big inspection of the diff.

mlyle:
Brief bench test shows things to be doing sane stuff...

Brief, initial bench test with this:

image

on lux looks good, though I need to reverse two of the motors.

mlyle:
https://streamable.com/dnqk4

mlyle:
OK...

I think it works.

I think I'm bad at 3d flight.

I don't really have a test vehicle for this anymore.

mlyle:
Sum of testing done:

* Static bench tests.
* Normal hover
* "Hold-down" tests, negative throttle on ground, rocking back and forth to confirm control authority.
* Inverted hover --- yeow scary
* Several attempted transitions back and forth where the right things seemed to be happening but the pilot did very badly.

mlyle:

LitterBugs:
Is it plausible to configure a switch that would set the leveling orientation to be normal or inverted?   (for auto-leveling modes)

mlyle:
@LitterBugs -- yes, and I was thinking I'm going to implement that so that I can actually test in my backyard.

It'll be a reprojection that will add 180 to roll, flip the sign on pitch and throttle.

LitterBugs:
Might also be interesting to have a variation when in horizon mode, that the throttle +- determines which way is up.....

mlyle:
OK uh.. I implemented that.. let's just say bad idea and leave it at that.  It worked, though.  :P

mlyle:
Yah.  I'm not going to build the reprojection.  I think 3D thrust should be saved for A) if you know what you're doing LOS, or B) for FPV

LitterBugs:
No worries.  Probably best to stick with the basics and add extra functionality later anyway.  Is it to a point ready for testing?

mlyle:
Yes, I've flown it successfully, it seems to work.

BUT-- be very careful.  This changes the two most dangerous things to change.  It changes actuator, which if it had bugs could spin motors uncommanded... and it changes manualcontrol, which changes how the sticks and control are interpreted.  I have read the diff, and I've flight tested.

Not compatible with d-shot currently though.

LitterBugs:
Trying to test on a Genuine SPR F3 Evo and having issues with the outputs tab.  Can not get it to save settings.  Keeps reverting back to default update rates and min/max values when the board restarts  I've tried multiple times, restarting just the board or restarting the GCS and board makes no difference.

jhitesma:
@LitterBugs - there's been a bit of discussion about that in IRC.  It appears to be unrelated to the 3D changes but something is definitely goofy right now.  I actually experienced it with neat on my DTFc when I first put my danaus style thing together.

It appears that saving from the oavobrowser is working, but saving from the GUI proper is sometimes not saving.  The outputs page appears to have it happen the most.

jhitesma:
I did "fly" this today...but decided to play it "safe" and try a powered descent in attitude mode first. Unfortunately I tried it from way too low of an altitude and it came down considerably quicker than I expected and was unable to react to it in time:

https://www.youtube.com/watch?v=P_36zXmtL5Q

It does seems to do what it's supposed to.  Bench testing went well, and it did reverse thrust when I commanded it to.  I just made a spectacularly poor decision about when and how much reverse thrust to command 🤣

LitterBugs:
Thanks, UAV browser did the trick.

mlyle:
To pick up the GCS output form fix:

LitterBugs:
Tested latest artifacts on SPRF3 Evo.  Output tab saves to FC now.

LitterBugs:
"Test outputs" on ouput tab motors move in correct directions, but when unchecking the test outputs box, motors go full negative rather than neutral/off   =-O

LitterBugs:
Getting same motors go full negative on disarm as well.   Oneshot125 on  BLHeli 14.9

LitterBugs:
Nevermind.  Somewhere along the line my ChannelDeadband settings got zeroed out.

LitterBugs:
Maiden and autotune.
http://dronin.org/tunes#?key=ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAiJH10gsM
Enabled Altitude hold module and Altitude Hold flight mode.  Enabled Hang time.  Bumped rates to my norm.  Verified Leveling and Horizon modes and motor directions working fine in both modes.  Pitch Flips, Roll flips, Yaw spins.  

Motors would not spin when arming in Altitude Hold mode.  Didn't have the guts to try switching to AH after arming in another flight mode.  May take props off and do more testing with GCS hooked up.

mlyle:
We disallow arming in althold.

mlyle:
Fixed althold based on report from @litterbugs.

mlyle:

mlyle:
rebased to next because of ... reasons

LitterBugs:
Testing latest artifacts...

LitterBugs:
Still will not descend in AltHold mode when configured for 3D motors.  Going lower than center on throttle makes quad rise.   Is throttle center for althold mode still the center or 3/4 throttle with 3D motor config?
Tried 3/4 just rises...

LitterBugs:
Inverted ground tests show motors/thrust reacting in correct direction to sticks when in negative thrust.  Two motors will go to zero thrust  under wrong conditions at close to zero throttle and in negative thrust.  Probably need to adjust center and deadband on those two motors rather than a firmware deficiency.

mlyle:

mlyle:

mlyle:

mlyle:
Failures were my fault-- I had a command shell in the artifacts dir.

mlyle:
jenkins, test this please

mlyle:
Flew a dshot flight test on a 95-size.  flew upside down-- took basically 100% throttle.  Had some weird pitching /rolling anomalies, both upside down and right-side up.  By end of flight I realized battery was hanging loose, which may have been the entire problem...

mlyle:
https://streamable.com/0ql4n

Need to make sure the initial spin-while-armed is positive direction, I think.

mlyle:

LitterBugs:
Testing this on a different build Seppuku FC with dshot ESCs.  Sinced shot handles the motor reversal via a message, are the output settings and channel deadband configured differently?

mlyle:
Yes, it's documented in the tooltip for deadband.  Calibrate neutral without 3D.  Then, deadband of 1 means "forward dshot 3d", deadband of 2 means "reverse dshot 3d" (you probably want 1).

Note that the ESC A) must be powered up before the flight firmware starts, and B) must support the dshot commands (check your specific firmware version).  We send the command 10 times early on for 3D mode and there is no acknowledgement-- if the ESC doesn't hear it or doesn't support it, negative 0 to 100% thrust will be instead positive 50 to 100% thrust and we'll never know.  It's kinda scary :/

LitterBugs:
I'm upgrading my Smoke-Stopper with a kill switch and pulling the props....

ESCs updated to 16.67....

LitterBugs:
Something does not seem right at all with my setup running dshot600.   Motors behave fine on outputs tab with very granular changes in speed.  But when I arm, its just about all or nothing other than idle speed.  Seems a bit better in the positive direction if I set  output deadband to 1, but negative goes full throttle, and in the positive direction and positive will not go very high.

LitterBugs:
Switching to Oneshot125 for the time being.  Going to try to head to the field.

LitterBugs:
In oneshot125, throttle is asymmetric when armed too.  Very touchy on positive throttle,  Smooth and controlled negative.

LitterBugs:
Outputs tab is smooth in both directions with OS125.

Defaulted all stabilization settings.  Seems like it is even in both directions now.  Maybe PIDs are not being applied in both directions or leveling is to blame on the bench......

mlyle:
@litterbugs that sounds like your ESCs are not receiving the dshot command--- you're powering up the FC by battery, right?

We send the command for dshot-3d at flight controller boot time, and don't/can't know that the ESCs are not on yet.  So if you plug in USB, then plug in ESC battery power they'll miss the command

mlyle:
PIDs are definitely applied in both directions, I flew this upside down with negative thrust ;)

LitterBugs:
Seems to be working fine in Oneshot125.  Just ran an AutoTune with some HQ 3D5x4.5 props.
http://dronin.org/tunes#?key=ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAiKfDnQkM
Heading to field.

I did power directly from battery and get different beep sequences when it was configured for Dshot 600.
Will look into Dshot later tonight.

LitterBugs:
Flew well in 3D acro OS125 out at the field.  Pilot....  not so much.  ;) 

The build I'm flying now has very consistent ESC timing in both directions even with OneShot125.
AirDTF Seppuku
Spedix IS30A with 16.67 BLHeliS
Center = 185.  Motors turn @ 189 and 181.  Deadband works good @ 10

LitterBugs:
Re-Tested Dshot600 / 3D on bench.  Had ESCs configured in 3D mode, Deadband = 1, Neutral =100.
Motors work as they should.  So my ESCs were not getting/Recognizing the 3D command at boot.

mlyle:
Force-push'd... to base on top of #1975 ... and added a couple new commits.

mlyle:

mlyle:
jenkins, test this please

mlyle:
Rebased on #1975

mlyle:
jenkins, test this please

mlyle:
With d-shot commands enabling 3d mode, and current bleeding edge 16.63 dshot-commands blheli-- https://streamable.com/ztfa6

mlyle:
Also https://streamable.com/pwbh7

mlyle:
Rebased.  Holy crap, 34 commits.  This is now in the critical path for A) python-XML-cleanup, B) motor-beacon-annunciator, and C) reprojection changes.  Yuck.

mlyle:

tracernz:
Needs solid testing on non-3d and non-quad stuff too since it touches critical path stuff quite heavily.

mlyle:
I'm not too worried about non-quad stuff, though heli feedback would be nice since this resulted in removing a couple of hacks related to it.  On fixed wing, etc, throttle/thrust work the same.

Nice cases for quads would be:

* [x] 3d acutators, 3d throttle, normal flight
* [ ] 3d actuators, 3d throttle, althold
* [ ] 3d actuators, normal throttle, normal flight
* [x] normal actuators, normal throttle, normal flight
* [x] normal actuators, 3d throttle, normal flight
* [x] ~~DONTCARE actuators, normal throttle, althold~~  Ran a set of althold scenarios in sim and feel pretty good about it based on just that-- 3D throttle, non-3D throttle, each reversed and not-reversed.  Right thing seems to happen across the board.

Knowing whether althold on helicopters still works would be nice.. @ekaszubski do you still live? ;)

mlyle:
Am curious about reverse thrust on planes, too.

ekaszubski:
@mlyle Hey, am still alive :) Glad to see you guys are still really active. I'll read over this and bench test it in a day or two for a basic sanity check. Need to catch up on what you guys have been working on, too.

Do these changes include reprojection e.g. for inverted althold? Or should I only be checking upright althold?

mlyle:
@ekaszubski awesome, thanks!  no bistability/reprojection in this... but big changes to throttle/thrust handling.

mlyle:
to get the UI hangtime bug fix:

expipiplus1:
While configuring a lizard95 (omnibus f3) with the setup wizard with the above artifacts I get a crash when clicking `next` after calibrating the gyros.

CC @LitterBugs

mlyle:
I am not good at this: https://streamable.com/9hc5j

ground/gcs/src/plugins/setupwizard/vehicleconfigurationhelper.cpp
```diff
@@ -622,9 +622,8 @@ void VehicleConfigurationHelper::resetVehicleConfig()
     MixerSettings *mSettings = MixerSettings::GetInstance(m_uavoManager);
 
     // Reset throttle curves
-    QString throttlePattern = "ThrottleCurve%1";
-    for (int i = 1; i <= 2; i++) {
-        UAVObjectField *field = mSettings->getField(throttlePattern.arg(i));
+    {
+        UAVObjectField *field = mSettings->getField("ThrottleCurve2");
         Q_ASSERT(field);
```
tracernz, line 627: needs `if (!field) { warning output, return, whatever }` Q_ASSERT is defined to nothing in rel builds. :/

flight/Libraries/math/smoothcontrol.c
```diff
@@ -117,21 +117,31 @@ void smoothcontrol_run(smoothcontrol_state state, uint8_t axis_num, float *new_s
 // Separate handling of thrust.
 void smoothcontrol_run_thrust(smoothcontrol_state state, float *new_signal)
 {
-	/* The -1 on zero throttle causes a huge blip when throttling up, so
+	/* The 0 on zero throttle causes a huge blip when throttling up, so
```
mlyle, line 120: I remain concerned about all this stuff.  Need to think about smooth control.

flight/Libraries/math/smoothcontrol.c
```diff
@@ -117,21 +117,31 @@ void smoothcontrol_run(smoothcontrol_state state, uint8_t axis_num, float *new_s
...
 		smoothcontrol_run(state, 3, new_signal, 1.0f);
 
 		// If prediction undershoots while original signal is positive
 		// bound it to zero.
-		if (*new_signal < 0)
-			*new_signal = 0;
+
+		bool new_sign = *new_signal > 0;
+
+		if (sign != new_sign) {
```
mlyle, line 138: Preserves 0 above; otherwise clips to a reasonable epsilon

flight/Modules/Actuator/actuator.c
```diff
@@ -100,24 +101,102 @@ static float motor_mixer[MAX_MIX_ACTUATORS * MIXERSETTINGS_MIXER1VECTOR_NUMELEM]
...
+int actuator_send_dshot_command(uint8_t cmd_id, uint8_t num_to_send,
+		uint16_t channel_mask)
+{
+	struct dshot_command new_cmd;
+
+	new_cmd.cmd_id = cmd_id;
+	new_cmd.num_to_send = num_to_send;
+	new_cmd.channel_mask = channel_mask;
+
+	if (__sync_bool_compare_and_swap(&pending_cmd.value, 0,
```
mlyle, line 178: This is an atomic in preparation for other modules to trigger dshot commands when disarmed (e.g. actuator beep)

flight/Modules/Actuator/actuator.c
```diff
@@ -100,24 +101,102 @@ static float motor_mixer[MAX_MIX_ACTUATORS * MIXERSETTINGS_MIXER1VECTOR_NUMELEM]
...
+static uint16_t desired_3d_mask;
+
+#define DSHOT_COMMAND_DONTSPIN 0
+#define DSHOT_COMMAND_NO3DMODE 9
+#define DSHOT_COMMAND_3DMODE   10
+
+#define DSHOT_COUNT_EXCESSIVE 12
+
+static bool actuator_get_dshot_command(bool armed) {
+	/* Finish off any current command. */
```
mlyle, line 142: The logic here is, if we have a command in progress, send it no matter what.  This allows a command to be sent at the arming edge and to continue sending.

flight/Modules/Actuator/actuator.c
```diff
@@ -341,7 +420,7 @@ static void post_process_scale_and_commit(float *motor_vect,
 	 */
 	float maxpoweradd_softlimit = MAX(
 			2 * actuatorSettings.LowPowerStabilizationMaxPowerAdd,
-			desired_vect[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE1])
+			fabsf(desired_vect[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE1]))
```
mlyle, line 421: The clip management logic is simplified by doing everything as we're positive thrust, and then inverting it all at the end.

flight/Modules/Actuator/actuator.c
```diff
@@ -715,6 +827,22 @@ static void actuator_task(void* parameters)
 				MIXERSETTINGS_MIXER1VECTOR_NUMELEM,
 				1);
 
+		/* At arming time, knock all 3d actuators into 3D mode.
```
mlyle, line 830: This is a tradeoff.  Ideally the end-user sets the ESC into the correct mode.

But it is *crucial* that if we're going to send 3D commands, the actuator be 3D, because small negative commands are otherwise very high throttle.  This could be a nasty surprise in flight.

Also, if the user doesn't set the ESC into the right mode, the neutral handling is all simplified.  So, meh.

flight/Modules/Actuator/actuator.c
```diff
@@ -761,24 +872,141 @@ static float collective_curve(float const input, float const * curve, uint8_t nu
 	return linear_interpolate(input, curve, num_points, -1.0f, 1.0f);
 }
 
+static float scale_channel_dshot(float value, int idx, bool active_command)
```
mlyle, line 875: Now when we're sending dshot, we don't have such a reliance on the MAX being a proper value, and neutral being in a legal range.  We really just calculate the correct values directly.

flight/Modules/Actuator/actuator.c
```diff
@@ -761,24 +872,141 @@ static float collective_curve(float const input, float const * curve, uint8_t nu
...
+			val = 0;
+		}
+
+		return val;
+	}
+
+	/* OK, this is more tricky.  Convert from a unipolar neutral value
+	 * to a "real one" as sent */
+
+	neutral -= 48;
```
mlyle, line 932: In the 3D case, use the same kind of configured neutral value as the non-3D case.  This means the neutral value is "half" of the amount sent over the wire.

flight/Modules/Actuator/actuator.c
```diff
@@ -761,24 +872,141 @@ static float collective_curve(float const input, float const * curve, uint8_t nu
...
+	if (val > 999) {
+		val = 999;
+	}
+
+	if (val < 0) {
+		/* Shouldn't happen, unless neutral val is invalid */
+		val = 0;
+	}
+
+	if (negative) {
```
mlyle, line 957: This kind of tomfoolery erodes our advantages in avoiding 1000.2000 "fake" ranges like BF, etc.  But because the dshot range is not continuous there is not really a super-sane thing to do.

flight/Modules/Actuator/actuator.c
```diff
@@ -761,24 +872,141 @@ static float collective_curve(float const input, float const * curve, uint8_t nu
...
+
 	float max = actuatorSettings.ChannelMax[idx];
 	float min = actuatorSettings.ChannelMin[idx];
 	float neutral = actuatorSettings.ChannelNeutral[idx];
+	float deadband = actuatorSettings.ChannelDeadband[idx] / 2.0f;
 
 	float valueScaled;
-	// Scale
+
+	if (!isfinite(value)) {
```
mlyle, line 984: NAN is now used to signal "don't spin".

flight/Modules/AltitudeHold/altitudehold.c
```diff
@@ -214,11 +214,7 @@ static void altitudeHoldTask(void *parameters)
 			AltitudeHoldDesiredGet(&altitudeHoldDesired);
 			altitude_error = altitudeHoldDesired.Altitude - position_z;
 
-			bool landing = altitudeHoldDesired.Land == ALTITUDEHOLDDESIRED_LAND_TRUE;
-
-			// For landing mode allow throttle to go negative to allow the integrals
-			// to stop winding up
-			const float min_throttle = landing ? -0.1f : 0.0f;
+			const float min_throttle = 0.00001f;
```
mlyle, line 217: This landing mode was very strange--- if you had the stick low, you could trigger landing mode, in which case.. it wouldn't listen to roll/pitch??  Didn't seem like a good idea, and conflicts with good scaling of the stick, so...

flight/Modules/CharacterOSD/charonscreendisplay.c
```diff
@@ -237,7 +238,7 @@ static void update_telemetry(charosd_state_t state)
 	SystemStatsFlightTimeGet(&state->telemetry.system.flight_time);
 	AttitudeActualGet(&state->telemetry.attitude_actual);
 	PositionActualGet(&state->telemetry.position_actual);
-	ManualControlCommandThrottleGet(&state->telemetry.manual.throttle);
+	StabilizationDesiredThrustGet(&state->telemetry.manual.thrust);
```
mlyle, line 241: OSDs are moved to using the "thrust" value which is far more useful (and can represent negative numbers across aircraft types).

flight/Modules/CharacterOSD/panel.c
```diff
@@ -329,7 +329,7 @@ void HORIZON_update(charosd_state_t state, uint8_t x, uint8_t y)
 }
 
 /* Throttle */
-STD_PANEL(THROTTLE, 7, "\x87%d%%", (int)MAX(0, state->telemetry.manual.throttle*100));
+STD_PANEL(THROTTLE, 7, "\x87%d%%", (int)MAX(-99, state->telemetry.manual.thrust*100));
```
mlyle, line 332: Bounded to 3 characters, so -99 .. 100

flight/Modules/ManualControl/failsafe_control.c
```diff
@@ -77,7 +77,7 @@ int32_t failsafe_control_select(bool reset_controller)
 	SystemSettingsAirframeTypeGet(&airframe_type);
 
 	StabilizationDesiredData stabilization_desired;
-	stabilization_desired.Thrust = (airframe_type == SYSTEMSETTINGS_AIRFRAMETYPE_HELICP) ? 0 : -1;
+	stabilization_desired.Thrust = 0;
```
mlyle, line 80: This is now consistent between helis and the rest of aircraft-- no longer is there a "special region" of -1..0 for multirotors.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -71,6 +71,10 @@
 #define RCVR_ACTIVITY_MONITOR_CHANNELS_PER_GROUP 12
 #define RCVR_ACTIVITY_MONITOR_MIN_RANGE 20
 
+// If the neutral point on the throttle channel is 35% over the configured
+// minimum, we assume they're providing bidirectional thrust.
+#define THRUST_BIDIR_THRESH 0.35f
```
mlyle, line 76: We infer what the user wants re: 3D thrust.  If throttle neutral is low, use the old unipolar behavior.  If throttle neutral is in the middle, they want bipolar.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -163,14 +169,80 @@ int32_t transmitter_control_initialize()
...
+
+	if (thrust < 0) {
+		/* Thrust is not intended to be bidirectional, squish the
+		 * bottom half of the range.
+		 */
+		thrust = 0;
+	}
+
+	if (!always_fullrange) {
+		/* TODO/consider: use the value from 3D flight, and remove
```
mlyle, line 194: I think this comment could be misplaced, will look at it.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -163,14 +169,80 @@ int32_t transmitter_control_initialize()
...
+		collective_is_thrust = false;
+		thrust_channel = MANUALCONTROLSETTINGS_CHANNELGROUPS_THROTTLE;
+	}
 
-	return thrust;
+	float minval  = settings.ChannelMin[thrust_channel];
+	float maxval  = settings.ChannelMax[thrust_channel];
+	float neutral = settings.ChannelNeutral[thrust_channel];
+
+	float throt_span = (neutral - minval) / (maxval - minval);
```
mlyle: This does OK with reversed throttle, too, I believe.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1115,25 +1186,20 @@ static void altitude_hold_desired(ManualControlCommandData * cmd, bool flightMod
...
+		const float DEADBAND = altitude_hold_deadband * 0.01f;
 
-		float const thrust_source = get_thrust_source(cmd, airframe_type, true);
+		float const thrust_source = get_thrust_source(cmd, true);
 
 		float climb_rate = 0.0f;
-		if (thrust_source > DEADBAND_HIGH) {
-			climb_rate = expo3((thrust_source - DEADBAND_HIGH) / (1.0f - DEADBAND_HIGH), altitude_hold_expo) *
+
+		if (thrust_source > DEADBAND) {
```
mlyle, line 1200: Unfortunately this deadband stacks with the thrust deadband.

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -1463,12 +1464,16 @@ void render_user_page(OnScreenDisplayPageSettingsData * page)
 
 	// Throttle
 	if (page->Throttle) {
-		ManualControlCommandThrottleGet(&tmp);
-		if (tmp < 0) {
-			tmp = 0;
+		StabilizationDesiredThrustGet(&tmp);
+
+		int throttle = (100 * tmp + 0.5f);
```
mlyle, line 1469: similar to charosd changes.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -520,26 +525,35 @@ static void stabilizationTask(void* parameters)
...
-			if ((this_systime - last_pos_thrust_time) <
+		} else if (last_nonzero_thrust_time) {
+			if ((this_systime - last_nonzero_thrust_time) <
 					1000.0f * settings.LowPowerStabilizationMaxTime) {
-				/* Choose a barely-positive value to trigger
+				/* Choose a barely-nonzero value to trigger
 				 * low-power stabilization in actuator.
 				 */
-				actuatorDesired.Thrust = THROTTLE_EPSILON;
+				if (last_thrust_pos) {
```
mlyle, line 552: The handling of last_thrust_pos is OK but I still occasionally am surprised to see motor thrust reversals on the ground.  It may be possible to improve this.  (Doesn't really -matter-)

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -598,7 +598,7 @@ bool vtol_follower_control_loiter(float dT, float *hold_pos, float *att_adj,
 		// Inverted because we want units in "Down" frame
 		// Doubled to recenter to 1 to -1 scale from 0-1.
 		// loiter_deadband clips appropriately.
-		down_cmd = loiter_deadband(1 - (cmd.Thrust * 2),
+		down_cmd = loiter_deadband(cmd.Thrust,
```
mlyle, line 601: It's nice to not have to recenter this anymore.

flight/PiOS/STM32/pios_servo.c
```diff
@@ -598,6 +599,20 @@ void PIOS_Servo_SetFraction(uint8_t servo, uint16_t fraction,
 	PIOS_Servo_SetRaw(servo, val);
 }
 
+bool PIOS_Servo_IsDshot(uint8_t servo) {
```
mlyle, line 602: Let's us force dshot-style scaling in actuator irrespective of min/max.  (And handle the weird 3D map)

flight/PiOS/STM32/pios_servo.c
```diff
@@ -683,6 +690,12 @@ static int DSHOT_Update()
 			case SYNC_DSHOT_600:
 			case SYNC_DSHOT_1200:
 				break;
+			case SYNC_DSHOT_DMA:
```
mlyle, line 693: Now we treat DMA/non-dma more consistently.  There's still duplicate code that can be eliminated.

flight/PiOS/STM32/pios_servo.c
```diff
@@ -704,7 +715,11 @@ static int DSHOT_Update()
 		PIOS_Assert(info->value < 2048);
 
 		uint16_t message = info->value << 5;
-		/* Don't set telem req bit for now */
+
+		/* Set telem bit on commands */
+		if ((info->value > 0) && (info->value < 48)) {
```
mlyle, line 720: This is a stupid, recent change in blheli.  It doesn't mean to actually send telem in this case-- they send it out at the same time to all channels.  It's uh, another "check bit" on commands, kind of?

flight/targets/aq32/fw/Makefile
```diff
@@ -272,7 +272,6 @@ CSTANDARD = -std=gnu99
 # Flags for C and C++ (arm-elf-gcc/arm-elf-g++)
 
 CFLAGS += -g$(DEBUGF)
```
mlyle, line 274: fast-math is removed everywhere because I've kept finding dangerous things with fast-math, and now we want to use NANs to signal no-spin.

mlyle (commented):
Went through a review pass.

mlyle:

ekaszubski:
@mlyle As of 0752b04 everything appears to work as expected for CPHeli except for collective output while disarmed

Tested:
- [x] Can't spin up motor while disarmed: works
- [x] Motor doesn't spin while armed, zero throttle: works
- [x] Motor spins while armed, non-zero throttle: works
- [x] Cyclic outputs respond to RC input while armed/disarmed: works
- [x] Tail output responds to RC input while armed/disarmed: works
- [ ] Collective output responds to RC input while armed/disarmed: can only move collective output when armed, but works otherwise
- [x] Cyclic outputs respond to vehicle motion while armed/disarmed: works
- [x] Tail output responds to vehicle motion while armed/disarmed: works
- [x] Collective output responds to vehicle motion while armed in Althold mode: works

Config: [d380-42-pr1953.uav.txt](https://github.com/d-ronin/dRonin/files/1425552/d380-42-pr1953.uav.txt)

Regarding arming/disarming:
- We allow servos to move while disarmed, but we do not allow motors to move while disarmed, correct?
- If so, I believe that on all helis, throttle should be labeled as a motor, whether or not the output connects to a nitro throttle servo or an ESC; that way we guarantee that nothing can spin up while the vehicle is disarmed.
- IMO all actuators that can be actuated without generating any forces should be enabled when the vehicle is disarmed.

mlyle:
Throttle was mostly protected correctly even if set as servo, but now should be 100%.

flight/Libraries/math/smoothcontrol.c
glowtape: It seems to work pretty reliable for me in general. The thrust thing before 3D was a simplistic wrapper because before negative was invalid. Not anymore obviously. Your current changes seem to continue to work as expected for normal flight.

In general I'd figure we need to smooth everything we can, otherwise things like the Smith predictor might seem moot. It would be interesting to get people to test this in general, I'd say specifically the psychic mode, which gets you linear interpolation type of smoothness with less of its delay (traded for some overshoot).

flight/Modules/Actuator/actuator.c
```diff
@@ -592,14 +713,11 @@ static void actuator_task(void* parameters)
 	FlightStatusConnectCallbackCtx(UAVObjCbSetFlag, &flight_status_updated);
 	ManualControlCommandConnectCallbackCtx(UAVObjCbSetFlag, &manual_control_cmd_updated);
 
-	// Ensure the initial state of actuators is safe.
-	actuator_settings_updated = false;
-	ActuatorSettingsGet(&actuatorSettings);
+	/* Prime dshot channels with 12 of the 'disarm' command */
+	actuator_send_dshot_command_now(0, 12, 0xffff);
```
glowtape, line 717: This a safety thing?

mlyle: Not really.  dshot ESCs like hearing the 0's a bunch before spinning, so we make sure to explicitly give them.  (for weird always-armed cases, etc...)  Also probably increases the odds in those cases of it locking on as dshot and not some other mode.

mlyle:
Flew "normal" actuators with both symmetric and unipolar throttle input without issue.

mlyle:
Flown 2D (posthrottle) on Colibri race (lux). Attitude, autotune.

LitterBugs:
mlyle, If you are still testing the artifacts from 19 days ago, I have flown 2D on omnibus F3 in Autotune, acro, horizon, and leveling.

mlyle:
@litterbugs -- I'm flying something slightly different-- 2 more commits on this branch, plus the first part of the sensor code change.

Flew on seppuku 2D with unipolar throttle: autotune, attitude, althold

flight/Modules/Actuator/actuator.c
tracernz: This is good. 👍

flight/PiOS/STM32/pios_servo.c
tracernz: Wut.

tracernz (commented):
Haven't read the GCS stuff carefully but it seems sane.

PR #1954, opened 2017-10-01
AQ32: Added OpenLog to UART1

yamahabest:
Per the forum thread: https://forum.dronin.org/forum/d/429-magnetometer-calibration


mlyle:
Thanks.  There's two of these open, I'm going to take the one that has one more thing -- sorry.

yamahabest:
Sure, np.

PR #1955, opened 2017-10-01
Update hwaq32.xml

ewingate:
Added Openlog and TBS Crossfire to UART1 options.


mlyle:
thanks! :)

jenkins, test this please

PR #1958, opened 2017-10-02
GCS:UAVOField: Create q(u)int8 QVariant from int

### LINKED ISSUE #1956: GCS: settings exports are blank?
Reported on the forum, seems to be true.

@tracernz -- any chance this relates to recent GCS uav infrastructure changes?  @pug398 says upgrade/import may be broken too but I haven't tried this.

tracernz:
quint8/qint8 is treat as a char by QVariant, and so QVariant::toString returns the (likely unprintable) char as a string.


mlyle:
Fixes the export half of the problem, but import seems messed up still.

PR #1961, opened 2017-10-04
gcs: output form: make reverse preserve neutral

### LINKED ISSUE #1960: Reverse on output pane breokn for motors

mlyle:
fixes #1960


mlyle:
Tested-- it's happy.

tracernz (approved):
Not the prettiest fix but I'm lazy and want to work on other stuff, and it works so let's take it. :grin:

tracernz:
In future I'd like to factor most of the logic (esp. min/max/neutral/reversing logic) common between InputChannelForm and OutputChannelForm... probably by a superclass, AbstractChannelForm. InputChannelForm and OutputChannelForm can then add the extra columns that are specific to each.

PR #1966, opened 2017-10-06
GCS: Roll Qt to 5.9.2 (LTS)

tracernz:
Plus required changes. Good on Linux, not tested yet on macOS/Windows.

TODO:
- [x] Test macOS
- [ ] Test Windows
- [x] Check Qt install script.. it needs additional packages selected I think

Fixes #1951

PR #1972, opened 2017-10-11
GCS: Qt 5.9.2 follow-up/nits

tracernz:
Follow up to #1966

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -175,7 +175,7 @@ endif
 ifdef WINDOWS
 	$(V1) "$(DL_DIR)/$(QT_SDK_FILE)" \
 		--script $(ROOT_DIR)/make/scripts/qt-install.qs \
-		dr_qt_path="$(QT_SDK_DIR)" \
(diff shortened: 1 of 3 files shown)


make/tools.mk
```diff
@@ -100,6 +100,21 @@ endif
 # Build openocd without FTDI (yes | no)
 OPENOCD_FTDI ?= yes
 
+
+# $1 URL
```
tracernz, line 104: I should use this for all the other downloads in tools.mk but I'm feeling lazy atm :stuck_out_tongue: . The only download if newer part is important for me installing Qt (long story... Qt picks a bad mirror by default..) so I needed to make it work on Windows and decided it would be useful for more than just Qt downloads hence the template.

jihlein:
Just moved to the latest next and can't build on Windows 7 64 bit.  Reloaded all the tools to get the latest versions, but there either is something wrong or I missed a step.  Using the old bash profile I get this error message:

image

And using the new bash profile I get this error:

image

PR #1975, opened 2017-10-19
stabilization: leveling modes: do atti changes in vehicle coords

mlyle:
Also factor cleaner.

Untested except some moments in the sim.

### DIFF
--- a/flight/Libraries/math/coordinate_conversions.c
+++ b/flight/Libraries/math/coordinate_conversions.c
@@ -73,11 +73,35 @@ void Quaternion2RPY(const float q[4], float rpy[3])
 	R23 = 2.0f * (q[2] * q[3] + q[0] * q[1]);
 	R33 = q0s - q1s - q2s + q3s;
 
+	/* If we're nearly nose-up or nose-down, consider yaw to be 0, encode
+	 * information in roll to deal with loss of freedom
+	 */
+	if (R13 >= 0.985f) {
+		if (R13 >= 1) {
+			R13 = 1;
+		}
+
+		rpy[2] = 0;
+		rpy[1] = RAD2DEG * asinf(-R13);
+		rpy[0] = 2 * RAD2DEG * atan2f(q[3], q[0]);
+
+		return;
+	} else if (R13 <= -0.985f) {
+		if (R13 <= -1) {
+			R13 = -1;
+		}
+
+		rpy[2] = 0;
+		rpy[1] = RAD2DEG * asinf(-R13);
+		rpy[0] = -2 * RAD2DEG * atan2f(q[3], q[0]);
+
+
+		return;
+	}
+
 	rpy[1] = RAD2DEG * asinf(-R13);	// pitch always between -pi/2 to pi/2
 	rpy[2] = RAD2DEG * atan2f(R12, R11);
 	rpy[0] = RAD2DEG * atan2f(R23, R33);
-
-	//TODO: consider the cases where |R13| ~= 1, |pitch| ~= pi/2
 }
 
 // ****** find quaternion from roll, pitch, yaw ********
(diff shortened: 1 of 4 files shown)


mlyle:
Flight tested on 450 size on sparky2

mlyle:
Flown as part of 3d branch on Lux

flight/Modules/Stabilization/stabilization.c
```diff
@@ -771,7 +786,7 @@ static void stabilizationTask(void* parameters)
 					// Blend from one rate to another. The maximum of all stick positions is used for the
 					// amount so that when one axis goes completely to rate the other one does too. This
 					// prevents doing flips while one axis tries to stay in attitude mode.
-					rateDesiredAxis[i] = rateDesiredAttitude * (1.0f-horizonRateFraction) + rateDesiredRate * horizonRateFraction;
+					rateDesiredAxis[i] = rateDesiredAttitude * (1.0f-horizon_rate_fraction) + rateDesiredRate * horizon_rate_fraction;
```
tracernz: spacing around - operator

tracernz (commented):
Looks okay. :+1: smoothcontrol api makes this code a lot harder to follow...

flight/Modules/Stabilization/stabilization.c
mlyle: you always bust me for touching lines with existing problems ;) fixing

mlyle:
Rewrote history-- fixed problem plus a couple whitespace things.

flight/Modules/Stabilization/stabilization.c
tracernz: :D

tracernz (approved):

mlyle:
OK, there is a numerical signularity at pitch of greater than 88 degrees, but what it does isn't too bad (judder around a bit), and only really shows up at all in angle/angle/angle mode.  (In angle/angle/rate the loss of control on one axis "wins" and keeps it out of the singularlity region in practice)

mlyle:
Added 3 new commits that should be reviewed, then I think this can safely merge.

mlyle:
(Also filed #1978, which has confused the heck out of me during this process)

mlyle:
Collapsed down those commits to 1

flight/Libraries/math/coordinate_conversions.c
```diff
@@ -73,11 +73,35 @@ void Quaternion2RPY(const float q[4], float rpy[3])
 	R23 = 2.0f * (q[2] * q[3] + q[0] * q[1]);
 	R33 = q0s - q1s - q2s + q3s;
 
+	/* If we're nearly nose-up or nose-down, consider yaw to be 0, encode
+	 * information in roll to deal with loss of freedom
+	 */
```
glowtape, line 78: Wouldn't it be of interest to implement a transition zone, where it fades between presumed and zero yaw based on angle? I mean, it's all an edge case to begin with, but there might be a large yaw error transitioning out of that extreme pitched attitude. If that yaw value is consumed by a controller, there might be an issue?

glowtape:
Hmm? Smoothcontrol shouldn't interfere much. Maybe you mean that raw_input stuff? That's because of some old loose plans to move stick scaling out of ManualControl to Stabilization. Right now it still contains scaled values, though.

flight/Libraries/math/coordinate_conversions.c
mlyle: Doesn't look like it.  Play with the function.  Arguably we should have a higher threshold than .985, but I can't make this be off by more than .8 degrees.  The thing is, once you're >80 degrees nose up, roll almost entirely = -yaw and vice versa

mlyle: Note that this is used for two main things-- A) converting attitude to R/P/Y which is not used by controllers anymore but is used by the PFD in GCS and the OSD, and B) converting an attitude error into R/P/Y errors.

In the case of A, .. there's a singularity there no matter what-- no matter what the solution is numerically unstable.  Tiny movements in orientation can create huge changes in roll & yaw (which are really the same parameter).

In the case of B, what this really means is, "if we can't tell the difference between roll & yaw, start off making the change entirely with roll".  And roll was chosen here because we might not control vehicle yaw.  Eliminating roll when we have no control of yaw would be weird.

mlyle:
https://streamable.com/j19ww

mlyle:
Another exercise of this code: https://streamable.com/wjj7f

mlyle:
merci beaucoup

mlyle:
https://streamable.com/ka3jv is some bails-to-level.  So sedate.

PR #1977, opened 2017-10-20
Flight: Fix debug build

tracernz:
Compiles and links now on everything except Sparky, SPRF3E, and PikoBLX targets (DTFc is fine), which all overflow flash:
Sparky: 15896 bytes
SPRF3E: 19764 bytes
PikoBLX: 7168 bytes

### DIFF
--- a/flight/Libraries/rscode/berlekamp.c
+++ b/flight/Libraries/rscode/berlekamp.c
@@ -248,7 +248,7 @@ Find_Roots (void)
     if (sum == 0) 
       { 
 	ErrorLocs[NErrors] = (255-r); NErrors++; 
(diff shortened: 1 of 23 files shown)


tracernz:
We now use -Og for debug rather than -O0 because -O0 overflows flash on most, if not all, targets.

mlyle:
Thx for whitespace cleanups.

mlyle:

mlyle (approved):

tracernz:
> Thx for whitespace cleanups.

I have my editor setup to do it automatically... but it bloats diff a bit when it actually makes changes to existing stuff.

PR #1980, opened 2017-10-23
python-api: add functionality to canonicalize XML files

mlyle:
Will bring in the actual changed XML when some other PRs commit/merge so there's less fixup.  (Also will require going through by hand to move some `<!--` style comments to field descriptions.

Comparison of uavodefs from generator, between canonicalized version and previous XML: https://gist.github.com/mlyle/c21f9229753ee8454e408bbf5191ff54

In the process, found a number of minor python API bugs which are corrected.


python/dronin/uavo.py
```diff
@@ -9,7 +9,12 @@
 
 import re
 from collections import namedtuple, OrderedDict
-import xml.etree.ElementTree as etree
+
+try:
+    import lxml.etree as etree
+except:
+    import xml.etree.ElementTree as etree
```
tracernz, line 16: ~~Doesn't this need to happen in the opposite order?~~ And `import lxml.etree as etree` can ideally just be `from lxml import etree`

tracernz:
Can you post the xml diff too please?

mlyle:
XML diff is 100% (it re-orders attributes, etc)

mlyle:
Here's an example -- https://gist.github.com/mlyle/df90fc5cc2d904b1fcca4fea1a34c522

1. elementnames/options are now all separate
2. attributes are in alphabetical order
3. indenting is different.
4. default float values are now 0.0 instead of 0 or whatever

mlyle:
This is towards replacing the UAV Object Generator.

Steps are--- A) get everything in a coherent, simpler syntax.  B) build SchemaNG and/or SchemaTron checking on the python API for parse-time validation, and then C) get the python stuff to generate code and substitute templates.

I also want to stick in a manual spell checker etc.

mlyle:
If you want I can put all the generated XML up somewhere.  It looks pretty and results in functionally identical results.

(Main reason I have done this is to confirm the python API can correctly parse all of the UAVO syntax and then reproduce the representation as used, though doing the normalization soon will be nice)

mlyle:
Can this please merge?

python/dronin/uavo.py
tracernz: Also.. should really only except ImportError?

tracernz:
> This is towards replacing the UAV Object Generator.

This would block some other paths I've been thinking of (unless we can embed python in GCS... desktop and mobile). :/

mlyle:
There are both licensing concerns that mean it should get replaced, and we don't want to require qt to build GCS.

If it's just about parsing at runtime, it seems like the UAVO C++ parser needs a complete refactor (and it would be cleaner without being optimized to generate output code and all of these special cases).

tracernz:
> There are both licensing concerns that mean it should get replaced, and we don't want to require qt to build GCS.

Uhhhh..?

> If it's just about parsing at runtime, it seems like the UAVO C++ parser needs a complete refactor (and it would be cleaner without being optimized to generate output code and all of these special cases).

Absolutely agree. 👍

mlyle:
> Uhhhh..?

Sorry, to build firmware, remember how that's been a constant problem (e.g. on pi).

> Absolutely agree. :+1:

Then let's move this way.

PR #1982, opened 2017-10-26
omnibusf3: hwdefs: fix AF mapping for PWM1/2

mlyle:
woops.  swore i'd tested oneshot, but it seems maybe i only tested dshot.


mlyle:
Just scoped and confirmed oneshot output is good after this change.

PR #1986, opened 2017-10-27
Make: Add GCS unit tests to all_ut_run

tracernz:
https://www.worksonmymachine.pro

### DIFF
--- a/ground/gcs/src/app/main.cpp
+++ b/ground/gcs/src/app/main.cpp
@@ -31,6 +31,7 @@
 #define COMMAND_LINE_CLEAN_CONFIG "reset-config"
 #define COMMAND_LINE_NO_LOAD "no-load"
 #define COMMAND_LINE_TEST "test"
+#define COMMAND_LINE_MULTIPLE_OK "multiple-ok"
 #define COMMAND_LINE_PLUGIN_OPTION "plugin-option"
 
 #include "utils/xmlconfig.h"
@@ -262,6 +263,11 @@ int main(int argc, char **argv)
         QStringList() << "p" << COMMAND_LINE_PLUGIN_OPTION,
         QCoreApplication::translate("main", "Passes an option to a plugin."),
         QCoreApplication::translate("main", "option_name=option_value"), "");
+    QCommandLineOption multipleInstancesOption(
+        QStringList() << "m" << COMMAND_LINE_MULTIPLE_OK,
+        QCoreApplication::translate("main", "Allows multiple instances to run at once"));
+    parser.addOption(multipleInstancesOption);
+
     // The options are passed to the plugin init as a QStringList i.e. >> iplugin::initialize(const
     // QStringList &arguments, QString *errorString)
     // These options need to be set in the pluginspec file (look in the Core.pluginspec file for an
@@ -294,9 +300,13 @@ int main(int argc, char **argv)
 
     RunGuard &guard = RunGuard::instance(QStringLiteral(GCS_PROJECT_BRANDING)
                                          + QStringLiteral("_gcs_run_guard"));
+
     if (!guard.tryToRun()) {
         std::cerr << "GCS already running!" << std::endl;
-        return 1;
+
+        if (!parser.isSet(multipleInstancesOption)) {
+            return 1;
+        }
     }
 
     // suppress SSL warnings generated during startup (handy if you want to use QT_FATAL_WARNINGS)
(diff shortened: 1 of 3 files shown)


mlyle:
I think-- at least in CI-- this should just run on Linux-- the chance of things getting stuck or the environment getting tainted seem too high on Windows and maybe MacOSX.

We could either add a new target that only the Linux builder runs, or we could automatically exclude it from all_ut_run on other platforms.

Don't we need to wrapper this in xvfb-run?

I think we want something like timeout 200 -k 2 xvfb-run $(GCS_BIN) -t all

Or maybe actually xvfb-run timeout 200 -k 2 $(GCS_BIN) -t all since i suspect it's GCS that needs the most robust hammer, not xvfb-run.

tracernz:
> We could either add a new target that only the Linux builder runs, or we could automatically exclude it from all_ut_run on other platforms.

Latter is fine.

> Don't we need to wrapper this in xvfb-run?

We don't want xvfb-run except on CI... so it needs to be in the CI invocation rather than the Makefile imo. Unless we look to see if CI env var is defined in the Makefile? (I don't even know if Jenkins defines that.. travis, appveyor etc. all do) In general I think things specific to CI should be in the CI build script rather than the Makefile etc.

> I think we want something like timeout 200 -k 2 xvfb-run $(GCS_BIN) -t all

You are concerned that something outside of the tests will hang (the tests themselves are already protected by timeout)?

mlyle:
> We don't want xvfb-run except on CI... so it needs to be in the CI invocation rather than the Makefile imo. Unless we look to see if CI env var is defined in the Makefile? (I don't even know if Jenkins defines that.. travis, appveyor etc. all do) In general I think things specific to CI should be in the CI build script rather than the Makefile etc.

Hmmm.  I don't really want the UI to pop up and consume a screen when I run the unit tests.  But if you want, we can make the xvfb-run only happen if DISPLAY is not defined.

I don't really want to add xvfb-run to the CI invocation, because it means that if we add other substantive tests we'll be consuming resources the whole time for X.

> You are concerned that something outside of the tests will hang (the tests themselves are already protected by timeout)?

Yah.  I want the best possible hammer.  It's nice that QTest sets up an alarm signal, but I really like the idea of something that progresses to SIGKILL.  Ideally it'd go into its own process group and we could whack the whole process group, but that's tricky.

e.g. if something called by a test sleep()'s it may cancel that alarm signal, depending on libraries etc.

Just about every time I have built a UI CI environment, something has eventually changed, and then I've ended up with hundreds of UI sessions, and ended up poring through process tables to figure out what to kill and how to stop it from happening again.

Another question is whether to pass `-a` to xvfb-run.  If we pass -a, we deal more healthily if a test session manages to get stuck, but it means we can more readily start accumulating processes, too.

mlyle:
I can append this onto your branch if OK

tracernz:
Sure. :+1:

mlyle:
Push'd.  Let's see how it goes.

mlyle:

tracernz:
The fb depth is needed on new arch too. :wink:

mlyle:
I wonder why it's not needed on my 17.04.

tracernz:
Not sure. I know default bash profile on Ubuntu has a _lot_ of stuff in it... maybe that sets it up?

build artifacts please

mlyle:

mlyle:

mlyle:

ground/gcs/src/app/main.cpp
```diff
@@ -294,9 +300,13 @@ int main(int argc, char **argv)
...
                                          + QStringLiteral("_gcs_run_guard"));
+
     if (!guard.tryToRun()) {
         std::cerr << "GCS already running!" << std::endl;
-        return 1;
+
+        if (!parser.isSet(multipleInstancesOption)) {
+            return 1;
+        }
     }
```
tracernz, line 310: What I had done (but not pushed) was just wrap both crash reporter and runguard in the same `if (parser.values(doTestsOption).isEmpty()) {`. This is fine too.

tracernz (commented):
I can't approve my own PR according to Github, but I approve. :+1:

ground/gcs/src/app/main.cpp
mlyle: Sorry that I conflicted.  Might be other reasons to want multi.

PR #1988, opened 2017-11-02
pios_i2c: add asynchronous transfer functionality

mlyle:
A little more is common between F3 and F4 now.


mlyle:
This is dumb, closing.

PR #1989, opened 2017-11-03
sensors: flatten into stabilization

mlyle:
Based upon previous sensor PR.

* Add simple scheduling of sensors based on next_time
* Move lis3mdl to new scheme
* sensors configuration should be flag based/use volatile
* Make simsensors pass through the normal sensor infrastructure
* Collapse sensors into stabilization task


flight/PiOS/Common/pios_sensors.c
```diff
@@ -108,14 +111,37 @@ bool PIOS_SENSORS_GetData(enum pios_sensor_type type, void *buf, int ms_to_wait)
...
+	if (sensor->next_time) {
+		uint32_t now = PIOS_Thread_Systime();
+		int32_t time_until = sensor->next_time - now;
+
+		if (time_until > ms_to_wait) {
+			return false;
+		}
+
+		if (time_until > 0) {
+			PIOS_Thread_Sleep(time_until);
```
tracernz, line 120: Would be a good idea to use PIOS_Thread_Period_Elapsed and make a new PIOS_Thread_Period_Sleep with the same args that sleeps until the period elapses rather than writing this code over and over in various places.

mlyle: Isn't this the opposite of elapsed?  I squinted at elapsed and couldn't figure out how to use it for this when I wrote the code.

tracernz: Perhaps a different API is required? I'd really like to get away from assuming system ticks == 1 ms if at all possible, which pretty much means moving all code handling systemticks into a common library.

mlyle: It's tricky, right?  We can't opaque everything because user code needs to do stuff with time.  Linux uses HZ as a scale factor everywhere and has jiffies and then has a much more complicated partially-opaquified timer api.

There's a few common cases:

1. Do something at interval correcting for runtime (PIOS_Thread_Sleep_Until)
2. Wait a fixed interval (PIOS_Thread_Sleep)
3. See if we have passed a predetermined time (PIOS_Thread_Elapsed)
4. See how much time has elapsed (Subtracting systimes?? PIOS_DELAY_DiffuS?? what about wrap, etc)

This wants 4 and then to do 2 based on it.  We don't have a great solution for 4 in the ticks space, and mixing delays and ticks is problematic.

tracernz (approved):

mlyle:
Rebased.  Need to think about i2c wakeups before this one merges

mlyle:
This is good to go, actually,  I had confused it with the stage 3 PR which i've not yet submitted-- it moves no I2C sensors so it is not blocked by the I2C issue.

PR #1993, opened 2017-11-19
UAVOs: restore limits

mlyle:
I screwed this up -- we had very few limits and I missed it in the diff.

### DIFF
--- a/python/dronin/uavo.py
+++ b/python/dronin/uavo.py
@@ -244,7 +244,7 @@ def canonicalize(subs, fields, is_settings):
                     'type' : field['type']
                     }
 
-            for attr in ( 'units', 'parent', 'limits' ):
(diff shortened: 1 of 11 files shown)


mlyle:
I guess we also lost category, which I'm inclined to remove because we've not really used it

tracernz:
Category is used in the UAVO browser (although not by default).

mlyle:
Yes-- which breaks it into hardware settings and everything else.

Do you really super want it back?  Because I can put it back, but it seems to add a *lot* of complexity to the uavo browser for little benefit.

If we're going to have a ton more UAVOs it might be nice, but we're up against pretty harsh limits on config flash.

tracernz:
No that's fine.

PR #1994, opened 2017-11-19
uavobjectgenerator: cleanup

mlyle:
* line endings
* remove category infrastructure which really complicated the uavobjectbrowser and was almost unused
* remove support for options attribute in the uavobjectgenerator-- in python we need to keep it for awhile because python needs to support old UAVOs.

Sorry that this series of changes was bumpier than expected.


mlyle:
With this one, it seems that when you save something in the UAVOgen it gets stuck blue.  So there's something wrong with turning that off.

tracernz (approved):
Good job on the line-endings. We should really make sure all sources are unix endings only (also add a line-ending to the last line in a few places where it's missing). :+1: 

re:Python, might be nice to have a flag/setting to choose whether to enable legacy stuff.

mlyle:
Rebased.  The highlighting thing was a false alarm/user error.  The right highlight actions still happen.

PR #1995, opened 2017-11-19
actuator annunciators: beep out of dshot ESCs

mlyle:
Based upon #1993 which should go first.

Unfortunately blheli enforces a time of silence between beeps which means this becomes a constant on-off beeping where you can't really hear the morse.

This works surprisingly well, though (almost as loud as real piezo with nicer tone, etc).  If there was just one beep on blheli without the 100ms silence it'd work perfectly.


flight/PiOS/inc/pios_modules.h
```diff
@@ -61,6 +61,13 @@ enum actuator_interlock {
 
 extern volatile enum actuator_interlock actuator_interlock;
 
+/* Ideally would have a better home for this constant */
+#define DSHOT_CMD_BEACON3 (3)
+
+extern int actuator_send_dshot_command(uint8_t cmd_id, uint8_t num_to_send,
+                uint16_t channel_mask);
+
+
```
tracernz, line 70: Ewww, none of this actuator stuff belongs here at all. :cry:

mlyle: We have no header, and worse-- no namespace-- for cross-module calls.  Thankfully we have very few.

A BeepDesired UAVO seems like overkill.

mlyle:
Rebased.

flight/PiOS/inc/pios_modules.h
tracernz: That's because they're supposed to be independent and only communicate through UAVOs (I guess you can think of them as IPC), or system calls in PiOS.

mlyle: Yes, I agree in general, though there are other queues and paths and indirected ways of communications.

The other actuator one should not be via uavo (the actuator-inhibit).  This one is debatable-- I can reframe it that way if you want, though it'll consume resources.

mlyle:
General tree health check

mlyle:
internet blip

flight/PiOS/inc/pios_modules.h
mlyle: So is this PR doomed?

tracernz: No. If we can't work around it nicely I guess it has to be that way. :/

PR #1996, opened 2017-11-22
unbreak build: canonicalize old UAVO XML

mlyle:
This is necessary because of the AndroidGCS  running old stuff through UAVOGen.  Only visible in artifact builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -494,7 +494,12 @@ $$(UAVO_COLLECTION_DIR)/$(1)/uavo-xml: $$(UAVO_COLLECTION_DIR)/$(1)/uavo-xml.tar
 	$$(V1) rm -rf $$@
 	$$(V1) mkdir -p $$@
 	$$(V1) tar -C $$(call toprel, $$@) -xf $$(call toprel, $$<) || rm -rf $$@
+	$$(V1) python/dronin-canonicalize $$(call toprel, $$@)/shared/uavobjectdefinition
+	$$(V1) rm -f $$(call toprel, $$@)/shared/uavobjectdefinition/vtxconfig.xml
+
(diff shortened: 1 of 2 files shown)


mlyle:

Makefile
```diff
@@ -494,8 +494,13 @@ $$(UAVO_COLLECTION_DIR)/$(1)/uavo-xml: $$(UAVO_COLLECTION_DIR)/$(1)/uavo-xml.tar
 	$$(V1) rm -rf $$@
 	$$(V1) mkdir -p $$@
 	$$(V1) tar -C $$(call toprel, $$@) -xf $$(call toprel, $$<) || rm -rf $$@
+	$$(V1) python/dronin-canonicalize $$(call toprel, $$@)/shared/uavobjectdefinition
+	$$(V1) rm -f $$(call toprel, $$@)/shared/uavobjectdefinition/vtxconfig.xml
```
tracernz, line 498: Why rm this particular one?

tracernz: P.S. I see we renamed it... should remove all the underscores and ugliness in the field names while we're breaking it.

mlyle: UAVO has the same name-- but before UAVO names and filenames were unrelated.. canonicalization makes sure the filename is correct.

mlyle: (so wire format, actual stuff is all the same, but we need to remove the "original" file in this case because the canonicalization tool isn't smart enough to).

tracernz: > UAVO has the same name-- but before UAVO names and filenames were unrelated..

Ick. Good job. 👍

mlyle:
Confirmed this fixed Linux.  Mac is still broken because #1996.

PR #1997, opened 2017-11-22
pios_sensors: remove redundant typedef

mlyle:
I guess we were broken in two ways; mac os x sim target didn't like this.


mlyle:

PR #2001, opened 2017-12-05
tools.mk: don't use = with define

mlyle:
GNU Make 3 doesn't support this and is still in wide use

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -104,7 +104,7 @@ OPENOCD_FTDI ?= yes
 # $1 URL
 # $2 save dir
 # $3 save filename
-define DOWNLOAD_template =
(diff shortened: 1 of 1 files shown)


mlyle:

tracernz:
Well... Debian stable, the slowest distro known to man, is at make 4.1... :stuck_out_tongue: Given that I wouldn't agree this is a bug.

tracernz:
Ahh, XCode 8.2.1: https://opensource.apple.com/source/gnumake/gnumake-131/NEWS.auto.html :snail: :snail: :snail:

mlyle:
Yup.  Those the = is mostly superfluous -- it's the default, you'd not usually want :=, and += is a weird thing.

PR #2004, opened 2017-12-07
simsensors: make callback oriented

mlyle:
Huge for performance / behaving properly without realtime.  17% of one
CPU on my laptop, down from 22%.


tracernz (approved):
Nice.

PR #2007, opened 2017-12-07
autotune: factor to periodic item instead of task

mlyle:
Saves resources, etc.  Tested in sim.  Going to bench test (I don't think this requires flight test at this juncture).


mlyle:
Benchtest sequence 1 (seppuku) successful

mlyle:
Benchtest sequence 2 (sparky) successful

tracernz (approved):
Neat, read this twice now and it seems good. 👍

PR #2011, opened 2017-12-08
rgbled: dim when usb power present

### LINKED ISSUE #1740: Decrease brightness of 2812 LED while powered via USB
The 2812 LED is incredibly bright, a fact Englishman is very proud of. This is great except over USB where it is very blinding.

mlyle:
Fixes #1740.  ~~Doesn't help Seppuku, because it doesn't know whether USB is present in firmware.~~  I lied, transfer_possible does the right thing on seppuku.

### DIFF
--- a/flight/Modules/System/rgbleds.c
+++ b/flight/Modules/System/rgbleds.c
@@ -318,6 +318,14 @@ void systemmod_process_rgb_leds(bool led_override, bool led_override_active,
 	RGBLEDSettingsRangeColorBlendModeOptions blend_mode =
 		rgbSettings.RangeColorBlendMode;
 
+#ifdef PIOS_INCLUDE_USB
+	if (rgbSettings.DimWhenUSB == RGBLEDSETTINGS_DIMWHENUSB_TRUE) {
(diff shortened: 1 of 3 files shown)


mlyle:
Should have brought this in long ago.  Slightly tweaked loader to make the light a little dimmer there, too.

tracernz:
Haha, I wrote this last night too, except I did /20 on USB which was much nicer on my eyes. 😁 The LED didn't go bright again after unplugging USB though, which I consider quite acceptable.

mlyle:
I wrote this... 8 months ago, and have held it back because I thought I had to do somethign with the seppuku target.  :P

mlyle:
Moved to 1/3rd brightness with rounding.  It's a tradeoff-- I think it's important that users be able to get their chosen color right when on USB.  So it's still bright enough to leave afterimages in your eye if you stare, but it's a lot less severe, and there's still 600,000 possible colors so hopefully a close approximation to the real color is reached.

tracernz (approved):

PR #2012, opened 2017-12-09
flight: make: dedup lots of makefiles

mlyle:
Specifically the "pios common" kind of stuff.

Minus lines are good.  Getting the base of Makefiles smaller simplifies further effort at dedup.

Mainly preparation to trying to factor STM32F3 & STM32F4 I2C common.

PR #2013, opened 2017-12-10
i2c: refactor common between F3 and F4

mlyle:
This depends upon #2012 .  It's mostly a change in the direction of cleanliness, though some of the F4 state machine stuff is breaking the implicit rules of the past state machine code style to make intention more obvious and ease further cleanup.  (That is, many state entry functions are named by **operation** rather than state, now).

Good test on Sparky and Sparky2.

This common factoring is in preparation for async-I2C-- to allow more complicated solutions without duplicating it in two places.


tracernz (approved):
Looks okay.

mlyle:
Rebased

tracernz:
I'm happy to merge this subject to a little test. I'm thinking external mag or baro might be a good idea (although I don't see any risk that's obviously specific to those scenarios).

tracernz:
Ext mag looks good on the scope (well, the wave rises rather sedately but this is i2c... 😛 ). 👍

PR #2014, opened 2017-12-10
usbhooks: factor common between F1 and F3

mlyle:
Factor common between F1 & F3.  F4 is still its own thing.  Requires #2012.


mlyle:
Rebased.

tracernz (approved):

tracernz:
Any reason F4 can't come in? It's underlying hardware would be more similar to F1 than F1 & F3 are? Although the F446 may be more similar to F3 since it's a newer design?

mlyle:
The underlying hardware, my impression is the actual F4 vs F1 USB libraries are not very similar.  Certainly the current implementation on usbhook is very different, so it takes analysis.  These were almost identical so easy to flatten.

tracernz:
> The underlying hardware, my impression is the actual F4 vs F1 USB libraries are not very similar. Certainly the current implementation on usbhook is very different, so it takes analysis.

Fair enough. 👍 

Tested okay with firmware and bootloader on Sparky and PipXtreme.

PR #2015, opened 2017-12-11
actuator: send dshot commands more times.

mlyle:
If there's lossy dshot wiring, we're much more likely to get a "good
streak" over this increased message count.  Of course, this is still
very bad...


tracernz (approved):
dshot makes me sad 😞

PR #2018, opened 2017-12-23
ChibiOS 17.6.3 integration test branch [no-ci]

glowtape:
Doesn't boot yet.


tracernz:
Did you run into a symbol name collision between ChibiOS gpio and stdperiph gpio? That was one of the issues I hit when previously looking at going to ChibiOS 3 (/2017).

flight/PiOS/ChibiOS/os/common/startup/ARMCMx/compilers/GCC/crt0_v7m.S
```diff
@@ -120,7 +120,7 @@
  * @brief   RAM areas initialization switch.
  */
 #if !defined(CRT0_INIT_RAM_AREAS) || defined(__DOXYGEN__)
-#define CRT0_INIT_RAM_AREAS                 TRUE
+#define CRT0_INIT_RAM_AREAS                 FALSE
```
tracernz: Please don't do this in ChibiOS source... define it somewhere in our sources.

flight/targets/revolution/fw/Makefile
```diff
@@ -245,7 +246,7 @@ CDEFS += -DUSE_$(BOARD)
 ifeq ($(ENABLE_DEBUG_CONSOLE), YES)
 CDEFS += -DPIOS_INCLUDE_DEBUG_CONSOLE
 endif
-CDEFS += -DSTM32F40_41xxx
+CDEFS += -DSTM32F40_41xxx -DSTM32F405xx
```
tracernz, line 249: Perhaps you can just define STM32F405xx... check CMSIS and StdPeriph to see what it does re: the legacy STM32F40_41xxx.

flight/PiOS/ChibiOS/os/hal/ports/STM32/STM32F4xx/stm32_registry.h
```diff
@@ -56,7 +56,7 @@
 
 #elif defined(STM32F405xx) || defined(STM32F415xx) ||                       \
       defined(STM32F407xx) || defined(STM32F417xx)
-#define STM32F40_41xxx
+// #define STM32F40_41xxx
```
tracernz: Defining STM32F405xx would be _strongly_ preferable to modifying Chibi sources.

flight/PiOS/ChibiOS/os/common/startup/ARMCMx/compilers/GCC/mk/startup_stm32f4xx.mk
```diff
@@ -6,7 +6,9 @@ STARTUPASM = $(CHIBIOS)/os/common/startup/ARMCMx/compilers/GCC/crt0_v7m.S
 
 STARTUPINC = $(CHIBIOS)/os/common/startup/ARMCMx/compilers/GCC \
              $(CHIBIOS)/os/common/startup/ARMCMx/devices/STM32F4xx \
-             $(CHIBIOS)/os/common/ext/CMSIS/include \
-             $(CHIBIOS)/os/common/ext/CMSIS/ST/STM32F4xx
+             $(PIOS_DEVLIB)/Libraries/CMSIS/Device/ST/STM32F4xx \
+             $(PIOSCOMMON)/Libraries/CMSIS/Include
+#             $(CHIBIOS)/os/common/ext/CMSIS/include \
+#             $(CHIBIOS)/os/common/ext/CMSIS/ST/STM32F4xx
```
tracernz: This is where you strike problems with upgrading ChibiOS. Their CMSIS impl. doesn't seem compatible with StdPeriph etc. and vice-versa.

flight/PiOS/ChibiOS/os/common/startup/ARMCMx/compilers/GCC/crt0_v7m.S
mlyle: I think this is chconf kinda stuff?

tracernz: Yep, although I can't find that in boardinfo or fw... did we get rid of it?

mlyle: PiOS/STM32/inc/chconf.h (a per-target one takes precedence but isn't necessary)

tracernz: Oh yep. That's good. :+1:

glowtape: OK, I'm gonna put it there. In regards to chconf.h, If I'm getting this to work, I'll probably define a bunch of macros so that PiOS compiles both on ChibiOS 2.6.6 on F3 and below and 17.6.3 on F4, and eventually F7, if we go there. SPRF3e is probably gonna blow again with new ChibiOS.

flight/PiOS/ChibiOS/os/hal/ports/STM32/STM32F4xx/stm32_registry.h
glowtape: That one stems from STM32F40_41xxx being defined in the Revo makefile. Seppuku curiously doesn't. Should probably remove the one in the Makefile.

flight/PiOS/ChibiOS/os/common/startup/ARMCMx/compilers/GCC/crt0_v7m.S
tracernz: Why use a different version on F3 and below?

flight/PiOS/ChibiOS/os/hal/ports/STM32/STM32F4xx/stm32_registry.h
tracernz: Yes

flight/PiOS/ChibiOS/os/common/startup/ARMCMx/compilers/GCC/crt0_v7m.S
glowtape: I'm gonna give it a try, but SPRF3e is really damn short on space in the firmware image. Chances are that it breaks compilation again. Plus testing PipXtreme and FlyingPio is kind of an issue for me. :)

glowtape:
OOOOOKKK...

I was including the GCC rules.mk file, assuming its necessity, which caused drama with the CORTEX_USE_FPU flag getting redefined, if I would keep using it in chconf.h like before. So I ditched it from chconf.h. That however created a wonky state where sometimes it is defined, and sometimes not. Apparently there's parts that do autoconfigure to the MCU, and some expect explicitely an USE_FPU flag set to yes. Which I didn't. The result was that the context switching assembly bit assumed them to be there, but not in the struct defined in another header. So during context switching, it was vpop-ing more registers than there were in the struct, and happened to hit the 0x55555555 stack fill and stick it into the program counter.

I've removed the offending rules.mk file, which seems unnecessary, because all necessary flags seem to get defined within our own makefiles, and things still work. Old ChibiOS apparently has one of these, too, which I found goes unused, too.

I have my Seppuku and Revo booting up just fine now. When I get time (prob. this weekend), I'm gonna start over with the port to get a clean one, also fixing previous points you both mentioned.

PR #2019, opened 2017-12-24
autotune: Calculate true yaw PIDs

mlyle:
Previously we had the 'iceetune' fudge because yaw beta calculation was
poor.  Now we still use the roll/pitch gains to calculate the filter
property but actually calculate pids based on yaw beta.

This has not been flown, and this needs to be evaluated carefully.

### DIFF
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -666,7 +666,7 @@ void AutotuneSlidersPage::compute()
         tuneState->outerKi = 0;
     }
 
-    for (int i = 0; i < 2; i++) {
+    for (int i = 0; i < 3; i++) {
         double beta = exp(tuneState->beta[i]);
 
         double ki;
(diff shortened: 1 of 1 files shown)


tracernz (approved):
Code looks good. Just need some people to fly it and report results... In 20/20 hindsight it would have been smart to have this in the uploaded JSON always. I still think that's a good idea even if the values aren't actually used.

P.S. not sure I would class this as a bug, but rather a feature request. 😛

ufoDziner:

ufoDziner:
Completed an autotune and did an initial LoS flight with success.  Will follow up with FPV results tomorrow.

tracernz:
Tempted to merge this and see if we get any negative feedback from upcoming testing cycle.

mlyle:
It's daring, but eh... why not?

ufoDziner:
Sorry for not following up.  I ran at least 10 packs with this and had no issues.  Flew well.

jhitesma:
Dusted off my hex to try @glowtape's LQG on it and noticed I got unusable yaw tune results.  Just dropped back to Wired to confirm that it's happening there as well.  

Previously I had no problems with AT yaw values on the hex, but now I'm getting results that are ridiculously high:

https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAmLyUgwoM

(and the two previous tunes as well)

Is it possible that this broke yaw AT for hexes?

mlyle:
Probably not for hexes, but perhaps for some aircraft (very small props?)

mlyle:
Can you please open an issue with the aircraft configuration so it doesn't get lost here

PR #2024, opened 2018-01-02
GCS: Use nullptr for null pointers

tracernz:
We require c++11 compilers so there's no reason not to go all nullptr. GCS is quiet lately so good time to avoid merge conflicts with this. It surprises me how small this diff actually is.

Replacements made by clang-tidy. Note that I didn't make replacements in ground/gcs/src/libs for libraries that we track upstream as it seemed pointless.


mlyle:
lgtm

PR #2026, opened 2018-01-08
osd_menu improvements plus added battery sub-menu

### LINKED ISSUE #1304: OSD: Adjust Battery Size from Menu
It is common to fly different sizes in the field.

c.f. https://forum.dronin.org/forum/d/141-osd-menu-add-ability-to-adjust-battery-mah

droney:
added battery settings menu; 
removed redundant 'if' statement in menu_fsm_event enum; 
no longer displays acro insanity setting out to six decimal places; 
no longer immediately returns to sub-menu if roll right held too long when exiting sub-menu; 
eliminated all _IDLE states which now immediately displays cursor in sub-menu without the pitch down; 
pidrate menu more closely matches GCS for I term upper limit and D term increment value;

edit by mpl: fixes #1304

### DIFF
--- a/flight/Modules/OnScreenDisplay/osd_menu.c
+++ b/flight/Modules/OnScreenDisplay/osd_menu.c
@@ -36,6 +36,8 @@
 #include "onscreendisplay.h"
 
 #include "flightstatus.h"
+#include "flightbatterystate.h"
+#include "flightbatterysettings.h"
 #include "homelocation.h"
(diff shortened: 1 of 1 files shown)


droney:
you know, it would be really nice to also have a camera angle setting capability. Would it be OK if I added that as well to one of the existing sub-menus? Perhaps under 'Flight Modes'???

mlyle:
LGTM

Issue #1879, opened 2017-07-09
Flight: Reclaim Init Task Stack


tracernz:
Memory is a bit tight on some targets. We could gain back 1 kiB by locating the init task stack adjacent to the end of the heap (would do this in linker script rather than mallocing it at runtime) and adding it to the heap when the init task deletes itself.

Questionable whether we could actually make use of it since modules would have started by then.

mlyle:
System task could be taught to start the rest of the modules tasks--- which would also have the effect of providing stronger ordering so we don't do config checks prematurely etc.

tracernz:
The only problem I see with that is that System is no longer a module, but an integral part of PiOS..

mlyle:
Yah, it's slightly unfortunate.  OTOH, main.c (and module start) just got factored out of targets in Neat.. having system call something to complete startup doesn't seem **too** unclean.

System task also does stuff like checks the malloc failed stuff during startup to do the boot fault things, and connects to object persistence-- which really should have a defined order with all the rest of "stuff" (otherwise some of these things could go off in the wrong order, e.g. someone tries to save something before System is listening).

tracernz:
It's not all that different to CorePlugin in GCS :stuck_out_tongue: (although the PluginManager does load and start all the plugins).

glowtape:
Any particular reason why we shouldn't just do it in main() and skip initTask? IIRC, from what I have seen with the ChibiOS patch, because we allow an idle thread, main() gets turned into a thread after chSysInit(). There's 1KB stack reserved for the main thread, anyway.

tracernz:
There would be no way to get that memory back (it would need to be quite a bit more than 1 k iirc). Afaik the os requires that thread to always exist. This only seems important on F3 (only 40 k) and PipX (20 k).

glowtape:
Hm, what do you mean get memory back? The process stack that gets used for the main thread is statically allocated in the linker script currently. So it's gone anyway.

The patch referred above skips the initTask thread and runs the stuff directly on main. My Seppuku quad fires up battery monitoring, logging, altitude hold, MSP bridge and Crossfire telemetry next to the necessary flight modules. The lowest I've seen unused process stack being reported after a bunch of reboots was 408 bytes, seems there's plenty of headroom for annoying module inits. We don't really need to spawn a separate thread for initTask.

[...]

PR #2027, opened 2018-01-09
pios: Initialize modules/tasks in main.

glowtape:
initTask seems redundant, considering the main function will be turned into a thread within chSysInit. Both have the same amount of stack allocated.

Ref: https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/Common/Libraries/ChibiOS/os/kernel/src/chsys.c#L103
 
--edit: Revo and SPRF3e boot up just fine with that. Should solve #1879.

### DIFF
--- a/flight/PiOS/STM32/chibi_main.c
+++ b/flight/PiOS/STM32/chibi_main.c
@@ -33,16 +33,10 @@
 #include "systemmod.h"
 #include "pios_thread.h"
 
-/* Local Variables */
-#define INIT_TASK_PRIORITY	PIOS_THREAD_PRIO_HIGHEST
-#define INIT_TASK_STACK		1024
-
 /* Function Prototypes */
 extern void PIOS_Board_Init(void);
 extern void Stack_Change(void);
-static void initTask(void *parameters);
-
-static struct pios_thread *initTaskHandle;
+static void initTask(void);
 
 /**
  * @brief   Early initialization code.
@@ -82,10 +76,7 @@ int main()
 	/* Brings up System using CMSIS functions, enables the LEDs. */
 	PIOS_SYS_Init();
 	
-	/* For Sparky2 we use an RTOS task to bring up the system so we can */
-	/* always rely on an RTOS primitive */	
-	initTaskHandle = PIOS_Thread_Create(initTask, "init", INIT_TASK_STACK, NULL, INIT_TASK_PRIORITY);
-	PIOS_Assert(initTaskHandle != NULL);
+	initTask();
 
 	while(1)
 		PIOS_Thread_Sleep(PIOS_THREAD_TIMEOUT_MAX);
@@ -129,7 +120,7 @@ void check_bor()
  *
  * Runs board and module initialisation, then terminates.
  */
-void initTask(void *parameters)
+void initTask(void)
 {
 	/* Ensure BOR is programmed sane */
 	check_bor();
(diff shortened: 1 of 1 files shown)


glowtape:
As per discussion in IRC, we ought to rename chibi_main eventually, if it's function wasn't supposed to be ChibiOS specific.

flight/PiOS/STM32/chibi_main.c
```diff
@@ -82,10 +84,14 @@ int main()
 	/* Brings up System using CMSIS functions, enables the LEDs. */
 	PIOS_SYS_Init();
 	
+#if !defined(PIOS_INCLUDE_CHIBIOS)
 	/* For Sparky2 we use an RTOS task to bring up the system so we can */
```
mlyle: Need to kill this comment

glowtape:
Removed comment and rebased.

mlyle:
OK.  This has tested sane without new chibios?  I can queue up a test sequence of this.

glowtape:
Yeah, I wrote this patch separately from it.

mlyle:
Tested sparky2

flight/PiOS/STM32/chibi_main.c
```diff
@@ -34,16 +34,18 @@
 #include "pios_thread.h"
 
 /* Local Variables */
+#if !defined(PIOS_INCLUDE_CHIBIOS)
```
mlyle: There's no non-chibios tasker on STM32.. I think all these can go away..

mlyle:
Tested dtfc

mlyle:
Tested pipxtreme/oplink

mlyle:
OK, so I have one small requested change (and I'm willing to be talked out of it) and @tracernz has self-requested review... otherwise I think this one is ready for merge.

flight/PiOS/STM32/chibi_main.c
tracernz: If the file is named chibi_main.c it makes sense to take them out. Alternately if it were named stm32_main.c they should stay. We can fairly easily do that in future if we bring in another OS again so agree these can go.

flight/PiOS/STM32/chibi_main.c
```diff
@@ -82,10 +84,12 @@ int main()
...
-	/* always rely on an RTOS primitive */	
+#if !defined(PIOS_INCLUDE_CHIBIOS)
 	initTaskHandle = PIOS_Thread_Create(initTask, "init", INIT_TASK_STACK, NULL, INIT_TASK_PRIORITY);
 	PIOS_Assert(initTaskHandle != NULL);
+#else
+	initTask(NULL);
+#endif
 
 	while(1)
 		PIOS_Thread_Sleep(PIOS_THREAD_TIMEOUT_MAX);
```
tracernz: Maybe we can do system mod stuff in this thread in future. It's already not a normal module.

tracernz (approved):

flight/PiOS/STM32/chibi_main.c
mlyle: Yup we have our plans for that in #1879, so that stacked with this would double our savings.

glowtape:
The #ifdef is gone, and also removed a straggling PIOS_Thread_Delete() I forgot. It needs to go anyway, if the systemmod stuff gets tied into this. Tested it for giggles, still the same.

Issue #1974, opened 2017-10-13
smoothcontrol: fixes needed


mlyle:
* operates on failsafe
* incorrect bounds for non-rate modes
* bounds aren't used

glowtape:
Oh hey, I ought to check the issues more often. Gonna look into this when I get up later today after my shift.

mlyle:
Need to audit this myself to close (with #1950 merged it has a chance of being OK)

mlyle:
OK

PR #2028, opened 2018-01-09
smoothcontrol,stabilization: Do bounds checking on the smoothed output.

glowtape:
I'm a smidge nervous about this, because I can't test all modes (yet) to see whether I inferred all stick scales correctly. Playing with rate and attitude mode, and switching back and forth between them, in my living room, the quad acts normal for whatever I can test in that space.

Also, turns out the axis reset happened late, which might have led to actuator blipping for one gyro update cycle under certain conditions (switching from high rates to attitude or a mode with -1..1 stick scale).
 
This deals with #1974.

### DIFF
--- a/flight/Libraries/math/smoothcontrol.c
+++ b/flight/Libraries/math/smoothcontrol.c
@@ -88,7 +88,7 @@ void smoothcontrol_set_mode(smoothcontrol_state state, uint8_t axis_num, uint8_t
 }
 
 // Processes the signal, if internal state allows it.
-void smoothcontrol_run(smoothcontrol_state state, uint8_t axis_num, float *new_signal, float limit)
+void smoothcontrol_run(smoothcontrol_state state, uint8_t axis_num, float *new_signal)
 {
 	PIOS_Assert(state && axis_num <= 3);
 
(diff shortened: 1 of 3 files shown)


glowtape:
Relies on #2029 for the HEX build to not blow up.

mlyle:
We have an existing set of bounding calls that should be used for this, probably.

But --- doesn't stabilization do existing bounding?  Can it be changed to do the bounding before/after (or just after, if this can be justified) smoothcontrol?

mlyle:
jenkins, test this please

glowtape:
From what I can tell, mostly just actual rates of the inner loops get bounded, and attitude in calculate_attitude_error. Anything that assumes -1..1 stick scale just expects it to be that. I can move the bound checks to stab.c. Makes probably more sense.

glowtape:
Alternatively, also just finally yank stick scaling from ManualControlCommand. It looks like it happens all over there in scale_stabilization(), which feeds into StabDesired and basta. It *looks* like it ought to be trivial.

mlyle:
Note that we need to be able to accept real-unit commands, though, because e.g. navigation spits them out.

tracernz:
> Note that we need to be able to accept real-unit commands, though, because e.g. navigation spits them out.

And we should always do this where possible in any case.

mlyle:
@tracernz --- I'm not sure.

Right now stabilization needs to do -some- stick scaling for e.g. horizon, coordinated flight, etc.  It is really bad to have stick scaling in both ManualControl and Stabilization.  @ekaszubski and I were talking about this before, and there was discussion of having the module driving StabilizationDesired providing units-- real world units, or "effort units".  ManualControl could send effort units and then the scaling and smoothing logic could be activated.

glowtape:
Gonna stick parts of this into #1950.

PR #2030, opened 2018-01-14
telemetry: further improvements

mlyle:
* Improved ground-side constants
* Fixed a dumb bug in F3 HID handling
* Increase flight-side buffers for received data a little bit
* The big one: add a request queue, ensure that requests take priority over all things telemetered.

This has behaved well in both local and radio tests-- before F3 doing weird things and objects being lost was 1 in 3 or 1 in 4 of connections, but now in dozens of tests have not been able to repro.  There is some remaining issue that does cause occasional loss of frames and re-requests, but never significant numbers of the permitted retries.

I also saw an anomaly with the RFM bind wizard once (where the FC & the oplink ended up set to different max link speeds), but in all subsequent tests it has been fine.

### DIFF
--- a/ground/gcs/src/plugins/uavtalk/telemetrymonitor.cpp
+++ b/ground/gcs/src/plugins/uavtalk/telemetrymonitor.cpp
@@ -44,9 +44,9 @@
 #define SESSION_RETRIEVE_TIMEOUT 20000
 // Number of retries for the session object fetching during negotiation
 #define SESSION_OBJ_RETRIEVE_RETRIES 3
-// Timeout for the object fetching fase, the system will stop fetching objects and emit connected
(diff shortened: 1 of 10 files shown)


ground/gcs/src/plugins/uavtalk/telemetrymonitor.cpp
```diff
@@ -427,6 +427,9 @@ void TelemetryMonitor::sessionObjUnpackedCB(UAVObject *obj)
 
 void TelemetryMonitor::objectRetrieveTimeoutCB()
 {
+    qInfo() <<
```
tracernz, line 430: Nice. 👍

tracernz (approved):
Very good work. 😁

mlyle:
Thx

PR #2034, opened 2018-01-19
GCS: Allow calibration of mag on boards without internal mag

### LINKED ISSUE #1896: BrainRE1 calibrate mag doesn't work

tracernz:
To determine if there is a mag, we check if the Magnetometer object is present, and if so, whether any of the fields have been changed since boot (non-default). Also fixed a memory leak.

Supersedes #1908 
Fixes #1896


mlyle:
LGTM

PR #2036, opened 2018-01-20
gcs,input: Always Armed isn't an option on flight anymore. Remove UI.

glowtape:
I figure it was removed with the rewrite of the arming state machine stuff a while ago.


mlyle:
Thx

PR #2037, opened 2018-01-21
Add LQG controller for rate mode and as attitude inner loop

glowtape:
Posted for @ufoDziner and anyone else that is brave.

LQG stabilization mode added. Need to configure a Stabilized1/2/3 with LQG per axis, it isn't added as flight mode currently. It is a rate controller only. Tried LQG/LQG/Rate outside and LQG/LQG/LQG inside, both work.

The LQR state weighting matrix is currently fixed to experimental values. My test platform is a 5" miniquad that currently tunes to 10.6-ish beta and ~14ms tau, and it gets a proper solution there. Anything similar to that should fly somewhat OK I'd figure.

I touched the ManualControlSettings UAVO, so if you flash it and boot it, the settings are gone, so export them before hand to reimport them later.

--edit: Also, you need to run autotune, so that SystemIdent fields get filled in. Adjusting Previous Tune is OK, if nothing changed on the quad.

--edit: Might also be an idea to smooth the RC signal. Done by setting via UAVO browser, StabilizationSettings.RCControlSmoothing.Axes/Thrust to Extended (setting I've been testing with.).


glowtape:
jenkins, build this, please

glowtape:
jenkins, build this, please

glowtape:

glowtape:
Current findings thanks to @ufoDziner is that yaw doesn't work entirely right yet. On my quad it seems OK, on his new 6" one, it goes apeshit. Probably a tuning issue setting Q2 on LQR.

glowtape:

glowtape:
OK, I made it somewhat configurable via UAVOs and dicked around some more in MATLAB, and a tiny bit outside. But I changed the defaults since the last artifact.

@ufoDziner Whenever you've a bout of boredom and want to test this one, the LQG configuration changed from the defaults in the build. Here's the current bunch of settings, which need to be changed in the UAVO browser:

LQGSettings.RTKF.R = 250
LQGSettings.RTKF.Q1 = 0.1
LQGSettings.RTKF.Q2 = 0.00001 (should say 1e-05 after hitting ENTER)
LQGSettings.RTKF.Q3 = 0.00000005 (should say 5e-08 after hitting ENTER)

LQGSettings.LQR.Q1 = 0.0001
LQGSettings.LQR.Q2 = 0.0001
LQGSettings.LQR.YawBetaOffset = -0.5
LQGSettings.LQR.YawQ2Coeff = 0.75

SensorSettings.LowpassOrder = 0

🎉 🎉 🎉  **Attention:** Don't fly any other mode with LowpassOrder to 0, it'll rattle your motors to pieces. That includes LQG/LQG/Rate. Means, take off in LQG/LQG/LQG, and if it/yaw doesn't rattle the quad like last time, maybe do a bunch of simple manoeuvers (maybe even some flips and punchouts), and then land in LQG. If it rattles anyway right after take-off, abort, I guess. Then hook it up to the GCS and set LowpassOrder back to 1, so you don't get surprised the next time.

I have to create a sidechannel for LQG to get the raw gyro data directly.

If you want a stronger rate response, you can raise LQGSettings.LQR.Q1 to 0.0002 or 0.0004.

glowtape:

ufoDziner:
Feels really smooth.  Seems to me that it stops flips more accurately.  Also tested with LQGSettings.LQR.Q1 to 0.00005 and Q2 to 0.001.  Didn't really notice much difference good or bad.

glowtape:
I think I'm going to patch in a gyro filter bypass later today. I'm just sitting at work bored right now and latently figured that your logs seem to be relatively devoid of noise on the regular gyro readings, which they shouldn't. Either you have a super smooth quad, or the gyro filter was accidentally on. Stacking delays probably doesn't help.

Also holy shit, GitHub mobile is terrible.

glowtape:
Also, with the bypass you could leave the filter running, so that it's safe to switch between LQG and acro mid-flight. Will probably help better for direct comparison.

ufoDziner:
Keep in mind that my FC is mounted on rubber bobbins.  They seem to smooth things out a ton.

glowtape:
Wow, that seems like they do take out a lot of the noise pretty effectively then. On my logs, the gyros look noisy as hell with hard mounting. Now I'm wondering whether the bobbins have an effect on the bounce. However instead of taking them out, dropping the R value by a lot is an option. Like this, it'll start relying way more on the actual gyro than the internal model. I've already suggested R=100 on IRC, maybe 25 or less is an option. I have to run these through MATLAB this afternoon.

ufoDziner:
I changed to hard mounts on a different build to test if the bobbins had an effect.  Didn't seem to stop the flips any more accurately with the hard mounts.  I haven't seen any negative effects with the bobbins.

glowtape:
Okay. Dropping R is however still worthwhile to explore, if your quad is relatively noise free regardless, to correct mispredicts. In the logs, the estimate of the RTKF got ahead of the gyro reading on the huge swings of the flips, both starting and stopping. It seems to think that your quad has more power than it actually reacts with.

That said, I have to figure out why it is happening for you.

glowtape:

ufoDziner:
This was reported via IRC, but figured I should add it here as well.  The biggest thing I noticed is that it's like the quad is in a soft leveling mode.  It tries to level itself sooner than in acro/rate mode.  Also the roll/pitch rates seem a bit more linear in acro, whereas this seems to have a steeper expo curve.  Still flew smoothly though.

mlyle:
@ufoDziner 

> The biggest thing I noticed is that it's like the quad is in a soft leveling mode. It tries to level itself sooner than in acro/rate mode

is CG low on the vehicle (e.g. do you have an underslung battery?)

If it's tending towards self-leveling with a low battery, it means that long term drift isn't being combated effectively.

ufoDziner:
@mlyle 
Yes.  Underslung battery.  Even in acro it tries to level, it's just much more prominent with this.

glowtape:

glowtape:
Updated the LQG default settings based on experimentation. The current set is:

LQGSettings.RTKF.R = 100
LQGSettings.RTKF.Q1 = 1
LQGSettings.RTKF.Q2 = 0.0002
LQGSettings.RTKF.Q3 = 0.0001

These changes will make sure the actual gyro will be considered more. Maybe this helps reducing the wandering (I have not yet had it, so I'm guessing here). Especially the process noise for bias got raised significantly, which should help more fighting the autoleveling.

(The problem with the RTKF is, there's a point where the estimates get too noisy and cause more drama than necessary, due to how aggressive the LQR can react to it.)

LQGSettings.LQR.Q1 = 0.0001
LQGSettings.LQR.Q2 = 0.001
LQGSettings.LQR.YawBetaOffset = 0
LQGSettings.LQR.YawQ2Coeff = 1

These changes for one disable the yaw fudging that was introduced because our autotune doesn't evaluate a direct torque beta that the original math wants. Seems to work fine on my quad, and still causes less of a ruckus than #2019. Also, the Q2 weight got raised, which will try to drive torque to zero some more, but not so much that it impacts performance noticeably. It should also keep the yaw axis under control some more.

(I was having a bunch of roll axis oscillations, but they could be easily reproduced with the PID controller in the same situation, so it seems the axis estimates aren't that correct.)

Not spinning a new artifact, since these are UAVO changes only.

I was considering LQI, but the matrices go from 2x2 to 4x4, which raises the computation time way too much. And it looks like it doesn't really do much more than what the RTKF spits out.

glowtape:
Separate sensor noise value for the yaw axis. Allows to drop R for pitch and roll considerably, so that even more gyro applies to the estimate.

glowtape:

glowtape:
So about LQG as inner loop for attitude. This one is tricky.

Situation is this, I've the default 55° on roll and my P-term on the outer loop is 7.24, and the I-term is 4. Initially, assuming I hammer the stick fast enough, the quad is supposed to reach around 400°/s. The main actuation from the inner loop seems like like 50ms-ish, so I figure the outer I-term doesn't do anything worthwhile.

The initial impulse from both loops seems to be just as strong with the current settings, the slopes overlap pretty nicely.

However, the PID-loop actually reaches the 400°/s, whereas the LQG-loop only goes to around 320-330°/s.

**But**... Considering the AttitudeActual starts moving, the proportional term starts pulling back. Around when the LQG stops accelerating, the attitude is at least 5° from level (should have screenshot it), maybe more since it may be a smidge behind due to logging AttitudeActual at only half the logging rate here currently (assuming it is a reason for delay, anyway). Anyhow, when LQG stops accelerating, the rate demand would be around 360°/s or a little less, because the error would be 50° or a little less instead of 55°.

On the first bit, PID is a tiny bit faster at reaching the attitude, then it crosses over and LQG keeps reaching target attitude faster. The argument that could be made that LQG is more accurate at following the outer loop's rate demands, while PID rides on overshoot at the very beginning, but then it swings back and hinders the outer loop.

Two things, I did hammer the stick, not put it on momentary switch, and second, it was windy.

White is PID, yellow is LQG, top gyro, bottom attitude. Tried to find events matching as close as possible. The LQG pulls back sooner compared to PID every time, on either side of the axis.
attitudelqg

glowtape:

ufoDziner:
Flew a couple of packs on this build today with a Bolt RC Kraken Worx 5".   Wandering is gone.  It's more floaty than PID in turns.  Also, in high speed high power turns there is a high frequency oscillation.  The LQG control rates also feel a bit more linear, so they're more aggressive near the center stick than I prefer.  Major improvement in flyability over the last build.  Still needs work for sure, but it's definitely going the right direction.  Great work!

glowtape:
Added RTKFEstimate UAVO, which dumps what the estimator thinks. Adjusted defaults some more after testing today.

Now for the fun part, this is gonna be long.

I've noticed the motors going kind of gung-ho on aggressive maneuvers, some sort of grindy noise, which I always thought was noise caused by aggressive braking and acceleration (to be honest, after further testing, it sounds almost exactly like the yaw rattle on low throttle drops). Attitude on LQG testing kind of made me rethink that.

The estimator works like that: Update state estimate based on gain matrix, then add control input _from the last loop cycle_ to it.

The wrinkle in this is that RC updates at a fraction of the rate than the stabilization loop runs, so RC commands look pretty much like a stair to the control law. If you swing the stick fast enough, the steps are pretty damn high, and in return, the initial error on value change is just as high. This causes the control law to demand instantly a huge change from the actuators. When you log ActuatorDesired, and the command rate is low enough or the logging rate is high enough, you can actually see it manifest as saw tooth pattern on the PID. Means it is pulsing the motors speeds, so long prop inertia allows it.

Now that the estimator and the LQR are in the loop, the unit delay causes some issues. The huge setpoint change creates a huge control demand, because of the rate error, to which the first LQR coefficient reacts. We assume nothing much is happening on torque (yet/anymore). Then the next loop, the estimator gets updated with the huge control demand, which creates a huge peak in estimated torque, to which the second coefficient of the LQR suddenly reacts and creates a huge opposing control demand (torque is always driven to zero, so it counters the rate part). That demand again influences the estimate, which again futzes with the controller response. Creating some sort of short feedback loop that dies down relatively quickly.

The torque is pretty much a differential term just like the D in PID. On the PID controller, we filter the D-term to avoid similar issues. While it's probably worthwhile to investigate, I'm not too fond of sticking a filter on the relevant part of the LQR, because both coefficients seem relatively finetuned to each other. The Kalman filter itself has a filtering function, but solely on the gyro input. It also tries to predict the rate change on estimated torque. The control input is what upsets it all, because it adds to it unfiltered.

The alternative is smoothing the control input. There's been a version of it in the firmware already, but deactivated by default. That version however is still stair-steppy, however literally takes the edge off of the outer corners of the stairs, by chamfering them. That however doesn't seem enough. No smoothing, normal and extended all cause the rattling motors here.

I've added most of #1950 to this, and with either Linear or Psychic mode, the rattling is pretty much gone. Either when doing huge stick swings in Attitude on LQG, and as well as entering and exiting flips and such in LQG acro.

I kind of suspect that the unsmoothed control input potentially also causes drama during hard FFF maneuvers, because the motors are getting held up in short bursts of grinding instead of properly doing their job, when doing fast and large stick motions.

I've enabled Psychic by default in this PR, because it served me pretty well in Summer 2017 with no noticeable issues. It uses the sort of predicts the existing Extended mode creates, but then interpolates between them. The predicts are to lower delay by the interpolation. It has theoretically the capability to overshoot, but from all logs I've looked at, human aren't able to swing or stop the sticks fast enough to create a hard edge, so this shouldn't be an issue. Linear mode just interpolates by trailing the control signal, this is the safest option, if you feel funny about the other mode.

@ufoDziner If you're going to install this, export the UAVO before flashing, then after flashing, reimport it. The stabilization settings get trashed due to small changes to the UAVO, reimporting the settings fixes them. Then reset the LQGSettings and make sure at least StabilizationSettings.RCControlSmoothing.Axes is set to either Linear or Psychic. You can do that for thrust, too, but the LQG isn't really concerned with that.

And I suppose using the Custom logging profile would be nice. It does 250hz with just +12% CPU.

(--edit: The control smoothing also affect PIDs.)

glowtape:

glowtape:
This latest bunch of settings come from experimentation. I'm going to freeze them for a while. They give generally PID like performance on my current acro quad and the 3S crapquad of mine, and don't cause any ruckus on large stick throws. For whoever wants to give this a try.

glowtape:
Rewritten history to compact all that mess, and rebased.

glowtape:
Current settings seem to work generally fine across the whole throttle band over here. Found that hangtime doesn't really like them. It also creates a rebound, but that's more like a fast jerk.

glowtape:
Re: Attitude rates

Over in #2046 I was reminded that attitude has a rate limit, which has a default limit of 350°/s. That actually explains the screenshot above in relation to Attitude on LQG. Seems that LQG is better at not overstepping the limit too much.

glowtape:
Added a configuration warning, that LQG will fall through to PID, if there's no system ident data. Running the autotune wizard does populate these fields.

Dicked around with the weights. I think I have the LQR part dialed in. The weights are tuned so that the linear gain is roughly 3.5-4x that of the typical P-gain, the quadratic gain then reels it in. Performance on my regular and crap quads is OK. Logs suggest that in rate mode, on my fast quad, with larger stick throws, that it follows the StabilizationDesired envelope better than PID. This will likely result in reports of different stick feel.

Also, tuned the RTKF weights a bit, so that the rate estimate gets corrected more, while the other two generate less noise. You don't want the quadratic term noisy, especially considering how high the gain in LQR is, and the bias term is essentially an actuator offset value.

All this together, my quads don't appear to wander, and in a hover scenario, the ActuatorDesired signal is cleaner in terms of noise than with PID (even with the higher linear gain compared to P-gain). And general performance is comparable.

Tried a bunch of zero throttle drops to generate propwash, it seems to handle it good over here. So does PID.

Also, rebased on current next (Wired + PR 2072).

glowtape:
At this point, I'm not sure what to add or do. At least not until I can put serious hours of flight time into this.

I'd love to build artifacts and post it to the forums, but I figure we should let the Wired announcement sit for a week or two first, before even considering it.

ufoDziner:

glowtape:
I've dropped the LQR weights another 33% for testing over here, to make huge setpoint changes not kick as much. It generally appears to still work good in extreme manoeuvers, but in the logs you can start to see how it deteriorates compared to the current weights in the PR. The top end of 900°/s flips start to slowly get slanted like in PID, and at very low authority, it slowly starts to rebound (although still less than PID).

The current weights are OK then.

We should also finally consider doing something about ManualControl smoothing defaults (to deal with kicks on either control law). I'm probably going to add a duty cycle variable for the linear interpolator in #1950, to have it work like Betaflight, where you can have it interpolate between the values at shorter intervals than control, so the risers of the stairs in the signal get slanted. Might probably be a good compromise to the current smoothing, which still kicks quite a bit.

(--edit: Actually I just checked, they go linear interpolation all the way by default.)

glowtape:

glowtape:
Changed the LQR weights so that the gains end up around 18% lower than before. Changed RTKF weights so there's 33% more gyro contribution for slightly lower delay.

Maybe this helps getting your propwash better under control, @ufoDziner.

Bias gets suppressed when the control signal is near clipping. The bias term gets multiplied by (1-x^4) in the estimator, where x is the actuator signal (-1..1). This makes flips in the hangtime actually feel right with the lowered gains. Without this, they get all wobbly due to bias build-up.

glowtape:

glowtape:

jhitesma:
Flew this on my hex earlier today.  As per my comments in IRC no complaints but didn't notice any real benefit either.   I was just doing some easy LOS flying in the front yard.  Motors ran cool on both PID and LQG and tune felt great.  Will try and get it out for a bit of FPV tomorrow.

However the yaw results from the tune were unusable.  @mlyle - I never got a chance to test fly before wired but remember you made a change to calculate yaw PID's differently.  Any chance that change may have made things unhappy for hexes?  Tune results here:  https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAmJGAgQoM  I'll give it a try on wired tomorrow just to see if it does any better but wanted to give you a heads up while I was thinking about it.

glowtape:
This branch has been rebased on Wired recently, so the results should be the same.

jhitesma:
Just tested on the piko and it works.  But it's crazy windy out so I could only do some basic hover tests in the living room.  Didn't notice any change in motor sound or temps.

glowtape:
Just creating some paper trail in regards to feeding the inverse mix of actuators into the RTKF, to have it predict based on what actuators are actually doing.

Turns out it's not so great of an idea. Both following graphs are 900°/s flips at hangtime.

Without inverse mixing:

noinv

With inverse mixing:

inv

Without the inverse mixing, it hits the target rate pretty easily with relatively low involvement of the actuators. With the inverse mix, as soon as motor clipping management gets involved, everything appears to go to shit. It looks after initiating the motion, the actuator module is already managing things on its own, due to bottom clipping I presume, thus the inverse mix result being way lower than the desired actuator action.

Feeding the uncorrected actuator demands in the RTKF doesn't seem to make the estimate get much more ahead or wrong than with the corrected values. However whatever happens in the whole stabilization loop, it causes the LQR to try so much harder, to which actuators reacts so much more. So it is interesting how the flight behavior falls out to be more correct with the small amount of mispredicting, than when you're trying to be more correct. These are high rate low actuator authority maneuvers. In regular flight coasting around, the inverse mixing didn't appear to be detrimental, so long you weren't cornering hard.

These findings should be kept in mind for #2044, since there might be a possibility that the Smith predictor might make things worse when we're near clipping, since adding some more might send things over the edge.

During regular flying, the RTKF seems relatively bang-on. The estimate usually ranges within the noise envelope of the signal, so control accuracy shouldn't be a problem.

acc

glowtape:
@ufoDziner If raising the neutral values didn't fix any of the issues and there's still propwash like effects present, here's a bunch of RTKF parameters I've just fiddled with. They seem promising and seem to give same flight characteristics.

LQGSettings.RTKF.R = 125
LQGSettings.RTKF.YawR = 200
LQGSettings.RTKF.Q1 = 1
LQGSettings.RTKF.Q2 = 0,00001 (1e-05)
LQGSettings.RTKF.Q3 = 0.0000005 (5e-07)

What it basically does, based on how the gains for the Kalman filter turn out, keep them the same as before, especially bias, to keep it from wandering, except for the gyro rates which get filtered some more.

If for some reason (pleeeease) all your issues appear to be gone, disregard these settings for now.

ufoDziner:

glowtape:

glowtape:
I removed the bias change, which turned out to be a bad idea. These new artifacts will reflect that.

glowtape:
@ufoDziner I have a bunch of more experimental RTKF settings, if you're bored.

lqg5

(Note the halving of the yaw LQR stuff. This should reduce noise from yaw direct torque without impairing performance. Similar to the halving of the roll and pitch costs a while ago.)

It lops off around 0.5ms of delay (1.2ms > 0.7ms) for roll and pitch and 0.3ms of delay (2.1ms > 1.8ms) from yaw in the corrector stage of the Kalman filter according to MATLAB. Those reductions in delay should theoretically be beneficial.

During hover, still only half the amplitude of the noise makes it through to the actuators (not too far off from before). Hope this translates similar to this to actual flight.

lqg4

I need to remove the softmounting when I can be assed and see if these noise reductions still happen in a similar fashion with hardmounting.

glowtape:
Upped RTKF Q3 to speed up the action of the bias term. Initially done trying to fix throttle chop attitude change experienced by @ufoDziner, which however turned out to be hangtime related (the leaky bucket stuff). Still looks beneficial.

Matched the yaw Qs of the LQR to the roll/pitch ones. The previous values were there, because they happened to keep working, but there's no reason why they should be different. The new ones should also reduce noise a bit, if there's odd stuff on gyro. If we go wider with this, and there are no objections to how yaw acts, I'm going to drop the YawQs. @mlyle seems to suffer from slight yaw drift, whereas the other axes seem stable (assuming acro mode was attempted, too). I'm not 100% of the cause for this, but given that the rate and torque costs on the YawQs stood in a different relation to each other than the other Qs, this is possibly an origin.

Bias limit was raised back to 0.3 from 0.15. I'm a bit wary about this, so long the leaky bucket time constant isn't configurable (or raised a little from default), given the hangtime antics ufo manages to induce. If bias builds up due to complete lack of authority, it needs to unload (then again, the PID I-term would do the same, but I'm not sure which one is faster). It was already 0.3 at the very beginning, based upon the idea that the PID's I-term can also potentially bring up to 30% effort on its own. It got eventually reduced to deal better with bias induced jerking during high rate maneuvers, something that I got in check with bias suppression while the axis is clipping (multiplying the RTKF term with (1-x^4), where x is effort -1..1, would probably profit from going (1-x^3.5), but that's computationally more annoying).

I went outside testing all this, I don't have any objections. Unless ufo identifies any more potential issues with his high speed maneuvering, and icee's yaw still drifts, I think this is almost final. Pending complaints from outside testers.

Going further from this:
- Testing how high I can push R on the RTKF, with current Qs, until things become detrimental, specifically keeping rate mode from drifting. This seems a good target for a slider to tame noisy quads.
- Hook the LQR costs up to TxPID. After figuring out the main solution right after boot, the LQGSolver keeps running in background on a low flame (one pass of covariance per ms, it can do a 1000 loops in a fraction of a second on the F4 from the looks of it), allowing changes to cost to reconverge to new control gains. Then figure out the effects of pushing things around to see how that could be turned into sliders.

--edit:
Current cost of running LQG, low flame LQGSolver and updating the UAVOs appears to be +2% CPU over PID from quick testing. On Seppuku though.

ufoDziner:

glowtape:
Some more stuff. Sampled multiple full stick throws in attitude mode. Max. rate limited to default 350°/s. Seems like LQG can muster up the same slope as PID, but isn't as overshooty.

Yellow = RateDesired.Roll
White = Gyro.X on PID
Red Left = Gyro.X on LQG with default LQR weights
Red Right = Gyro.X on LQG with lowered torque cost

lqg6

During the first downward slope, the PID seems to get closer to the desired rate (yellow), but that might be the same sort of over-acceleration that leads to the first overshoot. I guess it could be considered beneficial, because it's making up for things, but the LQG seems more true to rate, because the delay between desired rate and gyro seems constant. Example of two very similar events (top PID, bottom LQG) on their own:

lqg7
(This with lower torque cost and therefore slightly more overshooty LQG.)

I figure this qualifies as some evidence that the current weights point into a generally good direction. At least for the smaller quads.

But this also changes the feel of attitude. One might think LQG is slower, when in looks like PID gets an "advantage" via rate overshoot.

glowtape:
Tested my DVR today. This situation, that also happened the very first outside test flight is what has me so enthusiastic about it (note the amount of rebound LQG vs. PID): https://www.youtube.com/watch?v=NESAxbqnVsw

mlyle:

flight/Libraries/sanitycheck.c
```diff
@@ -121,6 +122,19 @@ int32_t configuration_check()
 			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_FAILSAFE:
 				// always ok
 				break;
+			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_LQG:
+			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_ATTITUDELQG:
```
mlyle: It would be nice for all of these to be LQGLeveling for consistency.  Also would be nice to catch it in stabilized banks -- e.g. if you have LQG/LQG/Acro or something (and factor this test out to a function).

flight/Libraries/sanitycheck.c
```diff
@@ -370,6 +386,9 @@ void set_config_error(SystemAlarmsConfigErrorOptions error_code)
 	case SYSTEMALARMS_CONFIGERROR_UNSAFETOARM:
 		severity = SYSTEMALARMS_ALARM_ERROR;
 		break;
+	case SYSTEMALARMS_CONFIGERROR_LQG:
+		severity = SYSTEMALARMS_ALARM_WARNING;
```
mlyle: Go ahead and make this fatal-- no sense allowing to fly with default PIDs.

flight/Modules/LQGSolver/lqgsolver.c
```diff
@@ -0,0 +1,117 @@
...
+#include "taskinfo.h"
+
+#include "lqg.h"
+
+#include "lqgsolution.h"
+
+#define STACK_SIZE_BYTES 600		// Reevaluate.
+#define TASK_PRIORITY				PIOS_THREAD_PRIO_LOW
+
+/* Hack things into pieces to give other threads more time. Seems like USB telemetry goes
```
mlyle: Can we do this in stabilization instead of a dedicated task?  Saves RAM, etc.  Just do a small number of iterations each stabilization cycle.  Also I don't really like more modules.

flight/Modules/Logging/logging.c
```diff
@@ -339,6 +343,8 @@ static void loggingTask(void *parameters)
 					register_default_profile();
 					break;
 				case LOGGINGSETTINGS_PROFILE_CUSTOM:
+					register_custom_profile();
```
mlyle: Do we really need this?  Or should we just change metadata so that the default profile logs this stuff.

ground/gcs/src/plugins/systemhealth/html/SystemConfiguration-Warning-LQG.html
```diff
@@ -0,0 +1,12 @@
+<html>
+  <head>
+    <title></title>
+    <meta content="">
+    <style></style>
+  </head>
+  <body style="color: black">
+    <h1>System Configuration: Warning</h1>
+    <p>At least one LQG-controller based flight mode has been configured, but no valid system identification has been data found. Please rerun the autotune wizard, or if necessary, or not done yet, the complete autotune procedure.</p>
+    <p>Trying to take off this way will not prevent flight, but stabilization will fall through to the PID-controller instead.</p>
```
mlyle: Hm.  I guess this lets you do e.g. lqg/autotune as flight modes and not have to reconfig.. but if you need to connect a computer anyways to apply autotune...  Probably should still mask arming and remove this line.

flight/Libraries/math/lqg.c
```diff
@@ -0,0 +1,505 @@
+#include "lqg.h"
```
mlyle: need our copyright notice.   Also, is there any source (e.g. sympy, matlab) that this came from that should be checked in too for license compliance?

mlyle (changes requested):
Looks good overall

flight/Modules/Logging/logging.c
glowtape: This was going to go away at some point, anyway. I have this in the test branch to allow faster logging without hitting the CPU limit.

flight/Libraries/math/lqg.c
glowtape: The MATLAB stuff was done just in some scratch file, with matrixes of syms and then run the formula and paste the output into a C file. I can redo these for inclusion, since they were trivial.

Will add the copyright notices.

glowtape:
- Axed LQGSolver and folded covariance convergence into stab.c
- Renamed AttitudeLQG flight mode to LQGLeveling. Left the stabilization mode untouched though, since it's Attitude for the other mode, too.
- It doesn't arm anymore if there's bad/no data and a LQG mode is configured.
- The logging hacks were reverted and logging of the RTKF and the LQR solution was added to the basic profile.

Rebased on latest next for the althold stuff.

Still need to add the MATLAB stuff.

Also, I did some testing with slightly lowered torque costs. Seemed beneficial. Would be nice to get some verification, to slip that maybe in. Starts braking a little later with huge setpoint changes. @ufoDziner , @jhitesma and uh what does LitterBug go with on GitHub?

LQGSettings.LQR.Q2 = 0.000075 instead of 0.0001
LQGSettings.LQR.YawQ2 = 0.000075 instead of 0.0001

glowtape:
I think for releasing it, we should let the Q2s at 0.0001, because with bad authority at hangtime, due to bad configuration, starts getting overshooty with the lower values. I'd still love to hear feedback on the tweaked one.

That cost factor should probably configurable using sliders in the autotune wizard, following up on the first wider tests.

mlyle:

flight/Libraries/alarms.c
```diff
@@ -272,7 +272,7 @@ static const char alarm_names[][10] = {
 // If someone adds a new alarm, we'd like it added to the array above.
 DONT_BUILD_IF(NELEMENTS(alarm_names) != SYSTEMALARMS_ALARM_NUMELEM, AlarmArrayMismatch);
 
-static const char config_error_names[][15] = {
+static const char config_error_names[][16] = {
```
mlyle: That 16 is the length of the longest one, right?

flight/Libraries/math/lqg.c
```diff
@@ -0,0 +1,515 @@
+/**
+ ******************************************************************************
+ * @addtogroup Libraries Libraries
+ * @{
+ * @addtogroup FlightMath math support libraries
+ * @{
```
mlyle, line 6: These need to be closed at the end of file.

flight/Libraries/math/lqg.h
```diff
@@ -0,0 +1,64 @@
+/**
+ ******************************************************************************
+ * @addtogroup Libraries Libraries
+ * @{
+ * @addtogroup FlightMath math support libraries
+ * @{
```
mlyle, line 6: Need to be closed at end of file.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -831,6 +947,27 @@ static void stabilizationTask(void* parameters)
...
+						rateDesiredAxis[i] = bound_sym(raw_input[i], settings.ManualRate[i]);
+
+						/* Compute the inner loop */
+						actuatorDesiredAxis[i] = lqg_controller(lqg[i], gyro_filtered[i], rateDesiredAxis[i]);
+						/* The LQG controller bounds data already, but whatever. */
+						actuatorDesiredAxis[i] = bound_sym(actuatorDesiredAxis[i], 1.0f);
+						break;
+					}
+#endif
+					/* Fall through to rate, if LQR gains haven't been solved yet, or if
```
mlyle, line 984: nice

flight/Modules/Stabilization/stabilization.c
```diff
@@ -813,6 +917,18 @@ static void stabilizationTask(void* parameters)
 
 		uint16_t max_safe_rate = PIOS_SENSORS_GetMaxGyro() * 0.9f;
 
+#if defined(PIOS_INCLUDE_LQG)
+		bool lqg_in_use = false;
+
+		for (int i = 0; i < MAX_AXES; i++) {
```
mlyle: So this basically means, if you enter into LQG in flight, it's going to briefly be flying acro instead?  Do we have an idea of what the bounds of this is, in flight, like on a F3?

It's a shame it's not easy to tell if it's in any flight mode or active stab bank.... sanitycheck knows, we don't.

flight/Libraries/sanitycheck.c
```diff
@@ -123,6 +139,12 @@ int32_t configuration_check()
 			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_FAILSAFE:
 				// always ok
 				break;
+			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_LQG:
+			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_LQGLEVELING:
+				if (!lqg_sysident_check()) {
```
mlyle, line 144: This still needs to get put into the stabilization bank code, right?

flight/targets/aq32/fw/pios_config.h
```diff
@@ -32,6 +32,7 @@
 #include <pios_flight_config.h>
 
 /* Major features */
+#define PIOS_INCLUDE_LQG
```
mlyle: This isn't really a PIOS_INCLUDE because it's not in PIOS right?

We typically have like

```
/* Flags that alter behaviors */

#define SYSTEMMOD_RGBLED_SUPPORT
#define SYSTEMMOD_RGBLED_VIDEO_HACK

#define CAMERASTAB_POI_MODE
```

But I question whether there should be a flag at all, since you're turning it on, on all targets, and it's something that people will expect to work.

flight/targets/sprf3e/fw/Makefile
```diff
@@ -49,11 +49,11 @@ OPTMODULES += VtolPathFollower
 OPTMODULES += FixedWingPathFollower
 #OPTMODULES += UAVOMavlinkBridge
 OPTMODULES += UAVOMSPBridge
-OPTMODULES += UAVOLighttelemetryBridge
+#OPTMODULES += UAVOLighttelemetryBridge
```
mlyle: One last rebase needed to get these commits out.

mlyle (approved):
Looks *great*.

Only action needed now is to rebase to take out the sprf3e commits (and answer the question about convergence time).  I'd leave the other stuff alone for a round of fixing after test.

Thanks so much for all your work on this.

flight/Modules/Stabilization/stabilization.c
glowtape: If there's values in SystemIdent, it starts to solve right after boot up. It does a set amount of cycles for RTKF (150) and LQR (1500), then it stops doing the math stuff and immediately returns.

When I moved the solver stuff into stab.c, I initially forgot to initialize the LQGSolution UAVO. So when it tried to store the solution data, it crashed. Took about a second to crash after hitting Boot in bootloader mode. Long story short, it has the LQR gains way before someone arms the quad. Would probably apply to F3, too.

flight/Modules/Stabilization/stabilization.c
tracernz: We should make a fallthrough macro so we can start to get ahead of GCC8 etc?

flight/targets/aq32/fw/pios_config.h
tracernz: Yep, should be STABILIZATION_LQG or something if it does stay.

glowtape: I put some defines there, to exclude some targets. I still wasn't sure the F3 can run it. The F3s can easily reach 80% as is according to the guys on IRC. I can rename the define for it to be proper.

flight/Libraries/alarms.c
tracernz: Bug fix? Surely the compiler should catch that though..

flight/Libraries/math/lqg.c
glowtape: Now I know what those weird comments at the end of files are even good for.

flight/Libraries/alarms.c
glowtape: Whoops, brainfart.

shared/uavobjectdefinition/lqgsolution.xml
```diff
@@ -0,0 +1,18 @@
...
+      <description/>
+    </field>
+    <field defaultvalue="0" elements="2" name="PitchK" type="float" units="">
+      <description/>
+    </field>
+    <field defaultvalue="0" elements="2" name="YawK" type="float" units="">
+      <description/>
+    </field>
+  </object>
+</xml>
```
tracernz: Missing linefeed

glowtape:
Force-pushed the rebase. Gonna add the changes/fixes requested above locally.

flight/targets/aq32/fw/pios_config.h
glowtape: Also, one F3 target was pretty damn low on memory. It was ufo's F3. Some OmnibusF3 clone with CharOSD I think. Dropping the LQGSolver task might have improved that situation though.

mlyle: We have also done half of the inittask thing since then.

flight/Modules/Stabilization/stabilization.c
mlyle: Ah, my bad.  Great.

mlyle:

mlyle:
Deleted lots and lots of "build artifacts please" and dead links to artifacts.

glowtape:
Since Inconceivable is going to have a lot of other improvements, how about going with the super-casual route with this? Just list it as short bullet point like it's nothing, with the requirement to rerun the autotune wizard in parentheses, then sit on our hands for a while before posting the big text?

mlyle:
Unprotecting those artifacts because #2118...   Respinning, trying again.

tracernz:
In future, it's best to do those kinds of things in one atomic commit. It gets messy, when we have to bisect for example, if intermediate states don't build + run.

glowtape:
I just posted them like that to not force push things again. If everything is A-OK, I will squash them into the appropriate commits, before merge.

glowtape:
Going to add _optional_ control term filtering on the RTKF, based on either global lowpass delay or tau. Since the whole shebang is a first order system, I think it might help, if we did something like that on the control term.

If there's a huge setpoint change, there's going to be a huge instant control demand, and if it's fed as is to the RTKF, it'll leap ahead in its estimate accordingly (because it adds torque directly scaled to u), even though the actual system still needs to get moving (tau et al). Visible for instance in this graph posted a while ago. Purple is the RTKF, note how the slope is always ahead of actual gyro:

noinv

glowtape:
Seems like I need to work some more on the default settings. When things seem really off balance for whatever reason, the torque estimate can go way overboard and make things terrible.

Here's an example. Top of the graph is @ufoDziner's problematic quad with T5050C props, doing flips on default RTKF weights. Bottom is the same quad with the Q2 (process noise on torque) bumped from 0.000003 to 0.00001, i.e. 3.3x.

Notice the cross-axis wobbling after flips, and the wobbling at the top rate of flips. Severely reduced.

wooo

(In the logs, there's still cross-coupling to the yaw axis, though. But it looks way saner.)

I need to find another set of weights and costs for the LQG, that improves the torque estimate a bit more than above, while still maintaining the noise advantage to some degree.

The Q2 bump was cross-checked on another quad, things stayed good over there. Noise magnitude of LQG vs. PID was 50% on the eye.

mlyle:
Any objections to merging this?

ufoDziner:
Not objection from me.  It's great for most of my flights.  In fact it's the only flight mode I'm using now.  Only anomaly is the weird back flip wobble with blheli_32 and aggressive props.  It's not horrible, but there on my builds anyway.

glowtape:
Apart from that one specific quad from ufo, the few guys who tried it, the experience ranged from neutral to mildly positive. I'm gonna squash the fix commits later today. The tricopters worked with surprisingly without much issues.

Whenever this one got merged, I'll run a quick optimization on the LQR covariance. The A11 term in the A matrix is always zero, the generated code doesn't consider that (doing so would save a bunch of multiplications and additions). I don't expect it to change anything, but it's a bit risky to fold it into this PR.

glowtape:
Squashed some stuff to reduce commit history size, verified things still work practically.

mlyle:
Ack dude, don't rewrite history right at the point I'm considering merging it.

It means I can't tell what's changed!

mlyle:
Skimmed diff from 53e374305ec1f29f486e5726587791c1dba926cb

glowtape:
Whoops. Thought the review was over with the consideration of a merge. I can push the previous version again, if that helps.

mlyle:
Nah, it's OK, I've re-read.

Rewriting history is great (makes git log prettier, makes the repo use less space, rebases are better than merging back and forth), but it means that the ability for other people to compare and make sure the right things go in goes away.  It's not always possible, but it's much better if it happens before review (especially for big nasty things).

PR #2042, opened 2018-01-26
stabilization: use antiwindup math for rate loop

mlyle:
Woops!  Seems we were not doing this before.

Also introduce a version of the PID calls that does both deriv setpoint
weighting and antiwindup.

Needs another round of careful verification but this looks correct to me.

### DIFF
--- a/flight/Libraries/math/pid.c
+++ b/flight/Libraries/math/pid.c
@@ -103,8 +103,8 @@ float pid_apply_antiwindup(struct pid *pid, const float err,
 		dterm = pid->lastDer +  dT / ( dT + deriv_tau) * ((diff * pid->d / dT) - pid->lastDer);
 		pid->lastDer = dterm;            //   ^ set constant to 1/(2*pi*f_cutoff)
 	}	                                 //   7.9577e-3  means 20 Hz f_cutoff
- 
- 	// Compute how much (if at all) the output is saturating
+
+	// Compute how much (if at all) the output is saturating
 	float ideal_output = ((err * pid->p) + pid->iAccumulator + dterm);
 	float saturation = 0;
 	if (ideal_output > max_bound) {
@@ -114,6 +114,7 @@ float pid_apply_antiwindup(struct pid *pid, const float err,
 		saturation = min_bound - ideal_output;
 		ideal_output = min_bound;
 	}
+
 	// Use Kt 10x Ki
 	pid->iAccumulator += saturation * (pid->i * 10.0f * dT);
 	pid->iAccumulator = bound_sym(pid->iAccumulator, pid->iLim);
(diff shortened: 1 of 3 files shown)


mlyle:

glowtape:
I've been doing quite a bunch of flips on my roll axis, and I don't even seem to get into the wind-up action. I just barely blip 100% on ActuatorDesired. My P-gain is 0.002something, so I should be clipping. I guess my stick throws are too slow for the quad to freak out about.

https://i.imgur.com/ToEHuE5.png

flight/Libraries/math/pid.c
```diff
@@ -168,6 +169,62 @@ float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const f
...
+float pid_apply_setpoint_antiwindup(struct pid *pid,
+		struct pid_deadband *deadband, const float setpoint,
+		const float measured, float min_bound, float max_bound)
+{
+	float dT = pid->dT;
+
+	float err = setpoint - measured;
+	float err_d = (deriv_gamma * setpoint - measured);
+
+	if(deadband && deadband->width > 0)
```
tracernz, line 181: Spacing + erroneous newline.

flight/Libraries/math/pid.c
```diff
@@ -168,6 +169,62 @@ float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const f
...
+	} else {
+		pid->iAccumulator += err * (pid->i * dT);
+		pid->iAccumulator = bound_sym(pid->iAccumulator, pid->iLim);
+	}
+
+	// Calculate DT1 term,
+	float dterm = 0;
+	float diff = (err_d - pid->lastErr);
+	pid->lastErr = err_d;
+	if(pid->d && dT)
```
tracernz, line 201: As above

flight/Libraries/math/pid.c
```diff
@@ -168,6 +169,62 @@ float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const f
...
+	{
+		dterm = pid->lastDer +  dT / ( dT + deriv_tau) * ((diff * pid->d / dT) - pid->lastDer);
```
tracernz, line 203: Extra space between ( and dT

flight/Libraries/math/pid.c
```diff
@@ -168,6 +169,62 @@ float pid_apply_setpoint(struct pid *pid, struct pid_deadband *deadband, const f
...
 }
 
+float pid_apply_setpoint_antiwindup(struct pid *pid,
+		struct pid_deadband *deadband, const float setpoint,
+		const float measured, float min_bound, float max_bound)
+{
+	float dT = pid->dT;
+
+	float err = setpoint - measured;
+	float err_d = (deriv_gamma * setpoint - measured);
```
tracernz, line 179: Why the parentheses?

tracernz (commented):
Do we ever want pid_apply_setpoint without anti-windup?

mlyle:
All of this is copy-paste from the previous implementations... to make a new variant (pid + setpoint + antiwindup) from the old (pid), (pid+setpoint), (pid+antiwindup) implementations.  So I'd rather not reformat because that'll obscure the common origin :/

Issue #1420, opened 2016-10-09
flyingpio: ppm is broken


mlyle:
Failsafes; intermittently flashes correct values according to @mhv-shared from IRC.  Have not tried myself.

xfce:
i change the pios_flyingpio.c's cmd->receiver_protocol = HWSHARED_PORTTYPES_SBUS
to cmd->receiver_protocol = HWSHARED_PORTTYPES_PPM
then no change, still only found the throttle stick in GCS's wizard ,
try both quitoxe and atifice branch

jihlein:
I've been looking into this and here's what I've found so far:
1)Made the changes in #2043 to correct RTC frequency.  Verified the RTC frequency with a Logic Analyzer to be 625 Hz.  PPM still does not work correctly.
2)Made a new time base definition TIM14 with the TIM period = 0xFFFF.  PPM still does not work correctly.
3)Changed the processor clock from 40 MHz to 48 MHz.  PPM still does not work correctly.

Interesting things to note:
1)After the above changes, PPM will sometimes lock, after waiting a while,  This wait has been observed to be as short as 56 seconds, or as long as 203 seconds, so it's not consistent.  And sometimes I've waited 5 plus minutes with no lock.  I'm not sure if all of the changes, some of the changes, or one of the changes above are responsible for this.
2)Once locked, however, I can cycle the RX power off and on with immediate lock recovery, and I can do this multiple times with no failure to immediately lock.
3)The PWM frame time in the code is 20 Msec.  I'm using a D4R-II with a 27 mSec frame time.  Don't know if this matters too much as it works fine on the other target boards I've tried, but it might be worth experimenting with.
4)I don't know if it's somehow related, but now that I'm paying more attention to this again, I'm occasionally seeing a "fpio: VERYBAS: Bad CRC response on data XXXX" report on the RPi console screen.
5)I've tried lowering the SPI clock rate from 15 MHz to 10 MHz with no change to the noted behaviors.
6)FWIW, same behaviors are noted whether or not I have the GPS and external mag connected and running.

To be continued..... :)

mlyle:
Confirmed working in #2140

PR #2043, opened 2018-02-03
stm32f0: rtc: change frequency to match other targets

mlyle:
This may solve PPM-related problems on flyingpi.

Untested.  Hat tip @jihlein for the question on IRC, which made me think of the supervisor-- I hadn't realized that PPM "started working eventually".

### DIFF
--- a/flight/PiOS/STM32F0xx/pios_delay.c
+++ b/flight/PiOS/STM32F0xx/pios_delay.c
@@ -32,9 +32,9 @@
 /* Project Includes */
 #include <pios.h>
 
-/* This will result in a value like 39062 -- about 16 bits */
-#define SYSTICK_HZ 1024
-#define RTC_DIVIDER 4	/* Must be power of 2; this results in 256Hz "RTC" */
+/* This will result in a value of 32000 -- about 15 bits */
+#define SYSTICK_HZ 1250
+#define RTC_DIVIDER 2	/* Must be power of 2; this results in 625Hz "RTC" */
 
 #define PIOS_RTC_MAX_CALLBACKS 3
 


mlyle:
@jihlein can you try this please?

jihlein:
@mlyle - no joy, in fact after those few times it took a while for PPM to sync, I haven't been able to get it to sync again.  Not sure what I did different those few times.

FWIW - I had tried something similar with the PIOS_PPM_Supervisor SUPEVISEOR_TICK_DIVIDER, but it didn't help either.  I also raised the clock value from 40 MHz to 48 MHz just to see what would happen, that didn't help either, not that I expected it would.

EDIT-Something deeper is not right, I made some mods to put the RTC Tick on a GPIO pin, logic analyzer says it's 312.5 Hz.  Investigating......

jihlein:
RTC Tick is fine, I used an LED toggle, on for one period, off for one period, and didn't account for that in my calculations.  Moral to the story - 6:00 am is too early to be working on this :)

mlyle:
Bummer, I was "sure" it would be this.

jihlein:
I was hopeful too.  I'll try to throw an SBUS receiver on it later, if that still works I think this is still a useful PR.

jihlein:
No time like the present, just threw an SBUS rx on it, and it still calibrates and works correctly on the bench.  No flight test though.

Not sure what else we can test here, so I'm inclined to say it has no ill effects, and matches the other targets, so it's a useful change.

mlyle:
Yah, I agree-- feel free to merge :D

mlyle:
Thx.  I will think more about PPM.  I've gotten stuck a few times looking at it, but I never realized that it can eventually work, which I feel like is a pretty big clue.

jihlein:
FWIW, I left some notes in #1420 detailing my observations so far.

tracernz:
Thx for merging! 😀

PR #2044, opened 2018-02-03
actuator: Delay compensation via Smith predictor.

glowtape:
Hope I did it right. Doesn't have any ill-effect at least in the living room. Can't test it outside yet, so I'm not sure what I'm supposed to be looking for.

Default delay is 3.2ms (for Seppuku x1.6 + 0.5, then rounded = 6 samples), the IIR has a default coefficient of 0.5, which creates a group delay of 1.4 samples. Should get in the ballpark of the 4ms mentioned before in IRC.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -97,6 +97,7 @@ static MixerSettingsMixer1TypeOptions types_mixer[MAX_MIX_ACTUATORS];
  */
 
 static float motor_mixer[MAX_MIX_ACTUATORS * MIXERSETTINGS_MIXER1VECTOR_NUMELEM];
+static float motor_mixer_inv[MAX_MIX_ACTUATORS * MIXERSETTINGS_MIXER1VECTOR_NUMELEM];
 
 /* These are various settings objects used throughout the actuator code */
 static ActuatorSettingsData actuatorSettings;
@@ -408,6 +409,12 @@ static void compute_mixer()
 #if MAX_MIX_ACTUATORS > 9
 	compute_one_token_paste(10);
 #endif
+}
+
+static bool compute_inverse_mixer()
+{
+	return matrix_pseudoinv(motor_mixer, motor_mixer_inv,
+		MAX_MIX_ACTUATORS, MIXERSETTINGS_MIXER1VECTOR_NUMELEM);
 }
 
 static void fill_desired_vector(
@@ -781,6 +788,14 @@ static void actuator_task(void* parameters)
 			SystemSettingsAirframeTypeGet(&airframe_type);
 
 			compute_mixer();
+
+			/* If we can't calculate a proper inverse mixer,
+			 * set failsafe.
+			 */
+			if (compute_inverse_mixer()) {
+				set_failsafe();
+				continue;
+			}
 
 			MixerSettingsThrottleCurve2Get(curve2);
 			MixerSettingsCurve2SourceGet(&curve2_src);
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -51,6 +51,7 @@
 #include "manualcontrolcommand.h"
 #include "pios_thread.h"
 #include "pios_queue.h"
+#include "pios_sensors.h"
 #include "misc_math.h"
 
 // Private constants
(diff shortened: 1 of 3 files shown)


flight/Modules/Actuator/actuator.c
```diff
@@ -1058,6 +1108,73 @@ static void set_failsafe()
...
+	}
+
+	/* Do inverse stuff. */
+	matrix_mul_check(motor_mixer_inv, m->iir, inv,
+			MIXERSETTINGS_MIXER1VECTOR_NUMELEM,
+			MAX_MIX_ACTUATORS,
+			1);
+
+	for (int i = 0; i < MIXERSETTINGS_MIXER1VECTOR_NUMELEM; i++) {
+		desired_vect[i] += m->mix * (inv[i] - desired_vect[i]);
```
glowtape: I'm hoping this one is the right way. I had the terms of the subtraction switched earlier and it made things audibly worse.

glowtape:
Tried autotuning with this, it makes the yaw readings way worse.

glowtape:
False alarm, I changed to a pristine set of props, issue went away.

tracernz:
This is targeted at post-wired right?

P.S. Any ideas for post-wired name? :)

glowtape:
Wasabi. --edit: Wait, no, this also starts with W.

glowtape:
The predictor currently also mixes thrust into command, since that's also part of the inverse. Do we even want that?

mlyle:
I think we do-- if know we're underperforming on thrust, we want to fix it right?

mlyle:
god damn wrong button.

flight/targets/sprf3e/fw/Makefile
```diff
@@ -50,13 +50,13 @@ OPTMODULES += AltitudeHold
...
+#OPTMODULES += UAVOMavlinkBridge
 OPTMODULES += UAVOMSPBridge
-OPTMODULES += UAVOLighttelemetryBridge
+#OPTMODULES += UAVOLighttelemetryBridge
 OPTMODULES += Battery
 OPTMODULES += ComUsbBridge
 OPTMODULES += Airspeed
-OPTMODULES += UAVOHoTTBridge
+#OPTMODULES += UAVOHoTTBridge
 OPTMODULES += UAVOFrSKYSensorHubBridge
```
tracernz: Will want to remove all these from HwSprf3e too https://github.com/d-ronin/dRonin/blob/next/shared/uavobjectdefinition/hwsprf3e.xml

glowtape:
On which config page should I stick the UI for this?

mlyle:
Let's prove it's good before adding UI

glowtape:
Seems like I need to get cracking on this. The other guys have implemented sorta Smith prediction on their throttle control.

glowtape:
So the loose idea I had anyway was an optionally purely filter based time delay (not in particular a fan of adding more noise than necessary :), as alternative to the sample based delay plus weak IIR. Separate values for throttle may be worth an experiment?

mlyle:
>  So the loose idea I had anyway was an optionally purely filter based time delay (not in particular a fan of adding more noise than necessary :), as alternative to the sample based delay plus weak IIR. Separate values for throttle may be worth an experiment?

Hm, I'm not really a fan.

1-- PID can effectively compensate for either pure-delay or exponential decay, but not both simultaneously.... the sample delay is what makes the smith predictor actually *do* something useful that doesn't mostly just commute with PID, the IIR is just to clean it up.

2-- throttle shaping, etc... how's this different from what smoothcontrol does?  We already put in maximal effort to deliver requested throttle with the clip management.

glowtape:
Yeah, was just a bunch of random ideas.

Rebased to fix conflicts. Added some minor massaging of the throttle signal, after noticing some motor grinding when on the ground with hangtime or very low throttle (took me a while to figure that one out).

glowtape:
What about a second order lowpass? The step response of one is kinda delay+IIR-ish, given the desired delays (4-7ms), if you squint your eyes. If it's remotely applicable, we could keep the noise down, which in turn would allow cranking up the contribution to the control signal.

mlyle:
Put back into inconceivable for now, just for consideration.

glowtape:

glowtape:
@ufoDziner These are the Smith predictor artifacts.

glowtape:
@ufoDziner I nearly forgot. You should probably set ActuatorSettings.SmithPredictorMix to 0.3 to start (lower than the default in this PR) after installing.

flight/Modules/Actuator/actuator.c
```diff
@@ -1217,12 +1217,11 @@ static void smithp_compensate(struct smith_predictor *m, float *desired_vect)
...
-			if ((t >= THROTTLE_EPSILON && (t+v) < THROTTLE_EPSILON) ||
-			    (t <= -THROTTLE_EPSILON && (t+v) > -THROTTLE_EPSILON)) {
-					v = 0;
-				}
-			/* Also bound throttle. */
-			desired_vect[i] = bound_sym(desired_vect[i] + v, 1.0f);
+			float r = t+v;
+
+			/* Don't cross zero via prediction. Also bound throttle. */
+			if (sign(t) == sign(r))
```
mlyle: THis isn't quite what we want, because if you overperform significantly and throttle is 10%, and it wants to lower it to -2%... we should still lower it to throttle epsilon.

mlyle:
Is this something we want to try and merge for inconceivable?  If so, it should lose the WIP tag and go through review.

glowtape:
We have the anti-windup changes in PID and then the LQG. It would be interesting to get people flying these changes before adding another influencer on control. IMO we should shove it off to Ludicrous. I'd probably want to run another round of logs eventually, to check its behavior at low and high end of throttle.

Quick-Flash:
I've ported some of this over to EmuFlight, but instead did the smith predictor on the gyro data and have had great results with it.

PR #2046, opened 2018-02-03
autotune widget: shift slow pole to control overshoot

mlyle:
On modern craft, it's just too aggressive.  Also eliminate the outer KP
soft-limit, since it was probably a reaction to the very same overshoot.

### DIFF
--- a/ground/gcs/src/plugins/config/configautotunewidget.cpp
+++ b/ground/gcs/src/plugins/config/configautotunewidget.cpp
@@ -635,7 +635,7 @@ void AutotuneSlidersPage::compute()
     // Set the real pole position. The first pole is quite slow, which
     // prevents the integral being too snappy and driving too much
     // overshoot.
-    const double a = ((tau + tau_d) / tau / tau_d - 2 * damp * wn) / 20.0;
+    const double a = ((tau + tau_d) / tau / tau_d - 2 * damp * wn) / 25.0;
     const double b = ((tau + tau_d) / tau / tau_d - 2 * damp * wn - a);
 
     CONF_ATUNE_QXTLOG_DEBUG("ghf: ", ghf);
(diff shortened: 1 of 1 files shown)


mlyle:
Flew on Vendetta and 150 size OG brain, to good results.  Will be measuring more.

IMO this one should be considered for wired, but it's super-last-minute.

glowtape:
On the mid to high throttle band, it's just fine as before. On the low throttle band, there's still rebound, but a little less. Tried damping 1.0 and sensitivity 1.3, that reigns it in just a tiny bit more. Tried the Smith predictor, it doesn't do anything noticeable in that case.

Reduced the I-gains on the 1.0/1.3 tune some more, seemed to get a little better.

With the non-stock damping and sensitivity, my outer loop tunes to P = 15 and I = 8, and it works out just fine. Thing flies like a Chihuahua on crack in attitude.

At this point, things look good enough, and I think trying to chase it completely away might be futile, because a) I'd like to see unstabilized cam footage of these ostensibly super awesomely tuned BF quads, and b) the LQG has the advantage of its Kalman filter, which integrates control output and kind of gets sort of a delay reduction via that. I'm going to stick the RTKF in front of the PID just to verify. --edit: Nope, that doesn't help.

mlyle:
@glowtape -- I think there's one more good thing to do, and that's to limit the integrator response when we're likely to be slope limited.  But it will have to wait a little.

Re: chihuahua on crack--- have you tried increasing maximum rate from the default 350? ;)  It's worth after an attitude check to touch motors and make sure they're not hot.

glowtape:
Oh I didn't know about that one. I've upped it to 600°/s and yea, it's more fun.

mlyle:
Rebased, jsut to get another couple recent changes in this branch.

ufoDziner:

mlyle:
Fine/good so far across 5 of my vehicles and 1 of @glowtape

glowtape:
My 3S crap quad also seems to be content with this.

mlyle:

mlyle:
Status: Before last  commit, flew on everything of mine, including giant quad.

@ufoDziner flew and got an uncommanded flip at non-standard noise setting in attitude mode.  Audited attitude loop and it looks OK, so my theory is that you can still run into problems with slew-rate limits on very aggressive things.

He is going to fly the old code at default noise sens to see if his vehicle is stable at this operating point, and I have a change to soften outer kP response beyond 7 (still more aggressive than what we had before).

glowtape:
That sounds strange, like the attitude loop threw the quad pretty hard, but then the motors didn't create an appropriate counter response fast enough, the axis flipped. Maybe a desync?

mlyle:
@glowtape -- autotune doesn't worry about any slope limitations-- it doesn't know for large signals how quickly it can decelerate an axis in practice.  It's purely focused on small-signal tuning.  So it's quite possible for outer loop kp to ask for high rates and then not ramp down quickly enough to stop them.

mlyle:
Update: the earlier testing was faulty of this, but then full-stick deflections with an aggressive tune (kp=18.9) burnt an ESC for @ufodziner... so probably better to use the new, safer outer values (would yield a much more sane 11.8 instead).

mlyle:
Recommended for merge.

PR #2048, opened 2018-02-04
flight: rename module.example to loadable.example

### LINKED ISSUE #2020: Flight: Rename example.module to example.loadable
s/module/loadable/ throughout it. We already have an incompatible definition of "[module](https://github.com/d-ronin/dRonin/tree/next/flight/Modules)" which will only result in pain.

mlyle:
fixes #2020.  While these are intended to fulfill the same role as
modules, having a different namespace seems to make sense.


tracernz:
Thanks! I know it's not a big deal, but it will make documentation, support, etc. a bit clearer.

PR #2051, opened 2018-02-04
telemetrymonitor: ensure we retrieve firmwareiap

### LINKED ISSUE #2047: Lux seems to not show hardware pane
Saying "Board detected, but of unknown type."

mlyle:
Before, nothing guaranteed we'd get it.  It telemetered only on change
(which only helps you if you are connected *very* early), and it's not a
settings object so we didn't retrieve it at the beginning of a session.

fixes #2047


mlyle:
With this, the hardware screen consistently shows up for me on F3.

tracernz:
Thanks!

tracernz (approved):

PR #2054, opened 2018-02-08
omnibusf3: support printing floats to OSD

mlyle:
Reported by yds on forum.

### DIFF
--- a/flight/targets/omnibusf3/fw/Makefile
+++ b/flight/targets/omnibusf3/fw/Makefile
@@ -281,4 +281,6 @@ LDFLAGS += -Wl,--fatal-warnings
 #Linker scripts
 LDFLAGS += -T memory.ld $(addprefix -T,$(LINKER_SCRIPTS_APP))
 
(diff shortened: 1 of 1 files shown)


tracernz:
From [newlib docs](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob_plain;f=newlib/README;hb=HEAD):
```
Floating-point support is split out of the formatted I/O code into
	 weak functions which are not linked by default.  Programs that need
	 floating-point I/O support must explicitly request linking of one or
	 both of the floating-point functions: _printf_float or _scanf_float.
	 This can be done at link time using the -u option which can be passed
	 to either gcc or ld.  The -u option forces the link to resolve those
	 function references.  Floating-point format specifiers are recognized
	 by default, but if the floating-point functions are not explicitly
	 linked in, this may result in undefined behavior for programs that
	 need floating-point I/O support.
```

tracernz (approved):

PR #2057, opened 2018-02-09
GCS:crashreporter: Fix git version info

tracernz:
- Hopefully make it work with the ancient git version on the mac builder, whilst sticking with plumbing commands
- Force mainwindow.cpp to recompile (unfortunately always for now)
    - Previously the version info only changed if crashreporterapp was rebuilt for some other reason

Fixes #1767

PR #2058, opened 2018-02-09
GCS: Make rule to generate version info, common version info for all

tracernz:
Builds on #2057 

- Generate a Makefile rule to build the version info since we don't always run QMake anymore
- Use the main GCS version info for crashreporterapp


tracernz:
build artifacts please

PR #2060, opened 2018-02-10
manualcontrol: make corner arming work.

### LINKED ISSUE #1962: Corner arming appears broken
Both in neat and new code.  Unfortunate that it uses the same arming/disarming command.

mlyle:
Fixes #1962 ...  Needs careful review.


tracernz:
Does anybody actually want corner arming? Another possible patch would be to remove it.

tracernz:
We should document exactly how corner arming should work somewhere in user docs, because it was never clear to me (I always thought outer-bottom corners should arm, and inner bottom corners disarm or something... which would seem a lot better to me).

tracernz (approved):
This looks good.

PR #2063, opened 2018-02-11
Flight:Airspeed: Add option to use raw ADC value

tracernz:
Can be "borrowed" for other use-cases, already plumbed through telemetry layers etc. Requested by IRC user us1m41 for his wind angle sensor.

### DIFF
--- a/flight/Modules/Airspeed/airspeed.c
+++ b/flight/Modules/Airspeed/airspeed.c
@@ -93,7 +93,7 @@
 static struct pios_thread *taskHandle;
 static bool module_enabled = false;
 volatile bool gpsNew = false;
-volatile bool settingsUpdated = false;
+volatile bool settingsUpdated = true;
 static uint8_t airspeedSensorType;
 static uint16_t gpsSamplePeriod_ms;
 
--- a/flight/Modules/Airspeed/airspeed.c
+++ b/flight/Modules/Airspeed/airspeed.c
@@ -82,7 +82,6 @@
 #define GPS_AIRSPEED_BIAS_KP                   0.01f //Needs to be settable in a UAVO
 #define GPS_AIRSPEED_BIAS_KI                   0.01f //Needs to be settable in a UAVO
 #define GPS_AIRSPEED_TIME_CONSTANT_MS        500.0f  //Needs to be settable in a UAVO
-#define BARO_TRUEAIRSPEED_TIME_CONSTANT_S      1.0f  //Needs to be settable in a UAVO
 
 #define BARO_TEMPERATURE_OFFSET 5
 
@@ -95,6 +94,7 @@ volatile bool gpsNew = false;
 volatile bool settingsUpdated = true;
 static uint8_t airspeedSensorType;
 static uint16_t gpsSamplePeriod_ms;
+static float tasFilterTau = 0.1f;
 
 #ifdef BARO_AIRSPEED_PRESENT
 static int8_t airspeedADCPin = -1;
@@ -133,6 +133,7 @@ static void doSettingsUpdate()
 
 	airspeedSensorType = settings.AirspeedSensorType;
 	gpsSamplePeriod_ms = settings.GPSSamplePeriod_ms;
+	tasFilterTau = settings.TASFilterTau;
 
 #if defined(PIOS_INCLUDE_MPXV7002)
 	if (airspeedSensorType==AIRSPEEDSETTINGS_AIRSPEEDSENSORTYPE_DIYDRONESMPXV7002) {
@@ -349,7 +350,7 @@ static void airspeedTask(void *parameters)
  #ifdef BARO_AIRSPEED_PRESENT
 		else if (airspeedData.BaroConnected==BAROAIRSPEED_BAROCONNECTED_TRUE) {
 			//No GPS velocity estimate this loop, so filter true airspeed data with baro airspeed
-			float alpha = delT/(delT + BARO_TRUEAIRSPEED_TIME_CONSTANT_S); //Low pass filter.
+			float alpha = delT/(delT + tasFilterTau); //Low pass filter.
 			airspeedData.TrueAirspeed = airspeed_tas_baro*(alpha) + airspeedData.TrueAirspeed*(1.0f-alpha);
 		}
  #endif
--- a/flight/Modules/Airspeed/airspeed.c
+++ b/flight/Modules/Airspeed/airspeed.c
@@ -1,9 +1,9 @@
 /**
  ******************************************************************************
  * @addtogroup Modules Modules
- * @{ 
+ * @{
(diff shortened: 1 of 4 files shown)


tracernz:
build artifacts please

flight/Modules/Airspeed/baro_airspeed_analog.c
```diff
@@ -124,6 +124,10 @@ void baro_airspeedGetAnalog(BaroAirspeedData *baroAirspeedData, uint32_t *lastSy
 		//not. This is something that will have to change on the ADC side of things.
 		baroAirspeedData->SensorValue=PIOS_MPXV5004_Measure(airspeedADCPin);
 
+	} else if (airspeedSensorType == AIRSPEEDSETTINGS_AIRSPEEDSENSORTYPE_RAWANALOG) {
+		// will return raw ADC reading in mV (typically in range [0, 3300])
```
mlyle: I think this is 0,4095

mlyle (approved):
This looks good, though any module that actually uses airspeed info (fixedwingpathfollower) will be surprised.  Need to come up with a generic mechanism to expose ADC data.

flight/Modules/Airspeed/baro_airspeed_analog.c
tracernz: Ahhh, I may have been thinking of GetVolts.

tracernz:
>This looks good, though any module that actually uses airspeed info (fixedwingpathfollower) will be surprised. Need to come up with a generic mechanism to expose ADC data.

Yep, I was thinking about that. Would need perhaps some kind of multiplexing system to select which modules etc. get their data from where?

mlyle:

us1m41:
Thank you for the attempt @tracernz .  Doesn't seem to be working as hoped and it could be many things.  I will likely reach back out to you sometime when you aren't busy trying to ship a release.  

Based on how you hacked this in, no other settings matter (calibration value or calibration zero point) correct?

at one point, I saw some values flash up on the X9D screen for a sample or so, but now nothing.  

(Obviously, the second batch or artifacts worked correctly using the .dmg)

us1m41:
Any chance I can interest anyone into taking a look at this again?

tracernz:
I'll take another look but can't promise anything in the next couple of days.

us1m41:
Awesome, thank you and understand the "time" thing!

I have been looking through the code, both in how the MPXV7002 calibration and reading are supposed to work and there are a couple interesting things.  I was trying to find a way to better pretend my sensor is a either the MPXV7002 or MPXV5004.  Both of these sensors are 0-5 volt sensors, but I thought the ADC could only take up to 3.3 volts?  now that I ave a better idea of how I think the calibration routines are trying to work, I will spend some time trying to fake being an MPXV5004 and see if I get signs of life.  

I also saw this comment in "baro_airspeed_analog.c" to be interesting as it lined up with some of the behaviors I briefly saw.

		//Moreover, due to the way the ADC driver is currently written, this code will return 0 more often than
		//not. This is something that will have to change on the ADC side of things."

mlyle:
Pushed to this (and sorry, rebased, since I screwed it up the first time / was on a prm/ branch).

Includes a fix needed to make this work (properly getting airspeed settings), a couple other fixes that improve this (ADC code improvement, and disabling GPS corrections to raw analog speed), and a few airspeed module cleanups.

mlyle:

mlyle:
One last dumb issue that only affected the brain target.  Corrected.

tracernz:
@us1m41 ^^

flight/Modules/Airspeed/airspeed.c
```diff
@@ -261,77 +258,78 @@ static void airspeedTask(void *parameters)
...
+							&air_STP) + airspeedErrInt * GPS_AIRSPEED_BIAS_KI;
+ #endif
+		} else
 #endif
 		{ //Have to catch the fallthrough, or else this loop will monopolize the processor!
-			airspeedData.BaroConnected=BAROAIRSPEED_BAROCONNECTED_FALSE;
-			airspeedData.SensorValue=12345;
-			
+			airspeedData.BaroConnected = BAROAIRSPEED_BAROCONNECTED_FALSE;
+			airspeedData.SensorValue = 12345;
```
tracernz, line 275: lol

flight/PiOS/STM32F4xx/pios_internal_adc.c
```diff
@@ -296,7 +295,8 @@ static int32_t PIOS_INTERNAL_ADC_PinGet(uintptr_t internal_adc_id, uint32_t pin)
 	}
 
 	/* return accumulated result and clear accumulator */
-	result = accumulator[pin].accumulator / (accumulator[pin].count ?: 1);
+	result = accumulator[pin].accumulator /
+		(accumulator[pin].count ? accumulator[pin].count : 1);
```
tracernz, line 299: IMO it was more readable before on one line. :stuck_out_tongue:

flight/Modules/Airspeed/airspeed.c
```diff
@@ -287,7 +287,7 @@ static void airspeedTask(void *parameters)
 		float delT = (lastSysTime - lastLoopTime) / 1000.0f;
 		lastLoopTime = lastSysTime;
 		if ( ((lastSysTime - lastGPSTime) > 1000 || airspeedSensorType==AIRSPEEDSETTINGS_AIRSPEEDSENSORTYPE_GPSONLY)
-				&& gpsNew) {
+				&& gpsNew && airspeedSensorType != AIRSPEEDSETTINGS_AIRSPEEDSENSORTYPE_RAWANALOG) {
```
tracernz: Nice. :+1:

tracernz (commented):
Seems I can't approve my own PR (despite only being the author of 1/7 commits). Anyway, looks good, nice improvements. :+1:

flight/PiOS/STM32F4xx/pios_internal_adc.c
mlyle: I broke it into a new line because the short-circuit ternary is a gnu extension that I'd rather we not use.

tracernz: Fair enough, agree we should avoid extensions where reasonable.

us1m41:
@mlyle and @tracernz 

Thank you for putting this together.  It works perfectly. 

It looks like OpenTx is electing to smooth/average this value somehow and I will work on this on my end (as I can see that the sensor value updates instantly in GCS).  Thank you very much for the help!

mlyle:
Pushed a change: Allow configuration of time constant for outer trueairspeed filter.  This is more and more of a hack.  Jenkins, build artifacts please.

tracernz:
Latest commit looks good. :+1:

PR #2068, opened 2018-02-12
agcs: bump versions for Wired

mlyle:
Also remove support for all versions older than artifice.


mlyle:
Will likely self-merge this in a couple hours if no vehement objections.

PR #2070, opened 2018-02-13
manualcontrol: fix non-corner arming

mlyle:


mlyle:
Confirmed OK with switch arming.

mlyle:
Confirmed OK with yaw arming.

mlyle:
Bench confirmed OK with corner arming.

mlyle:
Self-merge.

PR #2072, opened 2018-02-13
release tools: properly tag udev rules

mlyle:
Helpful for forthcoming releases, if we keep putting udev rules in the
artifacts like this.

### DIFF
--- a/package/release-asset-info.py
+++ b/package/release-asset-info.py
@@ -84,6 +84,8 @@ def main():
             label = "Windows (installer)"
         elif name.endswith('.dmg'):
             label = "macOS"
+        elif name.endswith('.udev'):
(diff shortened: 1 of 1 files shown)


tracernz:
That was a one-off thing I thought?

tracernz:
Or did we forget to package them? :o

tracernz:
The rules are contained in both Linux packages now so there should be no need to upload them separately. That said, this won't hurt so...?

tracernz:
See #1899 for history and reason for uploading them separately last time.

mlyle:
Alright, can just close it.  In the meantime I uploaded the udev rules because I used the last release as a template :P

PR #2073, opened 2018-02-15
pios: fix semicolons and assert statements

mlyle:
Prevents confusion or problems with any future implementations of
assert.

### DIFF
--- a/flight/PiOS/STM32/pios_i2c.c
+++ b/flight/PiOS/STM32/pios_i2c.c
@@ -225,7 +225,7 @@ static void i2c_adapter_reset_bus(struct pios_i2c_adapter *i2c_adapter)
 int32_t PIOS_I2C_CheckClear(pios_i2c_t i2c_adapter)
 {
 	bool valid = PIOS_I2C_validate(i2c_adapter);
-	PIOS_Assert(valid)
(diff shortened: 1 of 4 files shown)


tracernz:
It's unfortunate we can't really structure the expression in such a way that doesn't create an empty statement, with the bonus of failing compilation if the ; is missing. Some compiler flags will warn for this empty statement, including the ones Qt Creator's static analysis engine uses by default (it is handy for flight code sometimes too).

P.S. Not a bug. :stuck_out_tongue:

mlyle:
We can-- they should be do { optionalexpressionifenabled } while(0)

mlyle:
I went ahead and did the do/whiles and caught a few more.

tracernz (approved):
Ready to start merging to next again?

mlyle:
Yes :). Let's break some shit. :D

(Maybe another week before changing big build or tools things)

tracernz:
> (Maybe another week before changing big build or tools things)

I'm thinking about leaving Qt to later in the cycle, so we can have the most bugfixes while still only changing once (I'm planning to stay on 5.9). Not sure if we want to move gcc-arm. Will discuss in a few days when you're back.

PR #2074, opened 2018-02-16
Flight: Make ccache separate BUILD_CONFIG option as for GCS

tracernz:
Default action if FLIGHT_BUILD_CONF is unspecified is the same as before, and also FLIGHT_BUILD_CONF=release is the same. debug/default config on their own will not use ccache.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ include $(ROOT_DIR)/flight/targets/*/target-defs.mk
 GCS_BUILD_CONF ?= debug ccache
 
 # And the flight build configuration (debug | default | release)
-export FLIGHT_BUILD_CONF ?= default
+export FLIGHT_BUILD_CONF ?= default ccache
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -104,16 +104,6 @@ ifdef GCS_BUILD_CONF
...
-ifdef FLIGHT_BUILD_CONF
- ifneq ($(FLIGHT_BUILD_CONF), release)
-  ifneq ($(FLIGHT_BUILD_CONF), debug)
-   ifneq ($(FLIGHT_BUILD_CONF), default)
-    $(error Only debug or release are allowed for FLIGHT_BUILD_CONF)
-   endif
-  endif
- endif
-endif
-
```
tracernz, line 116: This check, with a better error message, is already in firmware-defs.mk

mlyle:

PR #2079, opened 2018-02-23
flightD: Fix serial driver baud rate selection

jihlein:
Two issues found wrt using serial devices on flightd:

1)When the serial port is initialized, there is no attempt made to setup the port baud rate with the value specified in the Module Settings UAVO.  Corrections made in PIOS_SYS.c.

2)The function that sets the serial baud rate does not seem to work.  Rewrote part if it, now seems to function properly.  Corrections made in PIOS_SERIAL.c.  I left my debug print statements in the code as they indicate what baud rates are being set to the various ports, but they can be removed from the merged code if desired.

Tested this with a pair of telemetry radios set at 57600 baud.  Prior to these changes, I could not establish communication between the FC and GCS, regardless of the baud rate settings for telemetry.  After these changes, with the default telemetry setting of 115200, I still could not get FC <-> GCS communication.  After setting the telemetry baud rate to 57600 and restarting flightD, the port was successfully set to 57600 baud and I was able to connect the FC <-> GCS over the telemetry radios.

The command line used for testing was:

./build/flightd/flightd -r-s /dev/spidev0 -d ms5611:0:4 -d bmx055:0:1 -d flyingpio:0:0 -S telemetry:/dev/ttyUSB0

I have not tested MSP or light telemetry, but the code is identical so it should function correctly.  I've not fully tested the GPS, there are some other issues preventing auto baud rate selection and auto configuration from working, which I have fixes for, and will create another pull request for.

-------------

EDIT: While working on the GPS pull request, I think I found a cleaner, simpler way to implement this.  Please hold off on review and let me revisit this PR.

### DIFF
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -1229,7 +1229,7 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
 		HwSharedSpeedBpsOptions speed) {
 	switch (speed) {
 		case HWSHARED_SPEEDBPS_1200:
-			PIOS_COM_ChangeBaud(com_id, 2400);
+			PIOS_COM_ChangeBaud(com_id, 1200);
 			break;
 		case HWSHARED_SPEEDBPS_2400:
 			PIOS_COM_ChangeBaud(com_id, 2400);
--- a/flight/PiOS/posix/pios_serial.c
+++ b/flight/PiOS/posix/pios_serial.c
@@ -215,31 +215,31 @@ void PIOS_SERIAL_ChangeBaud(uintptr_t serial_id, uint32_t baud)
 
 	switch (baud) {
 		case 1200:
-			printf("Setting Serial ID 0x%p to 1200 bps\n",
+			printf("Setting Serial ID %p to 1200 bps\n",
 					ser_dev);
 			cfsetispeed(&options, B1200);
 			cfsetospeed(&options, B1200);
 			break;
 		case 2400:
-			printf("Setting Serial ID 0x%p to 2400 bps\n",
+			printf("Setting Serial ID %p to 2400 bps\n",
 					ser_dev);
 			cfsetispeed(&options, B2400);
 			cfsetospeed(&options, B2400);
 			break;
 		case 4800:
-			printf("Setting Serial ID 0x%p to 4800 bps\n",
+			printf("Setting Serial ID %p to 4800 bps\n",
 					ser_dev);
 			cfsetispeed(&options, B4800);
 			cfsetospeed(&options, B4800);
 			break;
 		case 9600:
-			printf("Setting Serial ID 0x%p to 9600 bps\n",
+			printf("Setting Serial ID %p to 9600 bps\n",
 					ser_dev);
 			cfsetispeed(&options, B9600);
 			cfsetospeed(&options, B9600);
 			break;
 		case 19200:
-			printf("Setting Serial ID 0x%p to 19200 bps\n",
+			printf("Setting Serial ID %p to 19200 bps\n",
 					ser_dev);
 			cfsetispeed(&options, B19200);
 			cfsetospeed(&options, B19200);
--- a/flight/PiOS/posix/pios_serial.c
+++ b/flight/PiOS/posix/pios_serial.c
@@ -48,11 +48,8 @@
 #include <linux/serial.h>
 #include <sys/ioctl.h>
 
-#include "modulesettings.h"
-
 /* Provide a COM driver */
-void PIOS_SERIAL_ChangeBaud(uintptr_t udp_id, uint32_t baud);
-
+static void PIOS_SERIAL_ChangeBaud(uintptr_t serial_id, uint32_t baud);
 static void PIOS_SERIAL_RegisterRxCallback(uintptr_t udp_id, pios_com_callback rx_in_cb, uintptr_t context);
 static void PIOS_SERIAL_RegisterTxCallback(uintptr_t udp_id, pios_com_callback tx_out_cb, uintptr_t context);
 static void PIOS_SERIAL_TxStart(uintptr_t udp_id, uint16_t tx_bytes_avail);
@@ -167,7 +164,7 @@ int32_t PIOS_SERIAL_InitFromFd(uintptr_t *serial_id, int readfd,
 
 	*serial_id = (uintptr_t) ser_dev;
 
-	PIOS_SERIAL_ChangeBaud(*serial_id, HWSHARED_SPEEDBPS_9600);
+	PIOS_SERIAL_ChangeBaud(*serial_id, 9600);
 
 	return 0;
 }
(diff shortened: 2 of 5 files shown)


tracernz (approved):
Looks good.

jihlein:
@tracernz, @mlyle, don't merge this yet, I can't get the GPS to work.  I can't tell if I've got a code error or a hardware failure at the moment.

jihlein:
Okay the latest changes allow me to have both the serial telemetry radio and GPS connected and running.  I have a GPS that refuses to talk UBLOX at the moment, and that was the source of my confusion.  During troubleshooting that, I used a logic analyzer to verify the serial port baud rates, and they are correct, showing the proper timing for the given baud rate.  Going to see if I can figure out why this particular GPS will only work in NMEA mode, but for now I'm satisfied that this change works as expected.

jihlein:
Discovered that the GPS works fine in UBLOX mode, just not on the RPi/FlyingPi (tested on an AQ32 with Wired, and it worked fine in both NMEA and UBLOX modes).  Given that, I'll declare this PR ready for merge after a final review, and I'll continue to work on the GPS issue and cut a new PR when it's fixed.

mlyle:
Ok.  I really need to review.  Note that I have unlock GPS working with
current code.  I am unable to read code here, but it is my recollection the
actual GPS module does a proper changebaud etc itself in its start and
autobaud paths and shouldn't be done by a special path in the flyingpi.

jihlein:
Maybe it's something odd with my kernel install, because the current code will not set baud rates correctly, but with the proposed changes, it does.  Verified this behavior with the logic analyzer.

At the moment, serial telemetry is rock solid with 3DR telem radios.  GPS is iffy, sometimes works, sometimes doesn't.  The GPS is an older UBLOX 6 that I know works on STM boards.  I'm going to try a new out of the box UBLOX 8 shortly.

Curious to see what you find in review.

flight/PiOS/posix/pios_sys.c
```diff
@@ -232,6 +239,8 @@ static int handle_serial_device(const char *optarg) {
 		goto fail;
 	}
```
mlyle: This shouldn't be here, because of https://github.com/d-ronin/dRonin/blob/next/flight/Modules/GPS/GPS.c#L322
https://github.com/d-ronin/dRonin/blob/next/flight/Modules/Telemetry/telemetry.c#L903
etc

flight/PiOS/posix/pios_serial.c
```diff
@@ -206,48 +206,62 @@ void PIOS_SERIAL_ChangeBaud(uintptr_t serial_id, uint32_t baud)
...
+			cfsetispeed(&options, B115200);
+			cfsetospeed(&options, B115200);
+			break;
+		case 230400:
+			printf("Setting Serial ID 0x%x to 230400 Baud\n", (uint32_t)serial_id);
+			cfsetispeed(&options, B230400);
+			cfsetospeed(&options, B230400);
+			break;
+		default:
+			printf("Defaulting Serial ID  0x%x to 9600 Baud\n", (uint32_t)serial_id);
```
mlyle: It's probably good to support all the linux "built in" rates, but this loses the custom_divisor capability / the ability to set e.g. 100kbps for sbus.

The functionality we use on the old code works for me (will reconfirm when I'm back home tomorrow).  Does it complain in dmesg "flightd sets custom speed on ttyS0.  This is deprecated."?  (The functionality is deprecated, but it is in current Linus trees without a universal replacement, and this type of thing is *never* removed because it breaks userspace).

flight/PiOS/Common/pios_hal.c
```diff
@@ -1229,7 +1229,7 @@ void PIOS_HAL_ConfigureSerialSpeed(uintptr_t com_id,
 		HwSharedSpeedBpsOptions speed) {
 	switch (speed) {
 		case HWSHARED_SPEEDBPS_1200:
-			PIOS_COM_ChangeBaud(com_id, 2400);
+			PIOS_COM_ChangeBaud(com_id, 1200);
```
mlyle, line 1232: Good catch!

flight/PiOS/posix/pios_serial.c
jihlein: Here's what happens with the original pios_serial.c.  Things appear to error out when trying to set the baud rate for ttyUSB0, I can't even connect to the GCS via IP Network telemetry.

image

But if I remove the argument for the usb telem radios and try to run, flightd locks up, that's why I originally went down this path.  

Flightd does not talk to directly to the SBUS receiver from what I see, or is that a future enhancement?

flight/PiOS/posix/pios_sys.c
jihlein: Okay, I missed that, but it doesn't seem to be calling it correctly, as I see no evidence it has called PIOS_SERIAL_ChangeBaud in ...\PiOS\posix\pios_serial.c.  The GPS module does make the call correctly.  Confused......

jihlein: Could my issues/confusion stem from the fact I've got GPS hooked up thru the serial port, and the rf telemetry hooked up via the USB port, ttyAMA0 and ttyUSB0 respectively?

flight/PiOS/posix/pios_serial.c
mlyle: > But if I remove the argument for the usb telem radios and try to run, flightd locks up, that's why I originally went down this path.

Without the last -S you can't connect to it by ip address, TCP, over 9000?  It would probably be best to figure this out first.

It makes more sense that this is a USB device that you're not able to set the rate of this way-- almost all of the drivers in linux/drivers/usb/serial support it but some don't.  This is why it'd be good to use the builtin rate constants but still fallback to the current behavior for custom rates (which won't work on USB devices not supporting TIOCSSERIAL)

> Flightd does not talk to directly to the SBUS receiver from what I see, or is that a future enhancement?

There's no code in the sys init for it checked in, but it works now that RTC is somewhat sane in flightd-- I've tried it.

flight/PiOS/posix/pios_sys.c
mlyle: It should all still work (though the TIOCSERIAL baud rate thing for the USB device is an issue, still).

jihlein: Okay, the main issue I was experiencing is beginning to become clearer, I think.  Without the changes in lines 227 and 242, there is no call made to PIOS_SERIAL_ChangeBaud by the telemetry subsystem.  The com_driver for the posix system is setup in line 197, and I believe the telemetry module is already initialized.   GPS works correctly because the module is enabled in line 214, so PIOS_SERIAL_ChangeBaud is correctly called during GPS init and config.

mlyle: I still am not able to debug, but that still seems strange.

In flightd/fw/main.c initTask SYS_Init and SYS_Args are called before MODULE_INITIALISE_ALL and MODULE_TASKCREATE_ALL.

Ah, I think I see it.  Will open a PR for discussion.

mlyle: Does #2085 help / seem appropriate?

jihlein: Will review it later, and test it this evening.

jihlein:
Considering closing this PR.  #2085 fixes part of the problem.  The pios_serial.c changes are still required to change USB baud rates without an "ioctl-TIOCGSERIAL: Inappropriate ioctl for device" error being thrown.  I'll just put the pios_serial.c changes on my local branch for now so I can use the USB ports for telemetry.  Sound good?

We'll still need to catch the pios_hal change somewhere.....

mlyle:
@jihlein -- OK if I rewrite your branch and adjust it to only use the custom rate code if necessary and keep the 1200 baud change?

jihlein:
@mlyle - certainly, have at it!

mlyle:
OK, thanks!  This is my best cut at it, but please note it does not include the other change.

jihlein:
Well after all is said and done, narrowed this down to changes in 2 files.  Thanks for the pios_serial.c changes @mlyle.  I've rebased this with the next so it build clean with all the latest changes.  The testing I've done is:

1)Clean build of flightd on RPi3.  Verified proper calls to PIOS_SERIAL_ChangeBaud for ttyAMA0 and ttyUSB0 ports.  Verified that GCS connects both via TCP and Serial paths, and that GPS is functional.  Placed both serial RF telemetry and GPS at 38400 baud for testing purposes.  Here's a screen shot of flightd initialization:

image

2)Clean build of flightd on Windows 7.  Verified operation and localhost connection to GCS.

What I haven't tested:

1)Any of the non standard baud rates on either ttyAMA0 or ttyUSB0.

Here's what the bench test setup looks like:

image

mlyle:
Thanks!

PR #2083, opened 2018-02-25
uavsettings/upgrader: handle null settings well

### LINKED ISSUE #1697: Handling of upgrade of unitialized FC could be improved

mlyle:
Now allow the user to "get out" after the dialog shows up.  Fixes #1697.


mlyle:
This could be considered a bit of a hack, but if the return value is treated as whether the user has "completed the action" to the extent possible, it's sane.

mlyle:
jenkins, test this please

tracernz:
What does this do for the non-upgrader case? (harder to find that without a local checkout sorry)

mlyle:
No one cares about the return value otherwise-- only upgrader.

tracernz (approved):
Thanks for checking. :+1:

PR #2085, opened 2018-02-26
flightd/posix: swap telem ids

mlyle:
Otherwise, initialization of serial baud rate doesn't happen properly
because we use the "USB" handle which is assumed to have fixed rate.

Addresses some issues in #2079


mlyle:
I've confirmed this works for GCS TCP connection.  Not able to test serial now.

jihlein:
Works for me on both TCP and serial.  And FWIW  if I use the pios_serial.c changes from #2079, I can also use USB serial and successfully change baud rates.

jihlein:
Since it appears to be a simple change and it has tested correctly, ready for merge?

mlyle:
Yes please

tracernz (approved):

PR #2086, opened 2018-02-26
GCS:Eigen: Only use gcc Wno-int-in-bool-context flag on GCC

tracernz:
Fix up some warnings with clang as it doesn't have this flag.


tracernz:
build artifacts please

tracernz:
Need to nosy at builder logs to check this. Can't access port 8080 from here unfortunately.

tracernz:
Build output better now. :+1:

PR #2087, opened 2018-02-28
osd: show "arm.." for armed flag during arming hold

mlyle:
@ufoDziner had an issue where he'd see "ARMED" on the screen and
throttle up, and this would cancel the arming.  Now only display "ARMED"
after actual arm, and "arm.." before that.

### DIFF
--- a/flight/Modules/OnScreenDisplay/onscreendisplay.c
+++ b/flight/Modules/OnScreenDisplay/onscreendisplay.c
@@ -1170,8 +1170,11 @@ void render_user_page(OnScreenDisplayPageSettingsData * page)
 	// Arming Status
 	if (page->ArmStatus) {
 		FlightStatusArmedGet(&tmp_uint8);
-		if (tmp_uint8 != FLIGHTSTATUS_ARMED_DISARMED)
+		if (tmp_uint8 == FLIGHTSTATUS_ARMED_ARMED)
 			write_string("ARMED", page->ArmStatusPosX, page->ArmStatusPosY, 0, 0, TEXT_VA_TOP, (int)page->ArmStatusAlign, 0,
+					page->ArmStatusFont);
+		else if (tmp_uint8 == FLIGHTSTATUS_ARMED_ARMING)
+			write_string("arm..", page->ArmStatusPosX, page->ArmStatusPosY, 0, 0, TEXT_VA_TOP, (int)page->ArmStatusAlign, 0,
 					page->ArmStatusFont);
 	}
 


ufoDziner:
Sorry I'm so needy... but thanks!

mlyle:
Nah, this is a bug and has bothered me, too.

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -1170,9 +1170,12 @@ void render_user_page(OnScreenDisplayPageSettingsData * page)
 	// Arming Status
 	if (page->ArmStatus) {
 		FlightStatusArmedGet(&tmp_uint8);
-		if (tmp_uint8 != FLIGHTSTATUS_ARMED_DISARMED)
+		if (tmp_uint8 == FLIGHTSTATUS_ARMED_ARMED)
 			write_string("ARMED", page->ArmStatusPosX, page->ArmStatusPosY, 0, 0, TEXT_VA_TOP, (int)page->ArmStatusAlign, 0,
 					page->ArmStatusFont);
+		else if (tmp_uint8 == FLIGHTSTATUS_ARMED_ARMING)
+			write_string("arm..", page->ArmStatusPosX, page->ArmStatusPosY, 0, 0, TEXT_VA_TOP, (int)page->ArmStatusAlign, 0,
```
tracernz, line 1177: All the other modes etc. are uppercase?

mlyle: Yeah-- I wanted to make it obvious it's not armed, and not exceed 5 chars to not break peoples' layouts.  ARM.. won't look that different.  If you have a better suggestion...

ufoDziner: I know I'm not the expert here, so feel free to LMK if this isn't great.  I was just thinking that it would be neat to have it say "arm.." for the first half of the arming delay, and "..ing" for the second half.  Not sure how that would actually look on the OSD in real time, or if it's practical to program.  With this, it could be uppercase if that is a standard that should be adhered to.  Just a thought.

mlyle: @ufoDziner  --- it's a good idea but unfortunately at this place in the system we don't really know how arming works, other than it's in progress, and it's probably not worth too much effort to plumb additional information through.

tracernz: Fair enough. :+1:

tracernz (approved):

PR #2089, opened 2018-03-01
pios_servo: set maximum number of banks to 8

### LINKED ISSUE #2088: Quanton: Inport: Outputs+ADC boot loops
Crashes on quanton / boot loops

```
      <field name="InPort" values="Outputs+ADC"/>
```

reported here; 

https://forum.dronin.org/forum/d/501-bootloop-after-update-to-wired/7

mlyle:
Because Quanton in some modes has 8.  Never mind that we can't configure
more than 6.

Also cope with / avoid stack overflow in this case, and fix a small issue with channel masking.

Fixes #2088.

### DIFF
--- a/flight/PiOS/STM32/pios_servo.c
+++ b/flight/PiOS/STM32/pios_servo.c
@@ -868,7 +868,7 @@ void PIOS_Servo_Update(void)
 	// If some banks are oneshot and some are dshot (why would you do this?)
 	// get the oneshots firing first.
 	for (uint8_t i = 0; i < servo_cfg->num_channels; i++) {
-		if (channel_mask & (1 << 1)) {
+		if (channel_mask & (1 << i)) {
 			continue;
 		}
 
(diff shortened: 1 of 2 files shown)


flight/PiOS/STM32/pios_servo.c
```diff
@@ -336,12 +336,17 @@ int PIOS_Servo_SetMode(const uint16_t *out_rate, const int banks, const uint16_t
 				bank = j;
 		}
 
-		if (bank < 0)
-			bank = banks_found++;
+		if (bank < 0) {
```
tracernz: Nit pick.
`if (bank < 0 && banks_found < PIOS_SERVO_MAX_BANKS)`

flight/PiOS/STM32/pios_servo.c
```diff
@@ -868,7 +873,7 @@ void PIOS_Servo_Update(void)
 	// If some banks are oneshot and some are dshot (why would you do this?)
 	// get the oneshots firing first.
 	for (uint8_t i = 0; i < servo_cfg->num_channels; i++) {
-		if (channel_mask & (1 << 1)) {
+		if (channel_mask & (1 << i)) {
```
tracernz, line 874: This looks like a bugfix... but nothing in commit message about it?

mlyle: Fixed both, thanks :D

tracernz (approved):
Thanks. :+1:

PR #2090, opened 2018-03-02
Factor althold logic into stabilization

mlyle:
Has tested in simulator.

Remaining work items

* [x] Cleaning up config dirtiness
* [x] Making sanity check sane
* [x] Not building it on targets that'll never need it
* ~~Moving VTOLPathFollower to use this method of altitude control~~ [defer]
* [x] Sim testing
* [x] Bench testing
* [x] Flight testing


mlyle:
Tested on dtfc to not allow altitudehold module.

mlyle:
Tested on sparky2 to apparently do the right thing with thrust when held in hand.

mlyle:
Retested on DTFc to properly mask althold mode.

mlyle:
Tested on Revolution with no baro to properly mask althold mode.

mlyle:
Confirmed on seppuku w/ baro hacked out that the mode is sanitychecked masked...

mlyle:
Need to confirm on bench that F3 CPU use is OK in althold mode, since this does more work.  Need to flight test.

mlyle:
SPRF3E uses 68% CPU in leveling, 73% CPU in althold, so not too bad.  (definitely worse than before, but hey, sampling faster is good right?)

Confirm SPRF3E with baro works on bench, without baro configerrors.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -375,6 +392,190 @@ static void calculate_attitude_errors(uint8_t *axis_mode, float *raw_input,
...
+		last_nonzero_thrust_time = 0;
+
+		return 0.0f;
+	}
+
+#ifdef TARGET_MAY_HAVE_BARO
+	bool do_alt_control = false;
+	bool do_vs_control = false;
+
+	static bool old_vs_control;
```
tracernz: This confused me, this is just the previous state, not an old algorithm or something? Suggest prev_vs_control, as per other previous state vars.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -375,6 +392,190 @@ static void calculate_attitude_errors(uint8_t *axis_mode, float *raw_input,
...
+	} else if (flightStatus->Armed != FLIGHTSTATUS_ARMED_ARMED) {
+		last_thrust_pos = true;
+		last_nonzero_thrust_time = 0;
+
+		return 0.0f;
+	}
+
+#ifdef TARGET_MAY_HAVE_BARO
+	bool do_alt_control = false;
+	bool do_vs_control = false;
```
tracernz: vertspeed_control, alti_control?

flight/Modules/Stabilization/stabilization.c
```diff
@@ -375,6 +392,190 @@ static void calculate_attitude_errors(uint8_t *axis_mode, float *raw_input,
...
+
+			// Add ability to scale up the amount of compensation to achieve
+			// level forward flight
+			fraction = powf(fraction, (float) altitudeHoldSettings.AttitudeComp / 100.0f);
+
+			// Dividing by the fraction remaining in the vertical
+			// projection will attempt to compensate for tilt.
+			// If the fraction is starting to go negative we are
+			// inverted and should shut off throttle (but keep
+			// stabilization).
```
tracernz, line 530: This doesn't seem right for vehicles with symmetrical thrust?

flight/Modules/Stabilization/stabilization.c
```diff
@@ -375,6 +392,190 @@ static void calculate_attitude_errors(uint8_t *axis_mode, float *raw_input,
...
+				0	// don't use slope-antiwindup
+				);
+
+		AltitudeHoldStateData altitudeHoldState;
+		altitudeHoldState.VelocityDesired = vs_desired;
+		altitudeHoldState.Integral = vertspeed_pid.iAccumulator;
+		altitudeHoldState.AngleGain = 1.0f;
+
+		if (altitudeHoldSettings.AttitudeComp > 0) {
+			// Thrust desired is at this point the mount desired in the up direction, we can
```
tracernz: typo in amount :P

flight/Modules/Stabilization/stabilization.c
```diff
@@ -375,6 +392,190 @@ static void calculate_attitude_errors(uint8_t *axis_mode, float *raw_input,
...
+			// Project a unit vector pointing up into the body frame and
+			// get the z component
+			float fraction = attitudeActual.q1 * attitudeActual.q1 -
+				attitudeActual.q2 * attitudeActual.q2 -
+				attitudeActual.q3 * attitudeActual.q3 +
+				attitudeActual.q4 * attitudeActual.q4;
+
+			// Add ability to scale up the amount of compensation to achieve
+			// level forward flight
+			fraction = powf(fraction, (float) altitudeHoldSettings.AttitudeComp / 100.0f);
```
tracernz: space after cast

flight/Modules/Stabilization/stabilization.c
```diff
@@ -390,16 +591,29 @@ static void stabilizationTask(void* parameters)
 	FlightStatusData flightStatus;
 	SystemSettingsAirframeTypeOptions airframe_type;
 
+	volatile bool settings_flag = true;
+	volatile bool flightstatus_flag = true;
```
tracernz: Suggest, _updated, flag is very ambiguous.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -390,16 +591,29 @@ static void stabilizationTask(void* parameters)
...
 	SubTrimSettingsConnectCallbackCtx(UAVObjCbSetFlag, &settings_flag);
+#ifdef TARGET_MAY_HAVE_BARO
+	AltitudeHoldSettingsConnectCallbackCtx(UAVObjCbSetFlag, &settings_flag);
+#else
+	/*
+	 * Sanity check uses this to know whether althold mode is permitted,
+	 * so we'd better not have a baro if the target doesn't "support"
+	 * one.
+	 */
+	PIOS_Assert(!PIOS_SENSORS_IsRegistered(PIOS_SENSOR_BARO));
```
tracernz, line 619: Would be better to put this into the sanity check IMO.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -632,7 +810,7 @@ static void stabilizationTask(void* parameters)
 		uint16_t max_safe_rate = PIOS_SENSORS_GetMaxGyro() * 0.9f;
 
 		//Run the selected stabilization algorithm on each axis:
-		for(uint8_t i=0; i< MAX_AXES; i++)
+		for (uint8_t i = 0; i < MAX_AXES; i++)
 		{
```
tracernz: If you're cleaning it up, move the curly-bracket up where it belongs. :D

tracernz (commented):
Good work.

flight/Modules/Stabilization/stabilization.c
mlyle: The sanity check is in a library and is built for all targets.  It doesn't know if MAY_HAVE_BARO is true, so there's no choice but to do it here.

flight/Modules/Stabilization/stabilization.c
mlyle: OK i just moved these didn't I? :P  But fine :P

flight/Modules/Stabilization/stabilization.c
mlyle: :)

flight/Modules/Stabilization/stabilization.c
mlyle: It's from the existing althold module.  If we want to make althold work inverted, or make it make a pulse-of-negative-thrust when inverted until crossing level, we can think about it.  (Right now, with it being a unistable attitude thing, doesn't make much sense)

flight/Modules/Stabilization/stabilization.c
mlyle: OK

flight/Modules/Stabilization/stabilization.c
mlyle: OK

flight/Modules/Stabilization/stabilization.c
mlyle: :D

flight/Modules/Stabilization/stabilization.c
tracernz: Yup. Would be super awesome to get that working one day. 😃

flight/Modules/Stabilization/stabilization.c
tracernz: Makes sense. :+1:

mlyle:
Flown on Sparky2 successfully with default coefficients.  It's clearly superior to old althold in accuracy and smoothness (I guess doing it faster than 50Hz makes a difference).

Also flew conservative "vertical autotune" coeffs, which are awesome.

mlyle:
https://gist.github.com/mlyle/5dbe97aa964d3709db3a11c189471d70

mlyle:
Removed an inadvertent change disabling baro on seppuku (used during sanity test).

tracernz (approved):
Changes look good, thanks.

PR #2091, opened 2018-03-04
sparky2: fix airspeed module reference

mlyle:
Due to a long-ago rename, it was not getting built into the target.


tracernz (approved):
Good catch :+1:

PR #2092, opened 2018-03-04
pid: Schedule I-gain on antiwindup controller, based on P-term's proximity to clipping.

glowtape:
Reduces rebounding _a lot_, when actuators lack authority.

Only applied to the antiwindup version used in stab.c.

### DIFF
--- a/flight/Libraries/math/pid.c
+++ b/flight/Libraries/math/pid.c
@@ -76,6 +76,7 @@ float pid_apply(struct pid *pid, const float err)
  * @param[in] err The error term
  * @param[in] min_bound The minimum output
  * @param[in] max_bound The maximum output
+ * @param[in] aw_bound The absolute boundary for slope-related anti-windup.
  * @param[in] dT  The time step
  * @returns Output the computed controller value
  *
@@ -83,7 +84,7 @@ float pid_apply(struct pid *pid, const float err)
  *  chapter.
  */
 float pid_apply_antiwindup(struct pid *pid, const float err,
-	float min_bound, float max_bound)
+	float min_bound, float max_bound, float aw_bound)
 {
 	float dT = pid->dT;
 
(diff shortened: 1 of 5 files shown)


flight/Libraries/math/pid.c
```diff
@@ -190,7 +190,8 @@ float pid_apply_setpoint_antiwindup(struct pid *pid,
 		// If Ki is zero, do not change the integrator. We do not reset to zero
 		// because sometimes the accumulator term is set externally
 	} else {
-		pid->iAccumulator += err * (pid->i * dT);
+		float p = bound_sym(err * pid->p, 1.0f);
```
mlyle: This needs to be bounded and scaled based on the actual range, not assuming its -1..1 --- this would break nav.

mlyle (changes requested):

flight/Libraries/math/pid.c
glowtape: Are there any expectations regarding min_bound and max_bound, e.g. min_bound always zero or negative, and max_bound always zero or positive?

glowtape: Should we scale based on the larger boundary from zero?
```c
float s = MAX(fabsf(min_bound), fabsf(max_bound));
float p = bound_min_max(err * pid->p, min_bound, max_bound) / s;
```

mlyle: You could just let the caller pass in the absolute antiwindup bound.  0 means "don't do this step".

I'm not sure this logic is beneficial for all antiwindup use cases.  For example, the output of the antiwindup controllers [in navigation] is an attitude-- there's no comparable "slope" limitation.  Yes, it's good not to wind up the I term when you'd be exceeding the attitude limit, but we already have that in the present controller.

glowtape:
Pending outside testing.

flight/Libraries/math/pid.c
```diff
@@ -76,22 +76,29 @@ float pid_apply(struct pid *pid, const float err)
...
 
 	if (pid->i == 0) {
 		// If Ki is zero, do not change the integrator. We do not reset to zero
 		// because sometimes the accumulator term is set externally
 	} else {
-		pid->iAccumulator += err * (pid->i * dT);
+		if (aw_bound > 0) {
+			float p = bound_sym(err * pid->p, aw_bound) / aw_bound;
+			pid->iAccumulator += err * (pid->i * dT) * (1 - p*p);
+		} else
```
mlyle: nitpick but should be no newline before {

mlyle (changes requested):

PR #2093, opened 2018-03-04
actuators: Leaky bucket time constant configuration, bucket content logging.

glowtape:
Added the ability to configure the leaky bucket time constant via UAVO.

Also added logging of the bucket contents. After the throttle chop stuff we had with ufo's quad, it's probably prudent to record what hangtime's doing.

### DIFF
--- a/shared/uavobjectdefinition/actuatorcommand.xml
+++ b/shared/uavobjectdefinition/actuatorcommand.xml
@@ -14,5 +14,10 @@
     <field defaultvalue="0" elements="1" name="MaxUpdateTime" type="uint16" units="ms">
       <description/>
     </field>
+    <field defaultvalue="0" elements="1" name="LowPowerStabilizationReserve" type="float" units="s">
+      <description>
+        The power reserve available that can be added to actuators during negative clipping.
+      </description>
+    </field>
   </object>
 </xml>
--- a/shared/uavobjectdefinition/actuatorsettings.xml
+++ b/shared/uavobjectdefinition/actuatorsettings.xml
@@ -30,6 +30,9 @@
     <field defaultvalue="0.07" elements="1" limits="%BE:0.0:0.15" name="LowPowerStabilizationMaxPowerAdd" type="float" units="">
       <description>Maximum power to add to ensure stabilization at low power.  Always select a value well under hover power.</description>
     </field>
+    <field defaultvalue="0.3" elements="1" limits="%BE:0.1:1.0" name="LowPowerStabilizationTimeConstant" type="float" units="s">
+      <description>Sets the time constant for the power reserve management during low power stabilization.</description>
+    </field>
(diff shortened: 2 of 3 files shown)


glowtape:
Doing a bunch of roll flips at hangtime, exercising the bucket (second graph):

bucket2

(Kind of curious why the bucket's filling when pumping throttle before a flip. The actuators aren't clipping. Max power add is 10% on my quad. --edit: Nevermind, I see it does that via softlimit.)

mlyle (approved):

shared/uavobjectdefinition/actuatorcommand.xml
```diff
@@ -14,5 +14,8 @@
     <field defaultvalue="0" elements="1" name="MaxUpdateTime" type="uint16" units="ms">
       <description/>
     </field>
+    <field defaultvalue="0" elements="1" name="LeakyBucketContent" type="float" units="">
+      <description />
```
tracernz: ??

shared/uavobjectdefinition/actuatorsettings.xml
```diff
@@ -30,6 +30,9 @@
     <field defaultvalue="0.07" elements="1" limits="%BE:0.0:0.15" name="LowPowerStabilizationMaxPowerAdd" type="float" units="">
       <description>Maximum power to add to ensure stabilization at low power.  Always select a value well under hover power.</description>
     </field>
+    <field defaultvalue="0.3" elements="1" limits="%BE:0.1:1.0" name="LowPowerStabilizationTimeConstant" type="float" units="">
+      <description>Sets the time constant for the power reserve management during low power stabilization.</description>
```
tracernz: Would be nice to at least give an idea of the scale/units so somebody can read this description and then have some idea how to set the value.

shared/uavobjectdefinition/actuatorcommand.xml
mlyle: Yah description would be good, perhaps better name.  Good point.

mlyle: Units on both this and the other are "seconds"

glowtape:
Renamed LeakyBucketContent to HangtimePowerReserve, and added units. The leaky bucket stuff is percents, not seconds, tho. :P

tracernz:
I hate to nitpick even more, but I'm going to anyway. :grin:  You've got "LowPowerStabilizationTimeConstant" in one UAVO and "HangtimePowerReserve" in the other... :P Can you please make the second one "LowPowerStabilizationReserve" or similar?

glowtape:
Naming things is hard D:

Will do later today.

tracernz:
> Naming things is hard D:

You're not wrong! Pretty important to help people connect the dots though IMO.

mlyle:
Leaky bucket stuff is seconds times effort, and there's no "100" to make it
a percent.  Unit is seconds.

mlyle:
thx :D

PR #2096, opened 2018-03-07
adc: improve resolution, correctness

### LINKED ISSUE #2095: F3/F4 ADC: Should left-adjust results
Right now, we basically return values from 0 to 4095, which matches the resolution of the internal ADC.  However, we have oversampling and averaging.

We should have the ADC left align the result (return values from 0 to 65535) so that our averaging can increase the resolution beyond the 12 bits that the native ADC provides.  In turn, we can reduce the per-volt value by a factor of 16.

This would affect RSSI (which uses raw values, but we could just scale down by 16), and the change in #2063 (ditto).  Battery would benefit and get a bit more than 12 bit resolution.

mlyle:
Fixes #2095.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -296,7 +296,12 @@ int32_t transmitter_control_update()
 		case MANUALCONTROLSETTINGS_RSSITYPE_ADC:
 #if defined(PIOS_INCLUDE_ADC)
 			if (settings.RssiChannelNumber > 0) {
-				value = PIOS_ADC_GetChannelRaw(settings.RssiChannelNumber - 1);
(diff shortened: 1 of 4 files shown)


mlyle:
Confirmed on revolution that this works with existing scale etc / battery module behaves the same.

mlyle:
Confirmed on lux that this works with existing scale etc.

mlyle:
Confirmed on seppuku w/ baro that the right thing happens.

mlyle:
Confirmed on seppuku w/ baro hacked out that the mode is sanitychecked masked...

tracernz (approved):
Nice. :+1:

mlyle:
Woops there were two erroneous "test" messages here that applied to a different PR, removed.

PR #2102, opened 2018-03-09
gcs controller: support arming switch

mlyle:
And a couple other things-- small code cleanup, remove UDP-control functionality, increase resolution (because why not?).

Note a lot more diff shows than is real because the second commit (only) changed line endings, so probably better to review the first and third commits.


tracernz (approved):
Looks good. :+1:

PR #2103, opened 2018-03-10
hangtime: when enabling, calculate power add value

### LINKED ISSUE #2098: Display hover % throttle next to hangtime info
* Seen in version (release version or githash): 
* On flight controller (if applicable): 

Issue details:
Would be great to have the craft's hover % throttle displayed near the Max Power Add portion of the GCS, and include a tooltip that suggests the recommended Max Power Add.

### LINKED ISSUE #2099: Missing sensors is broken??
When TolerateMissingSensors is true, good_run is still nailed to false, and as a result the startup SENSORS alarm never gets cleared...  This got regressed at some point.

mlyle:
Use the hover throttle from autotune * .85, when the user checks the hangtime box.  Sanity checks the range.

This isn't exactly what #2098  stipulates, but it's a good way to do it IMO.

Fixes #2098

### DIFF
--- a/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
+++ b/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
@@ -1242,7 +1242,7 @@ bool ConfigTaskWidget::setWidgetFromVariant(QWidget *widget, QVariant value, dou
         if (scale == 0)
             label->setText(value.toString() + units);
         else
-            label->setText(QString::number((value.toDouble() / scale)) + units);
+            label->setText(QString::number(value.toDouble() / scale, 'g', 3) + units);
         return true;
     } else if (QDoubleSpinBox *dblSpinBox = qobject_cast<QDoubleSpinBox *>(widget)) {
         dblSpinBox->setValue(value.toDouble() / scale);
--- a/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
+++ b/ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
@@ -1223,10 +1223,17 @@ bool ConfigTaskWidget::setWidgetFromVariant(QWidget *widget, QVariant value, dou
                                             QString units)
 {
     units = units.trimmed();
-    if (!units.isEmpty() && !qFuzzyCompare(1 + 1.0, 1 + scale) && scale != 0)
-        units = applyScaleToUnits(units, scale);
-    if (!units.isEmpty())
-        units.prepend(' ');
+    if (!units.startsWith("%")) {
+        if (!units.isEmpty() && !qFuzzyCompare(1 + 1.0, 1 + scale) && scale != 0)
+            units = applyScaleToUnits(units, scale);
+        if (!units.isEmpty())
+            units.prepend(' ');
+    } else {
+        /* We have a lot of things like % / 100 in the unit set.
+         * Best to make them just %.  (Assume they'll use scale properly)
+         */
+        units = QString("%");
+    }
 
     if (QComboBox *comboBox = qobject_cast<QComboBox *>(widget)) {
         comboBox->setCurrentIndex(comboBox->findData(value.toString()));
--- a/ground/gcs/src/plugins/config/configstabilizationwidget.cpp
+++ b/ground/gcs/src/plugins/config/configstabilizationwidget.cpp
@@ -32,6 +32,7 @@
 
 #include "configstabilizationwidget.h"
 #include "manualcontrolsettings.h"
+#include "systemident.h"
 
 #include <coreplugin/iboardtype.h>
 #include <uavobjectutil/uavobjectutilmanager.h>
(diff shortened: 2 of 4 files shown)


jhitesma:
I like this one because the code actually makes sense to me and I learned a few things from it 👍 

I also like that it raises the limit from 15 to 20...I think we've been overly conservative with that and it's made a lot of casual users dismiss hangtime as ineffective.

But...I'd still like to see a display of the measured hover% added.  Though I know @mlyle will still protest this since quite a few people will see a percentage higher than the input box will allow and get frustrated by that and unfortunately I don't have a good suggestion on how to address that very valid concern.

tracernz (approved):
Looks good, but not sure how it relates to #2099?

glowtape:
Considering it is half-depending on autotune for the hover throttle, I think it might be an idea to add a checkbox in the autotune wizard somewhere down the road, too. Will probably make it way more visible and probably used.

ufoDziner:
@jhitesma Addressing your display concern; can't the percentage be displayed with a tooltip that suggests the recommended value?  That would explain to the concerned parties why the hangtime value is lower than hover throttle.

jhitesma:
@ufoDziner possibly.  But....people don't read things that are right in front of them already let alone hidden behind a tooltip.  (We seem to get a lot of people asking why something is red in the status display without realizing they can hover to get details in a tooltip.)

I guess this is starting to get off topic and I should probably open a new issue.  But what I'm really thinking is maybe it's time to rethink the restrictions/limits on hangtime.  I know you recently found that much higher percentages solved your issue with losing control authority at zero throttle, and the new leaky bucket stuff makes it even safer.  And I've also found that I regularly wind up setting hangtime percent and time higher than the limits in GCS.  

But normal users don't really know that they can go into the browser and increase things higher than the GCS GUI allows.  

I understand why we have limits and @mlyle's concerns over preventing potential flyaways.  But...at the same time I think we may be overprotective and turning people off as a result.  I know the local guy who's a heck of a flyer quit flying us because he wasn't happy with hangtime and was scared to increase the percentage higher than the GCS allowed.  And I've seen a number of other people on forums and FB express similar sentiment.

Maybe we need to offer a way to unlock higher limits in the GUI after the user acknowledges the potential risk of what they're doing and only if they have a sane failsafe setup configured?

ufoDziner:
@jhitesma Yah... It's off topic from my request, but I'm not sure that it is with the added functionality added in this PR.

Let me first state that rigs I'm testing on all hover from 12%-16% throttle.  So, it's easily possible for me to go over 100% on one of them currently, and both with the increase to 20%.  Yes, the higher hangtime fixed some issues I was having with throttle chops, and I was also unwilling to make the adjustment without consulting @mlyle first.  That being said, I agree that for racer types, the risk may be overly emphasized.  We don't fly huge (both in size and cost) craft.

So, maybe there is a checkbox that acknowledges the risk with the increased percentage?

Also, where does the flyaway risk come from?  Stick arming perhaps?  I just don't see where things could get too crazy with switch arming.  Maybe there is a limit to be set with stick arming?

Either way, I'd like to just have the percentage displayed for reference.  I like that the system is going to offer me a setting, but if I change it to something else, I'd like to have a reference (without having to dig in the uavo stuff) when I want to tweak it some more.

mlyle:
> Also, where does the flyaway risk come from? Stick arming perhaps? I just don't see where things could get too crazy with switch arming. Maybe there is a limit to be set with stick arming?

It basically all comes down to overtuned PIDs.  If you're overtuned and maxpoweradd is too high, you might not be able to descend (oscillations add enough power to climb).  Worse, if you're a novice and expect yanking throttle down to help things, it won't...

glowtape:
Messing about testing the LQG stuff for integration, I came to the conclusion that a proper hangtime setup is kind of mandatory, to convince freestyle pilots, so it should probably go into the autotune wizard to make sure it's on and properly configured.

I used to ride around with ~10%, because at some point in the past half of hover was suggested. Seems that changed with the leaky bucket. Going from 10% to 17.5% (around 80% of hover) makes a pretty noticeable difference, with both control laws.

Also assuming that people are even looking for it. The location of it in the settings is kinda unobvious, if someone is looking for something like airmode. And people never read manuals.

mlyle:
If people don't read the setup page, they're not going to get us to work.  The autotune wizard is already crowded with lots of related things-- if we want to coerce users to do stuff that doesn't mean "cram it into something they're probably doing anyways" :/

glowtape:
What about moving the checkbox and spinner from the Misc to the Basic page? --edit: Err, Advanced page to Basic that is.

mlyle:
Yah, that's an option, but it'll be in scroll.  Might be worth moving it to basic page, and **removing PIDS from basic page**.  That might be the kind of thing to do a request-for-comments on the forum.

Anyways I don't think this PR is the place to solve all the usability issues.

mlyle:
One (hopefully) last rebase and artifact build needed ... Then verification of those artifacts

glowtape:
Right smack in the face where rates are. That'll make sure it gets enabled.

glowtape:
The spinner for duration stays grayed out when enabling Hangtime.

glowtape:
Nevermind, UAVO mismatch, works now.

mlyle:
screenshot from 2018-03-12 13-48-37

(Fixed the capitalization to Observed Hover Power after this screenshot

ground/gcs/src/plugins/uavobjectwidgetutils/configtaskwidget.cpp
```diff
@@ -1223,10 +1223,17 @@ bool ConfigTaskWidget::setWidgetFromVariant(QWidget *widget, QVariant value, dou
...
-        units.prepend(' ');
+    if (!units.startsWith("%")) {
+        if (!units.isEmpty() && !qFuzzyCompare(1 + 1.0, 1 + scale) && scale != 0)
+            units = applyScaleToUnits(units, scale);
+        if (!units.isEmpty())
+            units.prepend(' ');
+    } else {
+        /* We have a lot of things like % / 100 in the unit set.
+         * Best to make them just %.  (Assume they'll use scale properly)
+         */
```
tracernz, line 1234: Feels like a hack, but meh. :/

tracernz (approved):

PR #2105, opened 2018-03-10
gcs: autotune: compute vertical autotune

mlyle:
Use information on hover throttle (presuming this is 1G acceleration)
and roll/pitch tau (assuming this corresponds to thrust tau) in order to
calculate a conservative (but still more aggressive than stock tuning)
vertical tune.

Produces similar results to the defaults on things that use 50% hover
power with 65ms tau.

### DIFF
--- a/ground/gcs/src/plugins/config/autotunesliders.ui
+++ b/ground/gcs/src/plugins/config/autotunesliders.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>521</width>
-    <height>439</height>
+    <width>561</width>
+    <height>557</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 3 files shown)


glowtape:
Looks OK to me, seems to do its job as intended.

While you're in there, might be an idea to set outerKi to -1 on line 754 to satisfy my slight OCD tendencies, so that the numeric field shows a bar instead of a zero when disabled, as all other ones do. :grin:

mlyle:
@glowtape I'm not sure I should.  In the others - means "don't change", where zero on outerKi means "set it to zero".

mlyle:
P.S. Thx for review

mlyle:
@pug398 don't know if you would find this interesting.  There's been improvements to althold (not yet used by the path follower, but soon) and now autotune calculates thrust.

tracernz:
This looks good, but I'd like to have another more careful read of it.

mlyle:
Rebased (conflict with LQG GCS-side work)

tracernz (approved):

glowtape:
Looks OK to me. Acts properly with the presence or lack thereof of hover throttle, i.e. enabling and disabling accordingly. Values don't look insane. Chances are I can get a quick practical test in between now and Friday, if you need it, crossing my fingers the Seppuku baro doesn't freak out (no foam).

glowtape:
OK, did a quick check outside. Obviously works. Interesting behavior though, on my miniquad. When I go quickly to min or max throttle, it jerks before it settles, i.e. when falling it goes nearly zero throttle for a very short moment before settling into a falling speed, or when rising, it gives a higher throttle push before settling. I suppose that's to be expected due to delay via signal processing.

mlyle:
> OK, did a quick check outside. Obviously works. Interesting behavior though, on my miniquad. When I go quickly to min or max throttle, it jerks before it settles, i.e. when falling it goes nearly zero throttle for a very short moment before settling into a falling speed, or when rising, it gives a higher throttle push before settling.

Not delay in signal processing.  If you ask to move from a climb to a descent, the quickest way is a 0 thrust coast maneuver (and then a rapid recovery when you're at the descent rate).  If you want to go from descending to level, the quickest way is a whole bunch of extra thrust, and when velocity reaches 0 to select hover power.

It's just a lot more aggressive than people doing these operations.

glowtape:
Ah OK.

I suppose this is ready to merge, given that tracernz already approved it?

tracernz:
Ack from me. :+1:

PR #2107, opened 2018-03-11
stabsettings: Increase yawPID limits

### LINKED ISSUE #2084: New Yaw autotune returns bad values on hexacopter
* Seen in version (release version or githash): wired
* On flight controller (if applicable): seppuku (pre-production green pcb)

Issue details:
I'm getting invalid yaw values from autotune on my hex after upgrading to wired and any artifacts build since wired.  I'm guessing this is due to #2019 but am not sure as I was busy and didn't have a chance to test that PR before it merged.

The last three tunes on this setup all returned unusable yaw numbers:
https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICAmP-shQsM
(and the two previous tunes)

This is on my hex with 2204 2300kv motors 18a KISS esc's and 6x4.5 props which has always returned good results previously.

One thing I have no yet eliminated is swapping to different props.  The tune graphs are considerably noisier than I'm used to seeing with this hex and the props are a bit beat up.  I'll try and dig up some fresh props to try again in the next few days.  But the roll/pitch tune results seem fine and in line with what I've seen before so I'm not convinced it's the props.

UAV's from current configuration attached:
[hex.uav.zip](https://github.com/d-ronin/dRonin/files/1757083/hex.uav.zip)

mlyle:
Before they were 100/500 PI.  Now they're 150/600.  Makes @jhitesma's
vehicle with 105/515 happy, and things a bit more aggressive than that.

Fixes #2084

### DIFF
--- a/shared/uavobjectdefinition/stabilizationsettings.xml
+++ b/shared/uavobjectdefinition/stabilizationsettings.xml
@@ -115,7 +115,7 @@
         <elementname>ILimit</elementname>
       </elementnames>
     </field>
-    <field defaultvalue="0.0035,0.0035,0.0,0.3" limits="%BE:0:0.01,%BE:0:0.05,, " name="YawRatePID" type="float" units="">
+    <field defaultvalue="0.0035,0.0035,0.0,0.3" limits="%BE:0:0.015,%BE:0:0.06,, " name="YawRatePID" type="float" units="">
       <description/>
       <elementnames>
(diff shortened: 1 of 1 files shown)


jhitesma:
Thanks!  I'll try not to build anything that requires even more aggressive yaw values ;)

PR #2108, opened 2018-03-11
sensors: adjust good_run / missing sensor logic

### LINKED ISSUE #2099: Missing sensors is broken??

mlyle:
Also some miscellaneous code cleanup.  Hopefully this is more intuitive
and does what is expected (there was previously a regression where
missing sensors would prevent arming even when PermitMissingSensors was
true).

Fixes #2099

### DIFF
--- a/flight/Modules/Stabilization/sensors.c
+++ b/flight/Modules/Stabilization/sensors.c
@@ -61,10 +61,10 @@ extern pios_i2c_t external_i2c_adapter_id;
 #include "coordinate_conversions.h"
 
 // Private constants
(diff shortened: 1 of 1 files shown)


mlyle:
Just the boring case has been tried so far (no missing sensors).  Confirmed that A) it still works, and B) when a point is dropped by non-realtime linux that we get the sensors alarm and it's held for awhile.  Need to play with missing mag / baro / etc.

Another notable thing this does-- if a mag stops reporting it actually asserts a SENSORS error.

tracernz (approved):

PR #2121, opened 2018-03-16
pios_flash: catch flash errors (enable interrupt)

### LINKED ISSUE #1562: Consider asserting/hanging on flash errors / ERRIE
As discussed / reported by @mindnever / #1560 --- if we have a null dereference bug we access flash, with undefined behavior.

If it's a write, it just doesn't happen, but also sets the flash error flag.  This causes us to hang on future flash programming operations (e.g. saving a config item).

It would be more helpful to crash immediately, because 1) we'd spot more bugs, and 2) the actual failure/crash would be more proximal to the actual problem code.

Also it's the eeriest flag.

mlyle:
fixes #1562

### DIFF
--- a/flight/PiOS/STM32F10x/pios_flash_internal.c
+++ b/flight/PiOS/STM32F10x/pios_flash_internal.c
@@ -62,6 +62,11 @@ static struct pios_internal_flash_dev *PIOS_Flash_Internal_alloc(void)
 	return(flash_dev);
 }
 
+
+void FLASH_IRQHandler() {
+	PIOS_Assert(0);
+}
+
 int32_t PIOS_Flash_Internal_Init(uintptr_t *chip_id, const struct pios_flash_internal_cfg *cfg)
 {
 	struct pios_internal_flash_dev *flash_dev;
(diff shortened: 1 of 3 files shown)


flight/PiOS/STM32F10x/pios_flash_internal.c
```diff
@@ -62,6 +62,11 @@ static struct pios_internal_flash_dev *PIOS_Flash_Internal_alloc(void)
 	return(flash_dev);
 }
 
+
+void FLASH_IRQHandler() {
```
tracernz, line 66: :+1: 
It would be neat to have a DEFINE that causes separate handlers like this to be built for all the unused/error IRQs, otherwise just weak link them all to a single default handler. I regularly find myself splitting them all out when debugging crashes that we encounter from time to time.

tracernz (approved):

flight/PiOS/STM32F10x/pios_flash_internal.c
mlyle: We could have a single generic one that grabs IPSR and puts it into a variable for inspection.

tracernz: You can dump IPSR out easy enough already, but then you need a table handy to look it up. :stuck_out_tongue:

PR #2122, opened 2018-03-16
Run system task in init task (to save task stack space)

### LINKED ISSUE #1879: Flight: Reclaim Init Task Stack

mlyle:
Fix #1879.

Saves a bunch of RAM.  Requires analysis of system task priority under this new scheme.  (Both posix & not).  Other small fixes.

Needs test on hardware (tested sim with and without realtime).

### DIFF
--- a/flight/PiOS/posix/pios_thread.c
+++ b/flight/PiOS/posix/pios_thread.c
@@ -34,8 +34,44 @@ struct pios_thread
 {
 	pthread_t thread;
 
-	char *name;
+	const char *name;
 };
+
+/**
+ * @brief   Creates a handle for the current thread.
+ *
+ * @param[in] namep        pointer to thread name
+ *
+ * @returns instance of @p struct pios_thread or NULL on failure
+ */
+struct pios_thread *PIOS_Thread_WrapCurrentThread(const char *namep)
+{
+	struct pios_thread *thread = PIOS_malloc_no_dma(sizeof(struct pios_thread));
+
+	extern bool are_realtime;
+
+	if (!thread) {
+		abort();
+	}
+
+	thread->thread = pthread_self();
+
+	thread->name = namep;
+
+	if (are_realtime) {
+		struct sched_param param = {
+			.sched_priority = 30 + PIOS_THREAD_PRIO_HIGHEST * 5
+		};
+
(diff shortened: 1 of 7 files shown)


glowtape:
Just tested this on a Revo for giggles, on top of my LQG branch. Seems to act normally. Without it, OSStackRemaining ends up at 456 bytes, whereas with it, it ends up at 272 bytes. So there's still some headroom.

glowtape:
Seppuku goes down to 248 bytes, fully armed.

mlyle:
Prios need work here.  Nailing things to HIGHEST at wrap-time seems sane.  But then we need systemmod's body to change its own priority to NORMAL.

glowtape:
Tested this on Seppuku, OG Brain, Revo and SPRF3e, they all work happily like nothing changed.

Code looks good too. Wish you'd placed a bracket on a newline so I can be a pain in the ass about it for a change, but alas.

flight/PiOS/posix/pios_thread.c
```diff
@@ -34,9 +34,57 @@ struct pios_thread
...
+	if (are_realtime) {
+		struct sched_param param = {
+			.sched_priority = 30 + PIOS_THREAD_PRIO_HIGHEST * 5
+		};
+
+		pthread_setschedparam(thread->thread, SCHED_RR, &param);
+	}
+
+#ifdef __linux__
+	pthread_setname_np(thread->thread, thread->name);
```
tracernz, line 70: Nice!. Can we do this for the other threads too? (unless I missed something, I don't think we do)

tracernz (approved):
Great. 👍

flight/PiOS/posix/pios_thread.c
mlyle: We do, actually :D

tracernz: Great. :D

PR #2123, opened 2018-03-17
flight: cleanup clock handling and some makefile stuff

### LINKED ISSUE #1454: Flight[ChibiOS]: Clock init is ineffective since PLL is running after bootloader
ChibiOS function `stm32_clock_init` attempts to setup the PLL, assuming that it is inactive. PLL is usually already activated by the bootloader when we jump to firmware. Writes to PLL configuration registers are ineffective when the PLL is active. 

Low priority since, as long as the bootloader clock configuration matches firmware, everything will be fine.

Relates to #1448.

mlyle:
1. ensure that it is possible to override pios includes in the board-info directory (we don't currently need this, but for awhile on this PR I did, and I think it's a good/related thing to have).
2. ensure that in cmsis_system we comply with datasheet recommendations on PLL configuration (namely, on 8MHz F4 targets, don't do / 10 * 420 / 2, but instead do / 8 * 336 / 2, which keeps the PLL divided reference above 1MHz).
3. factor chconf common, and make sure the values are the same as cmsis_system.  also disable clock init.  Fixes #1454


tracernz (approved):
Read this yesterday and I like it. :+1: Note that I didn't verify the PLL values but they look correct on quick eyeball.

jihlein:
I think this change broke the AQ32 target.  Flight time runs away at an unrealistic rate, attitude ball and heading tape spinning, altitude running away.  This occurs for a few seconds and then the firmware crashes.  Investigating.

glowtape:
Same happens on the Revo. I just compiled it to crosscheck.

glowtape:
I need to re-enable clock init for it to be sane.

jihlein:
Confirmed re-enabling clock init fixes AQ32 too.  Wonder if any of the F3 targets will be affected?  Can try a DTFc a little later today.

Can't find my spare DTFc, but Sparky seems to work okay.

mlyle:
Confirmed new loader (and firmware) is happy on 8MHz F4 targets--- revolution & seppuku

PR #2125, opened 2018-03-17
flight: in debug mode, skip targets that don't build

mlyle:
Allow to build all with debug, but "all" doesn't encompass these
flash-constrained F3 targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,8 +69,12 @@ include $(ROOT_DIR)/flight/targets/*/target-defs.mk
 # OpenPilot GCS build configuration (debug | release)
 GCS_BUILD_CONF ?= debug ccache
 
-# And the flight build configuration (debug | default | release)
-export FLIGHT_BUILD_CONF ?= default ccache
+# And the flight build configuration (debug | release)
+ifeq ($(DEBUG),YES)
+export FLIGHT_BUILD_CONF ?= debug ccache
+else
+export FLIGHT_BUILD_CONF ?= release ccache
+endif
 
 # Paths
 UAVOBJ_XML_DIR := $(ROOT_DIR)/shared/uavobjectdefinition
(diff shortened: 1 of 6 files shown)


flight/targets/omnibusf3/target-defs.mk
```diff
@@ -1,5 +1,6 @@
 # Add this board to the list of buildable boards
 ALL_BOARDS += omnibusf3
+NODEBUG_BOARDS += omnibusf3
```
tracernz, line 3: Maybe put a comment explaining why it's a NODEBUG board? (applies to the others too)

Makefile
```diff
@@ -950,8 +954,13 @@ all_$(1)_clean: $$(addsuffix _clean, $$(filter bu_$(1), $$(BU_TARGETS)))
 all_$(1)_clean: $$(addsuffix _clean, $$(filter ef_$(1), $$(EF_TARGETS)))
 endef
 
-# Some boards don't use the bootloader
+ifeq ($(filter debug, $(FLIGHT_BUILD_CONF)), debug)
+# Don't build boards that overflow with debugging
+ALL_BOARDS     := $(filter-out $(NODEBUG_BOARDS), $(ALL_BOARDS))
```
tracernz, line 959: Should probably print a warning to stderr with the list of boards that won't be built.

flight/targets/omnibusf3/target-defs.mk
mlyle: I think something like this, where it's the same file, providing the same set of variables for a large number of boards--- that it's a waste to document the same thing in every one.

A disproportional amount of the time we do that, the comments end up becoming lies, too.

Makefile
mlyle: so it's not easy to constrain this to cases where it makes a difference... I did this:

```
mlyle@midnight:~/dRonin$ FLIGHT_BUILD_CONF=debug make -j4 fw_sparky2
Makefile:961: Removing targets with inoperable debug mode: sparky pikoblx omnibusf3 sprf3e
make[1]: Nothing to be done for 'first'.
make[1]: Nothing to be done for 'flightlib_armhardfp'.
make[1]: Nothing to be done for 'flightlib_armsoftfp'.
make[1]: Nothing to be done for 'uavobjects_armsoftfp'.
make[1]: Nothing to be done for 'uavobjects_armhardfp'.
 CC        [fw|spk2]  flight/Modules/ManualControl/failsafe_control.c
```

mlyle:

mlyle:
(this is just another build test / i have altered jenkins to request ccache building again-- which was lost with the build system changes)

flight/targets/omnibusf3/target-defs.mk
tracernz: Problem is non-regular maintainers usually aren't interested in anything other than their own issue on their own board, and they'll be lacking this information.

mlyle: Yah, it's all a case of relative risks.  Chance of doc helping someone vs. cost of keeping it mamy places and chance of it rotting and misleading.

PR #2126, opened 2018-03-17
python: improvements to config handling

mlyle:
dronin-getconfig now outputs an XML file, new dronin-putconfig can store one to the FC, improvements to the telemetry class's handling of nacks and saving.  Infrastructure so that client programs can extend the auto-parsing without providing an entire telemetry argument parser.


mlyle:
jenkins, test this please

mlyle:
jenkins, test this please

mlyle:
jenkins, test this please

python/dronin-getconfig
```diff
@@ -57,15 +57,18 @@ def main():
...
         if val is not None:
-            print(val)
+            sobjs.append(val)
         else:
             missing.append(s._name)
 
-    for s in missing:
-        print("No instance of %s" % (s))
+    #for s in missing:
+    #    print("No instance of %s" % (s))
```
tracernz, line 69: We should use python's excellent logging class to do this stuff, then it's easy to switch on debug type stuff when needed. (to clarify, not asking you to go do that in this PR, just a suggestion for stuff we add, and stuff we clean up)

tracernz (approved):
Superb. :+1:

python/dronin-getconfig
mlyle: Yah, I've got that on my todo list for a big cleanup... but I'm trying to sprint towards integration tests and I kinda hate the python logging stuff (it definitely takes thought on how to instantiate it right in a framework like this etc)

PR #2127, opened 2018-03-17
macosx: improve build

mlyle:
1. Fix build with unit test --- remove unused matrix.
2. Remove all the build warnings in hidapi_mac, which I believe are the last ones that are not upstream from qwt.


mlyle:
jenkins, test this please

PR #2129, opened 2018-03-18
vtolpathfollower: improve config path

### LINKED ISSUE #706: vtolpathfollower should use correct dT
Similar to issues spotted in #633, vtolpathfollower needs to properly observe dT -- either use configured value or actual value or a filtered value.

mlyle:
1. Use user-configured dT, so tuning doesn't vary with update period.
(no verification is made that the task actually hits desired timing
though.  Fixes #706 
2. Use proper callback-flag based mechanism.
3. Only have one copy of the guidance settings, instead of two-- save a
little bit of ram.


tracernz (approved):
Magnifique :+1:

PR #2130, opened 2018-03-18
gcs: hidapi_linux: fix a crash seen in reports

mlyle:
Presumably a race on unplug.


mlyle:

tracernz (approved):

PR #2136, opened 2018-03-19
actuator: build motor gain functionality

mlyle:


mlyle:

glowtape:
The hangtime stuff would also need to be scaled, no?

--edit: Hmmm, no. I suppose it'll fall out automatically with the adjusted hover throttle.

ufoDziner:
Works on dshot600.

ufoDziner:

ufoDziner:
Latest artifacts are boot looping on Seppuku.

mlyle:
fixed, I think.  so stupid.

ufoDziner:
Tested 2 autotunes with MotorInputOutputGain at 1.0 and .7.  Tunes are very very close.  Slight wind difference during each which may be a contributing factor.  

https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA2IbPiwsM

https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA2N7ymwkM

I'm also getting a red x on the saved button when resetting the PIDs to default prior to an autotune.  They seem to save, but the button is showing that they don't.

glowtape:
The red X is fixed in another PR. It's a visual issue.

ufoDziner:
Tried on another rig.  Flies fine at 100%.  Adjusted down to 50% with the OSD (which worked fine).  Tried some flips, and the quad just continued to tumble.  Not nearly enough power being added to stop it.

mlyle:
@ufoDziner reports that tumbles were fixed by setting maxpoweradd appropriately for the new power scale.  Tuning appears independent of max power scale thanks to (wonky) actuator effort scaling.

mlyle:
Rebased.  GCS UI added.

image

mlyle:

ufoDziner:
First autotune with a HV high kv test rig.  Running 6s HV and 2207-2650 kv with 60% max power limitation.  Feels good LoS.  Will check FPV in the near future.

https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA2Onl0wgM

ufoDziner:

ufoDziner:
FPV'd with the power set at 65% to get similar performance to my "Regular" 1722kv motors (these are 2650kv).
Felt very similar.  I would not be able to tell the difference in a blind test.  DVR says that the high kv drew slightly more amps (84A vs 74A).  Both are identical builds except for the motors, and within 5g of each other.  Below are the tunes for the two quads I tested with.  

Regular 1722k: https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA2IHwlgoM

2650kv versionl: https://dronin-autotown.appspot.com/at/tune/ahFzfmRyb25pbi1hdXRvdG93bnIYCxILVHVuZVJlc3VsdHMYgICA2Onl0wgM

2 snaps up!

flight/Modules/OnScreenDisplay/osd_menu.c
```diff
@@ -216,7 +217,7 @@ enum menu_fsm_state {
 	FSM_STATE_STICKLIMITS_ROLLEXH,  /*!< Roll expo horizon */
 	FSM_STATE_STICKLIMITS_PITCHEXH, /*!< Pitch expo horizon */
 	FSM_STATE_STICKLIMITS_YAWEXH,   /*!< Yaw expo horizon */
-	FSM_STATE_STICKLIMITS_ACROINSANITYFACTOR, /*!< Acro insanity factor for AcroPlus */
+	FSM_STATE_STICKLIMITS_MOTORPOWER, /*!< Motor power gain scale */
```
tracernz, line 220: Should probably note the removal of Acro Insanity in commit message for searching purposes. Also should go in rel notes.

-e- How do you normally manage rel notes? Should we just have a wiki page for the rel that anyone can add to?

flight/Modules/OnScreenDisplay/osd_menu.c
```diff
@@ -1956,30 +1959,34 @@ void sticklimits_menu(void)
...
+			float motor_gain;
+			ActuatorSettingsMotorInputOutputGainGet(&motor_gain);
+			motor_gain *= 100;
+
 			data_changed = false;
-			sprintf(tmp_str, "Acro Insanity Factor: %d", (int)factor);
+			sprintf(tmp_str, "Motor power scale: %d", (int)motor_gain);
 			write_string(tmp_str, MENU_LINE_X, y_pos, 0, 0, TEXT_VA_TOP, TEXT_HA_LEFT, 0, MENU_FONT);
-			draw_hscale(225, GRAPHICS_RIGHT - 5, y_pos + 2, 0, 100, factor);
+			draw_hscale(225, GRAPHICS_RIGHT - 5, y_pos + 2, 50, 100, motor_gain);
```
tracernz, line 1970: Why a different min to GCS?

ground/gcs/src/plugins/config/configoutputwidget.cpp
```diff
@@ -483,6 +484,7 @@ void ConfigOutputWidget::refreshWidgetsValues(UAVObject *obj)
     }
 
     m_config->motorCurveFit->setValue(actuatorSettingsData.MotorInputOutputCurveFit);
+    m_config->motorPowerGain->setValue(actuatorSettingsData.MotorInputOutputGain * 100);
```
tracernz, line 487: This form is still not using uavo relations is it? :(

tracernz (approved):

flight/Modules/OnScreenDisplay/osd_menu.c
mlyle: Was the same as GCS, but now I've adjusted the uavo to match.

ground/gcs/src/plugins/config/configoutputwidget.cpp
mlyle: Nope.  And I was scared to do it partially here (is that OK to do, in general?)  Not to mention the form is generated at runtime so that it can't fully be done with auto uavo relations...

flight/Modules/OnScreenDisplay/osd_menu.c
mlyle: I usually read the commit log.  But this is pretty minor IMO and wouldn't make it, because it's not the total removal of the parameter-- just the removal in the OSD config.

PR #2138, opened 2018-03-21
gcs: configtaskwidget: improve label formatting

mlyle:
A) Detect scaling properly-- scale = 1 is also a no-op.
B) If we are scaling, show one figure after the decimal point
(eg %1f, not scientific notation etc).


mlyle:
This regressed in #2103, my bad.

tracernz (approved):

PR #2139, opened 2018-03-21
Radio path refactor / improvement (openlrsng)

mlyle:
This is getting close.  Telemetry goes across the link-- very, very slowly.  Remaining work:

* [x] What to transmit when source data invalid
* [x] TX telem retry path
* [x] Switch channels
* [x] Inhibit sysalarms widget when object not present
* [x] Binding management
* [x] Default TX configuration creation
* [x] Increased datarate mode
* [ ] ~~Robustness, delay hop when frame in progress~~ defer
* [x] Reliability evaluation at different data rates and channel counts
* [ ] Air-side radio support
* [ ] Final OpenLRS interoperability recheck
* [ ] Setup path / setup recommendations


mlyle:
mostly a benchmark of building

flight/PiOS/STM32/chibi_main.c
```diff
@@ -139,6 +141,19 @@ void initTask(void)
...
+
+	for (int i = 0; i < PIOS_SYS_SERIAL_NUM_BINARY_LEN; i += 4) {
+		randomize_addseed(
+				(serial[i]) |
+				(serial[i+1] << 8) |
+				(serial[i+2] << 16) |
+				(serial[i+2] << 24));
+	}
+
+	/* TODO: consider mixing in HRNG on F4 */
```
tracernz, line 157: Accel/gyro data would be a good source of entropy too, just not available this early. 😀

mlyle: Yes-- the problem is the RNG is not reentrant.  So once there's tasks it's not really safe to seed it anymore.

tracernz: Ack. You would have to do it in board_init after gyro driver is init'ed.

mlyle:
It's probably best to review this here, because of the massive number of commits.  It's looking pretty good- -at 57600+ radio telemetry is now **BETTER THAN DIRECT USB WAS BEFORE**.  I've done a bunch of improving of margins and interop testing, but more is needed.

The main things missing that would make this great are:

* Multiprotocol support (taranis -> oplink combined telemetry & control).
* Airside sbus/telemetry bridging support (oplink->oplink)
* Some minimal UI/defaults/documentation.
* Further work can be put into naming of functions, variables, etc, and factoring out the RFM22 specific stuff to support new radios.

Still, this is a sane place to pick up from, and subsequent PRs should be tidier and easier to review.

mlyle:
This will conflict (slightly) with idle task cleanup.

flight/PiOS/Common/pios_openlrs.c
```diff
@@ -1238,73 +1234,100 @@ static void pios_openlrs_task(void *parameters)
...
-			DEBUG_PRINTF(3, "T2: %d %d\r\n", time_till_timeout_us,delay_ms, time_since_packet_us);
+			delay_us = getInterval(&openlrs_dev->bind_data) +
+				packet_timeout_us -
+				time_since_packet_us;
+			DEBUG_PRINTF(3, "T2: %d %d\r\n", delay_us,
+					time_since_packet_us);
 		}
 
+		/* XXX: should this be +500 */
+		delay_ms = (delay_us + 999) / 1000;
```
tracernz: You want to round everything up (999), or to nearest (500)?

shared/uavobjectdefinition/openlrs.xml
```diff
@@ -5,6 +5,14 @@
     <logging updatemode="manual" period="0"/>
     <telemetrygcs acked="true" updatemode="onchange" period="0"/>
     <telemetryflight acked="true" updatemode="onchange" period="0"/>
+    <field defaultvalue="Disabled" elements="1" name="role" type="enum" units="">
+      <description>Whether to be a receiver or transmitter</description>
+      <options>
+        <option>Disabled</option>
+        <option>Rx</option>
+        <option>Tx</option>
```
tracernz, line 13: I find this to be a bit weird since either way you're doing both... Master/slave would be clearer?

flight/PiOS/Common/pios_openlrs.c
```diff
@@ -264,742 +278,364 @@ static void delay(uint32_t time)
...
+				}
 
-	rfm22_write(openlrs_dev, 0x70, 0x2C);    // disable manchest
+				if (bytes > 0) {
+					tx_buf[0] |= (0x37 + bytes);
+				} else {
+					// tx_buf[0] lowest 6 bits left at 0
+					tx_buf[1] = openlrs_status.LastRSSI;
+					if (FlightBatteryStateHandle()) {
+						FlightBatteryStateData bat;
```
tracernz: It's rather janky for this kind of stuff to be inside the radio driver. :/

flight/PiOS/Common/pios_openlrs.c
```diff
@@ -312,6 +312,33 @@ static bool pios_openlrs_bind_transmit_step(struct pios_openlrs_dev *openlrs_dev
 	return false;
 }
 
+static void pios_openlrs_config_to_bind_data(struct pios_openlrs_dev *openlrs_dev)
+{
+	OpenLRSData binding;
+	OpenLRSGet(&binding);
+
+	if (binding.version == BINDING_VERSION) {
```
tracernz: Should we not return an error if the version doesn't match?

flight/Modules/Telemetry/telemetry.c
```diff
@@ -796,6 +796,7 @@ static int32_t setUpdatePeriod(telem_t telem, UAVObjHandle obj,
  */
 static void gcsTelemetryStatsUpdated(telem_t telem)
 {
+	// XXX this is dumb.
```
tracernz: Doing telemetry handshake in telemetry seems... not dumb? Can you explain this more precisely?

flight/targets/pipxtreme/fw/pios_board.c
```diff
@@ -162,12 +162,14 @@ void PIOS_Board_Init(void)
 
 	ModuleSettingsSet(&moduleSettings);
 
+#ifdef PIOS_INCLUDE_OPENLRS
```
tracernz: Is the PipX useful for anything without olrs? It would seem to not be.

flight/PiOS/Common/pios_openlrs.c
```diff
@@ -971,9 +976,14 @@ static void pios_openlrs_tx_frame(pios_openlrs_t openlrs_dev)
 
 	pios_openlrs_do_hop(openlrs_dev);
 
-	int16_t channels[OPENLRS_PPM_NUM_CHANNELS] = {
-		800, 900, 1000, 1100, 1900, 2000, 2040
-	};
+	int16_t channels[OPENLRS_PPM_NUM_CHANNELS];
+
+	for (int i=0; i < OPENLRS_PPM_NUM_CHANNELS; i++) {
```
tracernz: spaces

flight/targets/pipxtreme/fw/pios_config.h
```diff
@@ -39,8 +39,9 @@
 #define PIOS_INCLUDE_OPENLRS_RCVR
 
 /* Supported receiver interfaces */
-#define PIOS_INCLUDE_PPM
 #define PIOS_INCLUDE_GCSRCVR
+#define PIOS_INCLUDE_PPM
+#define PIOS_INCLUDE_SBUS
```
tracernz: What's special about S.BUS?

flight/PiOS/STM32/chibi_main.c
tracernz: Turns out SRAM can be a good source of entropy if we want a secure PRNG: https://eprint.iacr.org/2013/304.pdf

flight/PiOS/Common/pios_openlrs.c
```diff
@@ -526,11 +528,17 @@ static bool binding_button_pushed(pios_openlrs_t openlrs_dev)
...
-		return !truth_value;
+	if (!openlrs_dev->cfg.bind_active_high) {
+		truth_value = !truth_value;
 	}
+
+	button_history[2] = button_history[1];
+	button_history[1] = button_history[0];
+	button_history[0] = truth_value;
+
+	/* Catch an edge, but also require 2 in a row. */
```
tracernz, line 612: AKA debounce. The usual method of counting cycles seems cleaner than an array IMO, especially if you want more cycles... which IMO we should require something like 100 ms.

```c
static int button_debounce = 0;
if (truth_value)
	button_debounce++;
else
	button_debounce = 0;

return button_debounce >= 3;
```

flight/Modules/System/systemmod.c
```diff
@@ -953,6 +953,7 @@ static int32_t eventPeriodicCreate(UAVObjEvent* ev, UAVObjEventCallback cb, stru
 	objEntry->evInfo.ev.obj = ev->obj;
```
tracernz, line 950: This should help a lot with other constrained links like BTLE, which I revived yesterday.

flight/PiOS/STM32/chibi_main.c
mlyle: Could be-- though lots of things init it etc by this point.

shared/uavobjectdefinition/openlrs.xml
mlyle: I think whatever sends control stuff is in the "tx" role historically.

flight/PiOS/Common/pios_openlrs.c
mlyle: Yah, this is pretty reliable in my test so far.  I can change variants if you really care.

flight/Modules/Telemetry/telemetry.c
mlyle: Committed an extended comment :D

flight/PiOS/STM32/chibi_main.c
tracernz: Yup. Would need to really do it in the bootloader, or make sure it the mem used wasn't touched by bootloader or anything prior. It's neat but not really something we need. :)

shared/uavobjectdefinition/openlrs.xml
tracernz: That's not so clear for the typical PipX role which doesn't involve control.

flight/PiOS/Common/pios_openlrs.c
tracernz: No, not really. It just seemed like a really weird way to do it. :stuck_out_tongue:

flight/Modules/Telemetry/telemetry.c
```diff
@@ -812,7 +812,9 @@ static int32_t setUpdatePeriod(telem_t telem, UAVObjHandle obj,
  */
 static void gcsTelemetryStatsUpdated(telem_t telem)
 {
-	// XXX this is dumb.
+	// XXX this is dumb.  It needs to not be connected across global
+	// objects but "projected" so that multiple telemetry sessions
+	// can be supported.
```
tracernz: Ack. :+1:

shared/uavobjectdefinition/openlrs.xml
mlyle: I think typical on the LP side involves control, and that's what lots of people want to do.  My objective with this is to get to something that we can snap in am odule slot and do useful stuff.

flight/PiOS/Common/pios_openlrs.c
mlyle: :D it looks like when I write HDL...

shared/uavobjectdefinition/openlrs.xml
tracernz: Whatever, let's not bikeshed.

tracernz (approved):

flight/PiOS/Common/pios_openlrs.c
mlyle: Kind of.  Ultimately the radio driver needs to be factored out of pios_openlrs.c.

This is a specified part of openlrs.  and it's frame and not stream oriented, so... :/

flight/PiOS/Common/pios_openlrs.c
mlyle: We could, but we'd just go straight back into awaiting bind.  We don't really have a great way for diagnostics from submodules to get out to the end user.

flight/targets/pipxtreme/fw/pios_board.c
mlyle: Not really, but I turned it off during re-bringup etc.

flight/PiOS/Common/pios_openlrs.c
mlyle: I think this had already gotten fixed by a later commit.

flight/targets/pipxtreme/fw/pios_config.h
mlyle: Will be using it as one way of tx-to-module comms.

flight/PiOS/Common/pios_openlrs.c
mlyle: I don't know.  :D  +500 made more sense to me, but it takes a lot of audit/justification to do something different than old code.

mlyle:
It builds now, so it must be perfect.

flight/PiOS/Common/pios_openlrs.c
tracernz: Yeah, I understand, and am okay with this for that reason. :)

flight/targets/pipxtreme/fw/pios_config.h
tracernz: Do we have enough resources to include some more serial protos? I would like SRXL so I can hook up a 2.4G rx to a PipX.

tracernz:
There is no other possible conclusion. :+1: I suppose I should put it onto some hardware before merging.

tracernz:
Tested successfully on standard 433 MHz hardware (PipX as Tx, Revo as Rx). AHI was smooth as butter! Throughput is really great.

This is really an amazingly good piece of work! Must have taken many, many hours!

I've not been successful binding 915 MHz hardware, but I'll continue trying that and see what I can find out.

mlyle:
\*blushes\*

915MHz is a big mystery.

tracernz:
915 MHz was my fault, I had missed setting the OpenLRS.role field on the PipX.

Tested ae9e8da successfully on two setups:
- 433 MHz
    - Chinese (Full) Revo clone, Chinese OPLink clone
- 915 MHz
    - Hand-assembled Sparky 2, Hand-assembled TauLink USB Stick

PR #2140, opened 2018-03-22
pios_tim/PPM:  correct F0 problem

mlyle:
1. The STM32F0xx implementation didn't work because it checked and reset the update event in a loop over multiple client device structures.  If the wrong client structure came up first, we lost the update event.  Boohoo.
2. ~~The STM32F1xx implementation didn't work~~  Pretty sure this was test methodology and I was smoking crack.  Apologies.
3. Also improved some comments, and used && instead of & (though & was safe here, it's subtle, and just askin' to be broken in the future).

We really need to unify these, the implementations vary a lot (e.g. the thresholding).

### DIFF
--- a/flight/PiOS/STM32F0xx/pios_tim.c
+++ b/flight/PiOS/STM32F0xx/pios_tim.c
@@ -250,7 +250,7 @@ static void PIOS_TIM_generic_irq_handler(TIM_TypeDef * timer)
 			}
 
 			/* Generate the appropriate callbacks */
-			if (overflow_event & edge_event) {
+			if (overflow_event && edge_event) {
 				/*
 				 * When both edge and overflow happen in the same interrupt, we
 				 * need a heuristic to determine the order of the edge and overflow
--- a/flight/PiOS/STM32F0xx/pios_tim.c
+++ b/flight/PiOS/STM32F0xx/pios_tim.c
@@ -166,25 +166,26 @@ out_fail:
 
 static void PIOS_TIM_generic_irq_handler(TIM_TypeDef * timer)
 {
+	/* Check for an overflow event on this timer */
+	bool overflow_event;
+	uint16_t overflow_count;
+	if (TIM_GetITStatus(timer, TIM_IT_Update) == SET) {
+		TIM_ClearITPendingBit(timer, TIM_IT_Update);
+		overflow_count = timer->ARR;
+		overflow_event = true;
+	} else {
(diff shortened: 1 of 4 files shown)


mlyle:
@jihlein there is some chance the F0 fix fixes #1420 ... will try to check myself relatively soon.

flight/PiOS/STM32F0xx/pios_tim.c
```diff
@@ -250,7 +250,7 @@ static void PIOS_TIM_generic_irq_handler(TIM_TypeDef * timer)
 			}
 
 			/* Generate the appropriate callbacks */
-			if (overflow_event & edge_event) {
+			if (overflow_event && edge_event) {
```
tracernz, line 254: good catch :+1:

flight/PiOS/STM32F10x/pios_tim.c
```diff
@@ -301,7 +300,7 @@ static void PIOS_TIM_generic_irq_handler(TIM_TypeDef * timer)
 				 * assume it happened before the overflow.
 				 */
 
-				if (edge_count < chan->timer->ARR / 2) {
+				if (edge_count < (chan->timer->ARR / 2)) {
```
tracernz, line 303: This is a nop, but still removes any doubt for the reader. :+1:

tracernz (approved):
These look good. :+1: The F0 overflow one would have caused occasional/unlucky blips.

jihlein:
@mlyle - I'll give this a test tonight on the FlyingPi.

jihlein:
Well, it's closer.  I can get thru the TX setup wizard on the FlyingPi as far centering all the controls, but when I press next to move to control extents, the wizard hangs.  I guess it's possible that's a different problem? 

Thought to check sBus, it hangs in the same spot.  So the good news is the PPM issue may be fixed, the bad news is the config wizard hangs for both sBus and PPM.  I'll try setting up PPM manually later.

This is on the LQG test branch for what it's worth.

mlyle:
@jihlein Well, that's exciting that PPM itself is maybe fixed. I haven't been testing on pi, but the input wizard worked for me just last night on openlrs... so will have to nail that down.

jihlein:
@mlyle - quick update, after deleting the "theflash.bin" file from the RPi and starting from a totally clean slate, I can get PPM to function correctly on the FlyingPi stack after building with this change.

Issue #1438, opened 2016-10-19
simulator dies when receiving NMEA GPS data


mhv-shared:
- Seen in version (release version or githash): next
- On flight controller (if applicable): simulator/flyingpi

Issue details: when configured with `GPSDataProtocol=NMEA`, the simulator silently exits almost immediately. This only occurs if the GPS is actually connected. Works fine when set to UBX.

mlyle:
Heh this is somewhat likely to affect more than flyingpi, given that almost no one uses NMEA.

tracernz:
I thought that too. I have an NMEA GPS on a quad, so I'll give it a try on a regular target in the next couple of days. I guess I can hook a GPS up to a serial port on my PC and use it from POSIX sim too?

mlyle:
that might work too yes :D

mlyle:
No repro on bench testing in flightd/sim, though I found a lot of minor things to fix.

tracernz:
Repro with MTK3329:
```
(gdb) run
Starting program: /home/mike/dev/dronin/build/flightd/flightd -S gps:/dev/ttyUSB0
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
Beginning simulation environment
[New Thread 0x7ffff74b6700 (LWP 15799)]
Initialization task running
Started thread (init) p=0x5555557df260
HW serial number-- hex: 000000000000000000000000
[New Thread 0x7ffff6cb5700 (LWP 15800)]
Started thread (pios_rtc_task) p=0x7ffff0000b20
Pseudo-RTC started
[New Thread 0x7ffff64b4700 (LWP 15801)]
Started thread (pios_tcp_rx) p=0x7ffff00056a0
tcp dev 0x7ffff0004e30 - socket 5 opened - result 0
Completed PIOS_Board_Init
[New Thread 0x7ffff5cb3700 (LWP 15802)]
Started thread (pios_serial_rx) p=0x7ffff00065a0
serial dev 0x7ffff0005d60 - fd 6/6 opened
Setting Serial ID 0x7ffff0005d60 to 9600 bps
serial dev 0x7ffff0005d60 - (path /dev/ttyUSB0)
SimSensorsInitialize: Using simulated sensors.
SimSensorsInitialize: starting SIMULATED sensors

Thread 2 "init" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff74b6700 (LWP 15799)]
[...]

PR #2141, opened 2018-03-24
GPS: Small changes for simulator, NMEA

mlyle:
1. Prevented invoking autoconfigure logic on NMEA even if autoconfigure is turned on.
2. Allow GPS to work on 64 bit architectures (e.g. linux-amd64)
3. Fix up debugging on NMEA to work nicely.

### DIFF
--- a/flight/Modules/GPS/GPS.c
+++ b/flight/Modules/GPS/GPS.c
@@ -215,6 +215,8 @@ static void gpsConfigure(uint8_t gpsProtocol)
 		}
 		break;
 #endif
+		default:
+			break;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mlyle:
@jihlein --- I opened this investigating #1438.  Unfortunately I could not repro #1438, other than it didn't like running on 64 bit x86-64 linux because of the pointer width thing.  A uBlox-6 NMEA GPS seemed to work fine with flightd.

tracernz:
Hooked up MTK3329, it SEGFAULTed on next (as per my reply on #1438). Worked fine with this PR! :beer:

flight/Modules/GPS/GPS.c
```diff
@@ -76,7 +76,7 @@ static void updateSettings();
 // ****************
 // Private variables
 
-static uint32_t gpsPort;
+static uintptr_t gpsPort;
```
tracernz, line 79: This is the bug fix. :+1:

tracernz (approved):
Nice work. :+1:

PR #2142, opened 2018-03-29
frsky: properly use detected cell count in bat telem

mlyle:


mlyle:

yds:
works better than having to preset NbCells :)

however, the GCS GUI does not allow setting NbCells < 1

mlyle:
Yah, you don't set the cells <1 on that page, you check the box that says "auto cell count detection"...

yds:
checking out this build again I noticed the _Live detected cell count_ gets it wrong in the GCS when the LiPo is plugged in _after_ being powered up via USB.  If I plug the LiPo in first, then connect to the GCS, the count is right, unplug the LiPo and the count drops to zero, plug it in again and the count is right.

probably doesn't matter out in the field, cuz it gets it right when booting off a LiPo..  but looks wrong in the GCS when the LiPo gets plugged in after the GCS has already booted the FC via USB power..  I've gotten count==1 and count==3 while testing with a 4S at 16v..

tracernz:
That’s board specific. I guess your USB supply is leaking back to whatever ADC input you use to measure battery.

mlyle:
That or it's related to rise time.

mlyle:
Added two changes to battery:

* Do cell detection after stable voltage for 1 second--- this should eliminate any rise-time (but not leakage) related effects.
* Sample at 4Hz instead of 2Hz-- so we can capture power usage more accurately and so that things seem more responsive.

tracernz (approved):

mlyle:
changed-- the holddown was not initialized

tracernz (approved):

glowtape:
Seems to LGTM.

Detects battery ~~voltage~~ cells fine, even on run down 3S and 4S batteries, USB plugged in beforehand or not. Recognizes different ~~voltages~~ cells during runtime while switching out 3S and 4S, while keeping it hooked up on USB.

PR #2146, opened 2018-04-02
lqg: Optimize LQR covariance function.

glowtape:
The A10 term in the state transition matrix of the LQR is always zero. Told MATLAB that and generated new math. Resulted in less multiplications and additions. Then I started factoring shit by hand.

Testing it on my computer, using gcc -O2, average runtime goes from 19.05s to 12.2s for 500 million iterations. So I'm definitely doing something over the optimizer, and it'll likely benefit ARM.

There's some factoring that could be done on nP[1][1] to cut out some more multiplications, but for some reason that started to affect float accuracy more than I'd like (probably IEEE-754 crap). As things are in this PR, the output still matches MATLAB's dlqr(), given sufficient iterations.

### DIFF
--- a/flight/Libraries/math/lqg.c
+++ b/flight/Libraries/math/lqg.c
@@ -257,34 +257,40 @@ void lqr_calculate_covariance_2x2(float A[2][2], float B[2], float P[2][2], floa
 	float nP[2][2];
 	const float R = 1;
 
-	nP[0][0] = (Q00*R + A00*A00*P00*R + B0*B0*P00*Q00 + B1*B1*P11*Q00 +
-		A10*A10*P11*R + A00*A00*B1*B1*P00*P11 - A00*A00*B1*B1*P01*P10 +
-		A10*A10*B0*B0*P00*P11 - A10*A10*B0*B0*P01*P10 + A00*A10*P01*R +
-		A00*A10*P10*R + B0*B1*P01*Q00 + B0*B1*P10*Q00 - 2*A00*A10*B0*B1*P00*P11 +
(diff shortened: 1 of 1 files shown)


mlyle:
Do you get any benefit from declaring those arguments restrict?

glowtape:
Shaves off half a second from the unoptimized version, (avg. of 18.6s), and _maybe_ 0.1s from the optimized one (avg. of 12.1s).

mlyle:
Do the underlying matlab changes need to get checked in?

glowtape:
It's just changing term A10 to zero in https://github.com/d-ronin/dRonin/blob/next/matlab/lqg/lqg.m#L44

I can add that when back home later today. The rest is done by hand on the MATLAB output.

glowtape:
Wrote a crappy little test program that compares the outputs of the current version and the optimized ones of random Beta and Tau. There's minor differences five decimals in on the calculated torque gain, which are likely due to IEEE-754 bullshit. On the output overhere with the default RNG seed, the max. difference is 0.000067 between the two versions. Assuming a torque gain of 1, this is a 0.0067% difference. Given the torque gain is usually somewhat larger than one, this seems a NothingBurger(tm).

Test app:
https://gist.github.com/glowtape/3b2fbf042a98b65634308ec38e5f69e7

Output:
https://gist.github.com/glowtape/0f5c286efb5ea6cc39be9a91f0112bbe

glowtape:
I went actual flying with that stuff this Sunday. It didn't blow my quad to pieces, nor did it fly unexpectedly like crap. Not that anything should have happened, since both variants spit out the same values, up to several decimals.

PR #2149, opened 2018-04-04
posix: pios_flash: various improvements

mlyle:
* Partially clean up unit test.
* Allow specifying the flash filename
* Properly lock datastructure on posix.


mlyle:
This had many annoyances with building it on OS X, so better try on windows too (will just be flightd since windows doesn't unit test, but fair 'nuff).

flight/PiOS/posix/pios_flash_posix.c
```diff
@@ -28,7 +27,14 @@ static struct flash_posix_dev * PIOS_Flash_Posix_Alloc(void)
 	return flash_dev;
 }
 
-int32_t PIOS_Flash_Posix_Init(uintptr_t * chip_id, const struct pios_flash_posix_cfg * cfg)
+static const char *PIOS_Flash_Posix_GetFName()
```
tracernz: PIOS_Flash_Posix_Filename() would be better? :P

flight/targets/flightd/fw/main.c
```diff
@@ -106,14 +106,14 @@ void initTask(void *parameters)
 
 	PIOS_SYS_Init();
 
-	/* board driver init */
-	PIOS_Board_Init();
+	/* Initialize the task monitor library */
+	TaskMonitorInitialize();
```
tracernz, line 110: I don't really like this going before Board_Init, and we need to make sure that _never_ happens on embedded hardware. Maybe we should just rename PIOS_Board_Init for flightd to something else to avoid any confusion.

mlyle: THis goes (close to) first thing in pios_board_init.  And should get factored out everywhere IMO.

Board init starts tasks that expect to be able to use the task monitor subsystem.

mlyle: I was looking into doing this on embedded and it looks like we already do ;)

https://github.com/d-ronin/dRonin/blob/next/flight/PiOS/STM32/chibi_main.c#L132

mlyle:
It looks like this PR is the critical path for everything, because faketick depends upon it, upon which integration tests depend.

flight/PiOS/posix/pios_flash_posix.c
```diff
@@ -107,32 +111,23 @@ void PIOS_Flash_Posix_Destroy(uintptr_t chip_id)
...
-	 *
-	 * Alternatively could choose to have a counter maintained using
-	 * atomics here.
-	 */
-	assert(!flash_dev->transaction_in_progress);
-
-	flash_dev->transaction_in_progress = true;
-#endif
+	if (PIOS_Semaphore_Take(flash_dev->transaction_lock,
+				PIOS_SEMAPHORE_TIMEOUT_MAX) != true) {
```
tracernz, line 117: I wonder if timeout should be a param... that kind of thing is annoying in c since you can't have a default though. :(

flight/targets/flightd/fw/main.c
tracernz: Ah. I guess UAVOs don't really depend too much on PIOS_Board_Init as long as they're not saved to flash, it's encouraging someone to do that one day though. :stuck_out_tongue:

tracernz (approved):

PR #2150, opened 2018-04-05
transmitter_control: don't arm on settings change

mlyle:
Before, if you had an arming switch in the armed position while you
configured the arming mode, it would immediately arm.  It's a minor
safety improvement to prevent this.


glowtape:
Code looks good to me. Verified working on my Revo.

Without PR: Always Disarmed, arm via remote, then change to Switch Arming, FlightStatus.Armed goes Arming then Armed.
With PR: Always Disarmed, arm via remote, then change to Switch Arming, FlightStatus.Armed stays Disarmed. I need to disarm and rearm the remote for actual arming on the quad.

Quad doesn't disarm now, if you configure Always Disarmed while armed. But who would get into that specific situation, anyway?

tracernz (approved):

PR #2151, opened 2018-04-05
posix: add "faketick" infrastructure

mlyle:
Now the clock can represent (monotonic) system time, or it can represent
a value controlled by the sensors subsystem.  The latter case allows us
to run the simulator without it ever "falling behind".

The sensor ticking is gated by control updates, so
that integration tests can get (somewhat) repeatable results.  It won't
be perfect because there will be nondeterministic scheduling between
different (non-realtime) tasks.

Depends/based on #2149.

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -828,7 +828,12 @@ static void stabilizationTask(void* parameters)
 					(dT_measured < dT_expected * 0.85f)) {
 				frequency_wrong = true;
 #ifdef FLIGHT_POSIX
-				printf("Stabilization: frequency wrong.  dT_measured=%f, expected=%f\n", dT_measured, dT_expected);
+				if (!PIOS_Thread_FakeClock_IsActive()) {
+					printf("Stabilization: frequency wrong.  dT_measured=%f, expected=%f\n",
+							dT_measured, dT_expected);
+				} else {
+					frequency_wrong = false;
+				}
 #endif
 			}
 		}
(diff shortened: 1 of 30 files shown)


mlyle:
posix wdg, queue, and normal sleeps etc adapted.  semaphores are the main thing that still aren't.

At this point-- in fake tick mode, can be flying a gcs rcvr flight, and ctrl-z GCS, wait a few seconds, and fg and pick up exaclty where you left off.

mlyle:
Now signal the status through the hwsimulation object, and tune up interval of gcsrcvr-related tick.  This is done.

mlyle:
This is rebased.  It's a bit of a mess now, though, because there's further faketick improvements in the integration test branch.  Still may make sense to land this one to lower the scope of review.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -828,7 +828,12 @@ static void stabilizationTask(void* parameters)
 					(dT_measured < dT_expected * 0.85f)) {
 				frequency_wrong = true;
 #ifdef FLIGHT_POSIX
-				printf("Stabilization: frequency wrong.  dT_measured=%f, expected=%f\n", dT_measured, dT_expected);
+				if (!PIOS_Thread_FakeClock_IsActive()) {
+					printf("Stabilization: frequency wrong.  dT_measured=%f, expected=%f\n",
```
tracernz, line 832: Should probably go to stderr

tracernz (approved):
Looks good. Locking looks correct. Not sure what mac gets upset about. :/

PR #2152, opened 2018-04-06
flight: preliminary "flipover" mode support

mlyle:
To configure, configure a quad with 3D output thrust, and a
stabilization bank with MANUAL/MANUAL/DISABLED/Flipover Mode Thrust
Reversed.  In this flight mode, hold the stick left, right, forward, or
back (or other directions, I guess ;) and apply a small pulse of
throttle to roll upright.

Completely untested.  Use caution.


ufoDziner:
I'm assuming that this is dshot dependent?
Is there a 3D setup page that can be referenced for throttle and motor configurations?
I've tested the 3D stuff a while ago, but had it setup for actual 3D flight and don't recall those settings at all either.  With a little guidance, I'm happy to test and provide example screenshots that can be included in documentation for others if there isn't some already (I couldn't find any).

mlyle:
No documentation, and I don't really want there to be yet.  Any 3D config (though I don't really recommend non-dshot ones).  Set ESCs to 3D, set channelDeadband to 1 or 2 to get correct rotation direction.  Input/throttle doesn't need to be 3D (for 3D, put neutral somewhere in the middle of span).

If we want to support this officially, we'd need to A) create a real flight mode for it, B) add a way to set the 3D mode of actuators on the GCS output screen.

ufoDziner:

glowtape:
So the loose idea I had was to create a flight mode, that forces all axes to manual, and then communicate it to Actuator somehow, that does different amounts of magic depending on whether there's 3D mode configured or not (demanding appropriate motor direction via DShot commands, when it detects a change from/to Flip Over thrust mode and no 3D configured). This PR is pretty much ways there.

Over in ManualControlCommand, the idea was to track the last valid non-FlipOver mode mode, and keep enforcing it, if someone tries to switch to Flip Over mode while armed. Mostly to prevent drama say if it's in the middle of a tri-switch and someone switches past it (the switch is going to hang for a very short moment in the middle position), or just plainly switches to it by accident.

ufoDziner:
Tested with 2 different quads.  One blheli_32, second with _s.  Both worked well.  FlipOver Mode Thrust Reversed worked well when inverted on the ground.  Regular FlipOver Mode didn't help at all.

I fly with camera reprojection, so I'm constantly in a special Stabilization Mode (ST3 is normal for me).  This will really need it's own mode name so it's not easily confused with other "ST#" modes.

ufoDziner:

mlyle:

ufoDziner:

ufoDziner:
New artifacts work really well.  Can easily tell when in the flipover mode, and it flips well.  No issues.  Thanks!!!

glowtape:

mlyle:
Rebased because of misc_math conflict with openlrs code.

flight/Libraries/math/misc_math.c
```diff
@@ -367,6 +367,27 @@ uint32_t randomize_int(uint32_t interval)
 	return rand_val % interval;
 }
 
+/**
```
mlyle, line 370: This is factored to misc_math, because it's now used by the actuator subsystem. (And because it's a generally applicable bit of code).

flight/Libraries/sanitycheck.c
```diff
@@ -211,30 +207,35 @@ DONT_BUILD_IF(MANUALCONTROLSETTINGS_STABILIZATION1SETTINGS_NUMELEM != MANUALCONT
  * @param[in] index Which stabilization mode to check
  * @returns SYSTEMALARMS_CONFIGERROR_NONE or SYSTEMALARMS_CONFIGERROR_MULTIROTOR
  */
-static int32_t check_stabilization_settings(int index, bool multirotor,
-		SharedDefsThrustModeStabBankOptions thrust_mode)
+static int32_t check_stabilization_settings(int index, bool multirotor)
 {
 	uint8_t modes[MANUALCONTROLSETTINGS_STABILIZATION1SETTINGS_NUMELEM];
+	uint8_t thrust_mode;
```
mlyle, line 213: This change is unrelated:   thrust_mode is factored in here to simplify enclosing call/switch statement.

flight/Libraries/sanitycheck.c
```diff
@@ -211,30 +207,35 @@ DONT_BUILD_IF(MANUALCONTROLSETTINGS_STABILIZATION1SETTINGS_NUMELEM != MANUALCONT
...
+			ManualControlSettingsStabilization3ThrustGet(&thrust_mode);
 			break;
 		default:
 			return SYSTEMALARMS_CONFIGERROR_NONE;
 	}
 
 	// For multirotors verify that nothing is set to "disabled" or "manual"
-	if (multirotor) {
+	if (multirotor &&
+			(thrust_mode != MANUALCONTROLSETTINGS_STABILIZATION1THRUST_FLIPOVERMODE) &&
```
mlyle, line 235: Manual mode is now allowed (encouraged, even) for flipover mode on multirotor.  So allow that case.

flight/Libraries/sanitycheck.c
```diff
@@ -136,6 +134,7 @@ int32_t configuration_check()
 			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_HORIZON:
 			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_AXISLOCK:
 			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_VIRTUALBAR:
+			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_FLIPOVERREV:
```
mlyle: Biggest set of changes is in sanity check, allowing the mode, allowing manual under specified circumstances on multirotor, allowing arming with the mode, etc.

flight/Modules/Actuator/actuator.c
```diff
@@ -480,6 +481,10 @@ static void post_process_scale_and_commit(float *motor_vect,
 	 */
 	float maxpoweradd = (*maxpoweradd_bucket) / hangtime_leakybucket_timeconstant;
 
+	if (flip_over_mode) {
```
mlyle, line 484: The core changes are down here in actuator to A) don't do any kind of clip management, and B) when clipping, ensure the motors do not spin, and C) to do some thresholding / deadbanding on axes before applying the mixer.  Stabilization selects whether it wants this behavior, and passes the information down.  Actuator acts upon it.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -568,6 +568,10 @@ static float calculate_thrust(StabilizationDesiredThrustModeOptions mode,
 			vertspeed_desired = desired_thrust;
 			break;
 
+		case STABILIZATIONDESIRED_THRUSTMODE_FLIPOVERMODETHRUSTREVERSED:
```
mlyle, line 571: Negate thrust, so that positive throttle produces negative thrust in the "reversed" mode.  This is what most people will want.  In a true 3D setup (with symmetric throttle) this is not necessary and the non-reversed mode can be used.

Positive thrust with flipover mode may be valuable, for cases where the quad is kinda pinned right-side-up.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -863,6 +867,10 @@ static void stabilizationTask(void* parameters)
 				stabDesired.ThrustMode, &flightStatus,
 				airframe_type, stabDesired.Thrust);
 
+		actuatorDesired.FlipOverThrustMode =
```
mlyle, line 870: This is the real work in stabilization (not much)-- just signal to actuator what we want.

flight/Modules/Actuator/actuator.c
```diff
@@ -704,6 +715,18 @@ static void normalize_input_data(uint32_t this_systime,
 
 	*stabilize_now = throttle_val != 0.0f;
 
+	if (*flip_over_mode) {
+		apply_channel_deadband(&desired.Pitch, 0.25f);
+		apply_channel_deadband(&desired.Roll, 0.25f);
+		apply_channel_deadband(&desired.Yaw, 0.25f);
```
glowtape, line 721: So any input is gonna be allowed? Not just the most prominent axis? Also no axis reversal hikinks (which don't matter IMO)?

mlyle: Yes, any input.  This is important, because e.g. +'s want to flip over diagonally.  BF has some weird code to detect diagonal situations.

The actual flight mode disables yaw, though.

flight/Libraries/math/misc_math.c
```diff
@@ -367,6 +367,27 @@ uint32_t randomize_int(uint32_t interval)
...
+ * @brief Apply deadband to Roll/Pitch/Yaw channels
+ */
+void apply_channel_deadband(float *value, float deadband)
+{
+	if (deadband < 0.0001f) return; /* ignore tiny deadband value */
```
tracernz, line 375: New lines required for these two ifs.

flight/Libraries/sanitycheck.c
```diff
@@ -146,16 +144,13 @@ int32_t configuration_check()
 				}
 				break;
 			case MANUALCONTROLSETTINGS_FLIGHTMODEPOSITION_STABILIZED1:
-				ManualControlSettingsStabilization1ThrustGet(&thrust_mode);
-				error_code = (error_code == SYSTEMALARMS_CONFIGERROR_NONE) ? check_stabilization_settings(1, multirotor, thrust_mode) : error_code;
+				error_code = (error_code == SYSTEMALARMS_CONFIGERROR_NONE) ? check_stabilization_settings(1, multirotor) : error_code;
```
tracernz, line 148: These are looooong lines. I wonder if it wouldn't be easier to read with if/else rather than ternary?

shared/uavobjectdefinition/actuatordesired.xml
```diff
@@ -23,5 +23,12 @@
     <field defaultvalue="0" elements="1" name="SystemIdentCycle" type="uint16" units="samples">
       <description/>
     </field>
+    <field defaultvalue="FALSE" elements="1" name="FlipOverThrustMode" type="enum" units="">
+      <description>TRUE if special "only spin some motors" behavior is requested</description>
```
tracernz: That's not very clear at all TBH.

tracernz (commented):
Looked at the first two commits... more to come later.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1101,6 +1099,9 @@ static void update_stabilization_desired(ManualControlCommandData * manual_contr
                                                STABILIZATIONDESIRED_STABILIZATIONMODE_ATTITUDELQG,
                                                STABILIZATIONDESIRED_STABILIZATIONMODE_LQG };
 
+	const uint8_t FLIPOVER_SETTINGS[3] = {  STABILIZATIONDESIRED_STABILIZATIONMODE_MANUAL,
+                                       STABILIZATIONDESIRED_STABILIZATIONMODE_MANUAL,
```
tracernz: The first column needs to be a tab, else the alignment is baroque :stuck_out_tongue:

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -634,65 +634,67 @@ void simple_artificial_horizon(float roll, float pitch, int16_t x, int16_t y,
 
 void draw_flight_mode(int x, int y, int xs, int ys, int va, int ha, int flags, int font)
 {
-	uint8_t mode;
+	SharedDefsFlightModeOptions mode;
```
tracernz, line 637: :+1:

flight/Modules/OnScreenDisplay/onscreendisplay.c
```diff
@@ -715,13 +717,17 @@ void draw_flight_mode(int x, int y, int xs, int ys, int va, int ha, int flags, i
 			write_string("TAB POI", x, y, xs, ys, va, ha, flags, font);
 			break;
 		}
+		}
```
tracernz: The indentation of the inner block is borked.

shared/uavobjectdefinition/shareddefs.xml
```diff
@@ -27,6 +27,7 @@
         <option>Acrodyne</option>
         <option>LQG</option>
         <option>LQGLeveling</option>
+        <option>FlipOverRev</option>
```
tracernz: Rev? Revision? Revert? Reverse? Revise? These fields are UI, so need to be in English. :wink: Preferably with spaces etc. but the others already don't.

flight/Libraries/sanitycheck.c
mlyle: I hate ternary :P  Still question whether it should be done this PR

shared/uavobjectdefinition/actuatordesired.xml
mlyle: How good do we want descriptions to be for internal objects (e.g. not configuration)?  Is the audience end-users, developers, etc?  It's also documented some when mode is used.

flight/Modules/OnScreenDisplay/onscreendisplay.c
mlyle: How so?

flight/Modules/ManualControl/transmitter_control.c
mlyle: It matches what's in the file already.

shared/uavobjectdefinition/shareddefs.xml
mlyle: It has to fit in a relatively narrow field places so I was going for not being the maximum width.  Do you have a preferred form?

flight/Modules/OnScreenDisplay/onscreendisplay.c
mlyle: (We tend not to indent for {} in case because otherwise you run out of tabstops reaaaal quick)

flight/Libraries/math/misc_math.c
mlyle: Just relocated this code... didn't change it.

flight/Libraries/sanitycheck.c
tracernz: Well, it's never gonna happen unless we fix it while we're working on it IMO. :stuck_out_tongue:

shared/uavobjectdefinition/actuatordesired.xml
tracernz: Ah, missed that it was not a setting. It's confusing because you are explaining one special mode by defining another equivalent special mode. :stuck_out_tongue:

flight/Modules/ManualControl/transmitter_control.c
tracernz: They all look very wrong with normal 8 spaces, as in code standard + how github shows it :(

flight/Modules/OnScreenDisplay/onscreendisplay.c
tracernz: In the diff for this PR is a counterexample :stuck_out_tongue_closed_eyes: https://github.com/d-ronin/dRonin/pull/2152/files/ef9fe7779342502133b63e00e8dd226f7ca3d013..35a076e2be96b621baa58b181ad05e69929c6f11#diff-4f4804c22880c52150730a11f3c4a5baR1177

shared/uavobjectdefinition/shareddefs.xml
tracernz: Looking at the other options in this same field I see no abbreviations (except LQG which shouldn't really be a flightmode):
Stabilized# rather than Stab#
AltitudeHold rather than AltHold
PositionHold rather than PosHold
ReturnToHome rather than RTH
etc... this is something I very much like about most of our UAVOs.

FlipReverse could work I think?

-e- FlipReversed matches the tense of e.g. Stabilized better

flight/Modules/OnScreenDisplay/onscreendisplay.c
mlyle: I'm saying the {} after a label.

I don't really favor the {} after a label, preferring (void) 0; to avoid bracket weirdness.  Would you prefer that?

flight/Libraries/sanitycheck.c
mlyle: Yah, I always try to make at least some things better in each PR, but ... if I were to fix everything, we'd get too much diff :P

mlyle:
A bunch of small commits for formatting/review comments.

flight/Modules/OnScreenDisplay/onscreendisplay.c
tracernz: I don't really care too much tbh. It just looks weird with two scopes closing at the same indentation.... Maybe we should just write everything in Python? :grimacing:

tracernz (approved):

flight/Modules/OnScreenDisplay/onscreendisplay.c
mlyle: I went with the (void) 0; trick to get rid of one.

PR #2154, opened 2018-04-07
Lots of small changes to hardware support

mlyle:
* Support panic blinks on targets with optional '2811 (before we'd crash and not blink)
* Revolution: Don't init internal I2C bus on cheapo clones.
* pios_mpu: Woops, don't enter high output rate modes on icm-20608
* Revolution: Don't config RX port by default-- this isn't necessary but it seems we shouldn't, just in case there's a conflict on subsequent crummy hardware.
* pios_bmp280: Make sure that the SPI version, counts alone, for "could have baro". This is another future trap averted.

### DIFF
--- a/flight/PiOS/Common/pios_hal.c
+++ b/flight/PiOS/Common/pios_hal.c
@@ -281,9 +281,9 @@ static inline void PIOS_HAL_Err2811(bool on) {
 		} else {
 			PIOS_WS2811_set_all(pios_ws2811, 0, 32, 0);
 		}
-	}
 
-	PIOS_WS2811_trigger_update(pios_ws2811);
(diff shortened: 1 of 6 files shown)


mlyle:

glowtape:
I think this LGTM. Tested on my Revo clone.

tracernz (approved):
The revo i2c thing makes it impossible to use internal baro/mag if you have a different memory chip? Did you see a specific issue that prompted it? Not so keen on that one, but the rest are good. :+1:

mlyle:
Yah, on Rev2 Omnibus F4 the I2C bus is stuck, unfortunately.  I have never seen I2C sensors on a omnibus-type-revo, and I've never seen an alternate flash on a revo pattern board.

PR #2155, opened 2018-04-07
vtolpathfollower: allow configuration of RTH height

mlyle:
Per discussion on forum.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -639,17 +639,17 @@ static void go_enable_rise_here()
 
 	float down = positionActual.Down;
 
-	float tmp;
+	float min_rise, rth_alt;
 
-	VtolPathFollowerSettingsReturnToHomeMinRiseGet(&tmp);
+	VtolPathFollowerSettingsReturnToHomeMinRiseGet(&min_rise);
+	VtolPathFollowerSettingsReturnToHomeAltitudeGet(&rth_alt);
+
 	// Set the target altitude for MIN_RISE above our current alt
-	down -= tmp;
-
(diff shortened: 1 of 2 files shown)


glowtape:
LGTM.

This is going to get UI eventually, so I assume we rely on the user to not enter stupid shit via the UAVO browser until then?

mlyle:
Yah, there's no UI at all for the path followers.

tracernz:
> This is going to get UI eventually, so I assume we rely on the user to not enter stupid shit via the UAVO browser until then?

If that's a concern, limits on the UAVO fields is the correct way to handle it.

mlyle:
UAVO browser doesn't enforce limits though, does it?  (And it's often good that it doesn't).

I view making the mainline UI as the time to figure out sane field limits for everything.  (All of vtolpathfollower needs it)

tracernz:
I couldn't remember if it shows you when you're out of limits or not. It should show you (e.g. by colour), but not enforce IMO.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -641,12 +639,17 @@ static void go_enable_rise_here()
 
 	float down = positionActual.Down;
 
+	float tmp;
```
tracernz: Can you name this `rth_alt` or similar? It'll make the code below a bit more readable.

mlyle: I can split it into two named vars

tracernz: Ahhh, yeah, didn't spot that it was used for multiple things. I think splitting is best tbh. Compiler should be able to figure out it can use the same register.

tracernz (approved):

PR #2157, opened 2018-04-08
pios_bmi160: Make taskless.

glowtape:
Becauuuuuuuuuuuse!

With current quad config, right after boot, CPU usage without this is 50% with random upticks to 51%, with the PR, it's 49% with random upticks to 50%, i.e. saves 1%.

### DIFF
--- a/flight/PiOS/Common/pios_bmi160.c
+++ b/flight/PiOS/Common/pios_bmi160.c
@@ -663,6 +663,8 @@ static bool PIOS_BMI160_callback_gyro(void *ctx, void *output,
 		return false;
 	}
 
+	/* Only fail, if sensor data readout fails (-2). Temperature readouts are assumed
+	   to fail intermittently only, if at all, and failures don't taint main sensor data. */
 	if (PIOS_BMI160_parse_data(dev) < -1) {
 		return false;
 	}
--- a/flight/PiOS/Common/pios_bmi160.c
+++ b/flight/PiOS/Common/pios_bmi160.c
@@ -494,7 +494,10 @@ bool PIOS_BMI160_IRQHandler(void)
 	return need_yield;
 }
 
-
+/**
+ * @brief Reads out SPI and parses the data
+ * @returns 0 on success, -1 on temperature readout failure, -2 on sensor readout failure.
+ */
 static int PIOS_BMI160_parse_data(struct bmi160_dev *dev)
 {
 	enum {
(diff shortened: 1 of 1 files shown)


mlyle:
Nice :D  Always great for Seppuku to be best supported :D

flight/PiOS/Common/pios_bmi160.c
```diff
@@ -495,165 +495,198 @@ bool PIOS_BMI160_IRQHandler(void)
...
 
+	dev->sensor_ready |= SENSOR_ACCEL;
+
+	// Get the temperature
+	// NOTE: We do this down here so the chip-select has some time to go low. Strange things happen
+	// When this is done right after readin the accels / gyros
+	if (dev->temp_interleave_cnt % dev->cfg->temperature_interleaving == 0){
 		if (PIOS_BMI160_ClaimBus() != 0)
-			continue;
+			return 0;
```
tracernz: Err, 0 is usually success.

flight/PiOS/Common/pios_bmi160.c
```diff
@@ -495,165 +495,198 @@ bool PIOS_BMI160_IRQHandler(void)
...
 
-		if (PIOS_SPI_TransferBlock(dev->spi_id, bmi160_tx_buf, bmi160_rec_buf, BUFFER_SIZE) < 0) {
+		uint8_t bmi160_tx_buf[BUFFER_SIZE] = {BMI160_REG_TEMPERATURE_0 | 0x80, 0, 0, 0, 0, 0,
+				0, 0, 0, 0, 0, 0, 0};
+		if (PIOS_SPI_TransferBlock(dev->spi_id, bmi160_tx_buf, bmi160_rec_buf, 3) < 0) {
 			PIOS_BMI160_ReleaseBus();
-			continue;
+			return 0;
```
tracernz: As above.

glowtape: I've interpreted the comment in front of the if clause, that this has the chance of being occasionally problematic, so we probably might not care that it fails here and there.

glowtape: Actually re-reading the diff, yeah, you're right. It's functionally different that way.

flight/PiOS/Common/pios_bmi160.c
```diff
@@ -494,166 +494,202 @@ bool PIOS_BMI160_IRQHandler(void)
...
 
-		accel_data.temperature = temperature;
-		gyro_data.temperature = temperature;
+	*next_call = 0;		/* Baby you know you can call me anytime */
 
+	if (PIOS_Semaphore_Take(dev->data_ready_sema, ms_to_wait) != true) {
+		return false;
+	}
+
+	if (PIOS_BMI160_parse_data(dev) < -1) {
```
tracernz: Maybe put a comment here saying something like "If temp readout fails (-1), we use the previous successful reading".

tracernz (approved):

tracernz:
Tested this with #2158 on Seppuku. It was happy with healthy sensors. Don't have a really good way to check sensor error paths since we killed the Disco targets.

PR #2158, opened 2018-04-09
pios_bmp280: Make taskless.

glowtape:
Saves some memory I guess. Works on my Seppuku. Don't have a target with the baro on I2C.

### DIFF
--- a/flight/PiOS/Common/pios_bmp280.c
+++ b/flight/PiOS/Common/pios_bmp280.c
@@ -39,8 +39,6 @@
 #include "pios_queue.h"
 
 /* Private constants */
-#define BMP280_TASK_PRIORITY	PIOS_THREAD_PRIO_HIGHEST
-#define BMP280_TASK_STACK		512
 
 /* BMP280 Addresses */
 #define BMP280_I2C_ADDR		0x76  // 0x77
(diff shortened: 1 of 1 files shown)


flight/PiOS/Common/pios_bmp280.c
```diff
@@ -540,37 +526,38 @@ int32_t PIOS_BMP280_Test()
...
+	read_adc_result = PIOS_BMP280_ReadADC();
+	PIOS_BMP280_ReleaseDevice();
 
-		float calc_alt = 44330.0f * (1.0f - powf(data.pressure / BMP280_P0, (1.0f / 5.255f)));
+	if (read_adc_result) {
+		return false;
+	}
 
-		data.temperature = ((float) dev->compensatedTemperature) / 256.0f / 100.0f;
+	struct pios_sensor_baro_data *data = (struct pios_sensor_baro_data *)output;
```
tracernz, line 552: What's the purpose of this var? Doesn't seem to add anything.

glowtape: Right, should probably put it directly in the function header.

tracernz: Ahhh, it's for the cast. No, it's fine as is. 👍

tracernz (approved):

tracernz:
Tested this with #2157 on Seppuku. It was happy with healthy sensors. Don't have a really good way to check sensor error paths since we killed the Disco targets.

PR #2160, opened 2018-04-13
gcs: Fix tooltip for Motor Power Limitation.

glowtape:
Used mostly the UAVO field description and replaced "Actuator limitation" with "Motor power limitation", so it matches the label next to the spinner. Previous text was the same as for curve fit.

Also, removed a line/paragraph break in the middle of the last sentence in the curve fit tooltip.


tracernz (approved):

mlyle:
Thx

PR #2162, opened 2018-04-15
pios_hal: handle RFM22 being missing on revo

mlyle:


tracernz (approved):
Indeed. Dirty cloners only putting half the hardware on board. :stuck_out_tongue:

PR #2164, opened 2018-04-17
pios_com: ensure that tx_start is always called

mlyle:
It's redundant often.  But USB sometimes does nothing for tx_start and
needs another kick, etc.


tracernz (approved):
Good catch. :+1: Hopefully `com_dev->driver->available` should clear the buffer in most cases though.

PR #2169, opened 2018-04-18
flight: f3: ADC related cleanups

### LINKED ISSUE #1002: Should remove pios_dma/pios_dma_hooks infrastructure
Only used on F3 by ADC driver.

mlyle:
Remove ADC hooks infrastructure in favor of simple interrupt dispatch.
Eliminate some of the nasty code on DTFc (forcing modulesettings on).
Some small code cleanups.

Fixes #1002 

Will require -very- careful verification.


mlyle:
Confirmed functional on DTFc.

mlyle:
Read my own diff, paying special attention to dma channels (please double check this too!) and it looks OK.

mlyle:
Looks good on omnibusf3.

mlyle:
Verified OK on SPRF3E  (These are the 3 I'm spot testing).

flight/PiOS/inc/pios_internal_adc.h
```diff
@@ -31,7 +31,12 @@
 #ifndef PIOS_INTERNAL_ADC_H
 #define PIOS_INTERNAL_ADC_H
 
-extern void PIOS_INTERNAL_ADC_DMA_Handler();
+#if defined(STM32F30X)
+/* TODO: transition other targets to work this way. */
+void PIOS_INTERNAL_ADC_DMA_Handler(uintptr_t internal_adc_id);
```
tracernz, line 36: Ideally a real pointer, but yeah. :stuck_out_tongue_winking_eye:

flight/targets/dtfc/fw/pios_board.c
```diff
@@ -255,17 +255,10 @@ void PIOS_Board_Init(void)
...
-		PIOS_HAL_CriticalError(PIOS_LED_ALARM, PIOS_HAL_PANIC_ADC);
-	if (PIOS_ADC_Init(&pios_internal_adc_id, &pios_internal_adc_driver, internal_adc_id) < 0)
-		PIOS_HAL_CriticalError(PIOS_LED_ALARM, PIOS_HAL_PANIC_ADC);
-
-	// Enable battery monitor module.
-	ModuleSettingsData modulesettings;
-	ModuleSettingsGet(&modulesettings);
-	modulesettings.AdminState[MODULESETTINGS_ADMINSTATE_BATTERY] = MODULESETTINGS_ADMINSTATE_ENABLED;
-	ModuleSettingsSet(&modulesettings);
+	uintptr_t unused_adc;
```
tracernz, line 258: If we're going to have the var, we may as well call it what it actually is right (i.e. internal_adc_id)? Otherwise, just change the API to accept+handle NULL and eliminate it.

tracernz (commented):
Looks good. I don't see the point in renaming internal_adc_id to unused_adc at all though. That's just going to be another bunch of diff one day if it does get used, and I can't really see the value in naming it unused anyhow?

flight/targets/dtfc/fw/pios_board.c
mlyle: I don't know what to call these.  We have the actual internal_adc internal adc device handle, and the wrapped PIOS_ADC device handle for the internal ADC.  Both of which are mostly unused.  I can cast them to void, which seems like the lesser of evils.

mlyle: Oh, it's not a retval.

mlyle: So yah, the whole reason I did this was we had both pios_internal_adc_id which was the actual internal adc id device, and the wrapped version of the same, which was internal_adc_id.  If anything, the thing you'd want to have broader scope / the pios name is the wrapped version, so the names are swapped.  But I dare not swap just the F3 targets, because all the F4 targets would then be different.

tracernz: So, this is the one you actually want to use with PIOS_ADC_* (which is the interface everybody should use other than PIOS_INTERNAL_ADC_Init). The only callers of any other PIOS_INTERNAL_ADC_* functions should be inside PIOS_ADC_*. I just had a quick look and found a bug:
https://github.com/d-ronin/dRonin/blob/cd7a75d880e031c72106747d779fb352bf645203/flight/Modules/System/systemmod.c#L761
That attempts to use a pios_internal_adc_dev as a pios_adc_dev.

mlyle: That code's already removed in this PR because no targets provided configuration for it.

mlyle: (it's also F4 specific; if we want to bring it back, we should have each pios_internal_adc driver provide an implementation rather than do target-specific configuration)

mlyle: Pushed-- removed the DevicePinGet call entirely, which was used only by flyingpio at this point.

tracernz: Might as well kill PIOS_ADC_Available since it's a companion to PIOS_ADC_DevicePinGet, and doesn't appear to be much use without a way to actually use the pin.

mlyle: Done, thx for pointing this out.

For multiple drivers, the way subdrivers are called each time to get index counts seems unsafe and should probably get improved, but out of scope.

tracernz (approved):

PR #2171, opened 2018-04-19
flightd: lots of cleanups

### LINKED ISSUE #1228: flyingpi: no way to configure receiver type

mlyle:
1. Allow specifying a receiver protocol on the command line instead of
it being hardcoded.
2. Static scope for some posix PIOS_SYS globals.
3. Rename first_arg to hw_argseen, because that's what it really means--
whether we've seen a hardware argument yet.
4. Don't make some non-SPI related options conditionalized on SPI.

Fixes #1228


tracernz (approved):
Great work :+1:

PR #2172, opened 2018-04-19
manualcontrol: improve arming (always disarmed)

mlyle:
Before, setting arming state to always disarmed would not actually
disarm, if you were already armed.  Ensure that it always speedily
disarms.


mlyle:
This tested satisfactorily in sim, and is relatively simple.

tracernz (approved):

PR #2173, opened 2018-04-20
gcs: qmessagebox needs a valid widget parent

### LINKED ISSUE #1285: Inaccessible warning popup
- Seen in version (release version or githash): Build #575   e009c5e14c7
- On flight controller (if applicable): airbot F4

Issue details:  Running setup wizard I hit "calibrate" before the board had reconnected.  An error window opened saying no board was connected - but it was behind the main window which was blocked by the wizard so I couldn't get to the error window.  I had to kill it from the windows task bar.  GCS crashed shortly after and sent a crash report.

Restarting GCS and trying to open help with no board connected I got another crash.

mlyle:
Make sure it's always sane.  Use the main window if we're not a widget
ourselves.

Fixes #1285.

There's a close cousin of issues with file dialogs, which I'll try to get to next.  Also, some unused calls and a couple of useless dialogs were removed.

### DIFF
--- a/ground/gcs/src/plugins/config/configplugin.cpp
+++ b/ground/gcs/src/plugins/config/configplugin.cpp
@@ -139,7 +139,7 @@ void ConfigPlugin::onAutopilotDisconnect()
   */
 void ConfigPlugin::eraseAllSettings()
 {
-    QMessageBox msgBox;
+    QMessageBox msgBox((QWidget *)Core::ICore::instance()->mainWindow());
     msgBox.setText(tr("Are you sure you want to erase all board settings?."));
     msgBox.setInformativeText(tr("All settings stored in your board flash will be deleted."));
     msgBox.setStandardButtons(QMessageBox::Ok | QMessageBox::Cancel);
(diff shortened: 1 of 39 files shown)


ground/gcs/src/plugins/config/configplugin.cpp
```diff
@@ -181,19 +181,16 @@ void ConfigPlugin::eraseFailed()
...
     disconnect(objper, &UAVObject::objectUpdated, this, &ConfigPlugin::eraseDone);
-    QMessageBox msgBox;
-    msgBox.setText(tr("Error trying to erase settings."));
-    msgBox.setInformativeText(
-        tr("Power-cycle your board after removing all blades. Settings might be inconsistent."));
-    msgBox.setStandardButtons(QMessageBox::Ok);
-    msgBox.setDefaultButton(QMessageBox::Ok);
-    msgBox.exec();
+
+    (void) ignored = QMessageBox::critical((QWidget *)Core::ICore::instance()->mainWindow(),
```
tracernz: oops

mlyle: Haha I'm losing patience for this crap :D

ground/gcs/src/plugins/config/configplugin.cpp
```diff
@@ -139,7 +139,7 @@ void ConfigPlugin::onAutopilotDisconnect()
   */
 void ConfigPlugin::eraseAllSettings()
 {
-    QMessageBox msgBox;
+    QMessageBox msgBox((QWidget *)Core::ICore::instance()->mainWindow());
```
tracernz, line 142: Discussed use of c-casts on IRC already. AFAICT these could all be static_cast since we can guarantee mainWindow returns the right type? Otherwise dynamic_cast.

tracernz (commented):
Quite a few instances of spaces after c-style casts, which is against the code style.

ground/gcs/src/plugins/config/configplugin.cpp
tracernz: If you _really_ don't want to do it, I guess I will clean them up later.

mlyle: I'd like to find a way to do this in bulk, across the whole GCS, rather than spend our time fiddling with it.

PR #2175, opened 2018-04-20
gcs: attitude: make sixpoint work on first try

### LINKED ISSUE #2094: GCS/Sparky2: Six point data collection timed out
* Seen in version (release version or githash): Release-20180212.2: 20180213-020e58fb
* On flight controller (if applicable): Sparky2

Issue details:
Using OSX, six point calibration always times out after the clicking Save Position for the first time.
Tested with a USB cable and Telemetry through a BT.

The message is: Six point data collection timed out.

mlyle:
Fixes #2094.

### DIFF
--- a/ground/gcs/src/plugins/config/configattitudewidget.cpp
+++ b/ground/gcs/src/plugins/config/configattitudewidget.cpp
@@ -332,6 +332,10 @@ void ConfigAttitudeWidget::updateCalibrationEnabled()
         m_ui->cbCalibrateAccels->setChecked(have_accel);
         m_ui->cbCalibrateMags->setChecked(have_mag);
     }
+
+    if (have_accel || have_mag) {
+        configureSixPoint();
+    }
 }
 
 /**


tracernz:
Seems like a more correct fix would be to just connect to [QCheckbox::stateChanged](http://doc.qt.io/qt-5/qcheckbox.html#stateChanged) here https://github.com/mlyle/dronin/blob/0caee6929d6f4a4b056cdd5cb96d2661c4b29d0f/ground/gcs/src/plugins/config/configattitudewidget.cpp#L136-L139 ? (I'm assuming the bug is that the clicked signal doesn't fire when the state is changed programatically, and that stateChanged will)

mlyle:
Problem is it never gets set up at all, even if the initial state is never changed.  (Default is both checked, and most targets want both checked).

tracernz (approved):
Feels a bit hacky, but works so meh.

Issue #2166, opened 2018-04-17
SetupWizard not currently loading on Windows


mlyle:
Probably a DLL circdep thing.

tracernz:
Got Windows VM up, confirmed rev cd7a75d8 is good, rev 7e1292ad is bad (as reported by @yamahabest, thanks!). Tried the dumb thing first and copied SetupWizard + Core plugins from good rev to bad rev, didn't help. Started setting up a dev environment with msys2 to bisect. Will continue after work.

tracernz:
Almost done bisecting this... one or two builds left to do.

tracernz:
```
Bisecting: 0 revisions left to test after this (roughly 0 steps)
[e31466c9b68962faa803d69cb82483c2b653d2d9] remove RFM22 code but leave OpenLRS impl in
```

PR #2178, opened 2018-04-20
GCS: Fail unit tests if plugins failed to load

tracernz:
Good for things like #2166 (although we don't currently run regular CI tests on Windows).

### DIFF
--- a/ground/gcs/src/app/main.cpp
+++ b/ground/gcs/src/app/main.cpp
@@ -451,11 +451,17 @@ int main(int argc, char **argv)
             QString msg(QCoreApplication::tr(
                         "Some problems were encountered whilst loading the following plugins: "));
             msg.append(plugins.join(QStringLiteral(", ")));
-            QMessageBox msgBox(QMessageBox::Warning,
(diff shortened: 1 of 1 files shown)


tracernz:
Note: should fail currently as the fix for above issue hasn't merged... let's check:
build artifacts please

tracernz:
Ahhh. Of course we don't build GCS tests on the release builder. Not sure this will be as useful as I had hoped but at least it will catch mac (and general *nix) loading issues.

tracernz:
(Note the fail here was unrelated/due to flight test)

mlyle:
Wow, that one on flight test looks like it actually *did* get stuck somehow.

mlyle:

mlyle:
Sucks that the error strings here didn't tell us of the true problem (though wine -did-).

Just an aside: I think on MacOS and Linux, we had libuploader in the search path and loading a plugin that depended upon it loaded the library... but didn't init the plugin.  This is bad because we could have called into the plugin before the plugin init code had all run.  Whereas on Windows it wasn't able to load it at all.  (So we shouldn't "fix" the load path issue).

PR #2179, opened 2018-04-20
assert: stop other threads

### LINKED ISSUE #1105: PIOS_Assert should maybe stop other threads
If something bad has happened it may impact other threads... and we can't be sure the system will operate safely if any given thread has stopped. Thoughts?

mlyle:
Fixes #1105


flight/PiOS/STM32/pios_debug.c
```diff
@@ -153,6 +157,15 @@ void PIOS_DEBUG_PinValue4BitL(uint8_t value)
  */
 void PIOS_DEBUG_Panic(const char *msg)
 {
+#ifdef PIOS_INCLUDE_RTOS
+	/* Stop all other threads (once we get to the infloop, at least)
+	 * by elevating our priority.  HIGHPRIO is a chibios constant,
+	 * so if we change RTOS we'll know this will need to change by
+	 * compile error.
+	 */
+	PIOS_Thread_ChangePriority(HIGHPRIO);
```
tracernz: Did you consider chSysHalt (not sure if our old rel has it)? http://chibios.sourceforge.net/docs3/rt/group__system.html#gaffab4052111407f092c7c65466a61450

-e- or maybe chSysDisable http://chibios.sourceforge.net/docs3/rt/group__system.html#ga8df460d6729d50d248f8d8a32497d727

mlyle: We could do that.  Couldn't do it this early in case PIOS_COM blocks.  I dunno, it seems like a wash.

mlyle: (Or even just PIOS_IRQ_Disable...)

tracernz: PIOS_IRQ_Disable seems like a good idea. That will kill RTOS (because PendSV + SVCall will be masked?) + interrupt handlers in one fell swoop right?

mlyle: Yes, and it will even work on flyingpio, so will do that.

mlyle:
Rewrote and force pushed.

tracernz (approved):

PR #2180, opened 2018-04-20
sprf3e: deal with crappy brushed clones

mlyle:
Make sure pins are nailed low during bootloader, since there's no
pulldowns!  Whatta kludge.  Even with this one of four motors twitches.

We should consider something like this globally in the bootloader across
all targets.

### DIFF
--- a/flight/targets/sprf3e/bl/pios_board.c
+++ b/flight/targets/sprf3e/bl/pios_board.c
@@ -56,7 +56,10 @@ void PIOS_Board_Init()
 			.GPIO_PuPd  = GPIO_PuPd_DOWN
 		};
 
-		/* Nail all servo channels as outputs, LOW */
+		/* Nail all servo channels as outputs, LOW.
+		 * Mostly for brushed clones of SPRF3E to ensure we don't
+		 * spin up (they should provide pulldowns, but meh.)
+		 */
 		channels[i].pin.gpio->BSRR = channels[i].pin.init.GPIO_Pin << 16;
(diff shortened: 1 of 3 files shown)


tracernz:
> We should consider something like this globally in the bootloader across
all targets.

And arguably in the assert handler, and default/busfault/etc handler. Driving pins hard on boards with alternative functions on the pins will cause bus-fighting though... We should be sure to only enable pull up/down resistors for those cases.

flight/targets/sprf3e/bl/pios_board.c
```diff
@@ -44,6 +44,24 @@ uintptr_t pios_com_telem_usb_id;
...
+	for (int i = 0; i < num_channels; i++) {
+		GPIO_InitTypeDef init = {
+			.GPIO_Pin = channels[i].pin.init.GPIO_Pin,
+			.GPIO_Speed = GPIO_Speed_2MHz,
+			.GPIO_Mode  = GPIO_Mode_OUT,
+			.GPIO_OType = GPIO_OType_PP,
+			.GPIO_PuPd  = GPIO_PuPd_DOWN
+		};
+
+		/* Nail all servo channels as outputs, LOW */
```
tracernz: Can you please note that this is only required for random brushed clones?

mlyle:
WDG will hopefully kick in for the other cases, at least, limiting the damage.  Comment improved.

tracernz (approved):

PR #2189, opened 2018-04-30
lqg: Expose control cost on LQR, remove beta offset for yaw.

glowtape:
Expose the control cost on the LQR, because why not, and also because I anticipate needing it for fixing yaw on brushed and eventually large copters. R defaults to 1, which was the constant used in lqg.c before.

Removed beta offset on yaw axis. This was a hack from the very beginning to imply a direct torque beta component, that TauLabs' newest EKF based autotune evaluates. It defaults to zero, anyway, and I don't intend to use it in future (it'd need a plant specific value anyway, instead of a static offset, to make sense).

### DIFF
--- a/shared/uavobjectdefinition/lqgsettings.xml
+++ b/shared/uavobjectdefinition/lqgsettings.xml
@@ -8,7 +8,7 @@
     <field defaultvalue="25, 75, 1, 0.000003, 0.000001, 0.3" name="RTKF" type="float" units="" elementnames="R, YawR, Q1, Q2, Q3, BiasLimit">
         <description>Weights for the Rate-Torque Kalman filter.</description>
     </field>
-    <field defaultvalue="0.000025, 0.0001, 0, 0.000025, 0.0001" name="LQR" type="float" units="" elementnames="Q1, Q2, YawBetaOffset, YawQ1, YawQ2">
+    <field defaultvalue="1, 0.000025, 0.0001, 1, 0.000025, 0.0001" name="LQR" type="float" units="" elementnames="R, Q1, Q2, YawR, YawQ1, YawQ2">
         <description>Weights for the LQ-regulator.</description>
     </field>
   </object>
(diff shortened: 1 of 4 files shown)


shared/uavobjectdefinition/lqgsettings.xml
```diff
@@ -8,7 +8,7 @@
     <field defaultvalue="25, 75, 1, 0.000003, 0.000001, 0.3" name="RTKF" type="float" units="" elementnames="R, YawR, Q1, Q2, Q3, BiasLimit">
         <description>Weights for the Rate-Torque Kalman filter.</description>
     </field>
-    <field defaultvalue="0.000025, 0.0001, 0, 0.000025, 0.0001" name="LQR" type="float" units="" elementnames="Q1, Q2, YawBetaOffset, YawQ1, YawQ2">
+    <field defaultvalue="1, 0.000025, 0.0001, 1, 0.000025, 0.0001" name="LQR" type="float" units="" elementnames="R, Q1, Q2, YawR, YawQ1, YawQ2">
```
mlyle: I'm concerned about import/export of this, that it will shuffle fields.  e.g. dumps look like

```
      <field name="GyroVar" values="9.99999974738e-06,9.99999974738e-06,9.99999974738e
-05"/>
```

Maybe rename the field?

glowtape: The change in element count (5 to 6) doesn't break import? If so, I can rename it to LQRegulator I guess.

glowtape: Nope it doesn't. Renaming.

tracernz:
That rename reduced the technical accuracy, but  I guess it's okay. :/ What happens if we just put R at the end and leave the field with it's correct name? It should then take it's default value on import I think 
@mlyle ?

glowtape:
Hmm? LQR means Linear Quadratic Regulator. Ignore the G's in the commit title, that's a typo, the actual rename is correct (I should probably fix the title).

As far as the R's to the end, I've removed the YawBetaOffset, which was in the middle anyway. So things would break.

mlyle:
@tracernz 

>What happens if we just put R at the end and leave the field with it's correct name? It should then take it's default value on import I think @mlyle ?

Unfortunately there was an element removed too, so it will just line them up.  Really the export format should contain the field names and they should get matched up.

PR #2190, opened 2018-05-02
crossfire: Fix failsafe timeout, didn't respect the 250ms implemented in #1937.

glowtape:
Move the supervisor reset from after the decoding to in front of it for Crossfire, SBus and IBus. Move it from within the loop to in front of it for DSM and HSUM. SRXL already did it that way.

Should lower the odd chance for a race between the supervisor and the actual processing of data, specifically when data is supplied in blocks (e.g. USART DMA) instead byte-wise. If block processing of up to 40 bytes or so doesn't happen within a supervisor tick, something's awfully wrong to begin with.

Also, Crossfire failsafe didn't actually get upped to 250ms, despite #1937.

### DIFF
--- a/flight/PiOS/Common/pios_crossfire.c
+++ b/flight/PiOS/Common/pios_crossfire.c
@@ -332,8 +332,8 @@ static void PIOS_Crossfire_Supervisor(uintptr_t context)
 	if (++dev->rx_timer > 1)
 		PIOS_Crossfire_ResetBuffer(dev);
 
-	// Failsafe after 50ms.
-	if (++dev->failsafe_timer > 32)
+	// Failsafe after 250ms.
+	if (++dev->failsafe_timer > 156)
 		PIOS_Crossfire_SetAllChannels(dev, PIOS_RCVR_TIMEOUT);
 }
 


glowtape:
I suppose this is interim until I figure out proper locking on STM32.

flight/PiOS/Common/pios_crossfire.c
```diff
@@ -333,7 +333,7 @@ static void PIOS_Crossfire_Supervisor(uintptr_t context)
 		PIOS_Crossfire_ResetBuffer(dev);
 
 	// Failsafe after 50ms.
```
mlyle: Is this right?

mlyle:
I am not sure moving the supervisor reset makes any difference?  The intervals should be the same, shouldn't they, between "beginning of frame processing to beginning of frame processing" to "end of frame processing to end of frame processing"? Can the RTC interrupt even fire while our interrupt handler is running?  Aren't we talking about a vanishingly small interval, too?

glowtape:
This is mostly just based on the theorycrafting about preemption from yesterday on IRC. Take S.Bus for instance, the idle detection is 5 RTC cycles, around 8ms, whereas D16 frame timing is 9ms. Depending on the offset when the RTC ticks in relation to frame reception in the RX code, there's a very minuscule chance it hits in the middle of a buffer copy and trashes the packet in flight, because rx_timer isn't zeroed yet.

In interrupt mode, it gets fed single bytes and therefore resets rx_timer all the time. Should DMA ever get merged, and be active (depending on free DMA streams per target), or anything else in future feed data in blocks, given the default buffer size and assuming the receiver doesn't have a weird cadence triggering idle detection early, we likely receive the frame in full and decode it at once. The copy part of e.g. 25-30bytes is probably 150 cycles/1us. In principle not worth fretting over, but shifting that statement mitigates that tiny window. It's a theoretical edge case.

Both RTC and USART IRQs have middle priority in the hwdefs. How does it work on STM32, the lower IRQ gets preferred treatment when priorities match? The RTC IRQ is lower than any USART one per vector table.

mlyle:
In either case, the behavior is undefined.  Compiler makes no promises about the order assignments happen in etc.

glowtape:
Interesting. You'd figure a for-loop would be some sort of demarcation you don't move instructions across. Well, in this case, it is moot then.

mlyle:
Should any of this be applied except the first and last commit?

glowtape:
Considering the undefined behavior stuff, best to leave things as they are.

Dropped these commits and squashed the Crossfire stuff. Renaming this PR.

mlyle:
Thx!

PR #2195, opened 2018-05-09
gcs: Fix Curve2 UI.

### LINKED ISSUE #513: gcs: curve2 source always shows "Throttle"
Regardless of actual curve2 source, the graph widget in gcs always says that the source of curve2 is "Throttle"

glowtape:
The curve 2 widget was never told it's actually curve 2, so it hid the dropdown and showed a label instead.

Also, the dropdown was never hooked up to the MixerSettings.Curve2Source to begin with.

Should deal with #513?

Fixes #513


mlyle:
Confirmed fixes #513--- thank you very much, that's one less thing to worry 'bout.

PR #2196, opened 2018-05-09
gcs,config: User interface OCDing.

glowtape:
Adjusted paddings, spacings and such, so that when you're switching between pages in the config widget, the tab controls and bottom button row line up on the pixel. Inner group boxes have been adjusted to line up as much as possible (titleless ones do with the titled ones in Ubuntu, don't in Windows).

Removed the remaining out-of-place black borders we inherited from OpenPilot.

The autotune groupbox now expands horizontally, so that it lines up with the hardware page, which is also just a groupbox (either board plugin, or now defaulthwsettings, too). The inner HTML control however doesn't expand beyond what it did before. The buttons don't expand beyond that either.

Stuck defaulthwsettings into a groupbox, so that boards without specific plugin forms (e.g. Revo) don't look entirely like everything's just puked onto a blank canvas.

(Would love to stick autotune and the hardware stuff into tab controls instead of groupboxes, for visual consistency, but I don't want to go through all board plugins and juggle controls around.)

Tested on Windows and Ubuntu.

### DIFF
--- a/ground/gcs/src/plugins/boards_brotronics/luxconfiguration.ui
+++ b/ground/gcs/src/plugins/boards_brotronics/luxconfiguration.ui
@@ -27,6 +27,18 @@
    <string>Form</string>
   </property>
   <layout class="QVBoxLayout" name="verticalLayout_2">
+   <property name="leftMargin">
+    <number>12</number>
+   </property>
+   <property name="topMargin">
+    <number>12</number>
+   </property>
+   <property name="rightMargin">
+    <number>12</number>
+   </property>
+   <property name="bottomMargin">
+    <number>12</number>
+   </property>
    <item>
     <widget class="QGroupBox" name="groupBox_2">
      <property name="title">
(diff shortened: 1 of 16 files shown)


mlyle:
Is this intended for inconceivable?

glowtape:
Yea sure, why not.

glowtape:
Some visual documentation.

dr_ui

1) Made sure the tab bars are all aligned between tabs that have them.
2) Removed these annoying borders. They are kinda redundant and look dumb (IMO).
3) Made sure the buttons are all aligned between tabs that have them.

dr_hwborder

Added a groupbox border around the defaulthwplugin, looking slightly nicer and visually in line with at least the Seppuku, SPRF3e plugins.

glowtape:
PS, the dropdowns are disabled because the Revo had a completely different branch installed, when taking the screenshot.

I just flashed it to this one and they're fine. I was confused about this screenshot after rechecking this PR just now. Apparently an UAVO version mismatch disables these? (Not sure why, because HwRevolution didn't change since quite a while).

mlyle:
hwrevolution, and shareddefs/hwshared underneath it, both changed for openlrs stuff.

PR #2198, opened 2018-05-15
python-api: default request callback needs to accept (unused) second parameter

nunohumberto:
Callbacks are constantly invoked with two parameters.
As such, default_cb needs a second parameter (although unused) to enable correct functionality when issuing object requests.

As it is right now, the telemetry thread will throw an exception when an object request is completed:
https://imgur.com/a/Xlm0kEY

### DIFF
--- a/python/dronin/telemetry.py
+++ b/python/dronin/telemetry.py
@@ -300,7 +300,7 @@ class TelemetryBase(metaclass=ABCMeta):
 
         response = []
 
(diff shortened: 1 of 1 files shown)


mlyle:
jenkins, test this please

mlyle:
@nunohumberto Thanks for spotting this!  Looks like we don't use the default callback anywhere and it was broken.

PR #2199, opened 2018-05-15
tests: request an object using sync interface

mlyle:
Catches the issue that #2198 fixes.

Hat tip to @nunohumberto for reporting / fixing the underlying issue; this just adds test coverage.

### DIFF
--- a/python/integrationtests-basic
+++ b/python/integrationtests-basic
@@ -90,6 +90,17 @@ class AAConfTests(SimulationTestCase):
         t_stream.save_objects(objs.values(), send_first=True)
 
 class SimpleSimTests(SimulationTestCase):
+    def test_retrieve_object(self):
+        t_stream = self.t_stream
+
+        sepp_obj = t_stream.uavo_defs.find_by_name("HwSeppuku")
+
+        hw_sepp = t_stream.request_object(sepp_obj)
+
+        self.assertIsNotNone(hw_sepp)
+        self.assertEqual(hw_sepp.USB_HIDPort,
+                sepp_obj.ENUM_USB_HIDPort['USBTelemetry'])
+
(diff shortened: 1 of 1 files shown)


python/integrationtests-basic
```diff
@@ -90,6 +90,17 @@ class AAConfTests(SimulationTestCase):
         t_stream.save_objects(objs.values(), send_first=True)
 
 class SimpleSimTests(SimulationTestCase):
+    def test_retrieve_object(self):
+        t_stream = self.t_stream
+
+        sepp_obj = t_stream.uavo_defs.find_by_name("HwSeppuku")
```
tracernz, line 96: We don't really have the API contract well defined but by the code, the onus is on the caller to check for success here.. find_by_name doesn't throw an exception when the object doesn't exist. It would sure be nice (and more pythonic) if it did, so we don't need error checking all through tests etc.

python/integrationtests-basic
```diff
@@ -90,6 +90,17 @@ class AAConfTests(SimulationTestCase):
...
+
+        hw_sepp = t_stream.request_object(sepp_obj)
```
tracernz, line 98: This would fail with a "NoneType has no member" deep down the callstack if sepp_obj is None.

python/integrationtests-basic
mlyle: Yah, I think I agree-- but probably out of scope for adding test coverage for this case, right?

(In general there is way too much magical "None" in the python API in retrospect)

mlyle: (next step is for us to actually define our own exceptions in the package, which is easier to do with pure py3)

tracernz: Yup, agree out-of-scope. :+1:

PR #2202, opened 2018-05-16
lqg: RTKF needs way way more cycles to stabilize with low beta.

glowtape:
I was graphing convergence of things, when I tried typical beta and tau from weak yaw axes, like those micro brushed and the large ones that seem to have yaw drift. Turns out that low beta requires a lot more cycles for the RTKF to stabilize. It was capped to 150 cycles, which works fine for quads with a beta of 10, but yaw drifty problem quads usually have a beta below 8, which seems to require like 2000-2500 cycles to get stable.

A typical value that causes problems is beta 7.5 tau 0.028s. When you cap the RTKF at 150 cycles, the Kalman gains for torque and bias are just 15% and 11% of what they'd be when stabilized. Bias and external torque is way way slower with that, which is what I'm suspecting is giving it a hard time with the yaw axis, while stabilizing higher beta pitch and roll just fine.

### DIFF
--- a/flight/Libraries/math/lqg.c
+++ b/flight/Libraries/math/lqg.c
@@ -46,8 +46,16 @@
 
 */
 
-#define RTKF_SOLUTION_LIMIT			150
-#define LQR_SOLUTION_LIMIT			1000
+#define SOLVER_MIN					100
+#define SOLVER_MAX					5000
+
+/*
+	Applies to torque and bias parts.
+
+	Beta 7.5 and Tau 50ms solves RTKF in 3077 iterations and the LQR is 1267 iterations.
+	Beta 10.5 and Tau 15ms solves RTKF in 298 iterations and the LQR in 156 iterations.
+*/
+#define SOLVER_DELTA_EPSILON		0.000000001f
 
 /* Bullshit to quickly make copypasta of MATLAB answers work. */
 #define P00 P[0][0]
@@ -99,8 +107,9 @@
 	K = PH' (R+HPH')^-1
 	P = (I-KH) P
 */
-void rtkf_calculate_covariance_3x3(float A[3][3], float K[3], float P[3][3], float Q[3][3], float R)
+bool rtkf_calculate_covariance_3x3(float A[3][3], float K[3], float P[3][3], float Q[3][3], float R)
 {
+	bool solved = false;
 	float nP[3][3];
 
 	nP[0][0] = Q00 + A00*(A00*P00 + A01*P10 + A02*P20) + A01*(A00*P01 + A01*P11 + A02*P21) + A02*(A00*P02 + A01*P12 + A02*P22);
@@ -119,9 +128,17 @@ void rtkf_calculate_covariance_3x3(float A[3][3], float K[3], float P[3][3], flo
 
 	float S = P00 + R;
 
+	float nK[2];
+
 	K0 = P00/S;
-	K1 = P10/S;
-	K2 = P20/S;
+	nK[0] = P10/S;
+	nK[1] = P20/S;
+
+	if (fabsf(K1 - nK[0]) <= SOLVER_DELTA_EPSILON && fabsf(K2 - nK[1]) <= SOLVER_DELTA_EPSILON)
+		solved = true;
+
+	K1 = nK[0];
+	K2 = nK[1];
 
 	P10 = P10 - K1*P00;
 	P11 = P11 - K1*P01;
@@ -134,6 +151,8 @@ void rtkf_calculate_covariance_3x3(float A[3][3], float K[3], float P[3][3], flo
 	P00 = P00 - K0*P00;
 	P01 = P01 - K0*P01;
 	P02 = P02 - K0*P02;
+
+	return solved;
  }
 
 struct rtkf_state {
@@ -160,10 +179,17 @@ struct rtkf_state {
  void rtkf_stabilize_covariance(rtkf_t rtkf, int iterations)
  {
  	PIOS_Assert(rtkf);
- 	for (int i = 0; i < iterations; i++) {
- 		rtkf_calculate_covariance_3x3(rtkf->A, rtkf->K, rtkf->P, rtkf->Q, rtkf->R);
+
+	if (rtkf->solver_iterations < 0)
+		return;
+
+	for (int i = 0; i < iterations; i++) {
+		rtkf->solver_iterations++;
+		if (rtkf->solver_iterations > SOLVER_MAX || (rtkf_calculate_covariance_3x3(rtkf->A, rtkf->K, rtkf->P, rtkf->Q, rtkf->R) && rtkf->solver_iterations > SOLVER_MIN)) {
+			rtkf->solver_iterations = -1;
+			break;
+		}
  	}
- 	rtkf->solver_iterations += iterations;
  }
 
 /*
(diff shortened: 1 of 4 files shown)


glowtape:

glowtape:
Convergence of the RTKF (torque and bias gains) on a high powered miniquad (beta 10.63 tau 14.4ms):

rtkffast

Same thing with a terrible yaw axis (beta 7.5 tau 28ms, those were values of LitterBugs ~~minibrushed~~ hex IIRC):

rtkfslow

mlyle:
1. Is it now possible to arm before this is done?
2. Would it be good to spot actual convergence and terminate?  (And have an alarm until it is done?)

glowtape:
Yeah, if you arm before it is done, it'll fallback onto PID, then switch as soon it is solved.

That said, 3000 cycles on MPU/1KHz is 3 seconds, on the BMI/1.6KHz it is 2. It does one iteration per loop in stab.c. Add a second for boot stuff, I'd expect it to be done easily before people arm.

glowtape:
Self-termination could be an idea. I have to run a bunch of tests for it on how to decide to cut off. If it's 2-4% off stabilized values wouldn't be so much of an issue than when it is like >85% off.

mlyle:
Are you sure the stabilization stuff is running at full speed when you're doing all this extra work?  I mean, i guess I can hold my nose and tolerate this...

Makes me nervous that failure to converge isn't spotted except by awful flight perf.

glowtape:
Yeah, I tested it on SPRF3e and it doesn't set an error on stab about dT_measured mismatching dT_expected, so it doesn't appear to overwhelm things. Either way, once things are considered solved, it doesn't call into the math anymore.

I made it adaptive now. It keeps solving both the RTKF and LQR until the difference on the relevant terms between iterations falls under a threshold.

Using test values beta 7.5 and tau 50ms, it goes on for 3077 iterations on the RTKF and 1267 iterations on the LQR. Using test values beta 10.5 and tau 15ms, it does 298 iterations on the RTKF and 156 iterations on the LQR. All using the current default process noise and LQR costs.

(There's a bunch of updated process noise values that I'd like to test on the forums, after this release has been out for a bit. These converge a bit quicker for the low beta stuff, too.)

I matched the output with MATLAB, and the difference is merely noise 7-8 digits after the comma.

mlyle:
Thanks-- I like this much better.

I notice old LQR max was 1000 and that this was still being exceeded on low beta-- do end-users maybe need to retest?  (I know yds did not get good results).

glowtape:
LQR wasn't so much the issue. Even with beta 7.5 and 50ms tau, it gets to 99.9% of convergence in 1000 iterations. With that commit, it gives a little bit more effort now. The Kalman filter was more the thing. In the aforementioned case, the gains for the torque and bias were over a magnitude too low with just 150 iterations (seen a 16x difference).

From what yds posted on IRC, there's a noticeable improvement with the static 3000 iterations artifact. There was mention of a yaw attitude shift with rolling left and right, something you've also noticed, that remains to be solved (altho the relevance of axislock needs to be tested, since neither LQG nor LQGLeveling do anything of the sort on yaw, as opposed to Acro and Leveling).

Another thing I need to evaluate is lowering the sensor noise value for yaw. It is currently higher than for pitch and roll, because of direct torque effects having an influence on the gyro. At least that was the thinking behind it.

flight/Libraries/math/lqg.c
```diff
@@ -46,8 +46,16 @@
...
+#define SOLVER_MIN					100
+#define SOLVER_MAX					5000
+
+/*
+	Applies to torque and bias parts.
+
+	Beta 7.5 and Tau 50ms solves RTKF in 3077 iterations and the LQR is 1267 iterations.
+	Beta 10.5 and Tau 15ms solves RTKF in 298 iterations and the LQR in 156 iterations.
+*/
+#define SOLVER_DELTA_EPSILON		0.000000001f
```
mlyle: Are you sure this is safe / can always be reached?  Only on really small floating point values can you get a (nonzero) difference this small.

flight/Libraries/math/lqg.c
```diff
@@ -160,10 +179,17 @@ struct rtkf_state {
...
- 	for (int i = 0; i < iterations; i++) {
- 		rtkf_calculate_covariance_3x3(rtkf->A, rtkf->K, rtkf->P, rtkf->Q, rtkf->R);
+
+	if (rtkf->solver_iterations < 0)
+		return;
+
+	for (int i = 0; i < iterations; i++) {
+		rtkf->solver_iterations++;
+		if (rtkf->solver_iterations > SOLVER_MAX || (rtkf_calculate_covariance_3x3(rtkf->A, rtkf->K, rtkf->P, rtkf->Q, rtkf->R) && rtkf->solver_iterations > SOLVER_MIN)) {
+			rtkf->solver_iterations = -1;
```
mlyle: It sure would be nice if "reached max" was not considered "solved" here.

flight/Libraries/math/lqg.c
```diff
@@ -46,8 +46,16 @@
 
 */
 
-#define RTKF_SOLUTION_LIMIT			150
-#define LQR_SOLUTION_LIMIT			1000
+#define SOLVER_MIN					100
+#define SOLVER_MAX					5000
```
tracernz: For future ref: always better to format anything except the line indentation with spaces.

flight/Libraries/math/lqg.c
glowtape: Seems to work out for all the random values I tossed at it (from beta 6 tau 150ms to beta 12 tau 5ms), assuming non-fastmath IEEE-754 works the same between ARM and x86.

flight/Libraries/math/lqg.c
glowtape: How to handle this? Warning but still let them arm?

mlyle: That would be one way-- or refuse to allow arm if they have any LQG flight modes and things are unsettled (sanity check), or whatever.  Just about anything is better than silently using known-bad coeffs.

flight/Libraries/math/lqg.c
mlyle: Basically float can be assumed to have 6 sigfigs; this number is 10^-9, so it means that this epsilon "works" on values that are less than 10^-3; otherwise it can be expected to do weird things.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -953,9 +954,23 @@ static void stabilizationTask(void* parameters)
...
+					/* Values don't converge in time, probably bogus. Light up the Christmas three. */
+					SystemAlarmsConfigErrorOptions err;
+					SystemAlarmsConfigErrorGet(&err);
+					if (err != SYSTEMALARMS_CONFIGERROR_LQG) {
+						err = SYSTEMALARMS_CONFIGERROR_LQG;
+						SystemAlarmsConfigErrorSet(&err);
+						AlarmsSet(SYSTEMALARMS_ALARM_SYSTEMCONFIGURATION, SYSTEMALARMS_ALARM_ERROR);
+					}
+				} else if (status == LQG_SOLVER_RUNNING) {
+					lqg_run_covariance(lqg[i], 1);
```
mlyle: Shouldn't the alarm be set here, too?

glowtape: OK, changed it. Works, too. With beta 6 and tau 250, the Seppuku needs a moment to converge, longer than for the GCS to connect, CONFIG is in error state and then clears up.

tracernz:
Nitpick: it's not really a configuration error... but the logic for handling that case is sound and proven so it's not a bad idea to use it anyway this late in the cycle.

mlyle (approved):
lgtm, does it work?

tracernz (approved):
Can't see any real issues, if it tests ok LGTM.

glowtape:
Yeah the config error is not entirely correct, but it's the cheapest way to prevent arming and showing an LQG warning on OSD.

glowtape:
Applying an artificially low SOLVER_MAX, to pretend being unable to converge to anything, it entirely prevents arming as expected, while lighting the LQG config error in OSD as well as GCS.

Applying artificially low beta to an axis with regular SOLVER_MAX, it takes a while to converge, lighting the LQG config error both in OSD and GCS, until it finished converging, when it clears it and allows arming.

The values it spits out are sane, compared them to MATLAB. Flew this PR earlier today, also no oddities.

As far as I'm concerned, this works.

PR #2203, opened 2018-05-16
revolution: support i2c BMP280 baros on clones

mlyle:
Needs test:

* [x] Base revo (I2C MS5611)
* [x] Acro revo
* [x] SPRF3E (I2C BMP280)
* [x] Omnibus F4 Pro early revs (SPI BMP280 with stuck I2C bus)
* [x] Omnibus F4 Pro current revs (SPI BMP280)
* [x] Seppuku (SPI BMP280)
* [x] Omnibus F4 Variant with I2C BMP280

### DIFF
--- a/flight/PiOS/Common/pios_bmp280.c
+++ b/flight/PiOS/Common/pios_bmp280.c
@@ -349,7 +349,8 @@ static int32_t PIOS_BMP280_ReadADC()
 }
 
 #ifndef PIOS_EXCLUDE_BMP280_I2C
-static int32_t PIOS_BMP280_I2C_Read(uint8_t address, uint8_t *buffer, uint8_t len)
+static int32_t PIOS_BMP280_I2C_Read(pios_i2c_t i2c_id,
+		uint8_t address, uint8_t *buffer, uint8_t len)
 {
 	const struct pios_i2c_txn txn_list[] = {
 		{
@@ -369,10 +370,11 @@ static int32_t PIOS_BMP280_I2C_Read(uint8_t address, uint8_t *buffer, uint8_t le
 		 }
 	};
 
-	return PIOS_I2C_Transfer(dev->i2c_id, txn_list, NELEMENTS(txn_list));
+	return PIOS_I2C_Transfer(i2c_id, txn_list, NELEMENTS(txn_list));
 }
 
-static int32_t PIOS_BMP280_I2C_Write(uint8_t *buffer, uint8_t len) {
+static int32_t PIOS_BMP280_I2C_Write(pios_i2c_t i2c_id,
+		uint8_t *buffer, uint8_t len) {
 	const struct pios_i2c_txn txn_list[] = {
 		{
 		 .info = __func__,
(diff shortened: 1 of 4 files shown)


mlyle:

flight/PiOS/Common/pios_bmp280.c
```diff
@@ -154,15 +154,6 @@ static int32_t PIOS_BMP280_Common_Init(const struct pios_bmp280_cfg *cfg)
 	dev->cfg = cfg;
 
 	uint8_t data[24];
-	uint16_t ref_digT1;
```
mlyle, line 157: This probe logic is factored into the I2C and SPI versions of the init call, before allocating a device structure.

flight/PiOS/Common/pios_bmp280.c
```diff
@@ -199,37 +185,6 @@ static int32_t PIOS_BMP280_Common_Init(const struct pios_bmp280_cfg *cfg)
 	return 0;
 }
 
-/**
```
mlyle, line 202: And the init calls are relocated to later in the file, to avoid needing prototypes and as many copies of the ifdefs.

flight/PiOS/Common/pios_bmp280.c
```diff
@@ -348,8 +303,31 @@ static int32_t PIOS_BMP280_ReadADC()
 	return 0;
 }
 
+static inline int32_t PIOS_BMP280_CheckData(const uint8_t *data,
```
mlyle, line 306: This call exists to confirm that we get good reads-- both that the data varies over the read (not stuck bus) and that reading the first few cal registers multiple times gets consistent results.

flight/PiOS/Common/pios_bmp280.c
```diff
@@ -348,8 +303,31 @@ static int32_t PIOS_BMP280_ReadADC()
...
+			return 0;
+		}
+	}
+
+	return -4;
+}
+
 #ifndef PIOS_EXCLUDE_BMP280_I2C
-static int32_t PIOS_BMP280_I2C_Read(uint8_t address, uint8_t *buffer, uint8_t len)
+static int32_t PIOS_BMP280_I2C_Read(pios_i2c_t i2c_id,
```
mlyle, line 329: These no longer use the globals, but instead take the device ids directly-- so that the device structure doesn't need to be allocated for early probe.

flight/PiOS/STM32/pios_i2c.c
```diff
@@ -277,6 +277,13 @@ int32_t PIOS_I2C_Init(pios_i2c_t *i2c_id, const struct pios_i2c_adapter_cfg *cfg
 	/* Initialize the state machine */
 	i2c_adapter_fsm_init(i2c_adapter);
 
+	if (PIOS_I2C_CheckClear(i2c_adapter)) {
```
mlyle, line 280: Every single target does this checkclear on the i2c (which can be removed in a further commit).  We do it here, before we've started the interrupt handlers, out of an abundance of caution and to simplify future factoring.

flight/targets/revolution/fw/pios_board.c
```diff
@@ -235,14 +235,11 @@ void PIOS_Board_Init(void) {
...
 #if defined(PIOS_INCLUDE_I2C)
-	if ((!is_modified_clone) && PIOS_I2C_Init(&pios_i2c_mag_pressure_adapter_id, &pios_i2c_mag_pressure_adapter_cfg)) {
-		PIOS_HAL_CriticalError(PIOS_LED_HEARTBEAT, PIOS_HAL_PANIC_I2C_INT);
-	}
-
-	if ((!is_modified_clone) &&
-			(PIOS_I2C_CheckClear(pios_i2c_mag_pressure_adapter_id) != 0)) {
-		PIOS_HAL_CriticalError(PIOS_LED_HEARTBEAT, PIOS_HAL_PANIC_I2C_INT);
-	} else if (!is_modified_clone) {
+	if (PIOS_I2C_Init(&pios_i2c_mag_pressure_adapter_id, &pios_i2c_mag_pressure_adapter_cfg)) {
```
mlyle, line 238: Now that PIOS_I2C_Init has less side effects when the bus is screwed up, things are much simpler--- just scream if the i2c init didn't work and it was expected to (on original revolution).  Other targets can use this cleanup in the future, too.

flight/targets/revolution/fw/pios_board.c
```diff
@@ -590,6 +587,16 @@ void PIOS_Board_Init(void) {
 	//I2C is slow, sensor init as well, reset watchdog to prevent reset here
 	PIOS_WDG_Clear();
 
+#if defined(PIOS_INCLUDE_BMP280) && !defined(PIOS_EXCLUDE_BMP280_I2C)
+	/* BMP280 I2C onboard support on clones */
+	if (is_modified_clone &&
```
mlyle, line 592: If we're a clone, have an unstuck I2C bus, and no baro so far-- probe on I2C for a BMP280.

LitterBugs:
Bench tested on Matek F405-Wing FC.  Previously this FC did not detect baro.  Baro was gray on Flight Data tab.

Baro is now green with this board when running the above artifacts.  Baro responds to pressure changes as well.

mlyle:
Bumping to ludicrous -- it is unclear the status of whether this is needed on anything otherwise supported by the omnibus target.  Almost all of this is good to have still even if not, but not worth picking up the risk for.

flight/PiOS/Common/pios_bmp280.c
tracernz: Good idea. :)

tracernz (approved):

tracernz:
Tested revo, seems happy.

PR #2204, opened 2018-05-16
python-api: typo on print() call on CRC mismatches

nunohumberto:
Typo on print() call throws NameError whenever CRC mismatches occur.


mlyle:
Thanks

Jenkins, test this please

mlyle:
Thanks for the catch :D

Issue #2188, opened 2018-04-28
Python API: Move to asyncio, async, await, etc.


mlyle:
Since we're on python 3, use the new syntactic sugar, sugar.

tracernz:
And use exceptions more for error handling? :stuck_out_tongue:

mlyle:
Hmmmm.  This is looking like a pain.

Using `yield from` looks like it requires way too much of a complexity/performance tax for the factoring improvements it can provide in uavtalk.

It looks like moving to an event loop with asyncio would provide some small factoring improvements.

PR #2206, opened 2018-05-18
uavtalk: move to generator class

mlyle:
This is cleanup-in-progress.  Small performance tax on some workloads (~5%).  It's working towards #2188.

Has received a degree of test:

* [x] Reading a log file
* [x] pipe-to-simulator
* [x] TCP to simulator
* [x] HID to flight board
* [x] Serial to flight board

### DIFF
--- a/python/dronin/telemetry.py
+++ b/python/dronin/telemetry.py
@@ -89,9 +89,7 @@ class TelemetryBase(metaclass=ABCMeta):
         self.filename = name
 
         self.acks = set()
-        self.req_obj = {}
-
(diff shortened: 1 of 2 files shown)


tracernz:
Looking good so far. :+1:

tracernz:
HID is good. Having problems with serial, but that may be the flight side as the board reboots...

tracernz:
It was GCS immediately booting the board when it halted! :man_facepalming: Serial okay too.

tracernz (approved):

Issue #2101, opened 2018-03-08
Need a warning or something for "Collective" on multirotor


mlyle:
At least two people have tripped over this so far.

tracernz:
Collective selected somewhere, or?

mlyle:
As an input channel-- it prevents throttle control.

tracernz:
That's rather a peculiar thing to select unless you actually have collective pitch (e.g. Curtis Youngblood quads).

mlyle:
Lots of people set e.g. all the channels to sbus and then map them, and we behave in a pretty unintuitive way then.

yds:
how about disabling **Collective** in **RC Inputs** if **Vehicle Type:** _Multirotor_ in **Mixer Settings**

mlyle:
No, because there are multirotors with collective.

tracernz:
Is this still valid with #2117? Seems you can't really accidentally configure it now.

mlyle:
#2117 helps mitigate it.  some people put the whole left column "sbus" and left it none, some people configured channels for everything because "why not".  But I think we should still consider doing more.

glowtape:
Move the collective channel to the bottom of the list, followed by a warning about helis and such?

PR #2207, opened 2018-05-18
sanitycheck: Add Throttle configuration alarm, when collective is configured for multirotor.

glowtape:
Generates a configuration warning, if collective is configured on a multirotor. Over in #2101 it is mentioned that collective on a multirotor is a rare but possible configuration, so the CFG:THROTTLE OSD error might be a little unfortunate for OSD targets, but how common is that?


mlyle (approved):

tracernz:
It's rather unfortunate to have an unmaskable warning on valid configurations. :/ 

-e- Wide variety of vehicle configurations is one of our strengths compared to other firmwares and I'd hate to see that go away.

glowtape:
We could add a checkbox to the Input page, that only shows on a multirotor+collective configuration, labelled ~~"I better not be an idiot by declaring a collective control on a multirotor"~~ "Collective on multirotor intentional", and automatically clears the collective when saving with this unchecked.

glowtape:
Or just a mere GCS side warning in the UI in that specific combo.

mlyle:
Yah I was thinking a GCS side warning, turning collective yellow, but it's tricky how the input form is made.

mlyle:
Thx, but superseded #2208

Issue #2200, opened 2018-05-15
F3/F4 targets have higher CPU usage, likely related to telemetry changes.


glowtape:
Several users have been mentioning higher CPU usage on the F3 targets lately. Most recent reporter Maquiabelo, where it seems to be causing a critical CPU alarm, and also claims to be unable to arm (presumably due to critical CPU). Two quads using a micro F3, running the Omnibus F3 target. 

The UAVOs are attached to this issue, but I haven't looked into it yet. [spc95.zip](https://github.com/d-ronin/dRonin/files/2006165/spc95.zip)

I'm seeing something similar. When I install an older build on my SPRF3e, it idles at 67%, and with next, it's at 80%. With erased settings both times (LQG, FlipOver and Motorgain don't seem to be responsible for the brunt of it).

I'm currently busy identifying a bunch of CPU time sinks, which may or may not happen only with the GCS attached. It seems like two commits are responsible for the bump on my SPRF3e, each worth around 5%:

https://github.com/d-ronin/dRonin/commit/b0b70c5b79feb6fa09a271bcbe21006cb4d64425 telemetry: throttled obj can only be in queue once
https://github.com/d-ronin/dRonin/commit/5388fe3e5a86f43d75de4d264dc63c2df8e2308c uavobjectmanager: improve timer math

The issue with latter one was identified over on IRC, altho it brings the additional CPU from 4-5% down to 1.5-2%, not nearly zero, so there's still a minimal effect on that one.

Former seems unclear.

Maquiabelo has an OSD on his Omnibus. I figure  its presence seems to send things over the edge together.

glowtape:
Eh, I spot this only now, https://github.com/d-ronin/dRonin/commit/5388fe3e5a86f43d75de4d264dc63c2df8e2308c is a small patch applying to https://github.com/d-ronin/dRonin/commit/b0b70c5b79feb6fa09a271bcbe21006cb4d64425 .

glowtape:
Yeah I don't get it. Right now on latest next, it idles at 77% with no config at all. When I revert the recent change 263720787679d638bc3257be998a5ae16f877497, which gives back 3%, and the other two changes above, I drop to 70%. I've been looking and prodding at it, and I don't get it.

glowtape:
OK, what I've noticed is that per TaskInfo (reverted vs next):

Actuator 12% -> 13%
Attitude 26% -> 29%
Stabilization 28% -> 30%
ManualControl "0%" -> 1%

[...]

PR #2210, opened 2018-05-20
Attempt to improve CPU use by declaring things const

mlyle:
Also just assume systime is milliseconds because the math before was faulty.

May improve #2200


glowtape:
L1690 in uavobjectmanager.c needs const.

Gives slight improvements, still not as far as reverting those commits mentioned in #2200.

mlyle:
OK, one more thing to try in here.

mlyle:
I'd advocate for this to merge-- with the PIOS_Thread time change we're back to where we were originally in CPU use.

PR #2213, opened 2018-05-29
Package:GCS: Remove libqjp2.so - JPEG2000 support

### LINKED ISSUE #2212: Ubuntu install broken on 18.04 - requires unavailable dependency (libjasper1)
* Seen in version (release version or githash): Release-20180212.2

Issue details:
The Ubuntu packaged deb requires libjasper1, which is not available in the Ubuntu 18.04 repos. Downloading it manually from 17.10 repos & installing works (for now). Unclear why libjasper1 isn't available in 18.04, but build deps should likely be changed.

tracernz:
It has a dependency on libjasper which is unmet on recent Debian based systems, due to security issues.

Fixes #2212


tracernz:
build artifacts please

tracernz:
aggh, didn't quote the echo and it had parens

build artifacts please

mlyle:
Thanks :D  It's a shame jpeg-2000 has always been rather impossible to use, because of poor software support, IPR problems, etc--- and it seems it's not ever going to get better.

PR #2217, opened 2018-06-05
transmitter_control: Don't forget about Altitude Hold.

### LINKED ISSUE #2216: AltitudeHold flight mode inoperative
* Seen in version (release version or githash):  Pre-Inconcievable and #2215 
* On flight controller (if applicable): Seppuku

Issue details:
Selecting Altitude hold mode causes motors to drop to Neutral, Throttle is ineffective, And beeper sounds .-. "R" code.

Also noticed that there is still an ALT.H indicator on the Flight Data screen in GCS although there is no longer an AltitudeHold Module.  

If FC is connected to GCS when selecting Altitude Hold mode, Flight mode shows as Failsafe, the RC Input indicator turns red.  Hovering over it gives the message: "Error Code: Undefined      This should not happen"

glowtape:
Fixes #2216.


mlyle (approved):
Thx much

PR #2218, opened 2018-06-29
Flight: Correct mag logic such that mag works without baro.

jihlein:
While working with flightd on a board with only a mag (no baro), a sensor error was displayed.  Investigation showed that the timeout variables for a mag sensor read were the same as for the baro sensor read, and not being correctly updated when no baro was present.  This PR creates dedicated mag variables and updates them correctly.

Bench verified with flightd, but should be fine on other targets.


jihlein:
Also tested on AQ32, works as intended.

mlyle (approved):
LGTM, thanks for the catch!

PR #2230, opened 2018-08-13
attitude: Implement sensor decimation.

glowtape:
Moves the collection of the gyro samples from the updateAttitudeINSGPS/updateAttitudeComplementary functions to the task function, and collects samples averaging them based on a decimator value calculated. Current default runtime for the attitude filter in this patch is 100hz, which drops the CPU usage in TaskInfo from 15% to about 2-3%. It still flies good (at least in the living room, rain and all).

PR for Maquiabelo, for artifacts.


glowtape:

glowtape:
Seems to work fine according to the guy on the forum triggering this PR.

PR #2232, opened 2018-11-01
Update translations.pro

ortas88:
prepares Polish translation

### DIFF
--- a/ground/gcs/share/translations/translations.pro
+++ b/ground/gcs/share/translations/translations.pro
@@ -2,7 +2,7 @@ TEMPLATE = aux
 
 include(../../gcs.pri)
 
-LANGUAGES = de es fr ru zh_CN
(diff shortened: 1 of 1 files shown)


tracernz:
If you run `make gcs_ts`, this generates the correct Polish translation files?

ortas88:
if you send me a gcs_ts file, I will do a translation into Polish


pt., 2 lis 2018, 03:52: Michael Corcoran <notifications@github.com>
napisał(a):

tracernz:
Things were a little broken I think (my fault!). Should be all fixed in #2233. The file you need to begin your translation is https://github.com/d-ronin/dRonin/blob/e431059f5ca4f974f39fbc0991f972b3c68b62c8/ground/gcs/share/translations/dronin_untranslated.ts I think. Rename to dronin_pl.ts and you should be ready.

PR #2233, opened 2018-11-04
GCS: Fix and update translations

tracernz:
I think I broke this a while back when merging upstream Qt Creator sources.
Remove mimetypes extract as we don't use this (and a file was missing to make it work).


tracernz:
build artifacts please

PR #2250, opened 2019-04-02
GCS:Config: Allow zero battery cells to disable auto-detect

fxbisto:
Allow a value of 0 cells so automatic cell detection isn't forced to report a battery fault when powered by USB alone.
Setting 0  is allowed in UAVO browser but applying changes in modules tab always updated FlightbatterySettings > NbCells to 1

### DIFF
--- a/ground/gcs/src/plugins/config/modules.ui
+++ b/ground/gcs/src/plugins/config/modules.ui
@@ -635,7 +635,7 @@
                     <bool>false</bool>
                    </property>
                    <property name="minimum">
-                    <number>1</number>
+                    <number>0</number>
                    </property>
                    <property name="objrelation" stdset="0">
                     <stringlist>


tracernz:
Setting BatterySettings.MaxCellVoltage to 0 should also disable auto-detect.

fxbisto:
Autodetect is the desired behaviour - Some flight controllers read a phantom voltage when powered by USB. Mine reports 0.01221v , which is clearly a negligible hardware issue.
Setting NbCells to 1 clamps autodetect into reporting a fault for any voltage below cell threshold - even when DetectedCellCount=0

battery_shows_faults

Changing to 0 releases this behaviour:

battery_no_fault

Currently there is disparity between using UAVO Browser > FlightbatterySettings > NbCells (min=0) and Configuration / Modules / Battery / Number of cells (where min=1)

glowtape:
If it's about the ability to arm with phantom voltages, there's a setting available in the UAVO browser. When you set FlightBatterySettings.CellVoltageThresholds.Alarm to 0V, you can arm at any voltage. The OSD warning will still come at 3.5V per (autodetected) cell.

fxbisto:
No, not at all. Perhaps I'm being a little picky here
It just seems to make sense that as we can select zero cells in UAVObject browser we should be able to do the same in the Module configuration page.
It bothers me that there's the little red square (which also triggers my warning LED) when there's not an error, just no battery connected. The battery module correctly identifies zero cells but forcing it to a minimum of assuming one cell generates that error unnecessarily in my mind.
If it's not a desirable change for anyone else then that's cool :)

tracernz:
test this please

PR #2243, opened 2019-03-17
disable '-Werror' on linux too. Fix issue #2205

rgorosito:
This solution was proposed by @tracernz on https://forum.dronin.org/forum/d/596-compiling-jihlein-s-gcs-fork-on-ubuntu-amd-2400g/ .
Tested on Fedora with GCC 8.3.1

PR #2251, opened 2019-04-02
Matek405 Target

jihlein:
Here are the target files for the the Matek405 target.  I've been flying this on my local branch for quite some time, on both the Matek405Ctr and Matek405Std boards,  I believe it should also work on the Matek405Osd and Matek405Mini boards, but it has not been tested on them.

I've not tested all the possible Uart port combinations, but I know for sure that SBus and FrSky Sport telemetry work.  As far as outputs, I've used SyncPWM and DShot600 with no problems, as does GPS and OpenLager.

The LED string shares a pin with the S6 output, there's a hardware UAVObject to select the function.  I did this so as to preserve the use if the I2C output pads, which in BetFlight, are shared between I2C and LED use.  The output waveform looks correct on a logic analyzer, but I've not tried it with a real LED string.

This target was largely derived from the AQ32 target files.  I also have a target for the KakuteF4V2 that I can submit if there's interest.  I've found both these boards to be of nice quality and good flyers.


fxbisto:
I have been testing the CTR which works as expected.
You noticed the weirdness when trying to use betaflight's DMA mapping as well I take it? I had updated your board_hw_defs.c to match betaflight's settings but it then won't work with DSHOT on any output other than PWM6. Multishot still worked fine though (I thought that also used DMA..?)
I've tried removing all the DMA definitions as pios_dmashot.c seems to assign any missing ones, that also works fine if not better!

My builds are only using 2 serial connections (rx3 for ibus + tx4 for vtxconfig) so I can't help any further regarding saturating UARTS I'm afraid.

I also have a quad with the F405-MINI so I could test that for you?

jihlein:
The DMA mappings work as is.  The Betaflight DShot implementation differs from dRonin if I recall correctly, and the DMA assignments from firmware to firmware don't match.

 I've been using DShot600 successfully on the STD.

Multishot does not use DMA.

Yes!  Please try the mini, would like to see if it works.

ufoDziner:

ufoDziner:
@jihlein I also have a mini that I'm going to try with a fixed wing this week.  Mine will be using CRSF protocol.

ground/gcs/src/plugins/boards_matek/matek405.cpp
```diff
@@ -0,0 +1,257 @@
...
+}
+
+//! Determine if this board supports configuring the receiver
+bool MATEK405::isInputConfigurationSupported(Core::IBoardType::InputType type)
+{
+    switch (type) {
+    case INPUT_TYPE_PWM:
+    case INPUT_TYPE_UNKNOWN:
+        return false;
+    default:
```
tracernz, line 104: I always preferred to have no default case here, so whenever a new protocol is added, it's impossible to forget about a board (the build will fail... although maybe not anymore since #2243). What do you think?

tracernz (commented):
First quick pass looks good. This seems to be a very popular board. Any ideas on its manufacturing longevity?

glowtape:
Dropping the gyro LPF configuration might be considered. We did that on the other targets, too, and just run at the highest bandwidth (188). The ICM settings will override the MPU ones in this patch currently.

@fxbisto The DMA settings for DShot don't need to match Betaflight. You really just want unused streams that work with TIMx_UP. It's possible that there's conflicts, because say we also use different DMA streams for ADC.

fxbisto:
@glowtape I figured that was the case as soon as I saw it conflicting. Good to know for the future if I plan on adding support for other targets I might own, although currently revo, omni, m405 are covering the majority of my FCs.

jihlein:
"The ICM settings will override the MPU ones in this patch currently."

Good catch.  I've got no issues with dropping the UAVO's to configure this and just setting the highest bandwidth by default.  (188 Hz for MPU6000 accel/gyro, 176 Hz for ICM gyro, 218 Hz for ICM accel).

I'll make this change if there are no objections.  The less knobs to twiddle the less chance for strange configuration issues........

This also brings up the question should we fix the gyro range at 1000 DPS and the accel range at 8G too?

ground/gcs/src/plugins/boards_matek/matek405.cpp
jihlein: Not a bad idea, but most, if not all, of the targets have the default case here, FWIW.

jihlein:
"Any ideas on its manufacturing longevity?"

Unfortunately no.  Matek has a selection of F4 and F7 boards, I'm not sure how popular the F7's are vs the F4's.

glowtape:
Is there even functional difference between the F4 and F7 ones? I have that F7mini of theirs here. The existence of that thing makes me question why there's still F4s, when they can make it fit on 20x20(ish?).

Support for F7 will take a while, until I can figure out that obscure memory corruption issue caused/unearthed by upgrading to a newer ChibiOS.

fxbisto:
> The existence of that thing makes me question why there's still F4s, when they can make it fit on 20x20

I believe that's down to the fact that betaflight have been unable to match their F4 performance figures with F7 boards until very recently, IIRC it was blamed on  a difference in the F7's HAL.
I'm hedging bets that F4's will be current for a good while yet (and hoping)

ufoDziner:
I seem to be getting an error.

flightData
hardwareSettings

fxbisto:
John, does your latest commit remove the 2000 deg/s option from Matek? If that's the case it would limit board support where freestyle is the intended flight as we are then limited to 850 deg/s max input

jihlein:
It does, but we can easily default it to 2000 DPS.

jihlein:
@ufoDziner - The Matek405Mini doesn't appear to have the baro chip, that's why it's getting hung up.  Let me see what I can do.

I thought PIOS_TOLERATE_MISSING_SENSORS should cover this, but apparently I don't understand how this define works.

I can make the use of the baro chip optional, defaulting to disabled, so the 405Mini would properly boot.  If you wanted to use it on 405Ctr or 405Std, you would have to enable it.

glowtape:
If you remove the Critical HAL error and just set it missing, it should do the job. Like e.g. in the Seppuku, which has an onboard BMP280, @mlyle chose to handle it that way:

https://github.com/d-ronin/dRonin/blob/3a866925d6148e33e0620c94c5940413574a7247/flight/targets/seppuku/fw/pios_board.c#L418

jihlein:
That's a cleaner solution, will re-work.  Thanks for pointing that out.

jihlein:
Latest commit cleans up I2C sensor handling, should work on the Matek405Mini now.

ufoDziner:

fxbisto:
Ace.
I've pulled it, hopefully flash my F405-MINI at some point tomorrow to give feedback

fxbisto:
Configuration bug - Stabilization.ui is limited to 850 deg/s
Now the drop-down menu for gyro range is removed it wasn't explicitly set, leaving matek405.cpp responsible.
I can't add to this PR of course but here's the correct setting https://github.com/fxbisto/dRonin-MQF/commit/e32fd39851a55446ea731327f04a126190c73274

jihlein:
Crud - set the 2000 DPS in pios_board.c, but overlooked matek405.cpp.  Will get corrected file pushed later.

jihlein:
"If you remove the Critical HAL error and just set it missing, it should do the job."

I'm still not sure this is working correctly.  I can't test with the baro, but I can test with the external mag.

Even after making the suggested changes, and setting the tolerate missing sensors uavobject to true, the sensors are still flagged with a warning and the system refuses to arm.  @Glowtape, can you take a look at the code changes and see if I'm missing something?  Thanks!

Edit:  If I change the missing sensor severity from warning to ok (sensors.c line 692) when tolerate missing sensors is set true, the system will ignore the missing sensor, and arm.  Is warning the intended severity level here or is it a bug and it should be severity level ok??

glowtape:
From what I can tell, it should only refuse to arm on an error alarm.

https://github.com/d-ronin/dRonin/blob/3a866925d6148e33e0620c94c5940413574a7247/flight/Modules/ManualControl/manualcontrol.c#L388

Did you happen to have an altitude hold mode configured during the test? E.g. by copying a config of a baro Matek to the one without?

jihlein:
I'm testing by enabling the external mag when one is not connected.  I don't have a MINI to test with, only a CTR which has the BMP280.

Edit: It does work as you pointed out, there was an error in my test setup.  No need to change anything in sensors.c

glowtape:
Hrm, I'm not seeing anything obvious. If PIOS_HAL_ConfigureExternalMag fails, it sets the sensor as missing itself. Otherwise, you're setting it in pios_board.c. Sanity check doesn't seem to care about whether mags are missing or not.

I tested a non-existing mag on the Seppuku, and it lets me arm with missing sensor tolerance, and bitches without. So it is supposed to work.

Is it a yellow warning or a grey-red hatched one?

jihlein:
It does work correctly, I was testing with the controller gadget rather than a real RC system, and for some reason that wasn't working as expected until I rebooted my computer.  That was why I couldn't get the board to arm.

After the reboot, I can arm with the sensor status at warning (yellow) with tolerate missing sensor set true.

fxbisto:
@jihlein Have you had a flight with all these changes enabled?
I compiled my branch this morning but couldn't get a hover going, it just kept oscillating as if there was a mismatch or delay to the gyro data.
Had to drop P gains to 1 in order that the oscillation gave me time get it to leave the ground, but it was less than a second before it got out of control. With the default gains of 20 it was very aggressive. I dropped maxpoweradd to 0% which gave me a little more time but as soon as you got to about 15% throttle it was game over.
On both X and Y axis, P gains at 20 or 1 had the same frequency of oscillations, about 4 Hz.
I tried calibrating, using gyrobias, upping the LPF from 55/1 to 100/1 and disabling, no help.
Motors all on correct outputs, prop direction set correctly, tried all motor protocols (was using oneshot125 to start with)

I've only managed to get fix it by reverting changes all the way back to https://github.com/d-ronin/dRonin/pull/2251/commits/20546ce8e81103ec7e6ab7859978cc12a517e9e0 

It's my own fault for not flashing and hovering after every commit! Will go back and cherry-pick, then test hover each commit to see where it starts to dick about. I've a huge hunch it's all went pear shaped at https://github.com/d-ronin/dRonin/pull/2251/commits/7c9def838609dd39cd77c2b54c38fda591154234

This is on the CTR board.
@ufoDziner How have you got on? You're using a MINI with the ICM, right?

ufoDziner:
@fxbisto Yes, I'm using the mini.  I haven't been able to fly yet though, only setup (no wind).  It was supposed to be blowing today, but the forecast was wrong.  :-/

jihlein:
Pushed the fix for the GCS at 2000 DPS, and one more I2C cleanup.

I haven't flown these revisions yet, maybe tomorrow on the Matek405Std.  Bench test on Matek405Ctr looks okay FWIW.

jihlein:
I’ve got the firmware loaded on the Matek405Std which is on an RCE V4 tricopter.  It has one shot escs, ext mag, and gps.  Everything up to the point of radio calibration is setup and working correctly.  I’ll get the radio setup done later, and test motors and the tail servo.  Then it’s ready for a test hop.

jihlein:
@fxbisto - just flew a quick test up with the Matek405 target.  It worked fine with one odd exception.  I tested it on a tricopter, on which I usually fly a build tailored to tricopters.  With my tailored build, it autotunes the tail just fine, but with the "basic" dRonin for the Matek405, there was a nan in the yaw data and it didn't autotune the yaw axis, the yaw pids where left at their default values.

As far as the gyros and accels go, it worked just fine with the default PIDs.  It was a little squirrely, but it was what exactly what I would expect before the autotune.

It's a little on the breezy side here, but I'm going to try the autotune again and see if I get a yaw tune.

As far as I'm concerned, the Matek405 target seems okay, but I'd like to see some reports of it on the Matek405Mini, and flying on a quad if at all possible before a final review and merge are done.

fxbisto:
That's good to hear, hopefully it was just something that didn't merge properly in my build folder.
My plan is to fork from a new account on my Win10 machine so it's two entirely different compilers. That might take a while though as I'm mostly a Linux guy, not up to scratch on compiling for Windows.

I hoped to have some F405-MINI flight footage for you this weekend but it's been sideways rain all day so far. It's a 465g AUW 220mm stretched X with 2600Kv 2306 motors and Hulkie 5055 props - if there's a build that's got the torque and response to amplify the overly-sensitive ICM gyro it's certainly that guy.

jihlein:
FWIW, I've tried a number of different things, but I can't get a successful autotune on yaw.  Both gain and bias are computing as nan.  I don't have a quad to test this with, going to move my Matek405Std back to my custom dRonin build and see if I can successfully autotune there.

image

jihlein:
With my custom build, the gyro set at 1000 DPS, autotune works fine.

image

Going back out to try 2000 DPS and see what happens.

Other than the tricopter specific routines, the only other significant change is my custom build still has all the gyro settings as options in the hardware pane.

jihlein:
And here it is with the gyro range as 2000 DPS:

image

Still works.  But given the changes in the noise values, I think somewhere, somehow, in the new Matek 405 branch I've messed up the gyro initialization.  Off to review the code........

jihlein:
There is one difference, in my branch, and the first rev of the Matek405 target, the accel lpf is 99 Hz.  When I removed the UAVO settings for the gyro/accel parameters in a later commit, I set the accel lpf to 218 Hz.  Rebuilding code with 99 Hz, recharging batteries......

jihlein:
Okay, think I fixed it.  Turns out I was setting up the default values of the gyro and accel low pass filters incorrectly.  It now autotunes all three axes correctly.

image

My tricopter code sets the accel lpf to 99 Hz, this Matek405 target is using 218.  I think 218, being closer to the 188 that's been used for so long on the MPU6000, and the fact that the ICM has a reputation for being slightly noiser, is probably a better choice.  Comments?

@fxbisto - I'm betting this is what caused your initial issues.  My tricopters are not significantly overpowered, so that's probably why I was able to fly okay with the bad lpf settings.

fxbisto:
Great find, I'll get testing tomorrow with luck.
Is there any chance you can you squash changes into one commit on another branch? Don't worry if not, I'll just comb through the changes manually and update accordingly - it's just that every time I cherry-pick and then revert or try a rebase it seems to need manually merging for some reason. I'm still learning Git but thought it might just be the number of commits?

fxbisto:
> 218, being closer to the 188 that's been used for so long on the MPU6000, and the fact that the ICM has a reputation for being slightly noiser, is probably a better choice.

I wonder whether there's any need to take it that high and risk less-filtered data; as you said the ICM's are notoriously more prone to spewing much noisier data than the MPU60x0.
I'm of the impression the stabilization code doesn't run at a high enough refresh rate to take advantage of the lower latency, if that's the case then 99Hz should be a better choice.

jihlein:
Squashed commits.  I also changed the accel lpf to 99 Hz, your comments make sense and that's actually what I've been flying on the Matek405Std for quite some time, so there's history with that value.

glowtape:
Interesting. AFAIK the autotune data only saves PID output and rates, so I'm wondering where the NaN came from. That said, I figure 99 Hz instead of max. frequency on lowpass would be fine. I doubt the tiny increase in delay would affect the complementary filter at all, considering how slow it is in comparison.

ufoDziner:

ufoDziner:
Can any of you confirm if the current pad is working please?  It doesn't appear to be working for me.

ufoDziner:
I'm also getting a couple of errors.

sensors
adc

jihlein:
Have you set TolerateMissingSensors to true in the SensorSettings UAVO?

I've never seen that ADC error on the CTR or STD boards (never seen it on any board actually), will need to look into that one.

ufoDziner:
@jihlein I don't usually mess with the UAVO stuff (I'm not that good).  So, I've not messed with that setting at all.

fxbisto:
> Can any of you confirm if the current pad is working please? It doesn't appear to be working for me.

Sorry I haven't had a chance to load it on the MINI yet, but I should get a moment in a few hours. I'll do my best

jihlein:
For the 405 Mini:

Wrt the sensor error, when TolerateMissingSensors is set to true, the system health sensor box will turn yellow.  This is due to the fact that the mini does not have the baro sensor that the ctr and std do.  With the yellow caution, the system should arm.  A yellow sensor health indication will be the normal operating state for the mini.  I've checked this operation using the ext mag setting on the ctr, but have not been able to verify it on the mini with the missing baro, as I don't have a mini to test with.

Wrt to the ADC error, I can get that error to appear if I select an ADC input that has not been defined for that board.  ADC0, ADC1, and ADC2 have been defined for the Matek405 target, being current, voltage, and RSSI respectively.  If I select ADC3 on the battery module setup page for voltage, I get the same ADC error you showed above.  Setting back to ADC1 clears the error.  If you've been running Betaflight on this board, it might be a good idea to erase the settings sector after flashing the board to dRonin.

I did not get a chance to test it last night, but I'm fairly certain my Matek405Std with the FCHUB-W power distribution board correctly reads current on my RCE V4 Tricopter.

fxbisto:
I've run out of time to experiment, but my F405-MINI won't pick up my receiver. It's currently soldered to UART4, maybe that has bearing on something.
When I get back I might try another UART.
Another interesting fact, I managed to stop this board from booting by selecting ibus on every UART, whereas the CTR still boots up just dandy, but throws a warning that more than one identical Rx is configured.

ufoDziner:
@fxbisto I'm fairly certain that I'm using uart4 with no issues. Crossfire protocol though.

fxbisto:
Thanks @ufoDziner 
One variable in my results is that I haven't used the artifacts above; I compiled my own. Perhaps there's something amiss there - I'm charging my Windows laptop and will try those instead to see if the issue persists

fxbisto:
Quick progress report-
Pre-compiled target from here works fine, even though the source should be pretty much identical to what was used to compile it here and on my machine. They have different git hashes so something, somewhere is different. I'll have to look into that before mentioning issues again.

Current sensor works fine, although one thing to mention is that the 184A shunts built into the MatekF405 product line need a calibration factor of ~17.9 It's pretty darn accurate there for me on the CTR and the FCHUB-6S (both same 184A) Before I set this up the ADC status box kept flickering between orange and green, and no result was reported in live data.

ufoDziner:
Got to fly with it this weekend, but in manual mode only.  So far so good.  I've got 3 more coming tomorrow that will be going in the other gliders.  We're hoping to test them within the next week or so, then high speed testing on the weekend of the 27th.

ufoDziner:
Flew about a dozen flights in manual mode with no issues.  Setting up 2 more gliders to fly next weekend as well.  Will test the stabilization then.

mlyle (approved):
LGTM.

PR #2252, opened 2019-04-04
Flight: Add CF Accel Cutoff (Experimental - Not for merge)

jihlein:
This PR adds an accel cutoff to the complementary attitude filter.  It's purpose is to prevent the bias calculations from running away when the aircraft is subjected to flight loads above and below 1 G.

There is a new UAVObject, AccCutoff, in the attitude settings that defines how sharp the cutoff is.  It defaults to 0.1.  An accel confidence factor is calculated from the cutoff and current magnitude of the accel vector.  The accel confidence ranges from 0 to 1, and is multiplied against accKi and accKp, effectively freezing the bias terms during dynamic maneuvering.

This has not been flight tested.  On the bench it does not do anything strange.  Comments and flight test help appreciated.  I used this concept on a different project quite a few years back.  It seemed to work as desired.  But as always, test carefully.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -766,18 +766,31 @@ static int32_t updateAttitudeComplementary(float dT, bool first_run, bool second
 		mag_err[2] = 0;
 	}
 
+	////////////////////////////////////////
+	
+	// G.K. Egan (C) computes confidence in accelerometers when
+	// aircraft is being accelerated over and above that due to gravity
+
+	#define ACCEL_ONE_G (9.8065f)
+	
+	float accel_confidence_decay = 1.0f / sqrtf(attitudeSettings.AccCutoff);
+	
+	float accel_confidence = bound_min_max(1.0f - (accel_confidence_decay * sqrtf(fabs(accel_mag / ACCEL_ONE_G - 1.0f))), 0.0f, 1.0f);
+	
+	////////////////////////////////////////
+	
 	// Accumulate integral of error.  Scale here so that units are (deg/s) but Ki has units of s
 	GyrosBiasData gyrosBias;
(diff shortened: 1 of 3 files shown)


tracernz:
build artifacts please

glowtape:
I need to test some other attitude changes this weekend, I'll slap this on, too.

fxbisto:
Awesome, I'll get involved too. Now that I've seen how it can fly without the constant gyro biasing it'll be a good comparison. Probably won't be for a couple of days

glowtape:
Just did a living room hover test. The 0.1 default in this patch is definitely too low. It never gets the chance to debias and I think it even started drifting a bit more. 0.25 works in the living room.

glowtape:
So I went just flying a pack outside, with fake gyro drift and BiasCorrectOn and 0.25 cutoff. At the end, there was still a bit of unnecessary gyrobias, based on the tilt in leveling, but it was reduced a lot compared to before. So this is worthwhile to explore.

jihlein:
That’s a great start.  I think we need to tune the cutoff for the smallest value that works in hover, then see what happens in fast forward and acrobatic flight.  I don’t know if I’ll get to this this weekend or not....

glowtape:
The origin of that idea/code is GPL 3.0, for reference. https://github.com/vedderb/QuadcopterSystem/blob/master/LICENSE

jihlein:
Added the accelConfidence factor to the AttitudeActual UAVO for analysis purposes.

mlyle:
We played with this exact idea before --- I cannot find the pull request.  Consensus was that it didn't work (at least how I tuned/implemented it).  Would love for that to be wrong / for gating attitude updates to make sense.

mlyle:

mlyle:
jenkins, test this please

PR #2257, opened 2019-06-29
make: support msys(2) build env

tracernz:
This looks to be a much cleaner way to do things on Windows, and will
make buildbot builds much easier.

### DIFF
--- a/make/system-id.mk
+++ b/make/system-id.mk
@@ -33,6 +33,9 @@ ifeq (MINGW, $(findstring MINGW,$(UNAME)))
   OSFAMILY := windows
   MINGW := 1
   export WINDOWS := 1
+else ifeq(MSYS_NT, $(findstring MSYS_NT,$(UNAME)))
+  OSFAMILY := windows
+  export WINDOWS := 1
 endif
 
 # report an error if we couldn't work out what OS this is running on


glowtape (dismissed):
Don't have a build system setup to test, but it looks good to me, since it just sets a bunch of vars.

--edit: Jenkins balks at it for whatever reason. I suppose the "else" statement isn't strictly necessary, since it's either MINGW or MSYS_NT?

tracernz:
Jenkins fatal problem is
```
/bin/sh: /Users/Shared/Jenkins/Library/Developer/org.jenkins-ci.slave.jnlp/workspace/dron/Nodes/osx/tools/Qt5.9.2/5.9.2/mingw530_32/bin/qmake: No such file or directory
```
Looks like all the tools got hosed.

make/system-id.mk
```diff
@@ -33,6 +33,9 @@ ifeq (MINGW, $(findstring MINGW,$(UNAME)))
   OSFAMILY := windows
   MINGW := 1
   export WINDOWS := 1
+else ifeq(MSYS_NT, $(findstring MSYS_NT,$(UNAME)))
+  OSFAMILY := windows
+  export WINDOWS := 1
 endif
```
tracernz, line 39: ~~Think I missed an endif here.~~

Nope, it's right.

make/system-id.mk
```diff
@@ -33,6 +33,9 @@ ifeq (MINGW, $(findstring MINGW,$(UNAME)))
   OSFAMILY := windows
   MINGW := 1
   export WINDOWS := 1
+else ifeq(MSYS_NT, $(findstring MSYS_NT,$(UNAME)))
```
tracernz, line 36: supposed to have a space before the parens

mlyle:

mlyle:
@tracernz --- jenkins CI (as opposed to artifact builds) are on mac.  It looks like it's trying to find qmake in a mingw (not clang64) directory on the mac build.

PR #2259, opened 2021-05-31
Updated QT url location in tools.mk

geosmall:
QT changed URL for QT package download, breaking 'make qt_sdk_install'

This pull updattes QT url location in tools.mk in case others want to fix.

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -118,7 +118,9 @@ endef
 .PHONY: qt_sdk_install
 
 # QT SDK download URL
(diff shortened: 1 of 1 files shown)


tracernz:
Are you still building on it? 😀The CI is gone so the check won’t complete, but this looks fine, and wouldn’t be tested by CI anyway.

geosmall:
Yes still messing about with adding modules for vtols over here https://github.com/jihlein/dronin/tree/myDronin.  Build includes GCS spin on Ubuntu 18.04 box.

